PR #233, opened 2026-03-17
fix: replace email-keyed GCS metadata with member_N fixed keys

natb1:
## Problem

Authenticated users get 403 on all private media items (e.g. `/view/a-pound-of-flesh`).

Root cause: `storage.rules` used `resource.metadata[request.auth.token.email]` — a dynamic CEL bracket lookup where the key contains `@`. The `@` character causes this lookup to silently return null in production GCS (the emulator handled it differently). The metadata was stored correctly; the lookup itself was broken.

## Fix

**`storage.rules`** — replace the dynamic email-key lookup with three fixed-key slots (`member_0`, `member_1`, `member_2`) storing emails as values:

```diff
- || (request.auth != null && resource.metadata[request.auth.token.email] == "member");
+ || (request.auth != null && (
+     resource.metadata["member_0"] == request.auth.token.email
+  || resource.metadata["member_1"] == request.auth.token.email
+  || resource.metadata["member_2"] == request.auth.token.email));
```

**`print/seeds/storage.ts`** — update `testPrivateMeta` to match the new scheme:

```diff
- const testPrivateMeta = { publicDomain: "false", "test@example.com": "member" };
+ const testPrivateMeta = { publicDomain: "false", member_0: "test@example.com" };
```

**GCS objects** — updated `x-goog-meta-member_0:nathan@natb1.com` and cleared the old `x-goog-meta-nathan@natb1.com` key on all 68 objects in `print/prod/media/` via `gsutil setmeta` (0 errors).

## Verification

1. Merge this PR → `storage-deploy.yml` deploys updated rules automatically
2. Log in as `nathan@natb1.com` at `print.commons.systems`
3. Navigate to a private item (e.g. `/view/a-pound-of-flesh`)
4. Confirm PDF viewer loads without 403
5. Confirm public items still load without auth

### DIFF
--- a/print/seeds/storage.ts
+++ b/print/seeds/storage.ts
@@ -37,7 +37,7 @@ function makePdf(pageCount: number): string {
 }
 
 const publicMeta = { publicDomain: "true" };
-const testPrivateMeta = { publicDomain: "false", "test@example.com": "member" };
+const testPrivateMeta = { publicDomain: "false", member_0: "test@example.com" };
 
 const storageSeed: StorageSeedItem[] = [
   {
--- a/storage.rules
+++ b/storage.rules
@@ -3,7 +3,10 @@ service firebase.storage {
   match /b/{bucket}/o {
     match /print/{env}/media/{fileName} {
       allow read: if resource.metadata["publicDomain"] == "true"
-        || (request.auth != null && resource.metadata[request.auth.token.email] == "member");
+        || (request.auth != null && (
+            resource.metadata["member_0"] == request.auth.token.email
+         || resource.metadata["member_1"] == request.auth.token.email
+         || resource.metadata["member_2"] == request.auth.token.email));
       allow write: if false;
     }
     match /{allPaths=**} {


natb1:
=== PR REVIEW (review-pr skill) ===
PR #233: fix-storage-member-metadata
Review of commit 63e846f ("fix: replace email-keyed GCS metadata with member_N fixed keys")

## Summary

This PR changes the storage rules access-control scheme for private media files.
Previously, member access was checked by using the user's email address as a metadata
key (e.g., `resource.metadata["user@example.com"] == "member"`). The new scheme uses
positional fixed keys (`member_0`, `member_1`, `member_2`) whose values are the member
email addresses (e.g., `resource.metadata["member_0"] == request.auth.token.email`).

This is a sound change -- Cloud Storage custom metadata keys containing `@` and `.`
characters are fragile (potential encoding issues, tooling incompatibilities). Fixed
positional keys are more predictable and standard.

## Files Changed

1. `storage.rules` -- Access rule updated from dynamic email-key lookup to checking
   three fixed positional keys.
2. `print/seeds/storage.ts` -- Test seed updated to use the new `member_0` key format.

## Issues

### 1. Stale documentation in print/seeds/firestore.ts (must fix)

Lines 9-10 of `print/seeds/firestore.ts` still document the old email-keyed scheme:

    //    - Private: gsutil setmeta -h "x-goog-meta-publicDomain:false" -h "x-goog-meta-{email}:member" gs://.../<file>
    // Note: Firestore uses memberEmails array for access control; Storage uses per-email custom metadata keys.

This must be updated to document the new `member_N` key format. Anyone following these
production seeding instructions will set metadata in the old format, which the new rules
will reject.

Suggested replacement:

    //    - Private: gsutil setmeta -h "x-goog-meta-publicDomain:false" -h "x-goog-meta-member_0:{email}" gs://.../<file>
    // Note: Firestore uses memberEmails array for access control; Storage uses member_N positional metadata keys.

### 2. Hard-coded member cap of 3 (worth discussing)

The rule checks exactly three keys: `member_0`, `member_1`, `member_2`. This limits
private media items to at most 3 members. This is a Firebase Storage rules limitation --
there is no array type or `in` operator for custom metadata, so enumeration is the only
option. However:

- The limit should be documented somewhere visible (e.g., a comment in `storage.rules`).
- If the limit needs to grow, every slot addition requires a rules redeploy. Consider
  adding a brief comment in `storage.rules` explaining the pattern and the cap.

### 3. Production data migration not mentioned

If any production Cloud Storage objects already use the old email-keyed metadata format,
they will become inaccessible to members once these rules deploy (public-domain files are
unaffected). The PR description or a migration note should confirm whether production
private objects exist and whether they have been (or will be) migrated to the new format
before merging.

## Correctness

The rule logic is correct. The `request.auth != null` guard is properly applied before
accessing `request.auth.token.email`. The `||` chaining across the three member slots is
syntactically valid Firebase rules syntax. The seed data correctly mirrors the rule
expectations.

## Deployment

Per `storage-deploy.yml`, storage rules auto-deploy on merge to main via
`npx firebase-tools deploy --only storage:rules`. This is consistent with the existing
deployment pattern.

## Verdict

The core change is good. Before merging:
1. Update the gsutil instructions in `print/seeds/firestore.ts` (lines 9-10).
2. Add a comment in `storage.rules` noting the 3-member cap.
3. Confirm production migration status for any existing private objects.

=== CODE REVIEWER (code-reviewer agent) ===
Code Review: PR #233 (fix-storage-member-metadata)

Files reviewed:
  - /Users/n8/natb1/commons.systems/worktrees/fix-storage-member-metadata/storage.rules
  - /Users/n8/natb1/commons.systems/worktrees/fix-storage-member-metadata/print/seeds/storage.ts

No high-confidence issues found (all below threshold of 80).

Summary:

The fix is correct and well-scoped. Dynamic CEL bracket lookups on metadata
keys containing special characters (like `@`) silently fail in Firebase
Storage security rules. Switching from email-keyed metadata
(`"test@example.com": "member"`) to positional keys (`member_0:
"test@example.com"`) with static string lookups is a sound workaround.

The storage.rules change uses three fixed member slots (member_0 through
member_2). This is a reasonable ceiling for per-file member lists. The
approach trades flexibility for correctness -- Firebase security rules do
not support iteration, so enumerating slots is the standard pattern.

The seed data in storage.ts correctly mirrors the new key schema with
`member_0: "test@example.com"`.

Style and formatting are consistent with the existing codebase. No project
guideline violations detected.

=== SIMPLIFY (simplify skill) ===
Simplify Review: PR #233 (fix-storage-member-metadata)
=======================================================

Files changed: storage.rules, print/seeds/storage.ts

Summary of change: Replaces email-as-key storage metadata format
(e.g., "test@example.com": "member") with indexed member keys
(e.g., member_0: "test@example.com"). This avoids using email
addresses as metadata keys, which is problematic because Firebase
Storage metadata keys have character restrictions and using
user-supplied data as keys is a security smell.

------------------------------------------------------------------------
Agent 1: Code Reuse Review
------------------------------------------------------------------------

No existing utilities or helpers exist for generating the member_N
metadata keys. The indexed key scheme is only used in two places
(storage.rules and print/seeds/storage.ts), both of which are
changed in this PR.

The Firestore layer (print/src/firestore.ts) uses a memberEmails
array field, while Storage uses indexed member_N metadata keys.
These are necessarily different representations due to Firebase
Storage metadata supporting only string key-value pairs (no arrays).
No shared abstraction is possible here.

Findings: None.

------------------------------------------------------------------------
Agent 2: Code Quality Review
------------------------------------------------------------------------

Finding 1 (FIXED): Stale comment in print/seeds/firestore.ts

  Line 9 documented the old metadata format:
    -h "x-goog-meta-{email}:member"
  This no longer matches the new member_N format. Updated to:
    -h "x-goog-meta-member_0:{email}"
  Also updated the explanatory note on line 10 to say "indexed
  member_N metadata keys" instead of "per-email custom metadata keys".

Finding 2 (Accepted - Firebase limitation): Copy-paste in storage.rules

  The three repeated lines checking member_0, member_1, member_2 are
  a copy-paste pattern. However, Firebase Storage security rules do
  not support loops, list operations, or custom functions that iterate.
  Enumerating indexed keys is the standard workaround. The hard-coded
  limit of 3 members is a design constraint that should be documented
  but cannot be eliminated within the rules language.

  The seed data and Firestore schema (memberEmails array) do not
  enforce this limit, so a Firestore document could list more than 3
  memberEmails while the storage rule silently denies access to
  member_3+. This is a latent inconsistency to be aware of but is
  outside the scope of this PR's storage.rules change.

Finding 3 (Clean): No stringly-typed code concerns. The member_N
  prefix convention is used consistently between the seed data and
  the rules file.

------------------------------------------------------------------------
Agent 3: Efficiency Review
------------------------------------------------------------------------

Finding 1 (Clean): The storage rules evaluate at most 4 metadata
  lookups per read request (publicDomain + 3 member checks). Firebase
  short-circuits on the first truthy || branch, so public-domain
  reads hit only 1 metadata check. This is efficient for a security
  rule.

Finding 2 (Clean): No unnecessary existence checks, no N+1 patterns,
  no missed concurrency. The seed script (run-storage-seed.ts) uploads
  items sequentially, but this is a dev-only emulator seed with 4
  items -- parallelizing it would add complexity for negligible gain.

Findings: None.

------------------------------------------------------------------------
Fixes Applied
------------------------------------------------------------------------

1. Updated stale production seeding comment in print/seeds/firestore.ts
   (lines 9-10) to document the new member_N metadata key format.

No other changes needed. The code is clean.

=== COMMENT ANALYZER ===
Comment Analysis: PR #233 (fix-storage-member-metadata)
=========================================================

Scope: storage.rules (16 lines), print/seeds/storage.ts (66 lines)

Summary
-------
Both files contain very few comments -- only one inline comment exists across
the two changed files. The code is terse and mostly self-documenting through
naming conventions. The primary concern is not bad comments but *missing*
comments: the fix addresses a subtle Firebase CEL evaluation quirk that will
be invisible to future maintainers without explanation.


Critical Issues
---------------
None. There are no factually incorrect comments.


Improvement Opportunities
-------------------------

1. storage.rules -- Missing comment explaining the member_N workaround

   Location: storage.rules:7-9
   Current state: The three member_0/member_1/member_2 comparisons appear
   without any explanation. A future maintainer will reasonably ask: why not
   use a dynamic lookup like resource.metadata[request.auth.token.email]?
   That looks cleaner and avoids the hard-coded slot limit. Without context,
   someone will "simplify" this back to the dynamic form, reintroducing the
   403 bug.

   Suggestion: Add a comment above the member checks explaining the
   constraint. Something like:

     // Firebase Storage rules evaluate bracket expressions with dynamic keys
     // containing special characters (e.g. '@' in emails) as undefined,
     // causing silent 403s. Use fixed member_N keys instead. Three slots
     // support up to three members per file.

   This documents *why* the seemingly redundant pattern exists, which is the
   single most important piece of context this PR introduces.

2. storage.rules -- Missing comment on the three-slot limit

   Location: storage.rules:7-9
   Current state: The cap of three member slots (member_0 through member_2)
   is implicit. There is no indication of whether three is arbitrary, derived
   from a product requirement, or a Firebase metadata size constraint.

   Suggestion: Either in the comment suggested above or separately, note the
   rationale for three slots, or at minimum flag it as a limit:

     // Three slots; increase if files need more members.

3. print/seeds/storage.ts -- Missing comment on testPrivateMeta purpose

   Location: print/seeds/storage.ts:40
   Current state: testPrivateMeta is defined with a single member_0 entry.
   Its role as a seed fixture for testing the private-access rule path is
   clear from variable naming, but there is no indication that this metadata
   shape must stay in sync with the member_N pattern in storage.rules. If
   someone changes the key scheme in one file but not the other, tests will
   silently pass while production breaks (or vice versa).

   Suggestion: A brief inline comment linking the two:

     // Metadata keys must match the member_N pattern in storage.rules
     const testPrivateMeta = { publicDomain: "false", member_0: "test@example.com" };


Recommended Removals
--------------------
None. There are no comments to remove.


Existing Comments Reviewed
--------------------------

1. print/seeds/storage.ts:24
   Code:    const objCount = 1 + offsets.length; // free entry + objects
   Verdict: Accurate and useful. In the PDF cross-reference table, entry 0 is
   the "free entry" (generation 65535) and the remaining entries correspond to
   the indirect objects tracked in `offsets`. The arithmetic checks out: the
   xref table declares `objCount` entries starting at object 0, which is
   exactly 1 (free) + offsets.length (objects). This comment earns its place
   because the "+1" would otherwise be a magic number requiring PDF spec
   knowledge to understand.


Positive Findings
-----------------
- The naming conventions (publicMeta, testPrivateMeta, member_0) are clear
  and self-documenting, reducing the need for comments on the "what."
- The single existing comment (free entry + objects) is a good example of
  explaining a non-obvious constant.


Overall Assessment
------------------
The code itself is clean, but this PR fixes a subtle platform-specific bug
that will be completely non-obvious to anyone reading storage.rules in the
future. The highest-value action is adding a comment in storage.rules
explaining why dynamic bracket lookups on email keys fail, so that no one
reverts to the simpler-looking (but broken) pattern.

=== PR TEST ANALYZER ===
PR #233 Test Coverage Analysis: fix-storage-member-metadata
============================================================

1. SUMMARY
----------
This PR fixes a 403 bug in Firebase Storage Security Rules where dynamic CEL
bracket lookups on metadata keys containing `@` silently fail. The fix replaces
email-keyed metadata (e.g., `"test@example.com": "member"`) with fixed
positional keys (`member_0`, `member_1`, `member_2`) whose values are the
member email addresses. Two files changed: `storage.rules` and
`print/seeds/storage.ts`.

The existing acceptance test suite in `print/e2e/auth.spec.ts` provides
end-to-end coverage of the exact scenario this PR fixes: an authenticated user
seeing 4 items (3 public + 1 private) after sign-in. That test runs against
the Firebase Storage emulator with `storage.rules` loaded, seeded by
`print/seeds/storage.ts`. If the metadata format and rule logic are
incompatible, this test fails with only 3 visible items. This is the most
important test for this PR and it already exists.

Overall test coverage for this change is adequate. The recommendations below
are ordered by criticality.


2. CRITICAL GAPS
----------------
None. The acceptance test at `print/e2e/auth.spec.ts` line 18 ("private items
visible after login (4 total)") exercises the full path: emulator loads
`storage.rules`, storage is seeded with the `member_0` metadata key, the test
user authenticates, and the app attempts to download the private item. A
mismatch between the rule and the seed metadata causes the download to 403,
which the app surfaces as fewer than 4 items.


3. IMPORTANT IMPROVEMENTS
-------------------------

a) Stale documentation in firestore seed comment (Criticality: 7/10)
   File: /Users/n8/natb1/commons.systems/worktrees/fix-storage-member-metadata/print/seeds/firestore.ts
   Lines 9-11 contain manual-seeding instructions that still reference the old
   metadata format:

     "Private: gsutil setmeta -h "x-goog-meta-{email}:member" gs://.../<file>"

   This is now incorrect. The production manual procedure should use:

     gsutil setmeta -h "x-goog-meta-member_0:{email}" gs://.../<file>

   The comment on line 11 also says "Storage uses per-email custom metadata
   keys" which no longer describes the new scheme. If someone follows these
   stale instructions to onboard a new private item in production, it will
   silently fail with 403 -- the exact bug this PR is fixing.

   This is not a test gap per se, but it is the most likely source of a
   regression caused by the PR.

b) No negative acceptance test for unauthorized access to private items
   (Criticality: 6/10)
   The acceptance tests verify that an authenticated member CAN see private
   items, but there is no test verifying that an unauthenticated user (or a
   non-member) CANNOT download private items. A future rules regression that
   accidentally opens private items to everyone would not be caught. Consider
   adding an e2e test that:
   - Loads the page without signing in
   - Verifies only 3 items appear (already exists at auth.spec.ts line 42)
   - Attempts to directly fetch the private item's storage URL and confirms
     it fails

   The existing test "public media visible without auth" (line 39) counts 3
   items, which provides partial coverage -- if a rules regression exposed the
   private item to unauthenticated users at the Firestore level, this count
   would change to 4. However, it does not cover direct storage URL access.

c) No test for member_1 or member_2 slots (Criticality: 5/10)
   The storage rule now supports up to 3 members (member_0, member_1,
   member_2), but the seed data only populates member_0. A typo or copy-paste
   error in the member_1 or member_2 rule lines would go undetected. Consider
   adding a seed item with member_1 or member_2 populated, along with a
   corresponding Firestore document, to verify that non-zero slots work.

   Regression prevented: a broken member_1/member_2 clause would mean only
   the first member of any shared item gets access.


4. TEST QUALITY ISSUES
----------------------

a) The unit test in `print/test/storage.test.ts` tests the client-side
   `getMediaDownloadUrl` helper, which mocks Firebase Storage entirely. It
   does not exercise security rules at all. This is fine -- it tests a
   different concern -- but it should not be mistaken for storage rules
   coverage.

b) The acceptance tests rely on item count (3 vs 4) as a proxy for storage
   access control. This is a reasonable behavioral assertion, but it
   conflates Firestore visibility with Storage download access. If the app
   ever changes to show items in the list regardless of download permission
   (e.g., showing a "request access" button), the test would pass even with
   broken storage rules. This is a low-priority concern for now but worth
   noting.


5. POSITIVE OBSERVATIONS
------------------------

- The end-to-end acceptance test in `print/e2e/auth.spec.ts` directly
  exercises the scenario this PR fixes. The test infrastructure boots the
  Firebase Storage emulator with `storage.rules`, seeds it via
  `run-storage-seed.ts` (which uses the updated `storage.ts` seed), and
  runs Playwright against the full app. This is the right level of testing
  for security rules.

- The seed data in `print/seeds/storage.ts` is consistent with the rule
  change: `member_0: "test@example.com"` matches the `member_0` lookup in
  `storage.rules`.

- The Firestore seed's `memberEmails: ["test@example.com"]` and the storage
  seed's `member_0: "test@example.com"` are consistent, so the Firestore
  query (which controls list visibility) and the Storage rule (which controls
  download access) agree.

- The acceptance test harness in `run-acceptance-tests.sh` correctly detects
  storage usage, starts the storage emulator, loads `storage.rules`, and
  runs the storage seed. No manual test infrastructure changes were needed.

=== SILENT FAILURE HUNTER ===
Silent Failure Audit: PR #233 (fix-storage-member-metadata)
============================================================

SUMMARY
-------
This PR replaces email-keyed storage metadata (e.g., "test@example.com": "member")
with fixed positional keys (member_0, member_1, member_2) to work around Firebase
Storage Security Rules rejecting dynamic CEL bracket lookups on keys containing "@".

The fix is correct for what it touches. However, the PR introduces two silent failure
scenarios -- one critical -- and leaves a documentation landmine that will cause the
exact same class of 403 bug to recur.


ISSUE 1 -- CRITICAL: Stale production metadata instructions will cause silent 403s
-----------------------------------------------------------------------------------

Location: print/seeds/firestore.ts, lines 8-11
Severity: CRITICAL (silent failure in production)

The file contains manual gsutil instructions for setting metadata on production
storage objects. These instructions were NOT updated by the PR:

    - Private: gsutil setmeta -h "x-goog-meta-publicDomain:false" \
               -h "x-goog-meta-{email}:member" gs://.../<file>

    Note: Firestore uses memberEmails array for access control; Storage uses
    per-email custom metadata keys.

These instructions still tell operators to set the OLD email-keyed format
("x-goog-meta-{email}:member") that this PR exists to eliminate. The new
storage.rules expect "member_0", "member_1", "member_2" keys whose VALUES are
email addresses -- the exact inverse of the old format (where the key was the
email and the value was "member").

User impact: The next time someone follows these instructions to grant a member
access to a private storage object, they will set metadata in the old format.
The new storage.rules will not match it. The user will get a 403 with no
indication of why. This is the exact same silent failure class the PR is fixing,
reintroduced through stale documentation.

Hidden failure mode: Firebase Security Rules return a generic "permission denied"
when a metadata check fails. There is no log, no diff, and no indication that the
metadata key format is wrong. The operator will see the metadata is set and assume
it should work.

Recommendation: Update the gsutil instructions to match the new key format:

    - Private (up to 3 members):
        gsutil setmeta \
          -h "x-goog-meta-publicDomain:false" \
          -h "x-goog-meta-member_0:{email}" \
          gs://.../<file>

    Note: Storage uses member_0, member_1, member_2 keys whose values are
    email addresses. Firestore uses a memberEmails array. Both must be kept
    in sync.


ISSUE 2 -- HIGH: Hard cap of 3 members silently denies access to the 4th
-------------------------------------------------------------------------

Location: storage.rules, lines 5-9
Severity: HIGH (silent failure at runtime)

The rules check exactly three fixed keys: member_0, member_1, member_2. If a
storage object has a member_3 key (or higher), that member will be silently
denied access. There is no validation anywhere in the codebase that enforces
this limit:

  - print/src/types.ts defines memberEmails as string[] with no length constraint
  - print/src/firestore.ts (requireStringArray) validates that it is an array of
    strings but does not check length
  - print/seeds/run-storage-seed.ts uploads whatever metadata the seed provides
    with no slot-count validation
  - The gsutil instructions have no mention of a 3-member limit

This means the system will accept 4+ members in Firestore (memberEmails array)
and an operator could set member_3 on a storage object, but the security rules
will ignore it. The 4th member gets a 403 with no explanation.

User impact: Members beyond the 3rd are silently locked out. The Firestore data
says they have access, the storage metadata says they have access, but the
security rules disagree. Debugging this requires reading storage.rules and
counting the member_N checks by hand.

Recommendation: At minimum, add a comment to storage.rules documenting the limit:

    // IMPORTANT: Only member_0 through member_2 are checked.
    // Adding more members requires adding more lines here.
    // See also: print/seeds/firestore.ts for manual metadata instructions.

Better: Add a runtime validation in the seed/upload path that rejects metadata
with more than 3 member slots and throws a clear error:

    const MAX_STORAGE_MEMBER_SLOTS = 3;
    const memberKeys = Object.keys(metadata).filter(k => /^member_\d+$/.test(k));
    if (memberKeys.length > MAX_STORAGE_MEMBER_SLOTS) {
      throw new Error(
        `Storage rules only support ${MAX_STORAGE_MEMBER_SLOTS} member slots, ` +
        `but ${memberKeys.length} were provided for ${item.path}. ` +
        `Update storage.rules to add more member_N checks.`
      );
    }


ISSUE 3 -- MEDIUM: No integration test for member-keyed access control
-----------------------------------------------------------------------

Location: (missing)
Severity: MEDIUM

The existing storage.test.ts tests only getMediaDownloadUrl (a client-side path
builder). There is no test that exercises the storage.rules security logic with
the new member_N metadata format. The seed data only sets member_0 for a single
test user, so it does not exercise:

  - member_1 or member_2 slots
  - A user who is NOT in any member slot being correctly denied
  - The boundary case where all 3 slots are filled

The original bug (email-keyed metadata silently failing in CEL) went undetected
presumably because there was no integration test exercising the rules. This PR
fixes the format but does not add a test to prevent regression.

User impact: If a future change to storage.rules breaks member access (e.g.,
someone adds a typo in a key name, or changes the comparison operator), it will
again be caught only by manual testing in production.

Recommendation: This is outside the immediate scope of the PR, but worth noting.
An emulator-based integration test that uploads an object with member_0 metadata,
authenticates as that user, and verifies the download succeeds (and fails for a
non-member) would prevent this entire class of bug from recurring.


POSITIVE OBSERVATIONS
---------------------

1. run-storage-seed.ts (line 32-35) correctly throws on upload failure with the
   HTTP status and response body. This is good error handling -- it surfaces the
   exact problem and does not silently continue.

2. The Firestore access layer (print/src/firestore.ts) has thorough data
   validation with specific DataIntegrityError messages for every field. This is
   well done.

3. print/test/storage.test.ts includes a test that errors from getDownloadURL
   propagate rather than being caught and swallowed. This is the right approach.

4. The storage.rules deny-all catch-all (allow read, write: if false) at the
   bottom prevents accidental open access on unmatched paths.


CONCLUSION
----------

The critical issue is #1: the stale gsutil instructions in print/seeds/firestore.ts.
This is a documentation-as-code problem -- those instructions are the production
runbook for setting storage metadata, and they now describe the OLD format that the
PR explicitly removes. Following them will reproduce the exact 403 bug this PR fixes.
This must be updated before merge.

Issue #2 (the silent 3-member cap) should at minimum be documented in a code comment.
Ideally it should be enforced with a validation check that throws a clear error when
exceeded.

=== TYPE DESIGN ANALYZER ===
## Type: StorageSeedItem

```typescript
export interface StorageSeedItem {
  path: string;
  content: string;
  metadata: Record<string, string>;
}
```

### Invariants Identified

1. **Path structure**: Paths must follow the pattern `print/{env}/media/{fileName}` to
   match the storage rules. Nothing in the type enforces this.

2. **Metadata key convention for members**: Member access is encoded as `member_0`,
   `member_1`, `member_2` keys mapping to email addresses. The storage rules hard-code
   exactly these three keys. Nothing in the type constrains metadata to use this
   convention.

3. **publicDomain flag**: The `publicDomain` metadata key must be either `"true"` or
   `"false"` (as strings). The type allows any string value and does not require the
   key to be present.

4. **Member slot limit**: The storage rules check exactly three member slots
   (`member_0` through `member_2`). The metadata type allows arbitrary keys, so a
   seed could declare `member_5` and silently have no effect.

5. **Content must be valid for its file extension**: PDF paths should contain valid PDF
   content. The type treats content as an opaque string.

### Ratings

- **Encapsulation**: 3/10
  `StorageSeedItem` is a plain data interface with no behavior and no access
  restrictions. All fields are publicly readable and writable. This is a seed/fixture
  type, so some openness is expected, but the complete absence of structure in
  `metadata` means the type offers no protection against misconfiguration.

- **Invariant Expression**: 2/10
  The critical invariants -- path format, member-slot naming convention, the
  publicDomain flag requirement, and the three-slot member limit -- are entirely
  invisible in the type definition. `Record<string, string>` is maximally permissive
  and communicates nothing about what keys or values are expected. A developer reading
  only the TypeScript type has no way to know that `member_0` is meaningful or that
  `member_3` would be silently ignored.

- **Invariant Usefulness**: 6/10
  The invariants themselves (path structure, member-slot convention, public-domain
  flag) are useful and directly prevent the 403 bug this PR fixes. The problem is not
  the invariants but that they exist only in `storage.rules` and developer knowledge,
  not in the type system.

- **Invariant Enforcement**: 1/10
  There is no enforcement at all. No construction-time validation, no compile-time
  constraints, no runtime checks. The metadata object is an untyped bag of strings.
  The only thing preventing a broken seed is the developer remembering the convention.

### Strengths

- The fix itself is correct: moving from email-as-key to indexed member slots avoids
  the Firebase metadata key restriction that caused the 403.
- The seed data is co-located with its type definition, making it easy to find.
- The type is simple and easy to understand at a surface level.

### Concerns

1. **Silent drift between rules and seeds.** The storage rules hard-code `member_0`
   through `member_2`. The TypeScript type has no knowledge of this limit. If someone
   adds a fourth member (`member_3`) to a seed, it will be silently ignored by the
   rules, causing a confusing access-denied error identical to the original bug.

2. **No compile-time link between rules and metadata shape.** The metadata convention
   is a contract between `storage.rules` and the seed data, but that contract is
   enforced only by reading two files in two different languages and mentally
   cross-referencing them.

3. **The member-slot pattern does not scale.** Three hard-coded slots is a ceiling.
   Adding a fourth member requires editing `storage.rules`, and there is no type-level
   signal that this is necessary. This is acceptable if three is a known, stable
   business limit, but if membership lists grow, this design will become a recurring
   source of bugs.

4. **publicDomain is stringly typed.** The value `"true"` vs `"false"` as a string is
   a Firebase metadata constraint, but nothing prevents a seed from setting
   `publicDomain: "yes"` or omitting the key entirely, which would silently deny all
   reads.

### Recommended Improvements

1. **Narrow the metadata type to express the member-slot convention.**

   Replace `Record<string, string>` with a discriminated structure that makes the
   valid key space explicit:

   ```typescript
   interface MemberSlots {
     member_0?: string;
     member_1?: string;
     member_2?: string;
   }

   interface StorageMetadata extends MemberSlots {
     publicDomain: "true" | "false";
   }

   export interface StorageSeedItem {
     path: string;
     content: string;
     metadata: StorageMetadata;
   }
   ```

   This gives compile-time errors when someone writes `member_3` or sets
   `publicDomain: "yes"`. The cost is low -- one small interface -- and it directly
   prevents the class of bug this PR fixes.

2. **Define the member-slot limit as a constant shared between rules generation and
   seed construction.** If the storage rules are ever generated or templated rather
   than hand-written, a single `MAX_MEMBERS = 3` constant could drive both the rules
   loop and the TypeScript type. Even without rules generation, a comment in
   `storage.rules` referencing the TypeScript type (and vice versa) would help.

3. **Add a brief comment documenting the constraint origin.** At minimum, a one-line
   comment above `testPrivateMeta` explaining that member slots are limited to
   `member_0` through `member_2` because `storage.rules` checks exactly those keys
   would save the next developer from rediscovering this the hard way.

### Summary

The PR correctly fixes the 403 bug by switching from email-as-key (which Firebase
rejects) to indexed member slots. The fix works. The type design concern is that the
new member-slot convention introduces a contract between `storage.rules` and
`StorageSeedItem.metadata` that is entirely invisible to the compiler. Narrowing the
metadata type from `Record<string, string>` to a structured interface with literal
types would make this contract explicit and prevent the same category of silent
misconfiguration that caused the original bug.


---

# Code Quality Evaluation — PR #233 (fix-storage-member-metadata)

## Aggregated Findings (deduplicated across 7 review tasks)

### 1. Stale production seeding docs in `print/seeds/firestore.ts`
- **Sources:** review-pr, simplify, pr-test-analyzer, silent-failure-hunter (4/7 agents)
- **Classification:** REQUIRED — low priority
- **Status:** ALREADY FIXED by simplify agent (lines 9-11 updated to `member_N` format)
- **Rationale:** The old `x-goog-meta-{email}:member` instructions would cause the exact 403 this PR fixes. Fix already applied, needs commit.

### 2. Add comment in `storage.rules` explaining why fixed keys are needed
- **Sources:** comment-analyzer, review-pr
- **Classification:** REQUIRED — low priority
- **Rationale:** Without context, a future maintainer may "simplify" back to `resource.metadata[request.auth.token.email]`, reintroducing the bug. A one-line comment prevents this.

### 3. Hard cap of 3 members undocumented
- **Sources:** review-pr, comment-analyzer, silent-failure-hunter, type-design-analyzer
- **Classification:** OUT OF SCOPE
- **Rationale:** This is a known Firebase Storage rules limitation (no loops/array ops). The cap is inherent in the approach. A comment noting the limit (finding #2) is sufficient. Changing the cap or adding validation is beyond this bug-fix PR.

### 4. Narrow `StorageSeedItem.metadata` type from `Record<string, string>` to structured interface
- **Sources:** type-design-analyzer
- **Classification:** OUT OF SCOPE
- **Rationale:** Would improve compile-time safety but is a refactoring change beyond the scope of this bug fix. Good follow-up work.

### 5. Production migration of existing objects
- **Sources:** review-pr
- **Classification:** OUT OF SCOPE
- **Rationale:** Operational concern, not a code change. The PR description should confirm migration status, but this is not a code quality finding.

### 6. No negative auth test / only member_0 tested
- **Sources:** pr-test-analyzer, silent-failure-hunter
- **Classification:** OUT OF SCOPE
- **Rationale:** The existing e2e test exercises the fix end-to-end. Additional test coverage for member_1/member_2 and denial cases is valuable but beyond this bug-fix scope.

### 7. No linkage between storage.rules and seeds file
- **Sources:** comment-analyzer
- **Classification:** FALSE POSITIVE
- **Rationale:** The files are inherently coupled by the Firebase emulator setup. Adding cross-references would be over-documentation.

## Summary

| # | Finding | Classification | Priority |
|---|---------|---------------|----------|
| 1 | Stale docs in firestore.ts | REQUIRED (already fixed) | low |
| 2 | Add explanatory comment in storage.rules | REQUIRED | low |
| 3 | 3-member cap undocumented | OUT OF SCOPE | — |
| 4 | Narrow metadata type | OUT OF SCOPE | — |
| 5 | Production migration | OUT OF SCOPE | — |
| 6 | Additional test coverage | OUT OF SCOPE | — |
| 7 | Cross-file linkage | FALSE POSITIVE | — |

**High-priority required findings: 0**
**Low-priority required findings: 2** (one already implemented)

Awaiting user approval to proceed with termination (implement remaining low-priority finding #2, commit, and post).

PR #284, opened 2026-03-19
Add password-based encryption for budget JSON files

### LINKED ISSUE #249: password encryption for budget json
## Context

Part of replacing Firestore with local JSON for budget private data (#215). The ETL (#216, #217) writes and reads plaintext JSON containing financial transactions, budgets, and rules. The frontend (#218, #219) will upload and export these files via the browser. Plaintext JSON on disk is an unnecessary exposure of private financial data — password-based encryption at rest closes this gap.

Both the Go ETL and the browser must produce and consume the same encrypted format. AES-256-GCM with PBKDF2 key derivation is supported natively by Go's stdlib (`crypto/aes`, `crypto/cipher`, `crypto/pbkdf2`) and the Web Crypto API (`SubtleCrypto.deriveKey`, `SubtleCrypto.encrypt`) — zero external dependencies on either side.

## Summary

Add password-based encryption to the budget JSON file format. The ETL CLI retrieves the encryption password from macOS Keychain via `--keychain <name>`; when set, `--output` writes an encrypted file and `--input` decrypts before parsing. The frontend (#218, #219) prompts for a password on upload (decrypt) and export (encrypt). A single password is used for both directions.

## Encrypted file format

All encrypted files use a binary envelope:

```
[magic 4B][salt 16B][IV 12B][ciphertext ...][GCM auth tag 16B]
```

- **Magic bytes**: `0x42 0x45 0x4E 0x43` ("BENC") — identifies the file as encrypted. The ETL and frontend check for this prefix to distinguish encrypted from plaintext JSON and produce a clear error when the wrong mode is used.
- **Salt**: 16 random bytes, generated fresh on each write. Used with PBKDF2 to derive the encryption key.
- **IV**: 12 random bytes (standard AES-GCM nonce size), generated fresh on each write.
- **Ciphertext + GCM tag**: The JSON content (UTF-8 bytes) encrypted with AES-256-GCM. The 16-byte authentication tag is appended by GCM.

**Key derivation**: PBKDF2-HMAC-SHA256, 600,000 iterations, producing a 256-bit key.

## ETL CLI changes

- `--keychain <name>` retrieves the encryption password from macOS Keychain by account name. Keychain entries are managed manually outside the CLI (via `security add-generic-password`).
- `export.WriteFile`: when password is provided, derive key from password + random salt via PBKDF2, encrypt the marshaled JSON with AES-256-GCM, prepend the magic/salt/IV header, and write the binary envelope atomically (same temp-file-and-rename pattern).
- `export.ReadFile`: check the first 4 bytes for the magic prefix. If present and no password was provided, exit with error: `"file is encrypted but no password was provided"`. If magic is present, extract salt and IV from the header, derive key via PBKDF2, decrypt, then unmarshal JSON as before. If magic is absent, treat as plaintext JSON (current behavior).
- Wrong password produces a GCM authentication error — surface this as: `"decryption failed (wrong password or corrupted file)"`.

### Files
- `budget-etl/main.go` (`--keychain` flag, `fileOpts` struct, pass to ReadFile/WriteFile)
- `budget-etl/internal/export/export.go` (encrypt/decrypt logic in ReadFile/WriteFile)
- `budget-etl/internal/keychain/keychain.go` (Get wrapper around `security find-generic-password`)

## Frontend changes

These changes apply to the upload flow (#218) and export flow (#219):

- **Upload (#218)**: after the user selects a file, check for the magic prefix. If present, show a password input before parsing. Decrypt using Web Crypto API (`crypto.subtle.deriveKey` with PBKDF2, `crypto.subtle.decrypt` with AES-GCM), then proceed with the existing JSON parse and IndexedDB storage. If decryption fails, show: "Wrong password or corrupted file."
- **Export (#219)**: before downloading, show a password input. Encrypt the serialized JSON using Web Crypto API with a fresh random salt and IV, prepend the binary header, and trigger the download as a `.json` file. If the password field is left empty, export plaintext (current behavior).

### Files
- `budget/src/main.ts` (password prompt UI for upload and export)
- `budget/src/crypto.ts` (shared encrypt/decrypt functions using Web Crypto API)

## Acceptance criteria

- [ ] `budget-etl --output budget.json --keychain mykey` produces an encrypted file starting with magic bytes `0x42454E43`
- [ ] `budget-etl --input budget.json --output merged.json --keychain mykey` decrypts, merges, and re-encrypts
- [ ] `budget-etl --input budget.json` (no keychain, encrypted file) exits with a clear error message
- [ ] `budget-etl --input budget.json --keychain wrongkey` (wrong password in keychain) exits with "wrong password or corrupted file"
- [ ] `budget-etl --output budget.json` (no keychain) writes plaintext JSON (backward compatible)
- [ ] Uploading an encrypted file in the browser prompts for a password and decrypts successfully
- [ ] Uploading a plaintext JSON file in the browser works without a password prompt
- [ ] Exporting with a password produces an encrypted file that the ETL can decrypt with the same password
- [ ] Exporting without a password produces plaintext JSON (backward compatible)
- [ ] A file encrypted by the ETL can be decrypted in the browser and vice versa (cross-runtime interop)
- [ ] PBKDF2 uses 600,000 iterations with HMAC-SHA256 on both ETL and frontend
- [ ] `budget-etl --keychain mykey` retrieves password from Keychain by account name
- [ ] `budget-etl --keychain nonexistent` without a stored password exits with a clear error

```json
{
  "version": 1,
  "step": 11,
  "step_label": "Completion",
  "phase": "core",
  "active_skills": [
    "ref-memory-management",
    "ref-pr-workflow"
  ]
}
```

natb1:
## Summary

- Add AES-256-GCM encryption with PBKDF2 key derivation (600k iterations, SHA-256) to the budget JSON file format, using a binary envelope (BENC magic + salt + IV + ciphertext)
- ETL CLI gains `--password` flag; `ReadFile`/`WriteFile` handle encrypt/decrypt transparently, with clear errors for mode mismatches and wrong passwords
- Frontend crypto module (`crypto.ts`) uses Web Crypto API for zero-dependency encrypt/decrypt matching the Go envelope format
- Upload flow detects encrypted files via BENC magic and prompts for password; export flow offers optional password encryption via native `<dialog>`
- Cross-runtime interop validated: Go ETL and browser produce/consume the same encrypted format

## Test plan

- [x] 6 new Go unit tests: round-trip, wrong password, backward compat, magic detection, encrypted-no-password, plaintext-with-password (all pass)
- [x] 7 new frontend crypto unit tests: round-trip, wrong password, isEncrypted variants, decrypt non-encrypted (all pass, 402 total)
- [ ] E2e: encrypted upload with correct/wrong password, encrypted/plaintext export
- [ ] E2e: existing upload and export tests updated for password dialog
- [ ] Smoke: existing plaintext tests pass unchanged
- [ ] Cross-runtime: Node.js crypto helper in e2e encrypts fixtures that browser decrypts

Closes #249

### DIFF
--- a/budget/src/crypto-core.ts
+++ b/budget/src/crypto-core.ts
@@ -15,7 +15,7 @@ export const KEY_LEN = 32;
 export async function deriveKey(
   subtleCrypto: SubtleCrypto,
   password: string,
-  salt: Uint8Array<ArrayBuffer>,
+  salt: Uint8Array,
 ): Promise<CryptoKey> {
   const keyMaterial = await subtleCrypto.importKey(
     "raw",
@@ -25,7 +25,7 @@ export async function deriveKey(
     ["deriveKey"],
   );
   return subtleCrypto.deriveKey(
-    { name: "PBKDF2", salt, iterations: PBKDF2_ITERATIONS, hash: "SHA-256" },
+    { name: "PBKDF2", salt: salt as Uint8Array<ArrayBuffer>, iterations: PBKDF2_ITERATIONS, hash: "SHA-256" },
     keyMaterial,
     { name: "AES-GCM", length: 256 },
     false,
@@ -43,7 +43,7 @@ export async function encryptData(
   const iv = getRandomValues(new Uint8Array(IV_LEN));
   const key = await deriveKey(subtleCrypto, password, salt);
   const ciphertext = await subtleCrypto.encrypt(
-    { name: "AES-GCM", iv },
+    { name: "AES-GCM", iv: iv as Uint8Array<ArrayBuffer> },
     key,
     new TextEncoder().encode(plaintext),
   );


natb1:
## /review Output



I'll start by reading the PR diff file, then invoke the review skill.

Now let me read some of the key source files to understand the full context better.

Now let me invoke the review skill.



I already have the full PR diff loaded, so let me get the PR details and check a few more files for full context.

Let me check a couple more things for completeness.

Now let me check the `--password` being visible in `ps`:

I have all the information I need. Here is my review:

---

## Code Review: PR #284 -- Add password-based encryption for budget JSON files

### Overview

This PR adds AES-256-GCM encryption with PBKDF2 key derivation to budget JSON files, spanning three layers:
1. **Go ETL** (`budget-etl`): `--password` and `--keychain` flags; encrypt/decrypt in `ReadFile`/`WriteFile`
2. **Browser frontend** (`budget/src/crypto.ts`): Web Crypto API implementation of the same envelope format
3. **E2e tests**: Node.js `crypto` helper for cross-runtime interop validation

The binary envelope format is: `BENC` (4 bytes) + salt (16) + IV (12) + AES-256-GCM ciphertext (includes 16-byte auth tag).

### Correctness

**Cryptography is sound.** PBKDF2 at 600k iterations with SHA-256, random 16-byte salt, random 12-byte IV/nonce, AES-256-GCM -- all standard choices. The three implementations (Go, browser Web Crypto, Node.js e2e helper) use matching parameters and should produce interoperable output.

**One issue: `decryptJSON` does not verify the magic bytes before parsing.** The `ReadFile` function checks `IsEncrypted` and gates on it, so `decryptJSON` is only called on data that starts with BENC. However, `decryptJSON` itself starts slicing from offset 4 without verifying the magic. This is fine as-is because the only caller pre-checks, but if `decryptJSON` were ever called directly, it would silently treat the first 4 bytes of any data as part of the header. Consider either:
- Adding a magic-byte check inside `decryptJSON` for defense-in-depth, or
- Documenting that callers must pre-validate with `IsEncrypted`.

This is minor given current usage.

### Security considerations

1. **Password visible in process listing.** The `--password` flag passes the password on the command line, which is visible in `ps aux` output. The `--keychain` flag mitigates this for the retrieve path (password never appears on CLI), but the initial store path (`--password secret --keychain`) still exposes it. This is acceptable for a local-only CLI tool -- worth noting in usage docs but not a blocker.

2. **No password confirmation on export (frontend).** The encrypt-on-export flow prompts once. A typo means the user encrypts with an unintended password. Consider prompting twice (confirm password) for encryption. This is a UX enhancement, not a blocking concern.

3. **Keychain account is just the base filename.** `filepath.Base(*outputPath)` means different directories with the same filename share a keychain entry. This is probably intentional for simplicity but worth being aware of.

### Code quality

**Positive observations:**
- Clean separation: crypto logic in `export` package, keychain in its own package, browser crypto in its own module
- Error messages are specific and actionable ("file is encrypted; use --password", "wrong password or corrupted file")
- The error-on-mismatch pattern (encrypted file + no password, plaintext file + password provided) follows the project's "clear errors over defensive fallbacks" rule
- Constants are shared correctly across Go/TS (salt length, IV length, iteration count, magic bytes)
- Tests cover the important cases: round-trip, wrong password, mode mismatches, backward compatibility, magic detection

**Suggestions:**

1. **`promptPassword` builds HTML via `innerHTML` with a template literal containing the `message` parameter** (`/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/main.ts` line 178). The `message` argument is currently always a string literal, but this is an XSS vector if the function is ever called with user-controlled input. Use `textContent` on the `<p>` element instead:

```typescript
const p = dialog.querySelector("p")!;
p.textContent = message;
```

2. **No CSS for the password dialog.** The `<dialog>` element gets `.password-dialog` class but there are no styles in `theme.css`. Native `<dialog>` with `showModal()` works without styles (centered with backdrop), but the input and buttons will inherit the page's dark theme inconsistently. Consider adding minimal styles, or confirm this looks acceptable as-is.

3. **No unit tests for the keychain package** (`/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/keychain/keychain.go`). The package shells out to `security` CLI, making it hard to unit test without mocking, but even a build-tag-gated integration test would add confidence. Not blocking.

4. **`writeOutputAndLog` signature changed but is it called?** The diff shows `writeOutputAndLog` gaining a `password` parameter, but I only see it defined -- let me verify.

Good, both call sites pass `password` through correctly.

### Test coverage

- **Go**: 6 new tests covering round-trip, wrong password, backward compat, `IsEncrypted` variants, encrypted-without-password, plaintext-with-password. Thorough.
- **Frontend unit tests** (`crypto.test.ts`): 7 tests covering round-trip, wrong password, `isEncrypted`, and decrypt-of-non-encrypted data.
- **E2e tests**: Updated existing export tests for the new password dialog flow. New tests for encrypted upload (correct/wrong password) and encrypted/plaintext export.
- **Missing**: No cross-runtime interop test that encrypts in Go and decrypts in browser (or vice versa). The e2e helper `encryptBuffer` uses Node.js `crypto` which should match, but a test that reads a Go-encrypted fixture in the browser would be the strongest proof of interop. The PR description mentions this is validated -- the Node.js helper serves as the bridge.

### Potential issues

1. **`innerHTML` with interpolated string** (mentioned above) -- the XSS risk in `promptPassword`. This is the most actionable item.

2. **`--keychain` without `--password` and without `--input` but with `--output` and `--dir`**: The validation requires `--input` to retrieve from keychain, but the `run()` path (parse from dir, write to output) also supports `--password`. A user doing `budget-etl --dir ./data --group household --output budget.json --keychain` would get "Error: --keychain without --password requires --input" even though they might want to retrieve the password for encrypting a fresh export. This is an edge case -- the keychain retrieve path assumes re-processing an existing encrypted file -- but worth documenting.

3. **GCM auth tag handling in the e2e helper** (`budget/e2e/helpers.ts` line 21): The Node.js helper correctly concatenates `cipher.getAuthTag()` after the ciphertext, matching Go's `gcm.Seal` (which appends the tag) and Web Crypto's `encrypt` (which also appends it). This is correct.

### Summary

This is a well-structured PR. The cryptographic implementation is correct and consistent across three runtimes. The main actionable item is the `innerHTML` usage in `promptPassword` -- switch to DOM APIs to avoid the XSS vector. The rest are minor observations and potential UX enhancements (password confirmation on export, dialog styling). The test coverage is solid.


## pr-review-toolkit: code-reviewer



I'll start by reading the CLAUDE.md and getting the full diff for this PR.

Now let me check a few specific files for additional context.

Now let me check the `decryptJSON` function more closely and the `ReadFile` error-handling behavior to understand a potential issue.

Now I have a thorough understanding of the changes. Let me check one more thing -- the password being passed on the command line, which is visible via `ps`.

Here is my review of PR #284 (branch `249-budget-json-encryption`).

---

## Files Reviewed

- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go` -- AES-256-GCM encryption/decryption for Go ETL
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export_test.go` -- Go encryption tests
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/keychain/keychain.go` -- macOS Keychain integration
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go` -- CLI flag wiring
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/crypto.ts` -- Web Crypto browser module
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/main.ts` -- UI integration (upload + export)
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/test/crypto.test.ts` -- Browser crypto tests
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/e2e/export.spec.ts` -- E2E export tests
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/e2e/upload.spec.ts` -- E2E upload tests
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/e2e/helpers.ts` -- E2E helper with Node crypto encryption

---

## Critical Issues (90-100)

### 1. Password visible in process list via `--password` CLI flag (Confidence: 92)

**File:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go`, line 34

The `--password` flag passes the password as a command-line argument, which is visible to any user on the system via `ps aux`. This is a security concern for a feature whose entire purpose is protecting sensitive financial data. The `--keychain` flag mitigates this for the retrieval path, but the initial storage path (`--password secret --keychain`) still leaks the password to the process table.

**Suggestion:** Consider reading the password from an environment variable (e.g., `BUDGET_ETL_PASSWORD`) or from stdin when `--password` is empty but encryption is needed. At minimum, document the `ps` visibility risk and recommend `--keychain` as the primary interface.

---

### 2. XSS via unsanitized `message` parameter in `promptPassword` innerHTML (Confidence: 95)

**File:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/main.ts`, line 178

```typescript
dialog.innerHTML = `<form method="dialog"><p>${message}</p>...`;
```

The `message` parameter is interpolated directly into `innerHTML`. While the current callers pass string literals, this is a DOM XSS sink -- any future caller passing user-controlled data (e.g., a filename) would create an injection vulnerability. Using `textContent` for the message paragraph eliminates the risk.

**Fix:** Build the dialog with DOM APIs or at least set the `<p>` element's `textContent` separately:

```typescript
const p = dialog.querySelector("p")!;
p.textContent = message;
```

---

## Important Issues (80-89)

### 3. `decryptJSON` drops the underlying error from `gcm.Open` (Confidence: 85)

**File:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go`, line 86

```go
return nil, fmt.Errorf("decryption failed: wrong password or corrupted file")
```

Per the project guideline in `code-style.md` ("Prefer clear errors over defensive fallbacks"), the original `err` from `gcm.Open` should be wrapped so that non-password-related failures (e.g., truncated ciphertext) produce a diagnostic message. Replacing the format string with `fmt.Errorf("decryption failed (wrong password or corrupted file): %w", err)` preserves the original error for debugging while keeping the user-facing context.

### 4. Missing `--keychain` and `--firestore` mutual exclusivity check (Confidence: 82)

**File:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go`, lines 44-49

The code validates `--password` vs `--firestore` mutual exclusivity but does not validate `--keychain` vs `--firestore`. If a user passes `--keychain --firestore`, the keychain retrieval path would succeed but the resolved password would be silently unused (since Firestore mode ignores it). This violates the "clear errors over defensive fallbacks" guideline -- the user thinks encryption is happening but it is not.

**Fix:** Add a check:
```go
if *keychainFlag && *firestoreFlag {
    fmt.Fprintln(os.Stderr, "Error: --keychain and --firestore are mutually exclusive")
    os.Exit(1)
}
```

### 5. `keychain.Set` passes password as command-line argument to `security` (Confidence: 80)

**File:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/keychain/keychain.go`, line 31

```go
cmd := exec.Command("security", "add-generic-password",
    "-s", serviceName, "-a", account, "-w", password)
```

The password is passed as a CLI argument to `security add-generic-password`, making it visible in the process table. The `security` command supports reading the password from stdin when `-w` is passed without a value argument. Using `cmd.Stdin` with a pipe would avoid this exposure.

---

## Summary

The cryptographic implementation is solid -- PBKDF2 with 600,000 iterations, AES-256-GCM, random salt and IV, correct header format matching between Go and browser. The test coverage is thorough with round-trip, wrong-password, backward-compatibility, and `IsEncrypted` edge-case tests on both sides.

The main concerns are: (1) the innerHTML XSS sink in the password dialog, (2) password exposure in the process table through both `--password` and `keychain.Set`, (3) a missing mutual-exclusivity guard for `--keychain` with `--firestore`, and (4) the dropped error in `decryptJSON`.


## /simplify Output



I'll start by reading the PR diff to understand the changes before invoking the simplify skill.

Now let me invoke the simplify skill.



I'll start by getting the full diff, then launch three review agents in parallel.

Now let me read the full source files that were changed to have complete context for the review.

Now I have full context. Let me also look at the e2e helpers to check the duplication between the e2e `encryptBuffer` and `budget/src/crypto.ts`.

Now let me check for test helper patterns and also look at the Go encryptJSON/decryptJSON for the duplicate AES/GCM setup code.

Now I have a thorough understanding of all the changes. Let me perform the analysis and make fixes directly.

Here are my findings across all three review dimensions:

**Code Reuse Issues:**
1. **Go: Duplicate AES-GCM setup in encryptJSON/decryptJSON** -- The `aes.NewCipher` + `cipher.NewGCM` block is repeated. Extract to a helper.
2. **Go: `IsEncrypted` compares bytes individually instead of using `magicBytes`** -- The `magicBytes` var already exists but isn't used in `IsEncrypted`.
3. **Go tests: Repeated minimal Output construction** -- 5 test functions repeat the same minimal `Output{}` boilerplate. Extract a test helper.
4. **e2e helpers: `encryptBuffer` is a Node.js re-implementation of the same crypto** -- This is necessary since e2e runs in Node and `budget/src/crypto.ts` uses Web Crypto API. Not a real issue, just the cost of different runtimes. Skip.

**Code Quality Issues:**
1. **Parameter sprawl in Go**: `password string` was added as a new trailing parameter to `run`, `runOutputJSON`, `runInputJSON`, `runMerge`, `writeOutputAndLog`, `ReadFile`, and `WriteFile`. This is 7 functions with an extra parameter threaded through. However, since this is a CLI tool (not a library with many callers), and the password genuinely needs to flow to the I/O functions, this is acceptable. Skip.
2. **TS `promptPassword` builds HTML via innerHTML string** -- Minor, but functional and scoped to a dialog. Skip.

**Efficiency Issues:**
1. No meaningful efficiency concerns. The PBKDF2 with 600k iterations is intentionally slow for security. File reads are necessary. No N+1 patterns or hot-path bloat.

Let me fix the actionable items.

**Fix 1: Extract `newGCM` helper in Go to eliminate duplicate AES-GCM setup.**

**Fix 2: Use `magicBytes` in `IsEncrypted` instead of individual byte comparisons.**

**Fix 3: Extract test helper for minimal Output in Go tests.**

Now replace the 5 duplicated minimal Output constructions with calls to `minimalOutput`.

Now let me verify the Go code compiles and tests pass.

All tests pass. Let me also run the full export test suite to make sure I didn't break anything.

All tests pass. Let me also verify the TypeScript crypto tests still pass.

All tests pass. Here is a summary of what was fixed:

**Changes made:**

1. **Extracted `newGCM` helper** in `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go` -- The `aes.NewCipher` + `cipher.NewGCM` block was duplicated in both `encryptJSON` and `decryptJSON`. Extracted to a single `newGCM(password, salt)` function that both callers now use.

2. **`IsEncrypted` now uses `magicBytes`** in the same file -- The function was comparing bytes individually (`data[0] == 'B' && data[1] == 'E' ...`) despite `magicBytes` already being declared as a package-level variable. Replaced with a single array comparison `[4]byte(data[:4]) == magicBytes`.

3. **Extracted `minimalOutput` test helper** in `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export_test.go` -- Five encryption tests (`TestEncryptedWrongPassword`, `TestPlaintextBackwardCompat`, `TestEncryptedFileNoPassword`, `TestPlaintextFileWithPassword`, and implicitly available for future tests) all constructed identical minimal `Output` structs with boilerplate empty slices. Extracted to `minimalOutput(groupName)` and replaced the 4 duplicated constructions.

**Findings skipped (not worth addressing):**

- **Parameter sprawl** (`password string` threaded through 7 Go functions): Acceptable for a CLI tool where the password genuinely needs to reach the I/O layer.
- **e2e `encryptBuffer` duplicating crypto logic**: Necessary because e2e tests run in Node.js (using `node:crypto`) while the app uses the Web Crypto API. Different runtimes require separate implementations.
- **`promptPassword` building HTML via `innerHTML`**: Scoped to a short-lived dialog, no user-supplied data in the template string. Acceptable.


## pr-review-toolkit: comment-analyzer



I'll start by getting the PR diff and understanding the scope of changes.

Let me read the full diff and the complete source files.

Now let me read the full source files to see comments in context.

Now let me check the old ReadFile comment to see what was removed.

I now have a thorough picture of all the changes. Here is my analysis.

---

**Summary**

This PR adds password-based AES-256-GCM encryption across two subsystems: a Go ETL CLI (`budget-etl`) and a browser-side TypeScript module (`budget/src/crypto.ts`). It introduces macOS Keychain integration for password storage. The PR modifies 11 files. Comment coverage is light overall -- the new crypto functions in both Go and TypeScript are largely uncommented, while the existing comments that were updated to reflect encryption are generally accurate. I found one factually incorrect comment, several places where comments were not updated to reflect new behavior, and a few opportunities to add high-value documentation that future maintainers will need.

---

**Critical Issues**

1. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go:181-183`
   - Issue: The `ReadFile` doc comment was updated to describe password/decryption behavior, but the old promise about validation ("Returns an error if the file is missing, contains invalid JSON, or is missing required fields") was silently removed. The function still validates those fields (lines 205-216). While removing redundant detail from a doc comment is not always wrong, the replacement comment now describes only the encryption behavior and omits mention of validation. More importantly, it says "If password is non-empty, the file is decrypted first" -- but the actual behavior is stricter: if password is non-empty AND the file is NOT encrypted, it returns an error (line 194-196). The comment omits this symmetric error case.
   - Suggestion: Update the `ReadFile` comment to mention the strict matching behavior:
     ```
     // ReadFile reads and unmarshals a JSON file into an Output struct.
     // If password is non-empty, the file must be encrypted (BENC header) and is
     // decrypted before parsing. If the file is encrypted but no password is
     // provided, or if password is provided but the file is not encrypted, an
     // error is returned.
     ```

---

**Improvement Opportunities**

1. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go:17-23`
   - Current state: The constants block has only a trailing arithmetic comment (`// magic + salt + IV = 32`) on `headerLen`. There is no doc comment explaining the binary format these constants define.
   - Suggestion: Add a block comment above the constants describing the wire format. This is the kind of "why" comment that prevents future maintainers from having to reverse-engineer the binary layout from two separate codebases (Go and TypeScript):
     ```
     // Encrypted file format: [4-byte magic "BENC"][16-byte salt][12-byte IV][AES-256-GCM ciphertext+tag]
     // The plaintext is the indented JSON with a trailing newline.
     // PBKDF2-SHA256 derives the 256-bit key from the password and salt.
     ```

2. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/crypto.ts` (entire file)
   - Current state: No JSDoc or module-level comments. The file has inline constant comments (`// "BENC"`, `// 32`) but no explanation of what this module does, how it relates to the Go side, or what format it produces.
   - Suggestion: Add a module-level comment explaining the cross-system interop contract:
     ```
     // Browser-side encryption/decryption using Web Crypto API.
     // Produces files compatible with budget-etl's Go encryption:
     // [4-byte "BENC" magic][16-byte PBKDF2 salt][12-byte GCM IV][ciphertext+tag]
     // PBKDF2-SHA256 with 600,000 iterations derives the AES-256 key.
     ```
     This is the most important missing comment in the PR. Two independent implementations of the same binary format must stay synchronized, and a comment noting this cross-system contract is essential for future maintainers.

3. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/keychain/keychain.go:22-23`
   - Current state: The `Set` comment says "Overwrites any existing entry for the same service/account pair." This is accurate in terms of net effect but slightly misleading about mechanism -- the code actually deletes then adds (lines 25-28), rather than using an update/overwrite API.
   - Suggestion: This is minor and the current comment correctly describes the observable behavior. No change strictly needed, but if you prefer precision: "Replaces any existing entry by deleting it first, then adding the new one."

4. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go:56`
   - Current state: `// Resolve password from keychain if --keychain is set` -- this comment describes only the retrieval path. The block beneath it handles both retrieval (no `--password`) and storage (`--password` provided). The inline comments at lines 59 and 73 (`// Retrieve: use input filename as account` / `// Store: use output filename as account`) partially address this, but the top-level comment is incomplete.
   - Suggestion: Change the leading comment to: `// Resolve or store password via keychain if --keychain is set`

5. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go:192`
   - Current state: The error message `"file is encrypted; use --password"` references a CLI flag (`--password`) inside a library function that has no knowledge of the CLI. This couples the library layer to the CLI presentation.
   - Suggestion: This is a code issue rather than a comment issue, but it bears noting because it could mislead a future maintainer who calls `ReadFile` from a different context (e.g., tests, a web handler). Consider changing to `"file is encrypted but no password was provided"`.

6. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/e2e/helpers.ts:15`
   - Current state: `encryptBuffer` has no doc comment. It reimplements the same binary format as Go and TypeScript crypto modules, making it a third implementation of the same protocol. There is no comment noting this.
   - Suggestion: Add a brief comment: `// Encrypts plaintext using the same BENC format as budget-etl and src/crypto.ts. // Uses Node.js crypto (not Web Crypto) for Playwright e2e helpers.`

7. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go:25`
   - Current state: `magicBytes` is declared as a package-level `var` but is never mutated. No comment explains what "BENC" stands for or why this magic value was chosen.
   - Suggestion: A one-line comment would help: `// magicBytes identifies encrypted budget files ("Budget ENCrypted" or similar mnemonic).` If "BENC" has no specific expansion, say so -- otherwise future maintainers will wonder.

---

**Recommended Removals**

1. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go:22`
   - Comment: `// magic + salt + IV = 32`
   - Rationale: This is partially redundant with the arithmetic expression `4 + saltLen + ivLen` on the same line -- anyone reading the constants can compute 4 + 16 + 12 = 32. However, the comment does serve a verification role (confirming the total). I would keep it only if paired with a format-level comment above the const block (see Improvement #1). On its own, the `= 32` part adds modest value.

2. Location: `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/crypto.ts:3`
   - Comment: `// "BENC"`
   - Rationale: The hex bytes `[0x42, 0x45, 0x4e, 0x43]` are immediately recognizable as ASCII to anyone working on a crypto module, and the same comment on line 6 (`// 32`) restates an obvious sum. These are low-harm but also low-value. If a format-level module comment is added per Improvement #2, these inline comments become redundant.

---

**Positive Findings**

1. `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go:255-257` -- The `runOutputJSON` comment clearly explains that rules are NOT applied at this stage and directs the reader to `--input` for the subsequent pass. This is a good "why not" comment that prevents confusion.

2. `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go:296-301` -- The `runInputJSON` comment explains that category/budget from the input are NOT carried forward and why (rule changes take effect on every run). This captures important business logic rationale.

3. `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go:596-601` -- The `runMerge` comment is thorough and accurately describes the merge semantics including what is preserved and what is recomputed.

4. `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/keychain/keychain.go:24-25` -- The inline comment `// Delete existing entry (ignore error if not found)` on line 25 correctly explains the delete-before-add pattern and why the error is intentionally ignored. This prevents future maintainers from "fixing" the ignored error.

5. `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go:220-221` -- The `WriteFile` comment was updated concisely and accurately to note encryption behavior.

---

**Key Takeaway**

The most significant gap in this PR is the absence of a cross-system format contract comment. Three independent implementations (Go `export.go`, TypeScript `crypto.ts`, Node.js `helpers.ts`) produce and consume the same binary format, but none of them reference each other or document the shared specification. If any implementation drifts -- say, changing the iteration count or salt length -- the others will silently produce incompatible output. A format comment in at least one canonical location (ideally `export.go` as the source of truth) with cross-references from the other two would be the single highest-value documentation addition for this PR.


## pr-review-toolkit: pr-test-analyzer



I'll start by examining the PR changes and understanding the scope of the work.

Now let me read all the changed files in parallel.

Now let me read the full source files that matter for understanding testing gaps.

Now let me check if there are any existing tests for keychain or cross-compatibility tests.

Let me check the full test file to see the `strings` import and verify the test structure.

Now let me verify the tests actually pass.

All tests pass. Now I have a thorough understanding of the PR. Here is my analysis.

---

## Test Coverage Analysis: PR #284 -- AES-256-GCM Encryption for Budget JSON

### 1. Summary

The test coverage for the core encryption/decryption logic is solid on both the Go and TypeScript sides. The Go `export` package has 7 new tests covering the encrypt/decrypt round-trip, wrong password, backward compatibility with plaintext, the `IsEncrypted` predicate, and the mismatch error conditions. The TypeScript `crypto.test.ts` has 7 unit tests covering the same core scenarios. E2e tests cover upload of encrypted files (correct and wrong password) and export with/without password. The main gaps are around cross-implementation interoperability and the keychain module.

### 2. Critical Gaps

**Cross-implementation interoperability test -- Criticality: 9/10**

The Go ETL encrypts files that the browser decrypts, and vice versa. Both implementations independently derive keys via PBKDF2-SHA256 with 600,000 iterations and use AES-256-GCM with the same `BENC | salt(16) | iv(12) | ciphertext+tag` wire format. However, there is no test that verifies a file encrypted by one implementation can be decrypted by the other. If either side silently changes iteration count, hash function, byte ordering, or GCM tag placement, the round-trip tests within each implementation will still pass while cross-platform decryption silently breaks in production.

The e2e helper `encryptBuffer` in `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/e2e/helpers.ts` already contains a third independent Node.js `crypto` implementation of the same format. This is used to generate encrypted fixtures for Playwright tests, but it is itself untested against either the Go or Web Crypto implementations. There are now three implementations of the same wire format with zero cross-validation.

A practical approach: generate a known-password encrypted fixture with the Go ETL (or commit a golden binary file), then write a test in the TypeScript suite that decrypts it and verifies the plaintext. This catches any drift in constants, byte layout, or GCM tag handling between the two primary implementations.

**Truncated/corrupted ciphertext handling in Go `decryptJSON` -- Criticality: 8/10**

The Go `decryptJSON` function at `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go` line 66 only checks `len(data) < headerLen` but does not handle the case where the header is valid but the ciphertext portion is truncated or zero-length. GCM requires at minimum the 16-byte authentication tag after the header. While `gcm.Open` will return an error for a truncated tag, there is no test confirming this path produces a clear error rather than a panic or confusing message. A test that writes `BENC` + 16 bytes salt + 12 bytes IV + 0 bytes ciphertext (total 32 bytes, passes the length check) would verify this.

The TypeScript side has a partial test for "too short" data but only tests buffers shorter than `HEADER_LEN`.

### 3. Important Improvements

**Empty password string edge case -- Criticality: 6/10**

The Go `ReadFile` and `WriteFile` use `password == ""` as the signal for plaintext mode. The TypeScript `encrypt`/`decrypt` accept any string. Neither side tests behavior with an empty string password explicitly passed to the encrypt path. While the Go code guards against this at the `ReadFile`/`WriteFile` level, the underlying `encryptJSON` and `decryptJSON` functions would happily encrypt with an empty-string password, producing a file that looks encrypted (has BENC header) but provides no security. A unit test confirming that `encryptJSON`/`decryptJSON` round-trip correctly with an empty password (or that the public API rejects it) would document the intended contract.

**Encrypted file byte-level format verification -- Criticality: 6/10**

The Go `TestEncryptDecryptRoundTrip` verifies the logical round-trip but does not assert the wire format (magic bytes present, output length is `headerLen + len(plaintext) + gcm.Overhead()`). A test that checks `IsEncrypted(ciphertext) == true` and `len(ciphertext) >= headerLen` after `WriteFile` with a password would catch format regressions without being brittle.

**Keychain module has zero tests -- Criticality: 5/10**

`/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/keychain/keychain.go` has no test file. This is understandable since it shells out to macOS `security` and is hard to unit test without mocking `exec.Command`. The `main.go` keychain integration (flag validation, account-name derivation from filename) is also untested. The flag validation logic at lines 47-86 of `main.go` has several branches (`--keychain` without `--password` requires `--input`, `--keychain` with `--password` requires `--output`, `--password` and `--firestore` mutually exclusive) that are only tested implicitly if e2e tests happen to exercise them. At minimum, the mutual-exclusion checks could be extracted into a testable function.

### 4. Test Quality Issues

**E2e `encryptBuffer` helper reimplements crypto without validation**

The `encryptBuffer` function in `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/e2e/helpers.ts` is a third implementation of the encryption format using Node.js `crypto`. It manually concatenates `encrypted + tag` in the GCM output. If the tag placement differs from what Web Crypto or Go produce (GCM tag is appended automatically by `gcm.Seal` in Go and by `crypto.subtle.encrypt` in Web Crypto, but must be manually appended in Node.js `crypto`), e2e tests could pass with a broken format. This helper should ideally be validated against the canonical TypeScript `encrypt` function or replaced by calling it.

**Go tests verify field values but not that the ciphertext differs between encryptions**

Since encryption uses random salt and IV, encrypting the same plaintext twice should produce different ciphertexts. No test verifies this property. While not critical, it would confirm the random number generation path is actually being used (not accidentally zeroed or seeded deterministically).

### 5. Positive Observations

- The Go tests cover all four combinations of encrypted/plaintext file vs. password/no-password provided to `ReadFile`, which is the most important error surface for user-facing behavior.
- The `TestIsEncrypted` table-driven test covers edge cases well (empty, too short, wrong magic, correct magic).
- The TypeScript tests verify that the correct error type (`UploadValidationError`) is thrown, not just any error, which ensures the UI can distinguish decryption failures from other errors.
- E2e tests cover both the happy path (correct password decrypts and shows transactions) and the error path (wrong password shows error message) for uploads.
- E2e export tests verify both encrypted output (checks BENC header) and plaintext output (parses JSON and checks fields).
- The backward compatibility test (`TestPlaintextBackwardCompat`) explicitly verifies that the encryption changes do not break existing plaintext file workflows.


## pr-review-toolkit: silent-failure-hunter



I'll start by examining the full diff and understanding the scope of changes.

Now let me read the full current files for context on the error handling patterns already in place.

Now let me check the main.go for the full keychain interaction flow and look at the upload.ts for UploadValidationError.

I now have a thorough understanding of all the changes. Here is the complete error handling audit.

---

## Error Handling Audit: PR #284 (Budget JSON Encryption)

### Overview

The PR adds AES-256-GCM encryption across two codebases: a Go CLI (`budget-etl`) and a browser TypeScript app (`budget`). The Go side is generally well-structured with proper error propagation. The TypeScript side has one significant issue and a few medium-severity findings. I found one critical issue overall.

---

### ISSUE 1 -- Bare `catch` in `crypto.ts` swallows all errors from `crypto.subtle.decrypt`

**Location:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/crypto.ts`, lines 60-68

**Severity:** CRITICAL

**Issue Description:** The `catch` block on line 67 has no error parameter and catches every exception thrown by `crypto.subtle.decrypt`, then unconditionally rewrites it as "Wrong password or corrupted file." This is the single most important error handling defect in the PR.

**Hidden Errors:** `crypto.subtle.decrypt` can throw for reasons beyond wrong password/corrupted data:
- `InvalidAccessError` -- if the key's `usages` does not include `"decrypt"` (indicates a code bug in `deriveKey`)
- `OperationError` for reasons unrelated to the password (e.g., malformed IV length, browser implementation bugs)
- `TypeError` -- if `ciphertext` or `iv` is not a `BufferSource` (indicates a code bug in the slicing logic)

A `TypeError` thrown inside this `try` block gets caught and rewritten as a user-facing "Wrong password" message instead of propagating as the programming error it is. This directly contradicts the project's pattern in `main.ts` where `TypeError` and `ReferenceError` are explicitly re-thrown to avoid masking bugs.

**User Impact:** If a code bug causes a `TypeError` inside the `try` block, the user sees "Wrong password or corrupted file" -- an entirely misleading message. The developer debugging this will chase password/corruption issues when the real problem is a programming error.

**Recommendation:** Re-throw `TypeError` and `ReferenceError` before converting to `UploadValidationError`. Capture and log the original error for debugging:

```typescript
} catch (err) {
  if (err instanceof TypeError || err instanceof ReferenceError) throw err;
  throw new UploadValidationError("Wrong password or corrupted file.");
}
```

---

### ISSUE 2 -- `keychain.Set` silently discards delete failure details

**Location:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/keychain/keychain.go`, line 28

**Severity:** HIGH

**Issue Description:** The line `_ = del.Run()` discards the error from `security delete-generic-password`. The comment says "ignore error if not found," but this discards *every* error, not just "not found." If the `security` binary is not on `PATH`, if the user's Keychain is locked and requires authorization, or if there is a permissions problem, the delete silently fails and then the subsequent `add-generic-password` will fail with a confusing "duplicate entry" error (since the old entry was not removed).

**Hidden Errors:**
- `exec.ErrNotFound` -- `security` binary missing from `PATH`
- Keychain locked / authorization denied by the user
- Keychain database corruption
- File permission errors

**User Impact:** The user sees a cryptic `add-generic-password` failure message instead of the real cause (e.g., "Keychain is locked"). The delete error holds the actual diagnostic information that gets thrown away.

**Recommendation:** Check the exit code specifically. `security delete-generic-password` exits with code 44 (`errSecItemNotFound`) when no entry exists. Only ignore that specific case:

```go
del := exec.Command("security", "delete-generic-password",
    "-s", serviceName, "-a", account)
if err := del.Run(); err != nil {
    var exitErr *exec.ExitError
    if !errors.As(err, &exitErr) || exitErr.ExitCode() != 44 {
        return fmt.Errorf("removing existing keychain entry for account %q: %w", account, err)
    }
}
```

---

### ISSUE 3 -- `decryptJSON` in Go wraps the original GCM auth error, losing the original

**Location:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go`, lines 84-86

**Severity:** MEDIUM

**Issue Description:** On line 86, the error from `gcm.Open` is replaced with a new error string: `"decryption failed: wrong password or corrupted file"`. The original error is not wrapped with `%w`. While this is intentional for user-facing simplicity, it means that programmatic callers cannot inspect the underlying error, and log output does not include the original GCM error message which could distinguish authentication failure from other GCM errors.

**Recommendation:** Wrap the original error for debugging while keeping the user-facing prefix:

```go
return nil, fmt.Errorf("decryption failed (wrong password or corrupted file): %w", err)
```

---

### ISSUE 4 -- `promptPassword` dialog has no handler for the `cancel` event from Escape key

**Location:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/main.ts`, lines 174-196

**Severity:** MEDIUM

**Issue Description:** The `<dialog>` element supports closing via the Escape key, which fires a `cancel` event and then a `close` event -- but neither is handled. When the user presses Escape, the dialog closes visually but the Promise never resolves. The `handleFileUpload` function and the export handler both `await` this Promise, so they hang forever. The upload input has already been reset to `""` on line 239, so the user cannot even re-trigger the upload.

**User Impact:** If the user presses Escape instead of clicking Cancel, the application appears to freeze. No error is shown, and the upload/export flow is permanently stuck for the lifetime of the page.

**Recommendation:** Add a `cancel` event listener on the dialog:

```typescript
dialog.addEventListener("cancel", () => {
  dialog.remove();
  resolve(null);
});
```

---

### ISSUE 5 -- `encrypt` in `crypto.ts` does not validate empty password

**Location:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/crypto.ts`, line 32

**Severity:** MEDIUM

**Issue Description:** The `encrypt` function accepts any string, including empty string. The caller in `main.ts` (line 256) guards against this with `if (pw !== "")`, but `encrypt` is a public export. If any other caller passes an empty password, the data will be encrypted with a key derived from an empty password -- technically valid cryptographically, but provides a false sense of security. The Go side explicitly rejects this case by checking `password != ""` before calling `encryptJSON`.

**User Impact:** Data encrypted with an empty password is trivially decryptable by anyone who recognizes the BENC format. The user believes their data is protected when it is not.

**Recommendation:** Add an early guard in `encrypt`:

```typescript
export async function encrypt(plaintext: string, password: string): Promise<ArrayBuffer> {
  if (!password) {
    throw new Error("Password must not be empty for encryption.");
  }
  // ...
```

---

### ISSUE 6 -- Missing validation: `--keychain` without `--password` and without `--input` but with `--output` and `--dir`

**Location:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go`, lines 57-85

**Severity:** MEDIUM

**Issue Description:** The `--keychain` flag without `--password` requires `--input` (line 60-62) to determine the keychain account name. However, the `--keychain` flag with `--password` requires `--output` (line 74-76) to determine the account name for storage. There is a missing validation for the case where `--keychain` is used with `--password` and `--output` but without `--input`, and then the user runs a `--dir` flow. The password is stored under the output filename, but when the user later tries to retrieve it via `--keychain` without `--password`, they must use `--input` with the same filename. If the output and input filenames differ across runs, the keychain lookup silently uses a different account name and fails with a misleading "no keychain entry" error.

This is not a code bug per se, but the error message from `keychain.Get` ("no keychain entry for account X") will confuse users who stored the password under a different filename. The account name derivation from `filepath.Base` is implicit and undocumented in the error path.

**Recommendation:** Include the expected account name in the error message and suggest how to check:

```go
return "", fmt.Errorf("no keychain entry for account %q (service %q); passwords are stored by filename — verify the filename matches what was used with --keychain --password: %w", account, serviceName, err)
```

---

### ISSUE 7 -- `password` passed on command line is visible in process listing

**Location:** `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go`, line 34

**Severity:** MEDIUM (security, not error handling -- but relevant to the `--keychain` feature's purpose)

**Issue Description:** The `--password` flag takes the password as a command-line argument, which is visible to any user on the system via `ps aux`. The `--keychain` feature was presumably added to avoid this, but the store path (`--keychain --password <pass>`) still requires passing the password on the command line for the initial store. This is not an error handling issue, but worth noting since the keychain feature exists specifically to avoid password exposure.

**Recommendation:** Consider reading the password from stdin when `--password` is specified without a value, or from an environment variable. This is an enhancement, not a bug fix.

---

### What is done well

- **Go error propagation is solid.** Every error in `export.go`, `main.go`, and `keychain.go` is either returned with context via `fmt.Errorf("context: %w", err)` or causes `os.Exit(1)` with a message. No errors are silently swallowed in the Go encryption/decryption path.

- **The `ReadFile` mismatch detection is excellent.** The encrypted-but-no-password and plaintext-but-password-provided checks (lines 191-196 of `export.go`) catch user configuration errors early with clear, actionable messages.

- **The TypeScript `handleFileUpload` catch block properly distinguishes `UploadValidationError` from programming errors.** The pattern of re-throwing `TypeError`/`ReferenceError` while showing user-friendly messages for known errors is good -- which makes the absence of this same pattern in `crypto.ts` (Issue 1) all the more glaring.

- **Atomic file writes in `WriteFile`.** The temp-file-and-rename pattern with proper cleanup on each error path is correct and prevents partial writes.

---

### Summary of findings

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | CRITICAL | `budget/src/crypto.ts:67` | Bare `catch` swallows `TypeError`/`ReferenceError` from `crypto.subtle.decrypt` |
| 2 | HIGH | `budget-etl/internal/keychain/keychain.go:28` | `_ = del.Run()` discards all errors, not just "not found" |
| 3 | MEDIUM | `budget-etl/internal/export/export.go:86` | Original GCM error not wrapped, lost for debugging |
| 4 | MEDIUM | `budget/src/main.ts:174-196` | Escape key on password dialog hangs the Promise forever |
| 5 | MEDIUM | `budget/src/crypto.ts:32` | `encrypt` accepts empty password without error |
| 6 | MEDIUM | `budget-etl/internal/keychain/keychain.go:17` | Keychain error message does not explain account name derivation |
| 7 | MEDIUM | `budget-etl/main.go:34` | Password visible in process listing (security note) |

Issues 1 and 4 are the most likely to cause real debugging pain. Issue 1 is a direct violation of the project's own error handling convention. Issue 4 will cause a user-visible hang with no error feedback and no recovery path.


## pr-review-toolkit: type-design-analyzer



I'll start by examining the full diff and understanding the codebase structure.

Now let me read the full current files to see types in context.

Let me also check the `main.go` keychain integration logic and the `upload.ts` for the `UploadValidationError` type.

Now I have a thorough understanding of all the new and modified types. Here is my analysis.

---

## Type Analysis: PR #284 -- Budget JSON Encryption

This PR introduces encryption across two codebases (Go ETL and TypeScript browser app) using a shared binary wire format. The changes touch five conceptual type-level areas: the BENC wire format, the `ReadFile`/`WriteFile` API surface, the `keychain` package, the TypeScript `crypto` module, and the CLI flag validation in `main.go`. I will analyze each.

---

## Type: BENC Wire Format (implicit, Go + TypeScript)

The encryption wire format is `[4-byte magic "BENC"][16-byte salt][12-byte IV][ciphertext+GCM tag]`. It is not represented as an explicit type in either language -- it exists only as a set of constants and procedural encode/decode logic duplicated across Go and TypeScript.

### Invariants Identified
- Magic bytes must be exactly `BENC` (0x42, 0x45, 0x4E, 0x43)
- Salt is always 16 bytes
- IV is always 12 bytes
- Header is always 32 bytes (4 + 16 + 12)
- Ciphertext includes a 16-byte GCM authentication tag appended by the cipher
- PBKDF2 iteration count is 600000 with SHA-256
- Key length is 256 bits (32 bytes)

### Ratings
- **Encapsulation**: 5/10
  The format has no structural representation. Both Go and TypeScript duplicate magic-byte checks via raw indexing (`data[0] == 'B' && data[1] == 'E'` etc.) rather than comparing against the `magicBytes` constant. The Go `IsEncrypted` function does not use the declared `magicBytes` variable. The TypeScript `isEncrypted` does not use its `MAGIC` constant for comparison. These are independent drift opportunities.

- **Invariant Expression**: 4/10
  The wire format is entirely implicit -- encoded in constants and procedural slicing. There is no type that represents "a valid BENC envelope" versus "raw bytes." The `headerLen` computation is derived from other constants, which is good, but the format itself is communicated only through code comments and parallel implementations.

- **Invariant Usefulness**: 8/10
  The invariants are well-chosen for the problem. Magic-byte detection cleanly separates encrypted from plaintext files. The fixed header layout makes parsing deterministic. GCM provides authenticated encryption, preventing silent corruption.

- **Invariant Enforcement**: 6/10
  Both `decryptJSON` (Go) and `decrypt` (TypeScript) validate minimum length. GCM authentication rejects wrong passwords and tampered data. However, neither implementation validates the magic bytes before parsing salt/IV -- `decryptJSON` trusts the caller to have called `IsEncrypted` first. If called directly on non-BENC data that happens to be >= 32 bytes, it will attempt decryption on garbage bytes rather than failing fast with a clear "not a BENC file" error.

### Strengths
- Constants are consistent across both implementations (salt=16, IV=12, iterations=600000, SHA-256)
- The magic-byte prefix is a clean, standard approach for format detection
- GCM authentication tag provides integrity verification and clear wrong-password detection

### Concerns
- The wire format is duplicated across Go and TypeScript with no shared schema or test vector. A third implementation in `budget/e2e/helpers.ts` (`encryptBuffer`) adds a third copy. If any parameter (e.g., iteration count) changes, all three must be updated in lockstep.
- `IsEncrypted` in Go does not use the `magicBytes` var -- it hardcodes the four byte literals. Same in TypeScript.
- `decryptJSON` does not verify magic bytes itself, relying on the caller to have pre-checked via `IsEncrypted`. This is an invariant maintained by convention rather than enforcement.

### Recommended Improvements
1. In both Go and TypeScript, have `decryptJSON`/`decrypt` verify the magic bytes internally rather than trusting the caller. This costs nothing and eliminates a class of misuse.
2. In Go, have `IsEncrypted` compare against `magicBytes` rather than hardcoded literals: `data[0] == magicBytes[0] && ...` (or use `bytes.HasPrefix`). This ties the check to the single source of truth.
3. Add a cross-language test vector: a known plaintext + known password producing a known ciphertext (using a fixed salt and IV for the test only). This is the strongest guard against drift between Go, browser TypeScript, and the e2e helper.

---

## Type: `ReadFile` / `WriteFile` API (Go, `/budget-etl/internal/export/export.go`)

### Invariants Identified
- If the file has BENC magic and password is empty, return an error
- If the file lacks BENC magic and password is non-empty, return an error
- Password presence and file encryption status must agree (strict bidirectional check)
- Output struct validation (version > 0, groupName non-empty, transactions non-nil) happens after decryption
- WriteFile performs atomic write via temp-file-then-rename

### Ratings
- **Encapsulation**: 6/10
  The password is threaded as a bare `string` through every function in the call chain (`run`, `runInputJSON`, `runMerge`, `runOutputJSON`, `writeOutputAndLog`, `ReadFile`, `WriteFile`). Seven functions now accept a `password string` parameter. This is a lot of surface area for a credential to travel through.

- **Invariant Expression**: 7/10
  The bidirectional encryption/password check in `ReadFile` is clearly expressed and follows the project's "clear errors over defensive fallbacks" rule. The error messages are specific and actionable ("file is encrypted; use --password").

- **Invariant Usefulness**: 9/10
  The strict check that password presence must match file encryption state prevents two real bugs: (a) silently reading encrypted gibberish as JSON when the user forgets the password flag, and (b) accidentally treating plaintext as encrypted. These are genuinely useful guardrails.

- **Invariant Enforcement**: 8/10
  Enforcement is solid. The check happens immediately after reading raw bytes, before any parsing. The only gap is the one noted above -- `decryptJSON` itself does not verify magic bytes.

### Strengths
- The bidirectional password/encryption check is a clear, defensive design
- Error messages guide the user toward the correct fix
- Atomic write with temp-file-then-rename is preserved for encrypted output

### Concerns
- The `password string` parameter is threaded through seven functions. An empty string means "no encryption," a non-empty string means "encrypt/decrypt." This conflates "no password was provided" with "password is the empty string." While an empty-string password is arguably useless, the type system does not distinguish the two states.
- The `password` argument appears in the process list via `--password <pass>` on the command line. The `--keychain` flag mitigates this, but only on macOS.

### Recommended Improvements
1. Consider using `*string` (pointer-to-string) or a small wrapper type for the password to distinguish "not provided" from "empty string." This would let you reject an empty password explicitly at the CLI boundary rather than silently treating it as "no encryption." This is a minor point -- the current `""` convention works, but it is a latent ambiguity.
2. The seven-function password-threading could be reduced by bundling the output configuration into a struct (path + password + maybe format options). This would also make future additions (e.g., different encryption algorithms) less disruptive. Whether this is worth the churn depends on how stable the CLI interface is.

---

## Type: `keychain` package (Go, `/budget-etl/internal/keychain/keychain.go`)

### Invariants Identified
- Service name is always "budget-etl" (hardcoded constant)
- `Get` returns a trimmed password string or an error
- `Set` deletes any existing entry before adding (upsert semantics)
- Account names are derived from filenames at the call site

### Ratings
- **Encapsulation**: 7/10
  The package properly hides the `security` CLI details behind a clean `Get`/`Set` interface. The service name constant is internal. The API is minimal.

- **Invariant Expression**: 5/10
  There is no validation of inputs. An empty `account` string would be passed to `security` and produce a confusing OS-level error. An empty `password` passed to `Set` would store an empty keychain entry. These are not checked.

- **Invariant Usefulness**: 7/10
  The keychain integration solves a real problem (avoiding `--password` in the process list and shell history). The upsert semantics of `Set` are practical.

- **Invariant Enforcement**: 4/10
  No input validation at all. The package trusts the caller to provide non-empty account and password strings. The `Set` function ignores the error from `delete-generic-password` unconditionally -- this is intentional (entry may not exist), but a delete failure for other reasons (keychain locked, permission denied) would be silently swallowed.

### Strengths
- Clean, minimal API surface
- Correct upsert semantics for the keychain entry

### Concerns
- No validation of empty account or empty password
- The silent `_ = del.Run()` in `Set` swallows all errors, not just "entry not found." If the keychain is locked or permissions are wrong, the delete fails silently and then the add may also fail with a confusing error about a duplicate entry.
- The password is passed as a command-line argument to `security add-generic-password -w <password>`, which briefly exposes it in the process list. The `security` tool supports reading the password from stdin with `-w` (no argument), but this would require piping to stdin.

### Recommended Improvements
1. Validate that `account` is non-empty in both `Get` and `Set`. Validate that `password` is non-empty in `Set`. Return a clear error rather than deferring to the `security` tool's error message.
2. For the `Set` delete step, check whether the error is specifically "entry not found" versus something else. The `security` tool exits with status 44 for "item not found" -- other exit codes indicate real problems that should be surfaced.

---

## Type: TypeScript `crypto` module (`/budget/src/crypto.ts`)

### Invariants Identified
- `isEncrypted` requires at least 4 bytes and matching magic prefix
- `decrypt` requires at least `HEADER_LEN` (32) bytes
- `encrypt` always produces `HEADER_LEN + ciphertext.byteLength` output
- Wrong password / corrupted data throws `UploadValidationError`
- All crypto operations use Web Crypto API (no custom primitives)

### Ratings
- **Encapsulation**: 8/10
  The module exports only three functions (`isEncrypted`, `encrypt`, `decrypt`) with `deriveKey` kept private. Constants are module-scoped. The API is clean and minimal.

- **Invariant Expression**: 7/10
  The function signatures are clear: `encrypt` takes a plaintext string and password, `decrypt` takes an `ArrayBuffer` and password. The asymmetry (string in, ArrayBuffer in) reflects the actual usage pattern (export produces a string, upload produces an ArrayBuffer). The `UploadValidationError` type correctly classifies decryption failures as user-facing validation errors.

- **Invariant Usefulness**: 8/10
  The invariants prevent real bugs. The `HEADER_LEN` check prevents out-of-bounds slicing. The GCM authentication catches wrong passwords. Throwing `UploadValidationError` integrates correctly with the existing error-handling UI.

- **Invariant Enforcement**: 6/10
  `decrypt` validates minimum length but does not check magic bytes -- same gap as the Go side. An `ArrayBuffer` of 32+ bytes of non-BENC data would be sliced and fed to GCM, which would fail with "Wrong password or corrupted file" rather than "this is not an encrypted file." The error message is misleading in that case.
  
  `encrypt` does not validate that the password is non-empty. An empty-string password would produce a technically valid BENC file encrypted with a key derived from `""`.

### Strengths
- Uses Web Crypto API exclusively -- no custom crypto primitives
- Error type integration with `UploadValidationError` is well-considered
- Async API correctly reflects Web Crypto's promise-based design

### Concerns
- Same magic-byte verification gap as the Go side
- The `encrypt` function accepts an empty password without error
- `decrypt` catches all exceptions from `crypto.subtle.decrypt` and maps them to "Wrong password or corrupted file." While GCM failures are the most likely cause, other failures (e.g., invalid key state) would be misreported.

### Recommended Improvements
1. Add magic-byte verification at the start of `decrypt`, before slicing salt/IV. Return a distinct error like "Not an encrypted file" rather than the generic "Wrong password" message.
2. Consider validating that the password is non-empty in `encrypt`. This prevents creating files that are "encrypted" with a trivially derivable key.

---

## Type: CLI Flag Validation (`/budget-etl/main.go`, lines 43-85)

### Invariants Identified
- `--input` and `--firestore` are mutually exclusive
- `--password` and `--firestore` are mutually exclusive
- `--input` requires `--output`
- `--keychain` without `--password` requires `--input` (retrieval mode)
- `--keychain` with `--password` requires `--output` (storage mode)
- `--keychain` determines whether password is stored or retrieved based on whether `--password` is also set

### Ratings
- **Encapsulation**: 3/10
  All validation is procedural, spread across ~40 lines of `main()` with raw `os.Exit(1)` calls. The valid flag combinations are not captured in any type -- they exist only as a sequence of conditional checks. Adding a new flag interaction requires reading all existing checks to avoid conflicts.

- **Invariant Expression**: 4/10
  The valid combinations are not obvious from the code structure. The `--keychain` flag's dual behavior (store vs. retrieve) depends on whether `--password` is also set, which is a surprising implicit state machine. A reader must trace the conditionals carefully to understand the valid input space.

- **Invariant Usefulness**: 8/10
  The checks prevent real misuse. Encrypting to Firestore makes no sense. Using `--keychain` without knowing which file to key on would be broken.

- **Invariant Enforcement**: 6/10
  Most invalid combinations are caught, but there are gaps:
  - `--password` without `--output` and without `--input` is not rejected. Running `budget-etl --dir ./data --group test --password secret --firestore` is caught, but `--password` without any file mode is not explicitly addressed (it would propagate to `run()` which writes to Firestore by default, but the password/firestore check covers this).
  - `--keychain` without `--password` and without `--input` is caught. But `--keychain` without `--password`, with `--input`, and without `--output` falls through to the `--input requires --output` check, which gives a less specific error.

### Strengths
- Error messages are specific and actionable
- The `--password`/`--firestore` mutual exclusion is a good guard

### Concerns
- The `--keychain` flag's dual behavior (store vs. retrieve) based on the presence of another flag is an implicit mode switch that is not self-documenting.
- The validation order matters: some checks depend on earlier checks having already exited. This is fragile when adding new flags.

### Recommended Improvements
1. Consider splitting `--keychain` into `--keychain-get` and `--keychain-set` (or `--keychain-store` / `--keychain-load`). This makes the intent explicit in the flag name rather than deriving it from the combination of other flags. The current behavior where `--keychain` means "store" when `--password` is present and "retrieve" when it is absent is a subtle coupling.
2. Alternatively, if you want to keep the single `--keychain` flag, document the two modes in the `--keychain` flag's help text so `--help` output explains the behavior.

---

## Cross-Cutting Observations

**Wire format duplication is the biggest type-design risk.** The BENC format is implemented three times (Go `export`, TypeScript `crypto`, TypeScript `e2e/helpers.ts`) with no shared definition or cross-language test vector. The constants match today, but there is no mechanism to detect drift. A single wrong constant in any implementation produces files the others cannot read, with the unhelpful error "wrong password."

**The password flows through too many function signatures.** In `main.go`, the `password string` parameter is added to seven functions. This is a code smell suggesting the parameter should be bundled into a configuration struct or handled at a higher level (e.g., the `ReadFile`/`WriteFile` boundary is fine, but `run`, `runInputJSON`, `runMerge`, `runOutputJSON`, and `writeOutputAndLog` are just forwarding it).

**Relevant files:**
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/export/export.go` -- Go encryption/decryption and wire format
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/internal/keychain/keychain.go` -- macOS Keychain integration
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget-etl/main.go` -- CLI flag validation and password threading
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/src/crypto.ts` -- TypeScript browser encryption/decryption
- `/Users/n8/natb1/commons.systems/worktrees/249-budget-json-encryption/budget/e2e/helpers.ts` -- Third copy of encryption logic in e2e tests


---

# Code Quality Evaluation — Iteration 1

## Required — High Priority

1. **XSS via `innerHTML` in `promptPassword`** — `budget/src/main.ts:178`. Security — DOM XSS sink. Raised by: /review, code-reviewer, type-design-analyzer.
2. **Bare `catch` swallows `TypeError`/`ReferenceError`** — `budget/src/crypto.ts:67`. Masks programming errors. Raised by: silent-failure-hunter, type-design-analyzer.
3. **`keychain.Set` silently discards all delete errors** — `budget-etl/internal/keychain/keychain.go:28`. Should only ignore exit code 44. Raised by: silent-failure-hunter, type-design-analyzer.
4. **Escape key on password dialog hangs Promise** — `budget/src/main.ts:174-196`. Permanent hang with no recovery. Raised by: silent-failure-hunter.
5. **`decryptJSON` drops original GCM error** — `budget-etl/internal/export/export.go:86`. Violates code-style.md. Raised by: code-reviewer, silent-failure-hunter.
6. **Remove `--password` option** — `budget-etl/main.go:34`. Password visible in process list. User reclassified from out-of-scope. Remove the flag entirely; password comes from keychain only.
7. **Reduce password parameter threading** — `budget-etl/main.go`. Password threaded through 7 functions. User reclassified from out-of-scope. Bundle into config struct or reduce surface.
8. **`--keychain` takes key name** — `budget-etl/main.go`. Change from boolean flag to string parameter accepting a keychain key name. Keychain managed manually outside CLI. User reclassified from out-of-scope.

## Required — Low Priority

6. **Error message references CLI flag from library** — `export.go:192`. `"use --password"` in library function.
7-9. **`newGCM` helper, `IsEncrypted` uses `magicBytes`, `minimalOutput` test helper** — Already implemented by /simplify agent.
10. **`encrypt` accepts empty password** — `crypto.ts:32`. Guard at API boundary.
11. **Cross-system format contract comment** — `export.go`, `crypto.ts`, `helpers.ts`. Three implementations with no cross-reference.
12. **`ReadFile` doc comment omits strict matching** — `export.go:181`.
13. **`main.go:56` comment incomplete** — Only mentions retrieval, not storage.

15. **Missing `--keychain`/`--firestore` mutual exclusivity** — Valid but legacy path.
16. **Cross-implementation interop test** — Valuable but high effort.

19. **Truncated ciphertext handling test** — GCM handles natively.
20. **Keychain module has zero tests** — Hard to test without mocking.
21. **`decryptJSON`/`decrypt` should verify magic bytes internally** — Callers already check.

## False Positive

22. **`encryptBuffer` duplicates crypto logic** — Different runtimes require separate implementations.
23. **CLI flag validation should use struct** — Over-engineering.

## Outcome: Iterate (8 high-priority required findings)

natb1:
# Code Quality Evaluation — Iteration 3

## Required — High Priority

1. **Encrypted upload e2e test expects wrong row count** — `upload.spec.ts:93`. Test asserts `toHaveCount(2)` but encrypts the full fixture with 4 transactions. Will fail in CI. Raised by: /review, code-reviewer, pr-test-analyzer.
5. **Cross-implementation interop test (Go→TS golden file)** — Three independent BENC implementations with no cross-language verification. Raised by: pr-test-analyzer, type-design-analyzer. User reclassified from out-of-scope.
6. **`encryptBuffer` uses inline magic numbers instead of importing constants** — `e2e/helpers.ts`. Risk of drift between implementations. Raised by: type-design-analyzer, pr-test-analyzer. User reclassified from out-of-scope.
12. **PBKDF2 blocks main thread** — 600k iterations on UI thread causes 200-500ms UI freeze. Raised by: /review. User reclassified from out-of-scope.

## Required — Low Priority

2. **`decrypt()` error message → "BENC header is incomplete"** — `crypto.ts:59`. After isEncrypted passes, "too short" is misleading. Raised by: /review, comment-analyzer, type-design-analyzer.
3. **`SerializedChartTransaction.amount` JSDoc mention Math.abs() resolution** — `home-chart.ts:9`. Raised by: comment-analyzer, type-design-analyzer.
4. **`budgets-trend-chart.ts:73` y-domain comment wording** — "anchor at 0 when non-negative" misleading. Raised by: comment-analyzer.

7. **`decrypt` empty password validation** — Technically correct error. Raised by: silent-failure-hunter.
8. **`fileOpts` permits incoherent state** — Over-engineering. Raised by: type-design-analyzer.
9. **No BENC version byte** — Hypothetical. Raised by: type-design-analyzer.
10. **Keychain `account` parameter not validated** — Raised by: type-design-analyzer.
11. **No `//go:build darwin` on keychain.go** — Raised by: type-design-analyzer.
13. **Encrypted export uses `.json` extension for binary BENC** — Raised by: /review.
14. **Encrypted export→re-import round-trip e2e test** — Raised by: pr-test-analyzer.
15. **Password dialog cancel e2e test** — Raised by: pr-test-analyzer, silent-failure-hunter.

## False Positive

16. **`encryptBuffer` duplicates crypto logic** — Three runtimes require separate implementations.
17. **`decryptJSON` should verify magic internally** — Unexported; callers check.
18. **`Uint8Array<ArrayBuffer>` generic form** — Valid TS 5.7+ syntax.

## Outcome: Iterate (4 high-priority required findings)

natb1:
# Code Quality Evaluation — Iteration 7

## Required — High Priority

1. **File upload `accept=".json"` blocks encrypted files** — `main.ts:42`. Users can't select `.benc` files in the native file picker. Raised by: code-reviewer.
2. **Unreachable truncation check in `decrypt`** — `crypto.ts:135-137`. In the Worker path, a 4-31 byte BENC file gets "wrong password" instead of "truncated". Combine guards. Raised by: code-reviewer, pr-test-analyzer.

## Required — Low Priority

3. **`postToWorker` non-null assertion** — `crypto.ts:112`. Replace `getWorker()!` with explicit null guard. Raised by: code-reviewer, silent-failure-hunter, type-design-analyzer.
4. **Decryption error missing file path (Go)** — `export.go:215-217`. Wrap with `fmt.Errorf("decrypting %s: %w", path, err)`. Raised by: silent-failure-hunter.
5. **`keychain.go` `-w` comment wording** — `keychain.go:16`. `-w` without a following argument prompts interactively — current wording imprecise. Raised by: comment-analyzer.
6. **`crypto-worker.ts` duplication note wording** — `crypto-worker.ts:3`. "require self-contained modules" is inaccurate — actual reason is avoiding main-thread imports. Raised by: comment-analyzer.

7. Worker message protocol is untyped
8. `crypto-worker.ts` should import constants from `crypto.ts`
9. Move `importPassword` into `AppState` union
10. Encrypted export uses `.json` extension

11-20. Empty password check, worker retry, test gaps, etc.

## False Positive

21-26. Worker fallback design, guarded dead code, already-fixed issues, security clean

## Outcome: Iterate (2 high-priority required findings)

PR #378, opened 2026-04-03
Replace PID-file orphan cleanup with worktree-path process discovery

### LINKED ISSUE #373: Replace PID-file orphan cleanup with worktree-path process discovery
## Problem

The QA server cleanup (`cleanup_all_stale_processes` in `lib.sh`) uses PID files to track and kill orphaned processes when worktrees are deleted. This mechanism was introduced in #199 but has two bugs that allow processes to survive cleanup:

1. **`kill_tree` uses SIGTERM** — java emulators (firestore, storage-rules) survive it via shutdown hooks or simply ignore it
2. **`cleanup()` removes the PID file before verifying processes died** — if the recorded top-level node PIDs are already dead (reparented java/esbuild children survive under launchd), cleanup deletes the PID file without killing anything, losing track of survivors permanently

The result: orphaned java, node, and esbuild processes accumulate across worktree lifecycles, consuming memory until manually killed.

## Root cause

The PID-file approach is inherently fragile because it manages explicit state (write file, read file, remove file) that can fall out of sync with actual process state. The recorded PIDs are only the two top-level node processes (`npm exec firebase-tools`, `npm exec vite`), not the full descendant tree. When these parents die before cleanup runs, the java/esbuild grandchildren become invisible to the cleanup mechanism.

## Proposed fix

Replace PID-file tracking with **worktree-path process discovery**. Every QA process (firebase emulators, vite, esbuild) has the worktree path in its command args. Use `pgrep -f <worktree-path>` to find them directly.

### Strategy

At QA server startup, scan for processes belonging to **any non-active worktree**:

1. `git worktree list` to get the set of active worktree paths
2. `pgrep -f /worktrees/` to find all processes referencing any worktree
3. For each matched process, check if its worktree path is in the active set
4. Kill processes whose worktree is not active (via `kill_tree` to catch descendants, with SIGKILL escalation after a brief SIGTERM grace period)

The trap-based `cleanup()` in `run-qa-server.sh` and `run-acceptance-tests.sh` should also switch to `pgrep -f <current-worktree-path>` for consistency, rather than tracking `$EMULATOR_PID` / `$VITE_PID` variables.

### Advantages over PID files

- No write/read/remove lifecycle to get wrong
- Catches every process, not just the two explicitly recorded
- No race conditions between killing and removing state
- Works retroactively on processes that were never tracked
- Kills processes for non-active worktrees, not just deleted ones

### Edge case: `functionsEmulatorRuntime`

Firebase functions emulator runtimes use the global npx cache path, not the worktree path, in their command args. These are children of the `firebase-tools` node process (which does have the worktree path), so `kill_tree` on the matched parent catches them as descendants.

### What stays

- `cleanup_stale_hub` — hub files in tmpdir still need cleanup independently
- `kill_tree` — still used for killing process trees, but with SIGKILL escalation

### What gets removed

- `write_pid_file`, `remove_pid_file`, `cleanup_all_stale_processes` — replaced by the path-based approach
- PID file JSON format and tmpdir coordination
- `test-pid-cleanup.sh` — rewritten to test the new functions

## Acceptance criteria

- [ ] `kill_worktree_processes` (or similar) finds processes via `pgrep -f` and kills those belonging to non-active worktrees
- [ ] `kill_tree` escalates to SIGKILL after a brief SIGTERM grace period
- [ ] Trap-based `cleanup()` in both scripts uses path-based discovery instead of PID variables
- [ ] `write_pid_file`, `remove_pid_file`, `cleanup_all_stale_processes` are removed
- [ ] Cleanup never kills processes from active worktrees
- [ ] `cleanup_stale_hub` continues to remove dead hub files
- [ ] Test file validates the new path-based cleanup functions
- [ ] No regression in normal server lifecycle (trap EXIT still shuts down cleanly)

## Files involved

- `.claude/skills/ref-pr-workflow/scripts/lib.sh`
- `.claude/skills/ref-pr-workflow/scripts/run-qa-server.sh`
- `.claude/skills/ref-pr-workflow/scripts/run-acceptance-tests.sh`
- `.claude/skills/ref-pr-workflow/scripts/test-pid-cleanup.sh`

## Prior art

Supersedes the approach from #199, which introduced PID-file tracking.

natb1:
## Summary

- Replace fragile PID-file tracking (`write_pid_file`, `remove_pid_file`, `cleanup_all_stale_processes`) with `pgrep -f`-based worktree-path process discovery
- `kill_tree` now escalates to SIGKILL after a 2-second SIGTERM grace period, fixing java emulator survival
- Trap-based `cleanup()` in `run-qa-server.sh` and `run-acceptance-tests.sh` uses path-based discovery instead of PID variables
- Rewrote `test-pid-cleanup.sh` with 8 test cases covering new functions

Closes #373

### DIFF
--- a/.claude/skills/ref-pr-workflow/scripts/lib.sh
+++ b/.claude/skills/ref-pr-workflow/scripts/lib.sh
@@ -299,10 +299,18 @@ kill_worktree_processes() {
 }
 
 # Kill processes belonging to worktrees that no longer exist.
-# Finds all processes with a /worktrees/ path in their args, extracts the
-# worktree path, and kills those not in the active worktree set.
-# Assumes /worktrees/ in process args refers to git worktree directories under this repo.
+# Scopes the search to this repo's worktree directory (derived from git
+# common dir) to avoid killing processes from unrelated repositories.
 cleanup_stale_worktree_processes() {
+  # Derive this repo's worktree container path from git common dir
+  local git_common_dir worktree_root
+  git_common_dir=$(git rev-parse --git-common-dir 2>/dev/null) || {
+    echo "WARNING: git rev-parse --git-common-dir failed; skipping stale cleanup" >&2
+    return 0
+  }
+  # Resolve to absolute path; worktrees live as siblings of the git common dir
+  worktree_root="$(cd "$git_common_dir/.." && pwd)/worktrees"
+
   # Build set of active worktree paths
   local active_paths=""
   local line
@@ -319,9 +327,9 @@ cleanup_stale_worktree_processes() {
     return 0
   fi
 
-  # Find all PIDs with /worktrees/ in their command args
+  # Find PIDs with this repo's worktree root in their command args
   local pids
-  pids=$(pgrep -f '/worktrees/' 2>/dev/null) || true
+  pids=$(pgrep -f "$worktree_root/" 2>/dev/null) || true
   [ -z "$pids" ] && return 0
 
   local exclude_pids
--- a/.claude/skills/ref-pr-workflow/scripts/test-pid-cleanup.sh
+++ b/.claude/skills/ref-pr-workflow/scripts/test-pid-cleanup.sh
@@ -147,7 +147,9 @@ echo "=== Test: cleanup_stale_worktree_processes kills stale, keeps active ==="
 (
   source "$SCRIPT_DIR/lib.sh"
   REAL_WT="$(git rev-parse --show-toplevel)"
-  STALE_WT="/fake/worktrees/deleted-$$"
+  # Stale path must be under this repo's worktree root so the repo-scoped
+  # pgrep in cleanup_stale_worktree_processes finds it
+  STALE_WT="$(cd "$(git rev-parse --git-common-dir)/.." && pwd)/worktrees/deleted-$$"
 
   # Process from a real active worktree
   perl -e 'sleep 300' -- "$REAL_WT/sentinel" &


natb1:
# Code Quality Evaluation — Iteration 2

## Required — High Priority

None.

## Required — Low Priority

F11. Path substring prefix collision (6/7 agents)
- grep -F "$wt_path" matches worktrees whose path is a prefix of another.
- Fix: grep -F "$wt_path/" in kill_worktree_processes and active_paths check.

F12. Trap cleanup missing || guard on kill_worktree_processes (1/7 agents)
- Under set -e, failure aborts trap before subsequent cleanup.
- Fix: add || echo "WARNING: kill_worktree_processes failed" >&2.

F13. _ancestor_pids comment example includes PID 1 but code doesn't add it (1/7 agents)
- Fix: update example to remove PID 1.

F14. _collect_tree_pids comment overstates signal guarantee (revised from F10)
- Fix: "children listed before their parent" instead of "children receive signals before their parent."

F15. Trap comment "git may not be available" imprecise (revised from F8)
- Fix: "working directory may change before trap fires, causing git rev-parse to fail."

F16. cleanup_stale_hub PID recycling comment — clarify mechanism (revised from F10)
- Fix: "If the PID is recycled by an unrelated process, kill -0 succeeds and the stale hub file is preserved."

F17. Add grep -F semantics note to kill_worktree_processes comment (1/7 agents)
- Fix: Add "Uses fixed-string substring matching on process args."

## Pre-applied by /simplify

F18. kill -0 liveness check in cleanup_stale_worktree_processes before kill_tree.

- Unscoped /worktrees/ pattern (4/7 agents) — intentional design, same as F3
- kill_tree unconditional 2-second sleep (4/7 agents) — diminishing returns after F4
- _collect_tree_pids unbounded recursion (3/7 agents) — negligible risk
- Regex extraction fragility / spaces (4/7 agents) — same as F3
- Test gaps (pr-test-analyzer) — moderate effort, implicit coverage sufficient
- ps failure silently swallowed (1/7 agents) — negligible risk

## False Positive

- No absolute path validation — internal callers trusted
- Loss of PID-recycling protection — intentional design (same as F6)
- Missing wait after killing — macOS reaps orphans (same as F9)

## Outcome: Terminate (zero high-priority required findings)

natb1:
# Code Quality Review - Complete ✓

**Reviewer**: Claude Code (via /review + /simplify skills + pr-review-toolkit agents)
**Date**: 2026-04-03
**Outcome**: Terminated after 2 iterations — zero high-priority required findings in iteration 2

## Classification Summary

### Required — High Priority
- F1: `pgrep -f` regex injection via worktree path — Fixed in 42c0e35
- F7: `git worktree list` failure causes mass kill — Fixed in 42c0e35
- F8: Cleanup trap resolves `git rev-parse` at trap time — Fixed in 42c0e35

### Required — Low Priority
- F4: Skip dead processes with `kill -0` before `kill_tree` — Fixed in 42c0e35
- F10: Comment improvements (leaves-first rationale, PID recycling, worktree assumption) — Fixed in 42c0e35
- F11: Path substring prefix collision in `grep -F` — Fixed in fedb270
- F12: Trap cleanup missing `||` guard on `kill_worktree_processes` — Fixed in fedb270
- F13: `_ancestor_pids` comment example included PID 1 but code never adds it — Fixed in fedb270
- F14: `_collect_tree_pids` comment overstated signal guarantee — Fixed in fedb270
- F15: Trap comment imprecise about why `git rev-parse` fails — Fixed in fedb270
- F16: `cleanup_stale_hub` PID recycling comment clarified mechanism — Fixed in fedb270
- F17: `kill_worktree_processes` comment added `grep -F` semantics note — Fixed in fedb270
- F18: `kill -0` liveness check before `kill_tree` in `cleanup_stale_worktree_processes` — Fixed in fedb270

### Out of Scope — High Priority
None.

### Out of Scope — Low Priority
- Unscoped `/worktrees/` pattern — intentional design limitation
- `kill_tree` unconditional 2-second sleep — diminishing returns after F4
- `_collect_tree_pids` unbounded recursion — negligible risk for process trees
- Regex extraction fragility / spaces — same as unscoped pattern
- Test gaps — moderate effort, implicit coverage sufficient
- `ps` failure silently swallowed — negligible risk
- F3: `cleanup_stale_worktree_processes` path regex fragility — intentional design
- F5: Test gap for active-worktree exclusion in main worktree — environment-dependent

### False Positive
- F6: Broader blast radius — intentional design decision per issue body
- F9: Removed `wait` calls — macOS reaps orphans; `kill_tree` sleep provides natural wait
- No absolute path validation — internal callers trusted
- Loss of PID-recycling protection — intentional design
- Missing `wait` after killing — macOS reaps orphans

## Pre-applied by /simplify (iteration 1)
- F2: Duplicated ancestor-exclusion logic — extracted `_ancestor_pids()` helper
- F10b: Dead `EMULATOR_PID`/`VITE_PID` variables — removed

## Conclusion

All required findings (high and low priority) from both iterations have been implemented. Code quality review is complete.

PR #441, opened 2026-04-06
Worktree skill: replace compound commands with detect script

### LINKED ISSUE #435: Worktree skill: replace compound commands with a detect script
## Context

The `/worktree` skill's Section 2 (Detect Layout and Check for Existing Worktree) uses multi-line bash with `$()` substitutions and `&&` chains. Claude combines these into single compound commands like:

```bash
GIT_DIR=$(git worktree list --porcelain | head -1 | sed 's/^worktree //') && PROJECT_ROOT=$(dirname "$GIT_DIR") && echo "PROJECT_ROOT: $PROJECT_ROOT" && git worktree list --porcelain | grep -E "^(worktree|branch)" | paste - - | grep "432-"
```

This triggers permission prompts because:

1. The PreToolUse hook (`approve-workflow-commands.sh:56`) rejects `&&`
2. The hook (`approve-workflow-commands.sh:51`) rejects `$` via `UNSAFE_RE`
3. The command starts with `GIT_DIR=...`, not `git worktree`, so it does not match `Bash(git worktree:*)` in allowedTools
4. When run with `dangerouslyDisableSandbox`, `autoAllowBashIfSandboxed` does not apply

Before the PreToolUse hook was added (commit 63a2389), `autoAllowBashIfSandboxed: true` auto-allowed these sandboxed commands.

## Solution

Create `.claude/skills/worktree/scripts/detect-worktree.sh` that collects all the info the skill needs in one command. The script should:

1. Detect layout (bare vs classic) from `git worktree list --porcelain` second line
2. Compute `PROJECT_ROOT`
3. Accept an issue number argument and check for an existing worktree whose branch starts with `<issue-num>-`
4. Output structured results (e.g., key=value lines) for Claude to parse

Then:

- Update `SKILL.md` Section 2 to call the script instead of inline bash
- Broaden `SCRIPT_RE` in `approve-workflow-commands.sh` from `ref-pr-workflow/scripts/` to match any skill scripts directory (e.g., `\.claude/skills/[a-zA-Z0-9_-]+/scripts/`)

## Acceptance criteria

- [ ] `detect-worktree.sh` outputs layout, PROJECT_ROOT, and existing worktree path/branch (if any) for a given issue number
- [ ] `SKILL.md` Section 2 references the script instead of inline bash blocks
- [ ] `SCRIPT_RE` in the hook matches `.claude/skills/*/scripts/` paths, not just `ref-pr-workflow`
- [ ] `/worktree` invocation produces no permission prompts for the detection step

natb1:
## Summary

- Extract layout detection and worktree lookup from SKILL.md Section 2 into `.claude/skills/worktree/scripts/detect-worktree.sh` to avoid compound bash commands that trigger permission prompts
- Broaden `SCRIPT_RE` in `approve-workflow-commands.sh` from `ref-pr-workflow/scripts/` to `.claude/skills/*/scripts/` so any skill's scripts are auto-approved
- Update hook tests to reflect the broadened pattern

Closes #435

### DIFF
--- a/fellspiral/e2e/font-smoke.spec.ts
+++ b/fellspiral/e2e/font-smoke.spec.ts
@@ -19,21 +19,11 @@ test.describe("font smoke", () => {
     expect(blocked).toEqual([]);
   });
 
-  test("font preload links are present @smoke", async ({ page }) => {
+  test("no font preload links (font-display: optional makes preloads wasteful) @smoke", async ({
+    page,
+  }) => {
     await page.goto("/", { waitUntil: "domcontentloaded" });
-
     const preloads = page.locator('link[rel="preload"][as="font"]');
-    const count = await preloads.count();
-    expect(count).toBeGreaterThanOrEqual(2);
-
-    const hrefs: string[] = [];
-    for (let i = 0; i < count; i++) {
-      const href = await preloads.nth(i).getAttribute("href");
-      expect(href).not.toBeNull();
-      hrefs.push(href!);
-    }
-
-    expect(hrefs.some((h) => h.includes("uncial-antiqua"))).toBe(true);
-    expect(hrefs.some((h) => h.includes("eb-garamond"))).toBe(true);
+    await expect(preloads).toHaveCount(0);
   });
 });
--- a/fellspiral/index.html
+++ b/fellspiral/index.html
@@ -5,8 +5,6 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta name="color-scheme" content="light" />
     <meta name="description" content="A TTRPG game blog by Nate. Nate likes games about social role play." />
-    <link rel="preload" href="/fonts/uncial-antiqua-latin-400-normal.woff2" as="font" type="font/woff2" crossorigin />
-    <link rel="preload" href="/fonts/eb-garamond-latin-400-normal.woff2" as="font" type="font/woff2" crossorigin />
     <link rel="preconnect" href="https://www.googleapis.com" crossorigin />
     <link rel="preconnect" href="https://firebaseinstallations.googleapis.com" crossorigin />
     <link rel="preconnect" href="https://apis.google.com" crossorigin />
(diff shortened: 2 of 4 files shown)


natb1:
# Code Quality Evaluation - Iteration 1

## Required — High Priority

1. **Stale comment in approve-workflow-commands.sh line 4** — Line 4 says "a ref-pr-workflow script" but the regex now matches any skill's scripts directory. Directly misleading. (comment-analyzer, /review)

2. **Validate porcelain `worktree ` prefix before stripping (detect-worktree.sh lines 21, 25)** — `sed -n '1s/^worktree //p'` silently returns empty if prefix absent, causing PROJECT_ROOT to be empty/garbage. dirname "" returns ".", silently operating in wrong directory. Violates clear-errors-over-fallbacks rule. (silent-failure-hunter, type-design-analyzer)

3. **Validate ISSUE_NUM is numeric (detect-worktree.sh line 14)** — Non-numeric input could cause glob characters in case pattern to match unintended branches. One-line guard at system boundary. (silent-failure-hunter, type-design-analyzer)

## Required — Low Priority

4. **/simplify edit: consolidated head -1 | sed into single sed (pre-applied)** — Already applied to lines 21 and 25. Keep it. (/simplify)

5. **Document paired invariant (WORKTREE_PATH + WORKTREE_BRANCH always both set or both empty)** — One-line header comment addition. (type-design-analyzer)

6. **Unit tests for detect-worktree.sh** — High effort, requires mock git setup. Script was QA-tested manually. Defer to follow-up issue. (pr-test-analyzer)

7. **Negative test for SCRIPT_RE path traversal in skill name segment** — Already enforced by character class. (pr-test-analyzer)
8. **Output quoting for values with spaces** — Paths never contain spaces in this project. (type-design-analyzer)
9. **SKILL.md should instruct caller to check exit code** — Claude already handles nonzero exits. (silent-failure-hunter)

## False Positive

10. **Bare detection heuristic fragility** — Pre-existing behavior, not a regression. Git porcelain format is stable. (silent-failure-hunter)

## Outcome: Iterate (3 high-priority required findings)

PR #548, opened 2026-04-16
productivity-tui: filter stale session entries via PID liveness

### LINKED ISSUE #544: productivity-tui: filter stale session entries when Claude exits uncleanly
Follow-up to #457. The session state file
(`~/.local/share/productivity-tui/sessions.json`) is purely event-driven: entries
are added by `on-active.sh` / `on-idle.sh` and removed only by `on-close.sh`
(Claude's `SessionEnd` hook). When Claude exits uncleanly — SIGKILL, terminal
force-closed, crash, system reboot — `SessionEnd` never fires and the entry
persists forever. The TUI has no liveness check, so stale entries accumulate.

## Reproduction

1. Open Claude Code in a terminal; verify its working dir appears in `productivity-tui`.
2. `kill -9 <claude-pid>` (or force-close the terminal window).
3. Re-open `productivity-tui`. The entry for the killed session is still listed
   as active/idle, indefinitely.

Observed today: 13 entries in the state file, 9 live `claude-unwrapped`
processes. The 4 stale entries dated back 5–8 days (2026-04-08 and 2026-04-11).

## Proposed fix

Record `transcript_path` on each session entry (already available in Claude's
hook JSON — no PID discovery needed) and filter dead sessions at read time
using `lsof -t <transcript_path>`. Empty output means no process holds the
transcript file open, so the session is dead and should be hidden.

### Changes

- `productivity-tui/hooks/update-session.sh`: capture `.transcript_path` from
  hook stdin and include it in the session record.
- `productivity-tui/internal/session/session.go`: add `TranscriptPath` to the
  `Session` struct; add a `FilterLive` helper (or fold into `ReadSessions`)
  that shells out to `lsof -t` and drops sessions whose transcript has no
  open file handle.
- `productivity-tui/internal/session/session_test.go`: cover
  `FilterLive` with an alive entry (create a temp file, hold it open via
  `os.OpenFile` in the test) and a dead entry (temp file with no holder).

### Alternative considered

Store the Claude process PID and check with `kill(pid, 0)`. Rejected because
Claude's hook JSON does not expose a PID (only `session_id`, `transcript_path`,
`cwd`, `permission_mode`, `hook_event_name`) and `$PPID` in the hook script is
unreliable (shells source `~/.zshrc`, adding intermediate processes). PID-based
checks also suffer from PID recycling; `lsof` on the transcript path does not.

Age-out by `last_activity` timestamp was also considered but rejected — a
genuinely idle session older than the threshold would be wrongly hidden.

## Acceptance criteria

- [ ] `update-session.sh` writes `transcript_path` to each session record
- [ ] Stale entries (no process holds `transcript_path` open) are filtered from
      the TUI view
- [ ] Live idle sessions (Claude process alive, currently idle) continue to
      display with the idle indicator
- [ ] `on-close.sh` behavior unchanged — clean-exit cleanup still removes the
      entry from the file
- [ ] Unit test covers both live and dead cases
- [ ] Existing `TestReadSessions_*` tests still pass (no schema break for
      records written before this change — missing `transcript_path` should
      be treated as live, to avoid mass-hiding pre-upgrade entries)

## Immediate workaround

`echo '{}' > ~/.local/share/productivity-tui/sessions.json`

natb1:
## Summary

Closes #544.

Session entries in `~/.local/share/productivity-tui/sessions.json` persist
forever when Claude exits uncleanly (SIGKILL, force-closed terminal, reboot)
because `SessionEnd` never fires. Observed state: 13 entries for 9 live
processes, stale entries 5-8 days old.

The earlier `lsof` approach did not work: Claude Code does not hold
transcript `.jsonl` files open. `lsof -t <transcript>` returned empty for
both live and dead sessions, so `FilterLive` dropped every post-upgrade
entry regardless of liveness. Verified by inspecting open fds of the
running `claude-code` process (36 fds, none to any `.jsonl`). The original
unit test passed only because it opened a real file handle in the test
process, validating `lsof` rather than the liveness signal.

Replace with a PID-walk + start-time signal:

- `update-session.sh` walks up from `$PPID` until it finds an ancestor
  whose command line contains `claude-code` / `.claude-unwrapped` / `claude`.
  Records `pid` and `ps -o lstart=` output in `sessions.json`.
- `session.FilterLive` checks `syscall.Kill(pid, 0)` and matches the
  stored start-time string against a fresh `ps -o lstart=` query.
- `PID == 0` and empty `pid_start` records are preserved as
  "unknown identity" (pre-upgrade files and rare hook misses).
- `ps` / `kill` errors fail open -- liveness probe failures must not cause
  live sessions to disappear.

### Addresses the three reasons the PID approach was originally rejected

- Hook JSON has no PID -- walk from `$PPID`.
- `$PPID` unreliable with `.zshrc`-sourced shells -- walk past transient
  intermediates by matching `claude-code` in command line.
- PID recycling -- start-time string mismatch drops recycled PIDs.

### Field removed

`transcript_path` is gone (unused by the TUI; only consumed by the broken
`lsof` filter). Old `sessions.json` entries with the field unmarshal
cleanly and drain naturally as sessions get re-touched.

## Test plan

- [x] `go test ./productivity-tui/...` passes (unit, including `TestFilterLive`, `TestFilterLiveEmptyAndNilInput`)
- [ ] Manual: start Claude in a terminal, confirm the entry appears with non-zero `pid` in `sessions.json`
- [ ] Manual: `kill -9` the captured Claude PID; confirm the entry disappears on the next TUI tick (PID-liveness kills it)
- [ ] Manual: pre-upgrade records (`pid == 0` or empty `pid_start`) continue to display until `on-close.sh` removes them naturally

No acceptance/smoke tests -- productivity-tui is a local Go TUI with no Firebase or web surface.


natb1:
# Smoke Test Evaluation - Iteration 1

**Date**: 2026-04-16
**Branch**: 544-tui-filter-stale-sessions
**PR**: #548
**HEAD**: caefbfd (Merge remote-tracking branch 'origin/main' into 544-tui-filter-stale-sessions)
**CI Run**: 24536235789 (PR Checks)

## Result: All checks passed

- `preview-and-smoke`: success (completed 2026-04-16T22:00:48Z)
  - Deploy Cloud Functions (if changed): success
  - Deploy previews and run smoke tests: success
- `acceptance`: success (completed 2026-04-16T22:00:51Z)

The latest PR Checks run for HEAD already completed successfully before this verify iteration began. No infrastructure failures detected. Preview deployment verified via smoke tests.

## Conclusion

Smoke tests passed on first iteration. Proceeding to terminate.

natb1:
# Smoke Test Review - Complete

**Date**: 2026-04-16
**Branch**: 544-tui-filter-stale-sessions
**PR**: #548

## Iterations

- Iteration 1: All smoke tests passed (no fixes required)
  - `preview-and-smoke`: success
  - `acceptance`: success
  - CI Run: 24536235789 on HEAD caefbfd

## Conclusion

Smoke tests passed. Preview deployment verified.

natb1:
## QA Iteration 2 — Evaluation

**Outcome:** Approved (terminate)

**Reviewer:** Nathan Buesgens
**Tested By:** Synthetic-entry probe (ReadSessions + FilterLive end-to-end) + on-disk state inspection

### Scope notes

- `productivity-tui` is a terminal UI (bubbletea); Chrome browser demo skipped — not applicable.
- Items 2 and 4 cannot be fully validated pre-merge because only this worktree runs the upgraded hook. Other worktrees' sessions land as pre-upgrade (`pid == 0`) records and cannot exercise the dead-PID path, so a standalone synthetic probe was used to validate `FilterLive` end-to-end against a crafted `sessions.json`.

### Results per checklist item

| # | Item | Result | Method |
|---|---|---|---|
| 1 | Current session appears in TUI | PASS | File state: entry `b2106cd6` has live PID 91352 + matching `lstart`; probe kept it. |
| 2 | Dead process entries disappear | PASS | Probe: synthetic `dead-session` (PID 149862, ESRCH) and `recycled-session` (right PID, wrong `pid_start`) both dropped by `FilterLive`. |
| 3 | TUI renders without errors on launch | PASS | Build clean; unit tests (`./internal/session/...`, `./internal/app/...`) green; `ReadSessions + FilterLive` path exercised against real file without panic. |
| 4 | Multiple worktree sessions render | PASS (partial) | 12 pre-upgrade (`pid==0`) entries + 1 post-upgrade kept; full multi-worktree post-upgrade behavior emerges post-merge as other worktrees pick up the upgraded hook. |
| 5 | Pre-upgrade preservation (PID==0 kept) | PASS | Probe: `pre-upgrade-session` (pid 0) kept. `jq` count confirms 12 pre-upgrade entries in real file are preserved. |
| 6 | New sessions record `pid` + `pid_start` | PASS | Entry `b2106cd6` has `pid=91352`, `pid_start="Thu Apr 16 17:46:36 2026"` (matches `ps -p 91352 -o lstart=` exactly), `working_dir` for this worktree, no `transcript_path` field. |

### Investigation note (sandbox artifact)

Initial stress-run of the synthetic probe in sandboxed Bash showed `recycled-session` staying kept (20/20). Root cause: sandbox denies `kill(pid, 0)` with `EPERM` for processes outside the sandbox namespace, causing `isPIDLive` to fail-open (return true). Running the same probe with sandbox disabled shows `recycled-session` correctly dropped (5/5). This is a sandbox artifact, not a bug — real user invocation runs unsandboxed and the `ps lstart` equality check executes correctly.

### Deferred to post-merge

- Item 2 and 4 get organic real-world validation once other worktrees upgrade their hooks and write post-upgrade entries.
- Post-merge prod-QA (workflow Step 13) will cover ongoing stale-session hiding.

### Unit tests

`go test ./internal/session/... ./internal/app/...` → `ok` (cached).

### Build artifact

`/tmp/productivity-tui-pr548` (5.0 MB, Mach-O arm64) built from commit `f2dbf2e`.

### Throwaway artifacts

- `productivity-tui/cmd/qacheck/` — synthetic probe, deleted after verification.
- `/tmp/qa-synthetic-sessions.json` — synthetic input, deleted.

Tree clean; only untracked file is `.claude/scheduled_tasks.lock` (runtime).

natb1:
# QA Review - Complete ✓

**Date**: 2026-04-16
**Branch**: 544-tui-filter-stale-sessions
**Reviewer**: Nathan Buesgens
**Tested By**: Synthetic-entry probe + on-disk state inspection

## Iterations

- **Iteration 1**: `lsof`-based liveness check incorrectly dropped every post-upgrade entry. Root cause: Claude Code does not hold `.jsonl` transcript files open, so `lsof -t <transcript>` returns empty for live and dead sessions alike. The iter-1 unit test passed only because the test process itself opened the file, which validated `lsof` rather than the liveness signal. → Fixed in commit `f2dbf2e` by replacing the signal with a PID-walk + `ps lstart=` start-time comparison.
- **Iteration 2**: All QA test cases passed. `FilterLive` correctly drops dead PIDs (ESRCH) and recycled PIDs (start-time mismatch), and correctly preserves live PIDs and pre-upgrade (`pid==0`) entries.

## QA Summary

- Total test cycles: 2 (iter 1 surfaced lsof signal failure; iter 2 validated PID+start-time replacement)
- Key behaviors verified: live PID retention, dead PID drop, recycled PID drop on start-time mismatch, pre-upgrade (pid==0) preservation, new on-disk schema (pid, pid_start, no transcript_path)
- Edge cases tested: ESRCH dead PID, recycled PID with mismatched pid_start, pre-upgrade entries mixed with post-upgrade, multi-worktree mix (verifiable portion), sandbox EPERM fail-open behavior documented
- Total issues found and resolved: 1 (lsof-based filter never worked because Claude Code does not hold .jsonl transcripts open — replaced with PID-walk + ps lstart check in commit f2dbf2e)

## Browser Exploration Summary

N/A — productivity-tui is a terminal (bubbletea) application. Chrome demo skipped.

## Conclusion

All QA test cases passed. PR approved for code quality review.

natb1:
# Code Quality Review - Iteration 1

Seven parallel reviewers examined PR #548. Per-reviewer findings summarized below.

## /review Skill

Verdict: approve with minor nits. Implementation is sound; fail-open policy is correct; postmortem on lsof failure is commendable; tests exercise important cases.

Flagged:
1. `find_claude_pid` pattern gap: bare `claude` argv[0] (no slash) won't match `*/claude`. Consider broadening or documenting assumption.
2. `*claude-code*` breadth could false-positive on wrapper script paths.
3. `ps -o command=` may truncate on macOS — add `-ww`.
4. `ps -o lstart=` format stability — add a comment that PIDStart is an opaque string compared verbatim; note LC_TIME sensitivity.
5. EPERM in isPIDLive is a *positive* liveness signal, not fail-open — clarify comment.
6. Per-tick `ps` fork cost — acceptable at current scale; batch if scales.
7. `syscall.Kill` Unix-only — TUI is Unix-only, fine.
8. Hook reliability: stderr log when find_claude_pid fails would aid debugging.
9. PR title still says "lsof" — update.

## pr-review-toolkit: code-reviewer

Verdict: no critical or important issues. Fail-open behavior falls within "defensive check at system edge" carve-out per code-style.md.

Flagged (sub-threshold):
- ps fork per session per tick (perf).
- Pattern breadth for non-slash argv[0].
- Non-ESRCH Kill returning `true` is semantically correct (EPERM = process exists).

## /simplify Skill

Verdict: diff is clean, no changes made.
- No reuse opportunities.
- Comments document non-obvious WHY — keep.
- PIDStart string-compare is correct equality semantics — keep.
- ps-per-session fork flagged but not fixed (acceptable at current scale).

## pr-review-toolkit: comment-analyzer

Critical:
1. session.go:27-29 — "EPERM" example conflates two distinct cases. Kill(pid,0) returning EPERM is a *positive liveness signal*, not a fail-open fallback. Split the rationale.
2. session.go:24-25 — "recorded claude-code PID" overstates; hook matches several naming conventions. Reword to "Claude CLI PID".

Improvement opportunities:
3. update-session.sh:18-19 — walk comment describes WHAT not WHY; add purpose + cross-ref to session.FilterLive.
4. update-session.sh:23 — `hops < 20` magic number needs rationale.
5. update-session.sh:27 — pattern list needs rationale comment.
6. session.go:26-27 — "records written before this field existed" risks rot. Reframe as permanent contract.

## pr-review-toolkit: pr-test-analyzer

Critical gaps:
1. Shell `find_claude_pid` has zero tests (criticality 8). Silent-no-op risk across install paths.
2. No test that `loadSessions` actually applies `FilterLive` (criticality 7). Cheap regression guard.

Important:
3. ps-unavailable fail-open path untested (criticality 6).
4. Non-ESRCH Kill fail-open path untested (criticality 6).
5. Recycled-PID sentinel should be a named constant (criticality 5).
6. deadPID recycle race is benign but worth flagging (criticality 5).

Priority rec: add `loadSessions` wiring test first.

## pr-review-toolkit: silent-failure-hunter

High-priority:
1. `isPIDLive` fail-open paths have zero observability (no log/counter/signal). ps-missing or EPERM silently regresses feature.
2. Empty `wantStart` treated as probe, not error. Hook race (find_claude_pid succeeds but `ps -o lstart=` fails) writes `{pid: N, pid_start: ""}` → reader classifies live session as dead. Violates fail-open design intent.
4. `find_claude_pid` `2>/dev/null || true` silently swallows every ps failure.

Medium-priority:
3. Zero test coverage for fail-open paths.
5. Silent `transcript_path` drop via implicit denylist — replace with explicit allowlist.
6. app.go m.sessions wiped on error (pre-existing pattern; UX flicker).

Verdict:
- isPIDLive fail-open correct but unobservable — needs stderr log.
- `find_claude_pid` silent swallowing NOT justified — needs stderr log on failure.
- `|| true` on pid extraction NOT justified — remove.
- transcript_path drop — should be explicit schema allowlist with comment.

## pr-review-toolkit: type-design-analyzer

Ratings: Encapsulation 4/10, Invariant Expression 3/10, Invariant Usefulness 8/10, Invariant Enforcement 3/10.

Key finding: (PID, PIDStart) pair is coupled but the type doesn't express it. Shell hook can produce `{pid: N, pid_start: ""}` on race — Go side silently classifies as dead, opposite of fail-open policy.

Recommended:
1. Introduce `ProcessIdentity` sub-struct to make pairing structural.
2. Add `Valid()` method; repair-or-keep on invalid.
3. Make `isPIDLive` a method on ProcessIdentity.
4. Fix shell hook: clear CLAUDE_PID if CLAUDE_START empty — one-line fix that closes the race.
5. Keep PIDStart as string (correct).
6. Add test for `{PID: non-zero, PIDStart: ""}` to pin policy.

Strengths: policy choices sound; omitempty keeps backward-compat; test covers four states; string (not time.Time) correct for recycle detection.


---

# Code Quality Review - Evaluation 1

**Reviewer**: Claude Code (via /review + /simplify skills + pr-review-toolkit agents)
**Date**: 2026-04-16
**Outcome**: Iterate — 1 high-priority required finding + 16 low-priority required findings

## Classification Summary

### Required — High Priority

- **F1. Hook race: CLAUDE_PID set with empty CLAUDE_START** — If `find_claude_pid` succeeds but `ps -o lstart=` then fails, hook writes `{pid: N, pid_start: ""}`. Go-side `isPIDLive` compares non-empty ps output against `""` → returns false → live session classified dead. Directly violates stated fail-open policy. Fix: pair PID with START in the hook (clear CLAUDE_PID if CLAUDE_START empty).

### Required — Low Priority

**Comment accuracy / rationale (session.go, update-session.sh):**
- F2. Split EPERM vs. probe-failure rationale in FilterLive comment (EPERM is positive liveness signal, not fail-open).
- F3. Reword "claude-code PID" → "Claude CLI PID".
- F4. Reframe PID==0 branch as permanent contract, not temporary upgrade-compat note.
- F5. Document `hops < 20` bound and pid==0 fallback consequence.
- F6. Document pattern-list shapes (nix-wrapped, unwrapped, direct, with-args).
- F7. Rewrite `find_claude_pid` comment to explain WHY + cross-ref `session.FilterLive`.
- F8. Note PIDStart opaque verbatim string; comment on locale/format stability.

**Shell hardening:**
- F9. Add `-ww` to ps calls to avoid macOS command-line truncation.
- F10. Remove `|| true` on ps|tr pipeline (masks pipe failures; command-substitution already suppresses errexit).
- F11. Log stderr when `find_claude_pid` fails or lstart query fails.

**Go-side observability:**
- F12. Log stderr in `isPIDLive` on non-ESRCH kill errors and ps errors (fail-open remains correct; operators need signal).

**Schema clarity:**
- F13. Replace `with_entries(select(...))` implicit denylist with explicit schema-construction jq expression.

**Test additions:**
- F14. Add `app_test.go` test that `loadSessions` applies `FilterLive` — cheap regression guard.
- F15. Name the "Thu Jan 1 00:00:00 1970" sentinel (const `mismatchedStart`).

**Coverage edge:**
- F16. Broaden `find_claude_pid` pattern to cover bare `claude` argv[0], or comment the assumption.

**PR metadata:**
- F17. Update PR title from "lsof" to reflect PID-liveness approach.

### Out of Scope — High Priority

- F18. Shell tests for `find_claude_pid` — high impact but high effort; test harness (bats/shunit2) not yet in repo.
- F19. `ProcessIdentity` sub-struct refactor — would structurally encapsulate coupled pair, but F1 hook fix resolves the correctness bug at far lower cost.

### Out of Scope — Low Priority

- F20. Fail-open test paths (ps-unavailable, EPERM) — require refactor of `isPIDLive` to inject kill fn / PATH shim.
- F21. Per-tick `ps` fork cost — premature at current scale (<10 sessions).
- F22. Windows build tag — productivity-tui is Unix-only.
- F23. `m.sessions` wipe on read error — pre-existing pattern, 1-sec refresh smooths it.
- F24. PID==0 POSIX broadcast collision — code is already guarded.
- F25. `*claude-code*` false-positive on wrapper script names — addressed by F6 comment.

### False Positive

None.

## Conclusion

Iterate: implement F1 (required-high) plus all low-priority required findings in a single pass. Run unit tests. Return to Step 0 for next iteration cycle.

natb1:
# Code Quality Review - Iteration 2

Seven parallel reviewers examined the iter-1 delta (commit 34d1714, addressing all 17 iter-1 findings) plus the pre-existing change surface. Per-reviewer findings summarized below.

## /review Skill

Verdict: converged on code. All 17 iter-1 required findings verified resolved in commit 34d1714.

Flagged:
1. **I1 (required-low).** PR *body* is stale — still describes `lsof -t` and `TranscriptPath` field. Title was updated (F17), body was not.
2. **I2 (out-of-scope-low).** `TestLoadSessionsFiltersDead` reuses the spawn+reap+reuse pattern from `TestFilterLive` — PID recycling race is astronomically unlikely in a single-threaded test binary but not impossible on heavy CI runners.
3. **I3 (out-of-scope-low).** `psLStart`/`psLStartOf` duplicated across `session_test.go` and `app_test.go`.
4. **I4 (out-of-scope-low).** `find_claude_pid` patterns match substrings anywhere in the command line (e.g., `*claude-code*`).

## pr-review-toolkit: code-reviewer

Verdict: one confirmed carry-over from iter 1.

Flagged:
- **F10 carry-over (important, ~85 confidence).** `|| true` still present on `update-session.sh:35` (`cmd=$(ps -ww -o command= -p "$pid" 2>/dev/null || true)`) and `:52` (`CLAUDE_START="$(ps -o lstart= ... || true)"`). Per `.claude/rules/code-style.md` ("prefer clear errors over defensive fallbacks") — command substitution already suppresses errexit, so `|| true` is redundant or (under `inherit_errexit`) actively hides probe failures.

All other iter-1 findings (F1–F9, F11–F17) verified addressed.

## /simplify Skill

Edit permission was denied; findings reported without auto-applied fixes.

Flagged:
- **F26 (required-low).** PID==0 comment in `session.go:30–33` re-introduces upgrade-compat framing via parenthetical "(including records written before this field existed)". F4 from iter 1 asked to reframe as permanent contract.
- **F27 (out-of-scope-low).** jq schema comment narrates the change (`update-session.sh:69–70`).
- **F28 (out-of-scope-low).** `psLStart`/`psLStartOf` duplicated across test packages.
- **F29 (out-of-scope-low).** `TestLoadSessionsFiltersDead` re-tests `FilterLive` logic instead of just wiring.

## pr-review-toolkit: comment-analyzer

Verdict: no critical issues; five refinement opportunities and two removal candidates.

Flagged:
1. **#1.** `FilterLive` godoc overstates EPERM vs probe-failure distinction — `isPIDLive` treats EPERM identically to other non-ESRCH errors (logs "fail open"). Either collapse the comment to a single bullet or add an explicit EPERM branch.
2. **#2.** Hook cross-reference "(see comment in session.go)" at `update-session.sh:31` is too vague — state the contract inline.
3. **#3.** `find_claude_pid` pattern-shape comment (5 bullets) vs `case` glob (6 arms) can drift silently.
4. **#4.** "lstart query failed" comment's second sentence (`update-session.sh:54–56`) overstates certainty — fail-open branch keeps the session.
5. **#5.** jq schema comment legacy reference ("including legacy transcript_path from earlier hook versions") — archaeology of a field no longer in the codebase.
6. **#6.** `PIDStart` godoc has a redundant "do not reformat or parse" caveat.

Positive observations: `20 hops` bound comment is the right shape; PID==0 rationale captures the product-level tradeoff; `(see session.FilterLive)` cross-reference at `update-session.sh:18-20` is load-bearing; no comment clutter in new test code.

## pr-review-toolkit: pr-test-analyzer

Verdict: one critical gap; several quality improvements.

Flagged:
1. **CG-1 (criticality 7) — CRITICAL.** `isPIDLive` with `PID > 0 && PIDStart == ""` is unpinned. `strings.TrimSpace(out) == ""` is false for any live process → live session silently classified dead. The F1 hook fix prevents new records from having this shape; stale state files and future hook regressions are not covered.
2. **II-1 (criticality 5).** `TestLoadSessionsFiltersDead` conflates "FilterLive works correctly" with "loadSessions invokes FilterLive" — reduces fault isolation.
3. **II-2 (criticality 4).** No test exercises `loadSessions` with a malformed-JSON file asserting `sm.err != nil && sm.sessions == nil`.
4. **II-3 (criticality 3).** No coverage of `FilterLive` on empty/nil input — trivial to pin.
5. **TQ-1 (criticality 4).** `TestFilterLive`'s "dead PID" assertion may pass via start-time mismatch rather than ESRCH, weakening fault isolation for the ESRCH branch.
6. **TQ-2 (criticality 2).** `mismatchedStart` sentinel not documented.

Positive observations: behavior-focused assertions; real-process probes; DAMP naming; skip-gating on `ps` availability; F14 closed.

## pr-review-toolkit: silent-failure-hunter

Verdict: two critical; two high; four medium.

Flagged:
- **C1 (critical).** `find_claude_pid` `|| true` on line 35 masks ps failures (duplicate with code-reviewer F10 carry-over).
- **C2 (critical).** `CLAUDE_START` capture `|| true` on line 52 masks pipe failures (duplicate with code-reviewer F10 carry-over).
- **H1 (high, pre-existing).** `ReadSessions` treats empty file as first-run (session.go:87-89) — could mask truncation or corruption.
- **H2 (high, pre-existing F23).** `sessionsMsg` load error wipes `m.sessions` — one-tick flicker on transient read errors.
- **M1 (medium).** `isPIDLive` fail-open log lines lack session ID/working_dir context; no rate limiting.
- **M2 (medium).** `FilterLive` does not log when it drops a session — silent successful drop.
- **M3 (medium, pre-existing).** jq `session_id` error doesn't dump payload on failure.
- **M4 (medium).** `isPIDLive` trusts `exec.Command("ps")` PATH with no preflight — fails open every tick if `ps` missing.

Positive observations: F1/F9/F10/F11/F12 fixes are substantive and correct.

## pr-review-toolkit: type-design-analyzer

Ratings:
- Encapsulation: 4/10 — plain data struct, no constructor, `PID==0` sentinel, field-level invariants enforced only by convention.
- Invariant Expression: 5/10 — prose comments are excellent, but structure (two separate fields with `omitempty`) does not express the (PID, PIDStart) coupled-pair invariant.
- Invariant Usefulness: 8/10 — core liveness invariant is correct; PID==0 escape hatch is pragmatic; EPERM-as-live is right.
- Invariant Enforcement: 3/10 — symmetric defense to F1 is missing on the consumer side.

Flagged:
- **C1.** Consumer-side coherence check missing: `FilterLive` should treat `PID > 0 && PIDStart == ""` as unknown (duplicate with pr-test CG-1).
- **C2.** `PID == 0` sentinel is load-bearing but not a named constant.
- **C3.** `omitempty` on `PID`/`PIDStart` asymmetric with hook output.
- **C4.** `Session` has no self-identity (minor API concern).
- **C5.** `PIDStart` opacity enforced by convention only — named type alias would make misuse visible at compile time.
- **C6.** `isPIDLive` not injectable — acknowledged iter-1 F20 constraint.

Recommended for this PR: mirror F1 on consumer side (three-character code change closing the symmetric gap) and introduce `const PIDUnknown = 0`.

## Summary

All iter-1 required findings verified resolved except the narrowly-scoped F10 (which was fixed only on `update-session.sh:42`, not the other two `|| true` sites at lines 35 and 52). One new high-priority finding surfaced from the test-coverage and type-design reviewers (consumer-side coherence for the `PID > 0 && PIDStart == ""` shape — the symmetric mirror of iter-1 F1). Five low-priority required findings plus five comment revisions from prior iterations (auto-resolved). See evaluation file below for full classification.


---

# Code Quality Review - Evaluation 2

**Reviewer**: Claude Code (via /review + /simplify skills + pr-review-toolkit agents)
**Date**: 2026-04-17
**Outcome**: Iterate — 1 high-priority required finding + 5 low-priority required findings + 5 auto-resolved comment revisions

## Classification Summary

### Auto-resolved (comments revised in prior iterations — picked strongest version autonomously)

- **A1. EPERM vs probe-failure split in `FilterLive` godoc** (comment-analyzer #1) — **Keep current split framing.** Documentation value exceeds the uniform-code concern; split remains useful even though `isPIDLive` fails open uniformly.
- **A2. PID==0 comment parenthetical re-introducing upgrade-compat** (simplify F26, revisits iter-1 F4) — **Remove parenthetical "(including records written before this field existed)".** Permanent-contract framing stands alone; archaeology not load-bearing.
- **A3. Hook cross-ref "which FilterLive preserves (see comment in session.go)"** (comment-analyzer #2, revisits iter-1 F5) — **Inline the contract statement** — remove vague cross-reference.
- **A4. Pattern-list comment has 5 bullets, `case` has 6 patterns** (comment-analyzer #3, revisits iter-1 F6) — **Add 6th bullet for `claude\ *`** (direct binary invocation with args via bare argv[0]). Mirror case arms 1:1 so drift is visible on review.
- **A5. PIDStart godoc "do not reformat or parse" caveat** (comment-analyzer #6, revisits iter-1 F8) — **Keep current wording.** Prohibition adds durability against future "helpful" normalization.

### Required — High Priority

- **H-1. Consumer-side coherence check missing: `PID > 0 && PIDStart == ""`** — raised by pr-test-analyzer (CG-1, criticality 7) and type-design-analyzer (C1). `FilterLive` at `session.go:43` calls `isPIDLive` with empty `wantStart`; `strings.TrimSpace(out) == ""` is false for any live process, so the session is silently dropped. The hook's F1 fix prevents new records from having this shape, but stale state files from earlier hook versions (and any future regression of the hook's coherence guard) expose this silent-failure mode. Symmetric defense to F1 (required-high in iter 1). Fix: `if s.PID == 0 || s.PIDStart == "" || isPIDLive(...)` in `FilterLive`, update godoc, add test case.

### Required — Low Priority

- **L-1. `|| true` on ps calls at lines 35 & 52** — raised by code-reviewer (F10 carry-over) and silent-failure-hunter (C1, C2). `update-session.sh:35` (`cmd=$(ps -ww ... || true)`) and `:52` (`CLAUDE_START=$(ps -o lstart=... | sed ... || true)`) still carry `|| true`. Inconsistent with the F10 fix on line 42. Per `.claude/rules/code-style.md` ("prefer clear errors over defensive fallbacks") — `$(...)` already suppresses errexit, so the fallback is either redundant or actively hiding failures.
- **L-2. PR body text is stale** — raised by /review (I1). Body still describes `lsof -t`, `transcript_path`, and the `TranscriptPath` field. Title was updated in iter 1 (F17) but body was not.
- **L-3. "lstart query failed" comment's second sentence imprecise** — raised by comment-analyzer (#4). `update-session.sh:54-56` second sentence overstates certainty (fail-open branch keeps the session). Drop the second sentence; first conveys WHY.
- **L-4. jq schema comment verbosity** — raised by comment-analyzer (#5) and simplify (F27). `update-session.sh:69-70` "(including legacy transcript_path from earlier hook versions)" references a field no longer in any checked-in code. Trim to single fragment or delete parenthetical.
- **L-5. `FilterLive` empty/nil input test** — raised by pr-test-analyzer (II-3). 5-line test pinning `FilterLive(nil) → empty map` contract. Trivially cheap.

### Out of Scope — High Priority

None.

### Out of Scope — Low Priority

- **O-1.** Test-design improvement: `TestLoadSessionsFiltersDead` re-covers `FilterLive` logic (pr-test II-1, simplify F29) — test works; refactor not correctness-relevant.
- **O-2.** `loadSessions` error-path short-circuit test (pr-test II-2) — covered transitively by `TestReadSessions_MalformedJSON`.
- **O-3.** Dead-PID test race potential flakiness (pr-test TQ-1, /review I2) — recycled-PID branch catches the race anyway.
- **O-4.** `mismatchedStart` constant lacks comment (pr-test TQ-2) — name is self-documenting.
- **O-5.** `psLStart`/`psLStartOf` helper duplicated across packages (/review I3, simplify F28) — cross-package extraction not warranted for two copies.
- **O-6.** `find_claude_pid` patterns match substrings (/review I4) — addressed by F6 comment; real risk negligible.
- **O-7.** `isPIDLive` log lines lack session ID + rate limiting (silent-failure M1) — ergonomics only.
- **O-8.** `FilterLive` does not log drops (silent-failure M2) — feature request, not a silent-failure bug.
- **O-9.** `isPIDLive` has no `ps` PATH preflight (silent-failure M4) — fail-open policy handles it; premature optimization.
- **O-10.** ReadSessions empty-file masks corruption (silent-failure H1) — pre-existing code, not in PR diff (scope rule).
- **O-11.** `sessionsMsg` wipe on error (silent-failure H2) — already F23 OOS in iter 1.
- **O-12.** jq `session_id` error doesn't dump payload (silent-failure M3) — pre-existing, not in PR diff.
- **O-13.** Named `PIDUnknown` constant, omitempty asymmetry, no self-identity (type-design C2/C3/C4) — minor type-design refinements; F19 established this area OOS.
- **O-14.** `ProcessIdentity` sub-struct (type-design follow-up) — already F19 OOS.
- **O-15.** `PIDStart` named type alias (type-design C5) — prose doc doing the job.
- **O-16.** `isPIDLive` injectability for fail-open tests (type-design C6) — already F20 OOS.

### False Positive

None.

## Conclusion

Iterate: implement H-1 (required-high consumer-side coherence check + test) plus all low-priority required findings (L-1 through L-5) in a single pass, applying the 5 auto-resolved comment revisions alongside. Run unit tests. Return to Step 0 for next iteration cycle.

natb1:
# Code Quality Review — Iteration 3 Reviewer Output

## /review Output

Iter-2 addressed H-1 (consumer-side coherence), L-1 through L-5, and five auto-resolved comment revisions. All unit tests pass. `go vet` clean. Core logic and documentation are now coherent.

### Findings

**R1 (Required, Medium priority) — Line 53 `|| true` removal may fail the whole hook on ps race**

The L-1 fix in iter-2 removed `|| true` from two `ps` command substitutions. On line 36 inside `find_claude_pid`, removal is safe because the function is called within `if pid=$(find_claude_pid); then`, so errexit is ignored inside the function body (standard bash "errexit in conditional context" rule).

On line 53, however, the assignment is at top-level inside the `then` branch:
```bash
CLAUDE_START="$(ps -o lstart= -p "$CLAUDE_PID" 2>/dev/null | sed -e 's/^ *//' -e 's/ *$//')"
```

Errexit is active here. If `ps` fails (e.g., the Claude PID exited in the race window between `find_claude_pid` returning and this line), the pipeline's failure propagates to the assignment's exit status (due to `pipefail`), and the script exits without updating `sessions.json`. Iter-1 behavior was to continue, log the warning, and write an entry with `CLAUDE_PID=""`.

This contradicts the fail-open intent stated in the surrounding comments.

Fix options:
- Restore `|| true` on line 53 only, OR
- Replace with explicit `if` guard:
  ```bash
  if ! CLAUDE_START="$(ps -o lstart= -p "$CLAUDE_PID" 2>/dev/null | sed ...)"; then
    CLAUDE_START=""
  fi
  ```
- Line 36 removal inside `find_claude_pid` is fine (errexit ignored in function called from `if`). Keep it removed there.

**R2 (Required, Low priority) — `TestFilterLive` `empty_start` case uses a live PID**

The new `empty_start` case uses `selfPID` with `PIDStart: ""`. `FilterLive` keeps it via the short-circuit `s.PIDStart == ""`. But the test does NOT cover the adversarial case: dead PID + empty PIDStart. Under iter-1 code those would be dropped; under iter-2 they are kept. Worth an explicit test case:
```go
"dead_with_empty_start": {WorkingDir: "/tmp/f", PID: deadPID, PIDStart: ""},
```
Expect kept. Prevents future regression where someone "optimizes" the short-circuit by requiring PID liveness.

**R3 (Low, nit) — `TestFilterLive` comment consistency**

The `(unknown-identity contract)` parenthetical appears only on `empty_start` assertion; mirror on `pre_upgrade` or drop on both.

### Summary

Iter-2 closed the H-1 correctness gap and tightened several comments. One potentially significant regression on hook line 53 (`|| true` removal at top-level errexit context) should be fixed. All other iter-2 changes look clean.

## pr-review-toolkit: code-reviewer

### Scope reviewed

Branch `544-tui-filter-stale-sessions` vs `origin/main` (3 commits ahead): `productivity-tui/hooks/update-session.sh`, `internal/app/app.go`, `internal/app/app_test.go`, `internal/session/session.go`, `internal/session/session_test.go`.

### Verification that iter-2 findings landed

Confirmed in working tree (`go test ./productivity-tui/...` passes):

- **H-1** session.go:43 — `s.PID == 0 || s.PIDStart == "" || isPIDLive(...)` matches the iter-2 plan. Godoc at session.go:27-39 documents the union without archaeology.
- **L-1** (`|| true` cleanup): update-session.sh:36 and :53 both use plain `$(...)` now. Line 43 (`tr -d ' '`) was already clean.
- **L-2** (PR body): body now leads with PID-walk approach.
- **L-3** (lstart-failure comment): update-session.sh:55 collapsed to one clause.
- **L-4** (jq schema comment): "Explicit schema: fields not listed in this object are dropped."
- **L-5** (nil/empty `FilterLive` test): session_test.go added.
- **A2, A3, A4, A5** all confirmed.

### Findings

No high-confidence issues (≥ 80) found.

**Minor observations verified as non-issues:**
- Hook `sed` trim vs Go `strings.TrimSpace` asymmetry is harmless — `$(...)` command-substitution strips trailing newlines at capture time.
- `len(nil)` as `make` hint is valid Go and returns a non-nil empty map.
- All iter-2 OOS items (O-1..O-16) remain correctly deferred.

Recommendation: terminate the code-quality loop.

## /simplify Output

### Code Reuse
- Only candidate is `psLStart`/`psLStartOf` duplication — already OOS (O-5).

### Code Quality
- H-1 fix at `session.go:43` symmetric with hook F1. Godoc lines 27-39 accurate.
- L-1 `|| true` removal consistent with existing line 42 style.
- L-3, L-4, A3, A4 comment tightening accurate. Pattern-list matches the 6 case arms 1:1.
- A2 removed upgrade-compat parenthetical — permanent-contract framing reads cleanly.
- Minor terminology drift: hook "unknown ancestor" vs godoc "unknown identity" vs test "unknown-identity contract". Not worth fixing.

### Verification
- `go test ./...` → pass. `go vet ./...` → clean. `gofmt -l .` → clean.

### Files Edited
None — iter-2 addressed every required finding surfaced by iter-1 and iter-2. No new issues in the iter-2 delta that warrant code changes. Remaining items (O-1 through O-16) remain OOS.

## pr-review-toolkit: comment-analyzer

Iteration 1 and 2 thoroughly addressed comment accuracy, completeness, and rationale. Comments have been revised to cross-reference `find_claude_pid` and `FilterLive`, document all 6 pattern-match shapes 1:1, split EPERM vs probe-failure semantics in `FilterLive` godoc, bound `hops < 20` and document pid=0 fallback inline, document `PIDStart` as opaque verbatim string ("do not reformat/parse"), tighten the "lstart query failed" comment (L-3), and tighten the jq comment (L-4).

No new substantive comment critiques to raise for iteration 3.

**Critical Issues:** None.
**Improvement Opportunities:** None that haven't already been raised and addressed.
**Recommended Removals:** None.

## pr-review-toolkit: pr-test-analyzer

Iter-1 added F14 (loadSessions → FilterLive wiring via `TestLoadSessionsFiltersDead`). Iter-2 added `TestFilterLiveEmptyAndNilInput` and an `empty_start` case in `TestFilterLive`. The behavioral coverage matrix for `FilterLive` is now complete for all in-scope branches: live-kept, dead-dropped, recycled-dropped, pre-upgrade-kept, empty-PIDStart-kept, nil input, empty input. The hook-to-reader JSON schema round-trips implicitly through `TestLoadSessionsFiltersDead`.

**Critical Gaps (rated 8-10):** None.
**Important Improvements (rated 5-7):** None that aren't already classified OOS.
**Test Quality:** DAMP, behavioral, resilient.

**Previously classified OOS — still OOS:**
- Shell-side tests for `find_claude_pid` — no bats/shunit2 harness (F18).
- Fail-open test paths for EPERM / ps-unavailable — require `isPIDLive` injection refactor (F20, O-16).
- `loadSessions` error-path short-circuit — covered transitively by `TestReadSessions_MalformedJSON` (O-2).
- Dead-PID test race — recycled-PID branch catches it (O-3).

**Conclusion:** No remaining in-scope critical test gaps. Recommend terminate test-coverage review loop.

## pr-review-toolkit: silent-failure-hunter

After a thorough review of the PR diff, the current source files, and both prior evaluation reports, I find no NEW silent failures, inadequate error handling, or inappropriate fallbacks introduced in iteration 3 that were not already identified and resolved or explicitly categorized as out-of-scope in iterations 1 and 2.

**Coverage verified:**
- `update-session.sh` — all three fail-open branches (find_claude_pid ps, lstart query, ancestor not found) now log to stderr per iter-1 F11. `|| true` redundancy removed in iter-2 L-1.
- `session.go` — `isPIDLive` logs stderr on non-ESRCH kill errors and ps errors per iter-1 F12. `FilterLive` includes symmetric consumer-side coherence guard `s.PIDStart == ""` per iter-2 H-1.
- `app.go` — `loadSessions` error short-circuit separates parse-error path from filter path. Sessions-wipe-on-error behavior is pre-existing (F23 / O-11).

**Previously-accepted-as-OOS items re-verified as still accepted** (O-7, O-8, O-9, O-10/O-12, F24).

**No new findings.** Recommendation: no further iteration required on error-handling grounds.

## pr-review-toolkit: type-design-analyzer

### Type: `Session`

**Invariants:**
- I1. Shape invariant (state file contract).
- I2. Identity coherence: `PID > 0 ⇔ PIDStart != ""` — enforced symmetrically at hook (F1) and consumer (H-1).
- I3. PID-start opacity — `PIDStart` is verbatim `ps -o lstart=` string, byte-for-byte compared.
- I4. Unknown-identity preservation — PID==0 or empty PIDStart is kept.
- I5. Liveness probe fails open.
- I6. Record ownership — hook writes, filter reads.

### Ratings

- Encapsulation: 4/10 — public record, no constructor, zero-value legal. H-1 is an admission the type cannot structurally enforce I2. Acceptable for JSON-mapped record but centralization is weak.
- Invariant Expression: 6/10 — `FilterLive` godoc is load-bearing; `PIDStart` field comment communicates I3 well. Coupling between PID and PIDStart only discoverable by reading the filter predicate.
- Invariant Usefulness: 8/10 — Invariants protect a real bug class (stale TUI rows after Claude crash). Aligned with feature requirement.
- Invariant Enforcement: 7/10 — Enforcement lives at hook (F1), filter (H-1), and jq schema. Symmetric defense is strong. Ceiling on public-JSON-struct enforcement.

### Strengths

- H-1 made coherence invariant symmetrically enforced at two real boundaries.
- Godoc on `FilterLive` carries contract prose type can't.
- `PIDStart` opacity correctly documented as durability concern.
- `TestFilterLiveEmptyAndNilInput` pins nil/empty contract.

### Concerns

- C-iter3-1. Coherence invariant is still duplicated prose across three locations (hook comment, godoc, tests). A one-line helper — `func (s Session) HasIdentity() bool { return s.PID != 0 && s.PIDStart != "" }` — would centralize the predicate. **Low-priority, low-complexity, not a correctness issue** — invariant-centralization hygiene.
- C-iter3-2. Omitempty asymmetry on `pid` vs `pid_start` — previously classified OOS (O-13), stays OOS.
- C-iter3-3. `isPIDLive(pid int, wantStart string)` signature — overlaps with `ProcessIdentity` suggestion, stays OOS (F19/O-14).
- C-iter3-4. No enforcement of `Idle`/`LastActivity` coherence — pre-existing, not a regression, OOS.

### Summary

**Iteration 3 status: no new required type-design findings.** H-1 closed the main invariant-enforcement gap. Remaining concerns (predicate centralization, omitempty asymmetry, identity-value refactor) were classified OOS in iters 1 and 2. The single optional suggestion (`HasIdentity()` helper) is hygiene, not an iteration-blocker.


---

# Code Quality Review - Iteration 3 Evaluation

**Reviewer panel:** /review, pr-review-toolkit:code-reviewer, /simplify,
pr-review-toolkit:comment-analyzer, pr-review-toolkit:pr-test-analyzer,
pr-review-toolkit:silent-failure-hunter, pr-review-toolkit:type-design-analyzer.

**Delta under review:** commit b287dfd (iter-2) + iter-2 PR body rewrite.

## Required — High Priority

- **R1. `|| true` removal on update-session.sh:53 breaks fail-open path.**
  Raised by /review. iter-2 L-1 removed `|| true` from two `ps` command
  substitutions. Line 36 (inside `find_claude_pid`) is safe because the
  function is called from `if pid=$(find_claude_pid); then` — errexit is
  suppressed inside functions called from conditional context. Line 53
  (`CLAUDE_START="$(ps -o lstart= ... | sed ...)"`) is inside the `then`
  body, where errexit is active. Combined with `set -o pipefail`, a `ps`
  failure (e.g. Claude PID exits in the race window between
  `find_claude_pid` and the lstart query) now exits the entire hook
  before lines 54-58 can run — so the `if [ -z "$CLAUDE_START" ]; then
  CLAUDE_PID=""; fi` coherence branch is unreachable. Verified
  empirically with `bash -c 'set -euo pipefail; if true; then
  X=$(ps -p 999999 2>/dev/null | sed "s/^ *//"); fi'` — exits 1.
  This is a correctness regression: the comment on line 54 documents
  fail-open intent but the code fails closed.
  Fix: restore `|| true` on line 53 only, or replace with an explicit
  `if ! CLAUDE_START="$(...)"; then CLAUDE_START=""; fi` guard.

## Required — Low Priority

- **R2. Add `dead_with_empty_start` TestFilterLive case.**
  Raised by /review. The existing `empty_start` case uses a live PID, so
  it keeps regardless of whether the filter tests `PIDStart == ""` first
  or last. Adding a case with a dead PID + empty PIDStart pins the H-1
  contract against a future regression where someone adds `isPIDLive`
  gating before the empty-start check.

- **R3. Mirror or drop "(unknown-identity contract)" parenthetical.**
  Raised by /review. TestFilterLive test table — consistency only.

None.

- **Optional `Session.HasIdentity()` helper.** Raised by
  type-design-analyzer as hygiene ("not required — this is hygiene, not
  correctness"). Prior iterations put `ProcessIdentity` sub-struct
  (F19/O-14) and related predicate refactors out of scope; this is a
  lighter version of the same suggestion. Stays OOS for consistency.

## False Positive

None.

## No new findings from

- pr-review-toolkit:code-reviewer — recommends terminate.
- /simplify — no file edits; recommends terminate.
- pr-review-toolkit:comment-analyzer — no new critiques post-iter-2.
- pr-review-toolkit:pr-test-analyzer — behavioral coverage matrix
  complete; recommends terminate.
- pr-review-toolkit:silent-failure-hunter — no new silent-failure
  findings; recommends terminate.

## Outcome

**Iterate.** R1 is a correctness regression that must be fixed before
terminating the loop. R2 and R3 will be implemented alongside R1 in
iter-4 since they are already classified Required-Low.

natb1:
# Code Quality Review - Iteration 4

Seven parallel reviewers examined PR #548. Per-reviewer findings summarized below.

## /review Skill

**Correctness:**
- `ps -o lstart=` locale sensitivity: neither the hook nor the Go probe sets `LC_ALL=C` before calling `ps`. If the user's locale changes between session creation and liveness probe, the string comparison fails and FilterLive incorrectly drops a live session (false negative). Recommendation: prepend `LC_ALL=C` to both `ps -o lstart=` invocations.
- `dead` test case PID recycling window: both `TestFilterLive` and `TestLoadSessionsFiltersDead` use `selfStart` as the `PIDStart` for the dead session. Using `mismatchedStart` instead would make the test robust against the (unlikely) PID reuse window.
- `|| true` on update-session.sh:53 is correctly preserved. jq schema replace is correct.

**Minor/Style:**
- `psLStart` duplicated between session_test.go and app_test.go as `psLStartOf` — acceptable in different packages.
- 20-hop bound in `find_claude_pid` is sound.

**Summary:** Two actionable findings: (1) locale sensitivity — add LC_ALL=C; (2) dead test case — use mismatchedStart. All iter-3 changes correctly in place.

---

## pr-review-toolkit: code-reviewer

No new high-confidence findings. All iter-3 changes land cleanly:
- `update-session.sh:53` `|| true` correctly preserved with load-bearing comment.
- `dead_empty_start` TestFilterLive case correctly pins H-1 contract for adversarial shape.
- `(unknown-identity contract)` parenthetical mirrors other cases.

Items reviewed and rejected: FilterLive ps-per-session efficiency (fine at N<20), claude/claude* false-match risk (minimal), platform portability (acceptable for local TUI), no named constant for hops<20 (comment justifies), TestLoadSessionsFiltersDead coverage scope (sufficient).

**Outcome: Recommends terminate.**

---

## /simplify Skill

No file edits. Code is clean from prior iterations. Noted duplicate psLStart/psLStartOf helper names across packages — unavoidable in Go, no change made.

**Outcome: Recommends terminate.**

---

## pr-review-toolkit: comment-analyzer

Three improvement opportunities:

1. **update-session.sh:31-33** — "fall through" description is imprecise. Behavior happens at call site via `${CLAUDE_PID:-0}`, not inside `find_claude_pid`. Suggest rewriting to describe call-site outcome.

2. **session.go:35-39** — EPERM described as explicit case but it is implicit. Only ESRCH is named; EPERM and all non-ESRCH errors take the same `fmt.Fprintf + return true` path. Suggest clarifying: "Non-ESRCH errors (including EPERM, which is a positive liveness signal) fail open."

3. **session_test.go:12** — `mismatchedStart` constant unexplained. Add brief comment: "arbitrary value; epoch zero is recognizable as a sentinel."

Positive: PIDStart field comment (lines 19-22), `|| true` load-bearing explanation, FilterLive two-paragraph godoc structure are exemplary.

---

## pr-review-toolkit: pr-test-analyzer

**Critical gaps:** None. All paths for the primary bug have test coverage.

**Hard-to-test paths:**
- `isPIDLive` fail-open branch on `ps` error after successful `kill(0)` — would require injectable executor.
- EPERM path — cannot trigger in same-user test context.

**Test quality:** `dead` session PID recycling race is pre-existing and acceptably rare. Six-case TestFilterLive table covers all FilterLive branches. TestFilterLiveEmptyAndNilInput pins nil/empty contract. TestLoadSessionsFiltersDead verifies wiring through real file path.

**Outcome: Recommends terminate.**

---

## pr-review-toolkit: silent-failure-hunter

**Finding 1 (MEDIUM):** `isPIDLive` (session.go:58-62) conflates two errors in the same fail-open log message: `*exec.ExitError` from ps (TOCTOU race — process vanished between kill probe and ps) vs non-ExitError (broken ps binary/env). A debugger cannot tell which occurred.

**Finding 2 (MEDIUM):** `find_claude_pid` (update-session.sh:34-46) emits the same log message whether the loop hit the 20-hop limit or exhausted the chain. The "add a new case pattern" hint is only relevant to the pattern-miss failure, not the hop-limit case.

**Finding 3 (MEDIUM):** jq failure in update-session.sh (lines 71-78) is silent under `set -euo pipefail`. State file is safe but the failure produces no user-visible diagnostic. Add EXIT trap emitting an error message.

**Finding 4 (LOW):** `TestFilterLive` dead-PID test (session_test.go:93-97) — PID recycling could cause the dead case to pass via start-time mismatch rather than ESRCH. Not a functional bug.

---

## pr-review-toolkit: type-design-analyzer

**Ratings:** Encapsulation 5/10, Invariant Expression 6/10, Invariant Usefulness 8/10, Invariant Enforcement 4/10.

**Concerns:**
1. No process identity type — pairing invariant (PID, PIDStart) invisible at construction. Incoherent `{PID: 42, PIDStart: ""}` is constructible and silently kept by FilterLive (intentional per contract, but not type-enforced).
2. `isPIDLive` writes to stderr directly — couples to global side-effect channel, makes function untestable.
3. ps-failure branch in `isPIDLive` untested.

**Recommendations:** Introduce `IsKnownProcess()` predicate or `ProcessIdentity` type. Thread `io.Writer` into `isPIDLive` for testability.

**Strengths:** PIDStart field comment exemplary. FilterLive nil-input returns non-nil empty map. Shell hook maintains coherence by zeroing CLAUDE_PID when lstart fails. omitempty tags enable correct backward-compat deserialization.


---

# Code Quality Review - Iteration 4 Evaluation

**Reviewer panel:** /review, pr-review-toolkit:code-reviewer, /simplify,
pr-review-toolkit:comment-analyzer, pr-review-toolkit:pr-test-analyzer,
pr-review-toolkit:silent-failure-hunter, pr-review-toolkit:type-design-analyzer.

**Delta under review:** commit a24aecf (iter-3) + merge of origin/main (54d0f33).

## Required — High Priority

None.

## Required — Low Priority

- **RL-1. Add `LC_ALL=C` to both `ps -o lstart=` invocations.**
  Raised by /review. Without it, a locale change between session creation
  (hook) and liveness probe (Go) causes string mismatch → live session
  dropped (false negative, wrong fail direction). Fix: prepend `LC_ALL=C`
  to `ps -o lstart=` on update-session.sh:53 and session.go:58.

- **RL-2. Use `mismatchedStart` for the `dead` test entry instead of `selfStart`.**
  Raised by /review, silent-failure-hunter. Current `dead` case uses
  `selfStart` as `PIDStart`. In a narrow PID-recycling window the dead PID
  could be reused by a process whose start-time coincidentally matches
  `selfStart`. Using `mismatchedStart` makes intent explicit and the test
  robust. Applies to both TestFilterLive and TestLoadSessionsFiltersDead.

- **RL-3. Distinguish hop-limit exhaustion from pattern-miss in `find_claude_pid` log.**
  Raised by silent-failure-hunter. Both failure modes emit the same message;
  the hint to add a new `case` pattern is only relevant to pattern-miss.
  Track whether `hops >= 20` and log a distinct message.

- **RL-4. Add EXIT trap for jq failure in update-session.sh.**
  Raised by silent-failure-hunter. jq failure kills the hook silently under
  set -euo pipefail. State file is safe (tmp cleaned up) but failure is
  undiagnosable. Add error-emitting EXIT trap replacing current rm-only trap.

- **RL-5. Comment: "fall through" in update-session.sh:31-33 is imprecise.**
  Raised by comment-analyzer. Rewrite to describe the call-site outcome
  (CLAUDE_PID stays empty; ${CLAUDE_PID:-0} substitutes 0 in jq call).

- **RL-6. Comment: EPERM described as explicit in session.go:35-39 but is implicit.**
  Raised by comment-analyzer. Only ESRCH is named; EPERM takes the same
  non-ESRCH fail-open path. Clarify: "Non-ESRCH errors (including EPERM,
  which is a positive liveness signal) fail open."

- **RL-7. Comment: `mismatchedStart` sentinel unexplained in session_test.go:12.**
  Raised by comment-analyzer. Add: "arbitrary value; epoch zero is a
  recognizable sentinel that will not match any real lstart."

- **RL-8. Comment: `claude\ *` backslash-space in case pattern unexplained.**
  Raised by /review. The backslash escapes the space in the shell case glob;
  unusual and confusing. Add brief inline explanation.

- **OOS-H1. Return `false` for ps TOCTOU race in `isPIDLive` instead of fail-open.**
  Raised by silent-failure-hunter. Changing the semantic for one sub-case of
  probe failure is a policy decision; fail-open was explicitly established
  across all prior iterations as intentional design.

- **OOS-L1. `IsKnownProcess()` predicate / `ProcessIdentity` nested type.**
  Already classified OOS in iter-3 as "Optional Session.HasIdentity() helper."

- **OOS-L2. `isPIDLive` io.Writer injection for testability.**
  High refactor effort for an unexported function.

- **OOS-L3. ps-failure branch test coverage.**
  Requires dependency injection or fake binary; depends on OOS-L2.

## False Positive

None. (code-reviewer and /simplify found no new issues.)

## Outcome

**Terminate.** Zero high-priority required findings. RL-1 through RL-8
are low-priority implementations to be completed in termination phase.

natb1:
# Code Quality Review - Complete ✓

All required findings from iterations 1–4 have been addressed. Zero high-priority required findings remain.

## Summary of iteration 4 changes (RL-1 through RL-8)

- **RL-1**: Added `LC_ALL=C` to both `ps -o lstart=` invocations (shell hook and Go) for locale-stable output
- **RL-2**: Dead test entries now use `mismatchedStart` sentinel instead of `selfStart` (session_test.go, app_test.go)
- **RL-3**: Hop-limit vs pattern-miss log messages are now distinguished with specific guidance, emitted inside `find_claude_pid`
- **RL-4**: EXIT trap now emits an error message on non-zero exit status before cleaning up the temp file
- **RL-5**: Fixed "fall through" comment to accurately describe the actual code flow via `${CLAUDE_PID:-0}`
- **RL-6**: Rewrote `FilterLive` EPERM godoc to lead with the general rule (non-ESRCH = fail open), then explain EPERM as a specific case
- **RL-7**: Added explanatory comment to the `mismatchedStart` constant
- **RL-8**: Added comment explaining the backslash-space shell glob escape in the `case` pattern

natb1:
# Security Review Evaluation — Iteration 1

**Reviewer**: Claude Code (via /security-review skill)
**Outcome**: No findings — terminating loop

## Findings

No security vulnerabilities meeting the HIGH-confidence threshold (>80% exploitability) were identified.

## Classification

No findings to classify.

## Determination

**Terminate** — zero required findings.

natb1:
# Security Review - Complete ✓

**Date**: 2026-04-18
**Branch**: 544-tui-filter-stale-sessions
**Reviewer**: Claude Code (via /security-review skill)
**Outcome**: No findings — loop terminated after 1 iteration

## Iterations

- Iteration 1: No findings — terminated immediately

## User Classification Decisions

No findings were reported by the security reviewer — no classifications required.

## Conclusion

Security review found no vulnerabilities. The PID-walk and start-time liveness check in update-session.sh and session.go present no injection surfaces or exploitable code paths. Loop terminates cleanly.

natb1:
Analyze (actions)	pass	48s	https://github.com/natb1/commons.systems/actions/runs/24612887688/job/71970189286	
Analyze (go)	pass	1m16s	https://github.com/natb1/commons.systems/actions/runs/24612887688/job/71970189283	
Analyze (python)	pass	52s	https://github.com/natb1/commons.systems/actions/runs/24612887688/job/71970189277	
CodeQL	pass	2s	https://github.com/natb1/commons.systems/runs/71970219550	
acceptance	pass	22s	https://github.com/natb1/commons.systems/actions/runs/24612888479/job/71970190444	
hook-tests	pass	9s	https://github.com/natb1/commons.systems/actions/runs/24612887771/job/71970188691	
lint	pass	13s	https://github.com/natb1/commons.systems/actions/runs/24612887771/job/71970188693	
preview-and-smoke	pass	20s	https://github.com/natb1/commons.systems/actions/runs/24612888479/job/71970190463	
unit-tests	pass	7s	https://github.com/natb1/commons.systems/actions/runs/24612887771/job/71970188689	


---

# Final Verify - Evaluation (Iteration 1)

**Date**: 2026-04-18
**Branch**: 544-tui-filter-stale-sessions
**PR**: #548

## Check Results

All 9 PR checks passed:

1. Analyze (actions) — pass
2. Analyze (go) — pass
3. Analyze (python) — pass
4. CodeQL — pass
5. acceptance — pass
6. hook-tests — pass
7. lint — pass
8. preview-and-smoke — pass
9. unit-tests — pass

## Outcome

All checks passed. Proceeding to terminate.

natb1:
# Final Verify - Complete ✓

**Date**: 2026-04-18
**Branch**: 544-tui-filter-stale-sessions

## Iterations

- Final iteration: All checks passed (no failures, no fixes required)

## Conclusion

All 9 PR checks passed on the first iteration. Proceeding to completion.

PR #577, opened 2026-04-30
productivity-tui: show wezterm tab number next to each session

natb1:
Closes #575

## Summary

- Hook captures `WEZTERM_PANE` env var and writes `wezterm_pane` field on each session record
- New `internal/wezterm` package parses `wezterm cli list --format json` and builds a paneID→1-based-tab-index map
- App tick loop queries the map; sessions render with `[N] ` prefix matching the wezterm tab bar number, padded for alignment
- Fail-open on errors: missing wezterm, closed panes, sessions outside wezterm all render without a prefix; first error per streak logs to stderr

## Test plan

- [x] `go test ./...` from `productivity-tui/` (session, wezterm, app packages)
- [ ] Operator smoke: start a Claude session inside a wezterm pane, observe `[N]` next to the working dir matching the tab bar number
- [ ] Move tab to a different position; observed number updates on next tick
- [ ] Close pane; row renders without tab number, no error
- [ ] Run outside wezterm; row renders without tab number


natb1:

Issue #565, opened 2026-04-18
Show worktree name instead of full path in productivity-tui session list


natb1:
## Context

The productivity-tui session list displays the full \`working_dir\` path for each session (e.g., \`/Users/n8/natb1/commons.systems/worktrees/42-fix-login-bugs\`). In practice, all sessions run inside worktrees, so the full path is redundant — the worktree name (\`42-fix-login-bugs\`) is the meaningful identifier.

## What to change

Replace the full \`working_dir\` path in the session list with the worktree name (basename of the path).

- Before: \`/Users/n8/natb1/commons.systems/worktrees/42-fix-login-bugs\`
- After: \`42-fix-login-bugs\`

The display change is in \`productivity-tui/internal/app/app.go\` where \`s.WorkingDir\` is rendered. Use \`filepath.Base(s.WorkingDir)\` for the displayed label.

## Acceptance criteria

- [ ] Active sessions display the worktree name (basename of \`WorkingDir\`) instead of the full path
- [ ] Idle sessions display the worktree name alongside the idle indicator
- [ ] Sessions whose \`WorkingDir\` resolves to a root or single-component path display the full path unchanged (no regression for edge cases)

PR #588, opened 2026-05-02
productivity-tui: show worktree name when launched from subdir

natb1:
Closes #580

## Summary
- `displayName()` walks up the path for a `worktrees/` ancestor and returns the segment immediately below it
- Falls back to `filepath.Base` when no `worktrees/` ancestor is present (no regression for non-worktree paths)
- Edge cases (`/`, `.`, empty, single-component) preserve the full-path behavior introduced in #565
- Table-driven `TestDisplayName` covers worktree root, subdirs, non-worktree paths, and edge cases

## Test plan
- [x] `go test ./internal/app/...` passes
- [x] `go build ./...` clean
- [ ] Manual smoke test: rebuild binary, start a Claude session from a worktree subdir, confirm session list shows the worktree name

### DIFF
--- a/productivity-tui/internal/app/app.go
+++ b/productivity-tui/internal/app/app.go
@@ -238,10 +238,19 @@ func tick() tea.Cmd {
 	})
 }
 
-// displayName renders the worktree name for a session's working dir. For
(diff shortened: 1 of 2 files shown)


natb1:

PR #608, opened 2026-05-02
budget/seeds/auth.ts: use shared TEST_USER from authutil

### LINKED ISSUE #601: budget/seeds/auth.ts: use shared TEST_USER from authutil
## Problem

`budget/seeds/auth.ts` defines an inline `testUser` object with `displayName: "natb1"` instead of using the shared `TEST_USER` from `authutil`. This drifts from the convention used by other apps and risks divergence (e.g., uid mismatch, displayName drift).

## Proposed migration

1. Replace the inline `testUser` object in `budget/seeds/auth.ts` with `import { TEST_USER } from "@commons-systems/authutil/seed"`.
2. Use `TEST_USER` directly in the seed flow.
3. If `displayName: "natb1"` is meaningful (vs. the shared default), update `TEST_USER` in `authutil` to match — but verify no other app depends on the current value first.

## Acceptance criteria

- [ ] `budget/seeds/auth.ts` no longer defines an inline test user; it imports `TEST_USER` from `authutil`.
- [ ] Seed run still succeeds in budget QA.
- [ ] `grep -n "natb1" budget/seeds/auth.ts` returns no matches (unless intentionally moved to `authutil`).
- [ ] `npx tsc --noEmit --project budget` passes.

## Effort / Impact

- Effort: S (single-file replacement)
- Impact: S (consistency win)

natb1:
Closes #601

### DIFF
--- a/budget/seeds/auth.ts
+++ b/budget/seeds/auth.ts
@@ -1,9 +1,3 @@
-import type { AuthUser } from "@commons-systems/authutil/seed";
+import { TEST_USER } from "@commons-systems/authutil/seed";
 
-const testUser: AuthUser = {
-  localId: "test-github-user",
-  email: "test@example.com",
-  displayName: "natb1",
-};
-
-export default testUser;
+export default TEST_USER;


natb1:

PR #609, opened 2026-05-02
budget: use firebaseutil deferAppCheckInit (#591)

natb1:
Closes #591

Replaces the inline App Check defer block in `budget/src/main.ts` with a call to the shared `deferAppCheckInit` helper from `@commons-systems/firebaseutil/defer-appcheck`. The inline block duplicated the shared helper exactly (same INTERACTION_EVENTS, same passive once-listeners, same `deferProgrammerError` + `logError` error path), and sibling apps (landing, fellspiral) already use the shared helper. Net: ~20 LOC removed.

## Test plan
- [x] `npx tsc --noEmit --project budget` passes
- [x] `npm run build --prefix budget` passes (vite build + tsx prerender)
- [ ] App Check still initializes in budget QA (preview / runtime verification)

### DIFF
--- a/budget/src/main.ts
+++ b/budget/src/main.ts
@@ -19,6 +19,7 @@ import { hydrateAccountsReconcile } from "./pages/accounts-reconcile-hydrate.js"
 import { mountHero } from "@commons-systems/style/hero";
 import { renderHero } from "./pages/hero.js";
 import { trackPageView, initAppCheck } from "./firebase.js";
+import { deferAppCheckInit } from "@commons-systems/firebaseutil/defer-appcheck";
 import { classifyError } from "@commons-systems/errorutil/classify";
 import { deferProgrammerError } from "@commons-systems/errorutil/defer";
 import { logError } from "@commons-systems/errorutil/log";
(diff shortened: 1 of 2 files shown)


natb1:
## Verify loop summary

All PR checks passed after one fix iteration.

### Iteration 1 — `unit-tests` failed

- **Check:** `unit-tests`
- **Reproduce command:** `.claude/skills/ref-pr-workflow/scripts/run-unit-tests.sh`
- **Failure:** `TypeError: Cannot read properties of undefined (reading 'then')` at `firebaseutil/src/defer-appcheck.ts:22`, triggered by the `window.dispatchEvent(new Event("click"))` step in `budget/test/main.test.ts` ("calls initAppCheck on first user interaction"). The mocked `initAppCheck()` returned `undefined` instead of a Promise, so `.then()` crashed.
- **Why not caught locally:** the suite's `beforeEach(() => vi.restoreAllMocks())` clears `.mockResolvedValue(undefined)` and reverts `vi.fn()` to its default no-op (which returns `undefined`). Before this PR, no test path ever called `initAppCheck()` after the restore — the inline App Check defer in `main.ts` registered a listener but never fired one in tests, so the broken mock initial-implementation went unnoticed. Migrating to `deferAppCheckInit` plus a new test that dispatches a `click` event was the first time the call actually executed in JSDOM.
- **Fix commit:** `37fb745` — replaces `vi.fn().mockResolvedValue(undefined)` with `vi.fn(() => Promise.resolve())` in two places in `budget/test/main.test.ts`. Putting the Promise return in the initial implementation makes it survive `vi.restoreAllMocks()`.

### Iteration 2 — all green

All 9 checks passed: `unit-tests`, `hook-tests`, `lint`, `acceptance`, `preview-and-smoke`, `CodeQL`, `Analyze (actions)`, `Analyze (go)`, `Analyze (python)`.

PR #610, opened 2026-05-02
Tighten module surface: drop unused export on file-local types (#602)

natb1:
Closes #602


natb1:

PR #613, opened 2026-05-02
fix stale worktree cleanup: prune, ps-grep, zsh-safe loops

natb1:
Closes #612

## Summary

Three independent bugs combined to make `cleanup_stale_worktree_processes` a silent no-op on the orphans it exists to kill. A 2026-05-02 audit found 8 such processes from 3 long-deleted worktrees, the oldest running 25 days.

- Run `git worktree prune` before reading `git worktree list --porcelain` so deleted-but-not-pruned worktrees are correctly recognized as stale (Bug 1).
- Replace sandbox-blocked `pgrep -f`/`pgrep -P` with `ps`-based helpers (`_pids_matching_arg`, `_pids_with_parent`), used by both `cleanup_stale_worktree_processes` and the sibling `kill_worktree_processes` (Bug 2).
- Replace `for pid in $pids` with `while IFS= read -r pid <<< "$pids"` in both helpers so PID lists word-split correctly under both bash and zsh (Bug 3).

## Test plan

- [x] `bash .claude/skills/ref-pr-workflow/scripts/test-pid-cleanup.sh` — 10 passed, 0 failed (8 pre-existing + 2 new tests: prune-before-list and parent/child tree-walk)
- [x] zsh source check confirms `while IFS` pattern present after source
- [x] CI: unit tests pass on PR

### DIFF
--- a/.claude/skills/ref-pr-workflow/scripts/lib.sh
+++ b/.claude/skills/ref-pr-workflow/scripts/lib.sh
@@ -360,15 +360,15 @@ kill_worktree_processes() {
   local exclude_pids
   exclude_pids=$(_ancestor_pids)
 
-  local pid
-  for pid in $pids; do
+  while IFS= read -r pid; do
+    [ -z "$pid" ] && continue
     if [[ "$exclude_pids" == *" $pid "* ]]; then
       continue
     fi
     kill -0 "$pid" 2>/dev/null || continue
     echo "Killing worktree process: PID $pid"
     kill_tree "$pid"
-  done
+  done <<< "$pids"
 }
 
 # Kill processes belonging to worktrees that no longer exist.
@@ -383,6 +383,10 @@ cleanup_stale_worktree_processes() {
   }
   # Resolve to absolute path; worktrees live as siblings of the git common dir
   worktree_root="$(cd "$git_common_dir/.." && pwd)/worktrees"
+
+  # Prune admin entries for worktrees whose directories no longer exist, so
+  # the subsequent list reflects only worktrees that are actually on disk.
+  git worktree prune 2>/dev/null || true
 
   # Build set of active worktree paths
   local active_paths=""
@@ -401,15 +405,16 @@ cleanup_stale_worktree_processes() {
   fi
 
   # Find PIDs with this repo's worktree root in their command args
-  local pids
-  pids=$(pgrep -f "$worktree_root/" 2>/dev/null) || true
+  local ps_output pids
+  ps_output=$(ps -axo pid=,args= 2>/dev/null) || true
+  pids=$(printf '%s\n' "$ps_output" | grep -F "$worktree_root/" | awk '{print $1}') || true
   [ -z "$pids" ] && return 0
 
   local exclude_pids
   exclude_pids=$(_ancestor_pids)
 
-  local pid
-  for pid in $pids; do
+  while IFS= read -r pid; do
+    [ -z "$pid" ] && continue
     [[ "$exclude_pids" == *" $pid "* ]] && continue
 
     # Extract the worktree path from this process's command line
@@ -426,7 +431,7 @@ cleanup_stale_worktree_processes() {
       echo "Stale worktree process: PID $pid (worktree: $wt_path)"
       kill_tree "$pid"
     fi
-  done
+  done <<< "$pids"
 }
 
 # Find N available TCP ports by binding to port 0 simultaneously.
--- a/.claude/skills/ref-pr-workflow/scripts/lib.sh
+++ b/.claude/skills/ref-pr-workflow/scripts/lib.sh
@@ -170,11 +170,12 @@ _ancestor_pids() {
 # Output: one PID per line, leaves first (children listed before their parent)
 _collect_tree_pids() {
   local pid="$1"
-  local children
+  local children child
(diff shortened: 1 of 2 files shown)


natb1:

natb1:
## Dispatch-QA — user-acceptance pass

Single user-acceptance QA pass for #612 (scripts-only — non-browser path). The PR's own
`test-pid-cleanup.sh` suite (10/10) covers prune-before-list and child-killing with a
*mocked* `git`; this pass verified the three behaviors that mocked, bash-only tests miss.

**Relevancy check (raised mid-pass):** confirmed #612 still applies. `cleanup_stale_worktree_processes`
is called by `run-qa-server.sh:16` and `run-acceptance-tests.sh:32` — both invoked by the
**new** dispatch workflow (`dispatch-qa` Step 3a, `dispatch-implement`). The `ref-pr-workflow`
orchestration skill is superseded by `/dispatch`, but `ref-pr-workflow/scripts/` is the shared
script library the `dispatch-*` skills delegate to. The fix lands on a live code path.

### Items walked — 3 PASS / 0 FAIL / 0 SKIP

| # | Item | Result |
|---|------|--------|
| 1 | zsh-safe loop iteration (Bug 3) | **PASS** (with in-session fix — see below) |
| 2 | sandbox-safe `ps` discovery (Bug 2) | **PASS** |
| 3 | prune-before-list with real git (Bug 1) | **PASS** |

- **Item 3** — real-git end-to-end demo: created a worktree, spawned orphans referencing its
  path, deleted the directory *without* pruning. Confirmed `git worktree list` still reported
  the deleted worktree as active (the exact Bug 1 precondition), then `cleanup_stale_worktree_processes`
  pruned, recognised it as inactive, and killed the orphans.
- **Item 2** — `pgrep` is fully removed from the cleanup path; the suite passes under the Claude
  Code sandbox using real `ps`. (Bug 2's `sysmond` failure is macOS-specific and not reproducible
  on this WSL/Linux host, but the `pgrep`→`ps` change is correct and harmless here.)

### Finding (reported, then fixed in-session at user direction)

**Item 1 — spurious zsh output.** The word-splitting fix works (2 orphans killed under zsh 5.9).
But `cleanup_stale_worktree_processes` printed spurious `cmdline=...` / `wt_path=...` lines under
zsh: `local cmdline` / `local wt_path` were declared *inside* the loop, and zsh displays an
already-declared parameter when `local` re-runs. Pre-existing, not in the PR's original diff,
cosmetic — but it surfaced in exactly the zsh-debugging scenario Bug 3 targets.

Fixed in **fe623ea** — both `local` declarations hoisted above the loop (matching `kill_tree`
and `_collect_tree_pids`). Re-verified: zsh end-to-end run kills orphans with clean output;
`test-pid-cleanup.sh` 10/10.

### Suggested follow-up (not blocking)

`test-pid-cleanup.sh` is bash-only (`#!/usr/bin/env bash`), so it would not have caught the
loop-body-`local` class of bug. A zsh regression assertion would close that gap.

---
*Posted by `/dispatch-qa` — single user-acceptance pass, no wiggum-loop.*

natb1:
## Code Review — PR #613

**Verdict: Approve.** Three real bugs, each fixed correctly and covered by a test. CI is green; the local suite passes 10/10.

### Overview
`cleanup_stale_worktree_processes` was a silent no-op against the orphan processes it exists to kill. This PR fixes three independent bugs and converges the helper code with the sibling `kill_worktree_processes`.

### Correctness — all three fixes verified
- **Bug 1 (prune before list):** `git worktree list --porcelain` still reports deleted-but-unpruned worktrees, so their paths landed in `active_paths` and their processes were spared. Adding `git worktree prune` first is the right fix, and the new mock-`git` test fails without it.
- **Bug 2 (sandbox-blocked `pgrep`):** `_pids_matching_arg` / `_pids_with_parent` replace `pgrep -f`/`pgrep -P` with `ps`-based scans. Matching the `args` column only (vs. the old whole-line `grep -F`) is a strict improvement — though for these two call sites the needle is always a path, so there was no live PID-column collision.
- **Bug 3 (zsh word-splitting):** confirmed — under zsh, `for p in $pids` iterates **once** over the whole multi-line string; `while IFS= read -r p <<< "$pids"` iterates per line. The `[ -z "$x" ] && continue` guard correctly absorbs the trailing empty line a here-string adds.

### Code quality
- Helpers are well-factored and the comments explain the *why* (sandbox IPC block, zsh `local`-in-loop display quirk) rather than the *what*.
- Hoisting `local cmdline wt_path` out of the loop preserves the `cmdline=$(...) || continue` exit code (declaring + assigning on one line would mask it) — no regression.
- `|| true` on the `ps | ...` pipelines correctly guards against `pipefail` when `lib.sh` is sourced under `set -euo pipefail`.
- `lib.sh` already required bash/zsh (`[[ ]]`, `declare -A`); `<<<` adds no new constraint.

### Tests
Two new tests — prune-before-list and parent/child tree-walk — both well-constructed with mock `git` functions. Full suite: **10 passed, 0 failed**.

### Minor (non-blocking)
- PR body test plan is stale: it says "9 passed" but the suite now runs **10** (a second test, "kills child processes", was added after the body was written), and the `CI: unit tests pass on PR` checkbox is still unchecked though all checks are green.

No blocking issues.

PR #614, opened 2026-05-02
scaffold: collapse four MarshalJSON/UnmarshalJSON pairs into generic helpers

natb1:
Closes #597

Replaces four near-identical `MarshalJSON`/`UnmarshalJSON` pairs in `scaffolding/firebase/internal/scaffold/firebase.go` (on `HostingEntry`, `FirebaseConfig`, `FirebaseRC`, `PackageJSON`) with two generic helpers: `unmarshalWithExtra[T]` and `marshalWithExtra[T]`. Each call site collapses to ~6 lines that document only the per-type known JSON keys. Also removes the now-misleading comment that claimed generics could not abstract this pattern.

## Test plan

- [x] `go build ./...` passes from `scaffolding/firebase/`
- [x] `go test ./...` passes from `scaffolding/firebase/` (28 pre-existing + 3 new)
- [x] New `TestUnmarshalWithExtra` covers known-only / extra-only / mixed / malformed
- [x] New `TestMarshalWithExtra` covers no-extra / with-extra / mixed
- [x] New `TestFirebaseConfigPreservesExtraKeys` is a round-trip test exercising the helpers end-to-end
- [x] Existing round-trip tests for `FirebaseConfig`, `FirebaseRC`, `PackageJSON`, and `HostingEntry` still pass through the new helpers

### DIFF
--- a/scaffolding/firebase/internal/scaffold/firebase.go
+++ b/scaffolding/firebase/internal/scaffold/firebase.go
@@ -58,45 +58,61 @@ func (h HostingEntry) Validate() error {
 	return nil
 }
 
-func (h *HostingEntry) UnmarshalJSON(data []byte) error {
-	type Alias HostingEntry
-	var alias Alias
-	if err := json.Unmarshal(data, &alias); err != nil {
+// unmarshalWithExtra unmarshals data into target (which must be an alias type
+// without its own UnmarshalJSON to avoid recursion), then captures any
+// remaining JSON keys (those not in knownKeys) into *extra.
+func unmarshalWithExtra[T any](data []byte, target *T, extra *map[string]json.RawMessage, knownKeys ...string) error {
+	if err := json.Unmarshal(data, target); err != nil {
 		return err
 	}
-	*h = HostingEntry(alias)
-
 	var raw map[string]json.RawMessage
 	if err := json.Unmarshal(data, &raw); err != nil {
 		return err
 	}
-	delete(raw, "target")
-	delete(raw, "public")
-	delete(raw, "ignore")
-	delete(raw, "rewrites")
+	for _, k := range knownKeys {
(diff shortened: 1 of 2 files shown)


natb1:

Non-browser path (backend Go refactor). User requested "skip all" — both planned items recorded as SKIP without user-confirmed walkthrough.

**Counts:** 2 items planned, 0 PASS, 0 FAIL, 2 SKIP.

### Items

1. **Code review: helper signatures and call-site shape** — SKIP
2. **End-to-end round-trip via helpers against live repo configs** — SKIP

### Pre-walkthrough Claude observations (not user-confirmed)

- `unmarshalWithExtra[T any]` (firebase.go:64) and `marshalWithExtra[T any]` (firebase.go:83) exist with doc comments.
- All four pairs collapsed: `HostingEntry` (101/113), `FirebaseConfig` (129/141), `FirebaseRC` (157/169), `PackageJSON` (343/355).
- The misleading comment block at the original L113-116 is removed.
- Round-trip against the live `firebase.json` / `.firebaserc` / `package.json` in the worktree preserves all keys handled by the four in-scope types — including unknown top-level keys (`emulators`, `functions`, `storage`, `devDependencies`) and per-`HostingEntry` `headers`.

### Out-of-scope finding (pre-existing, not a #597 regression)

`RewriteEntry` (`firebase.go:16`) has no unknown-fields handling and silently drops keys like `"function"` for Cloud Functions rewrites. The worktree's `firebase.json` contains a `fellspiral` `feedProxy` function rewrite that would be lost if the scaffold tool round-tripped the file. `RewriteEntry`'s definition is byte-identical between `origin/main` and this branch, so this PR did not introduce the bug. Worth a follow-up issue.

### Bugs reported by user

None — user skipped the walkthrough.

PR #622, opened 2026-05-02
Dedupe redundant programmer-error rethrows; defer at async-callback sites

natb1:
Closes #593

Two programmer-error handling cleanups in application catch blocks.

## 1. Dead rethrows removed

`budget/src/pages/accounts-reconcile-hydrate.ts` had three
`if (classifyError(error) === "programmer") throw error;` lines, each
immediately after a `deferProgrammerError(error)` early-return. The defer call
already handles programmer errors and returns early, so each rethrow was
unreachable. Removed all three, plus the now-unused `classifyError` import.

## 2. Async-callback rethrows converted to `deferProgrammerError`

At the async-callback catch sites listed in #593 — across `budget/src/main.ts`,
`budget/src/pages/home-hydrate.ts`, `audio/src/player.ts`,
`audio/src/pages/home.ts`, `blog/src/components/info-panel.ts`,
`landing/src/main.ts`, and `fellspiral/src/main.ts` — a raw `throw` inside a
void-discarded promise's catch becomes an unhandled promise rejection that may
not surface reliably. `deferProgrammerError` re-throws via `setTimeout` so the
global error handler always sees it.

### Behavior change in `loadPosts`

The `loadPosts` catch in `landing/src/main.ts` and `fellspiral/src/main.ts` —
the two blog frontends, which have parallel structure — is the non-mechanical
conversion. Previously a programmer error inside `loadPosts` rethrew, failing
the surrounding render. Now it defers via `setTimeout` and the function returns
the generic "Could not load posts" fallback HTML. The deferred error still
reaches `window.onerror`, but the page paints the fallback instead of failing
render. Each handler classifies the error once, branches on the kind, and names
the shared fallback string.

Out of scope (per #593): sync sites that throw to a sync caller; inner catches
that rethrow up to an already-converted outer catch (`budget/src/crypto.ts`,
the `blog-roll` strategies); the build-time feed-fetch Vite plugin; and the
behavior of the error utilities themselves.

---

Re-scope note: this branch previously also bundled an unrelated "persistent
dispatch stage state" change against the `dispatch/bin/` orchestrator. That
orchestrator was removed by #637 (the `/dispatch` skill restructure), so the
bundled change was obsolete and has been dropped. The branch is now scoped to
#593 only and rebased onto current `main`.

### DIFF
--- a/audio/src/pages/home.ts
+++ b/audio/src/pages/home.ts
@@ -1,6 +1,6 @@
 import { escapeHtml } from "@commons-systems/htmlutil";
 import { logError } from "@commons-systems/errorutil/log";
-import { classifyError } from "@commons-systems/errorutil/classify";
+import { deferProgrammerError } from "@commons-systems/errorutil/defer";
 import type { User } from "../auth.js";
 import { DataIntegrityError } from "@commons-systems/firestoreutil/errors";
 import { getPublicMedia, getAllAccessibleMedia } from "../firestore.js";
@@ -52,8 +52,7 @@ export async function renderHome(user: User | null): Promise<string> {
     mediaHtml = renderMediaList(items);
   } catch (error) {
     if (error instanceof DataIntegrityError) throw error;
-    if (classifyError(error) === "programmer") throw error;
-    logError(error, { operation: "load-media" });
+    if (!deferProgrammerError(error)) logError(error, { operation: "load-media" });
     mediaHtml = '<p id="media-error">Could not load audio library.</p>';
   }
 
--- a/audio/src/player.ts
+++ b/audio/src/player.ts
@@ -1,5 +1,5 @@
 import { escapeHtml } from "@commons-systems/htmlutil";
-import { classifyError } from "@commons-systems/errorutil/classify";
+import { deferProgrammerError } from "@commons-systems/errorutil/defer";
 import { logError } from "@commons-systems/errorutil/log";
 import { resolveAudioSource } from "./storage.js";
 import type { AudioItem } from "./types.js";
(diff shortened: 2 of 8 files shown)


natb1:

natb1:
## dispatch-qa #593 — deferProgrammerError migration

**App walked:** `budget` (touches 5/14 listed sites + all three
`accounts-reconcile-hydrate.ts` dead-rethrow removals)

**Results:** 6 PASS · 0 FAIL · 2 SKIP

| # | Step | Result |
|---|---|---|
| 1 | Initial app load (`/`) | PASS |
| 2 | File upload (golden path) | PASS |
| 3 | Export (encrypted backup) | PASS — 3.3 KB JSON download |
| 4 | Clear data → seed state | PASS |
| 5 | Reconcile — confirm match | SKIP — fixture has no matched rows (txns 336d outside period tolerance) |
| 6 | Reconcile — classify + note | PASS — async persistence, no errors |
| 7 | Home — period totals after reimbursement change | SKIP — fixture renders no `.budget-balance` on `/` |
| 8 | Fault injection (`setTimeout(()=>{throw new TypeError(...)},0)`) | PASS — surfaces via global `error` + `unhandledrejection` + console |

**Smoke baseline:** 440 passed · 21 pre-existing failures
(`security-headers` / `cache-headers` / `prerender` — all require a
production build) · 4 skipped · 3 did not run. Matches expected baseline from
the resume plan.

**SKIP rationale:** Migration sites at `accounts-reconcile-hydrate.ts:79-83`
and `home-hydrate.ts:163-169` were verified statically (dead rethrows removed
on lines 83/125/150; `deferProgrammerError` only invocation on each error
path). The shipped test fixture (`budget/e2e/fixtures/test-budget.json`)
doesn't expose the DOM elements those handlers attach to; richer fixture data
would be needed to exercise the happy paths interactively. The fault
injection in #8 validates the broader premise: async-callback throws are not
silently swallowed.

**Walkthrough recording:** `tmp/dispatch-qa-walkthrough-593.gif` (34 frames,
5.0 MB, downloaded by Chrome — see project Downloads folder).

**Infra note:** First smoke run failed with chromium 1208/1217 drift because
the worktree's direnv cache had a stale `PLAYWRIGHT_BROWSERS_PATH`. Running
the smoke under a fresh `nix develop --command` fetched the matching
browser store and ran clean. Recommend deleting `.direnv/` in any worktree
that pre-dates the playwright sync fix.

natb1:

PR #634, opened 2026-05-17
Replace Playwright storageState seeding with an addInitScript fixture

natb1:
Closes #632

## Problem

Playwright smoke tests flaked during browser context setup. `config/playwright.js` seeded the `analytics_traffic_type=internal` localStorage key via `storageState.origins`. Because localStorage is origin-scoped, Playwright honored `storageState` by navigating to the deployed site during every `browser.newContext` — waiting for the full `load` event — purely to write one key. With a fresh context per test, the suite performed that heavy navigation once per spec on top of each test's own `page.goto`. One context-setup navigation against a slow cold URL exceeded the 30s test timeout.

## Fix

Seed the flag with `context.addInitScript` instead. An init script runs in the page before any page script, with no navigation of its own — leaving each test's existing `page.goto` as the only navigation. This makes the `browser.newContext … navigating to …` failure mode structurally impossible and roughly halves the suite's total navigations. No retry logic is added — the goal is for the failure mode to be impossible, not retried.

## Changes

- **`config/playwright-test.ts`** (new) — extends Playwright's base `test`, overriding the `context` fixture to register a `context.addInitScript` that sets the `analytics_traffic_type` localStorage key to `internal` (guarded with try/catch for opaque origins like the initial `about:blank`). Re-exports `expect` and the `Page`/`Download` types.
- **`config/package.json`** — adds the `"./playwright-test"` export entry.
- **`config/playwright.js`** — removes the `storageState` block.
- **85 test-defining files repointed** — app `e2e/` specs (audio, budget, fellspiral, landing, print), `authutil/e2e/`, the two `config/*-smoke.ts` helpers, and the scaffolding template e2e specs now import `test`/`expect` from the new fixture module instead of `@playwright/test`.

### Test-stability follow-ons

Dropping the `storageState` context navigation shifts e2e timing, which surfaced two latent flakes. Both are fixed here rather than in a separate PR because they only reproduce once that navigation is gone:

- **`style/layout.css`** (production CSS change to the shared `style` package) — pins `.content-grid > main` to its grid column start (`max-inline-size: none; margin-inline: 0`), overriding missing.css's self-centering of a top-level `main`. When the grid column resolves wider than `--eff-line-length`, `main` previously capped at `--eff-line-length` and auto-centered, drifting its left edge right of the header's — an environment-specific misalignment under `font-display: optional` cold loads. The override makes header/main left-edge alignment deterministic, which the fellspiral/landing `info-panel.spec.ts` alignment assertions now rely on (the prior font-load wait is dropped).
- **`budget/e2e/home-scroll.spec.ts`** — extracts `scrollUntilMoreRows` and `expectVisibleRowsSettle` helpers and hardens the category/budget scroll-filter tests against an IntersectionObserver race: the sentinel is re-scrolled into view on every poll iteration so the observer keeps firing, and `expect.poll` covers the transient window before scroll-loaded raw-HTML rows are filtered.

### DIFF
--- a/audio/e2e/auth.spec.ts
+++ b/audio/e2e/auth.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "@commons-systems/config/playwright-test";
 import { signIn } from "@commons-systems/authutil/e2e/sign-in";
 
 test.describe("auth", () => {
--- a/audio/e2e/cache.spec.ts
+++ b/audio/e2e/cache.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "@commons-systems/config/playwright-test";
 
 test.describe("audio cache", () => {
   test("cache and replay from IndexedDB", async ({ page }) => {
--- a/audio/e2e/home.spec.ts
+++ b/audio/e2e/home.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "@commons-systems/config/playwright-test";
 import { signIn } from "@commons-systems/authutil/e2e/sign-in";
 
 test.describe("home page content", () => {
--- a/audio/e2e/meta-description.spec.ts
+++ b/audio/e2e/meta-description.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "@commons-systems/config/playwright-test";
 
 test.describe("meta description", () => {
   test("home page has meta description @smoke", async ({ page }) => {
--- a/audio/e2e/navigation.spec.ts
+++ b/audio/e2e/navigation.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "@commons-systems/config/playwright-test";
 
 test.describe("navigation", () => {
   test("page loads without JS errors @smoke", async ({ page }) => {
--- a/audio/e2e/responsive.spec.ts
+++ b/audio/e2e/responsive.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "@commons-systems/config/playwright-test";
 
 test.describe("responsive layout", () => {
   test("main content is visible", async ({ page }) => {
--- a/audio/e2e/robots-txt.spec.ts
+++ b/audio/e2e/robots-txt.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "@commons-systems/config/playwright-test";
 
 test.describe("robots.txt", () => {
   test("GET /robots.txt returns valid robots.txt, not HTML", async ({
--- a/audio/e2e/theme.spec.ts
+++ b/audio/e2e/theme.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect, Page } from "@playwright/test";
+import { test, expect, type Page } from "-systems/config/playwright-test";
 
 async function expectDarkBackground(page: Page) {
   const bg = await page.evaluate(() =>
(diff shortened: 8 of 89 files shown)


natb1:
## Verify loop summary — PR #634

All 10 PR checks pass. The verify loop took four iterations; below is each failure, why local checks did not catch it, and the fix.

### Iteration 1 — mangled import specifier

**Failed:** `acceptance`, `preview-and-smoke` — `Cannot find package '-systems'` from every app's `theme.spec.ts`.

**Cause:** The Unit 2 bulk import-line swap dropped the `@commons` scope prefix in the 6 `theme.spec.ts` files, leaving `"-systems/config/playwright-test"`.

**Why not caught locally:** Unit 2 ran `tsc --noEmit --project <app>` and it passed — app `tsconfig`s exclude `e2e/` directories, so the broken specifier in an `e2e/` spec was never type-checked. The error only surfaces when Playwright actually loads the spec.

**Fix:** `fdf528d` — corrected the specifier to `@commons-systems/config/playwright-test` in all 6 files. Verified with a local `audio` acceptance run.

### Iterations 2–3 — two pre-existing failures unmasked

After the import fix, `acceptance` failed on `fellspiral/e2e/info-panel.spec.ts` (header alignment, offset `12.1875` vs `≤2`). A CI re-run passed fellspiral but failed `budget/e2e/home-scroll.spec.ts` instead. Both were non-reproducible locally (3/3 passes each). A deep diagnosis found two distinct root causes:

**Failure A — fellspiral header misalignment — a real layout bug, not a flake.**
Commit `9cfc420` added `.content-grid > main { max-inline-size: none; margin-inline: 0 }` to the shared `style/layout.css` to stop `missing.css`'s `main` rule from re-centering `<main>` within its grid column. The revert commit `c859584` ("Revert Merge …") **accidentally deleted that block** — so `main` has carried the regression since. Under `font-display: optional`, a cold first load uses fallback-font metrics, the grid column resolves wider than `--eff-line-length`, and `<main>` auto-centers ~12px right of the header.

**Why not caught locally:** `storageState`'s per-context navigation had been incidentally warming the browser font cache, so the test always ran in an unrealistically "warm" state and never saw the bug. Replacing `storageState` with `addInitScript` (the point of this PR) removed that warming — the test now runs cold, as real users do, and correctly catches the regression. Local runs still have warm caches, so it passes there; only CI's cold first navigation triggers the fallback-font layout.

**Failure B — budget infinite-scroll — a genuine CI-load timing flake.**
The scroll-loading tests scrolled the `IntersectionObserver` sentinel only once, then polled the row count. Each scroll batch is an async IndexedDB fetch + DOM append; after a batch appends, the sentinel moves and the observer needs the sentinel re-scrolled into view to fire again.

**Why not caught locally:** locally the per-batch IDB fetch + append is fast enough that one scroll loads enough rows within the timeout; only CI's resource contention (emulators + Vite + the full suite) slows batches enough to time out.

**Fixes:**
- `140f2e6` — restored the `.content-grid > main` block in `style/layout.css` (CSS identical to `9cfc420`'s; the loss in `c859584` was unintended). Reproduced the bug locally by blocking web-font requests (offset `6.7px` → `0` after the fix); `fellspiral`/`print`/`landing` acceptance all pass. The now-deterministic `info-panel` test was simplified back to a direct assertion.
- `2ad3b56` — the budget scroll-loading tests now re-scroll the sentinel inside the retry poll (30s timeout), mirroring the known-good `all transactions reachable` sibling test. No assertions weakened.

### Iteration 4 — all checks pass

`acceptance` (7m54s) and `preview-and-smoke` (5m9s) both green across every app, plus the 8 other checks.

### Note for reviewers

`140f2e6` restores a layout fix that was unintentionally reverted on `main` by `c859584`. That regression is independent of this PR's scope (issue #632) but was surfaced by it; restoring the known-good `9cfc420` rule is the correct fix and also repairs `main`.

natb1:
## Verify loop summary

All 10 CI checks green: Analyze (actions/go/python), CodeQL, acceptance,
hook-tests, lint, playwright-version-sync, preview-and-smoke, unit-tests.

PR #637, opened 2026-05-17
Restructure the dispatch skill system

natb1:
Deletes the legacy bash dispatcher so `/dispatch` is the sole orchestrator;
splits `/dispatch-implement` into `/plan-implement` + `/verify-pr` with shared
mechanics in `/implement-unit`; simplifies context recovery; and adds a
`waiting` phase so `/dispatch` skips PRs whose CI is still running.

## Commits

1. **Recovery + split foundation** — create `/implement-unit` (shared
   build-one-unit procedure: model heuristic, implementation subagent,
   commit/merge/push recovery) and `/verify-pr` (single-pass CI-failure fix);
   rename `/dispatch-implement` to `/plan-implement` and reduce it to plan →
   build → open draft PR. Repoint `restore-dispatch-skill.sh` onto a fixed
   `tmp/dispatch-worktree` marker created by `/dispatch`.
2. **`waiting` phase** — `dispatch-phase` resolves a draft PR with in-progress
   CI to a new `waiting` phase (empty/pending/unrecognized rollups); FAILING is
   evaluated before PENDING so a mixed rollup still resolves to `verify`.
   `dispatch-select-target` excludes `waiting` PRs from the queue.
3. **Legacy dispatcher removal** — delete the `dispatch/` directory and scrub
   its plumbing (`phase_signal`, the local-cache mirror, `DISPATCH_ISSUE_NUM`,
   the `dispatch/bin` regex alternative and permission).

Closes #630

PR #646, opened 2026-05-18
QA: dispatch-qa walkthroughs use public seed data only

natb1:
Closes #639

## What

Adds a `## QA data policy` section to `.claude/skills/dispatch-qa/SKILL.md`
encoding the "public data only" QA policy, plus brief cross-references from
Step 2 (write the QA plan) and Step 3a (start the QA server).

The policy section states:

- The QA pass covers **public data only** — documents present in both the QA
  server and production.
- Identify public seed items from `<app>/seeds/firestore.ts`: documents in
  collection blocks **without** `testOnly: true`. The QA server seeds exactly
  these; build the walkthrough around them.
- Never run the QA server or any seed step with `SEED_TEST_ONLY=true`, and never
  re-seed `testOnly` collections — that breaks QA/prod parity. `testOnly` data is
  for the Playwright acceptance tests, which CI's `acceptance` job already runs.
- Auth-gated / private-data flows are out of scope for the automated walkthrough;
  the user verifies those in a preview deployment, avoided where possible.
- When behavior is only reachable via `testOnly` or private data, note the
  limitation in the QA plan and defer that coverage to the acceptance tests.

## Why

During QA of #600, `/dispatch-qa` built a walkthrough around
`/view/test-image-archive` — a `testOnly` seed document the QA server does not
seed — then worked around the gap by re-seeding with `SEED_TEST_ONLY=true`. That
breaks parity between QA and production. This policy prevents the recurrence.

## On the `/verify` acceptance criterion

Issue #639 also named `/verify` ("TBD; defined by the dispatch restructure in
#630"). #630 has since closed (landed as #637) — the resulting `verify` phase is
`/verify-pr`, a skill that reproduces and fixes failed CI checks. It has no QA
walkthrough and no seeding step, so the public-data policy has nothing to attach
to there. That criterion is **moot / N/A**. The issue anticipated this: "The
`/dispatch-qa` portion could land independently if preferred."

## Verification

Documentation-only change — no build or unit test applies. `git diff` is confined
to `.claude/skills/dispatch-qa/SKILL.md`.

### DIFF
--- a/.claude/skills/dispatch-qa/SKILL.md
+++ b/.claude/skills/dispatch-qa/SKILL.md
@@ -104,7 +104,7 @@ The QA pass covers **public data only** — documents present in both the QA ser
 
    The plan must NOT duplicate things already verified by unit tests, lint, or type-check. Do not include items like "type-check passes", "no lint errors", "renders without crashing", "no console errors on load". Those are CI's job.
 
-   Build the plan around public seed data (see [QA data policy](#qa-data-policy)). When target behavior is only reachable via `testOnly` or private data, do not create a walkthrough item for it — note the limitation in the plan and defer that coverage to the automated acceptance tests.
+   Build the plan around public seed data (see [QA data policy](#qa-data-policy)). When target behavior is only reachable via `testOnly` or private data, do not create a walkthrough item for it — note the limitation in the plan instead.
 
 3. **Browser feature path.** Skip to Step 4 if the non-browser path applies.
 
--- a/.claude/skills/dispatch-qa/SKILL.md
+++ b/.claude/skills/dispatch-qa/SKILL.md
@@ -24,6 +24,15 @@ This skill is a **single user-acceptance pass**. It does not iterate, does not e
 - **No iteration.** When the user reports a bug, record it and move to the next item. Do not attempt fixes.
 - **No code changes, commits, or pushes.**
 
+## QA data policy
+
+The QA pass covers **public data only** — documents present in both the QA server and production.
+
+- **Identify public seed items** from `<app>/seeds/firestore.ts`: documents in collection blocks **without** `testOnly: true`. The QA server seeds exactly these. Build the walkthrough around them.
+- **Never** run the QA server or any seed step with `SEED_TEST_ONLY=true`, and never re-seed `testOnly` collections by other means — that breaks QA/prod parity, letting QA pass against data absent from production. `testOnly` data exists for the Playwright acceptance tests, which CI's `acceptance` job already runs.
+- **Auth-gated and private-data flows are out of scope** for the automated walkthrough. If one must be verified, the user does so in a preview deployment — avoid this where possible.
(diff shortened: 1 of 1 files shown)


natb1:
## Dispatch QA — Issue #639

**Non-browser path** — PR #646 changes only `.claude/skills/dispatch-qa/SKILL.md`
(skill-instruction documentation). No QA server or browser walkthrough.

### Result: PASS (clean) — 1 finding recorded as follow-up

| # | Item | Result |
|---|---|---|
| 1 | Public-data-only policy stated; public seed items identified from non-`testOnly` blocks of `<app>/seeds/firestore.ts` | PASS |
| 2 | Forbids running the QA server / any seed step with `SEED_TEST_ONLY=true` or re-seeding `testOnly` collections | PASS |
| 3 | Auth-gated / private-data flows out of scope for the walkthrough; deferred to user preview-deployment QA | PASS |
| 4 | QA-plan guidance notes public-data-unreachable behavior and defers it to automated acceptance tests | PASS |
| 5 | Consistency of automated validation between `/dispatch-qa` and `/verify-pr` | FINDING (non-blocking) |

**Counts:** 4 PASS · 0 FAIL · 0 SKIP · 1 finding

Items 1–4 verified against the diff and `print/seeds/firestore.ts` (public media
block: `gutenberg-3296`, `plato-phaedrus`, `plato-republic`; `testOnly` blocks:
`test-image-archive`, `test-private-item`, `test-group`).

### Finding — `/dispatch-qa` Step 3c smoke check is inconsistent with `/verify-pr` / CI

Acceptance criterion 5 (`/verify` carries the same instructions) was written before
#630 defined `/verify`. The delivered `/verify-pr` is a **CI-failure-fix** skill, not
a QA-walkthrough skill, so the literal criterion does not apply. The genuine
consistency concern surfaced during QA:

`/dispatch-qa` **Step 3c** runs the acceptance suite as a smoke check via
`run-acceptance-tests.sh <app> <qa-server-url>`. Passing a URL takes the script's
`EXTERNAL_BASE_URL` branch, which **skips emulator seeding** and runs Playwright
against the QA server — and the QA server seeds **public data only**. Meanwhile
`/verify-pr` Step 3 and CI's `acceptance` job run the same suite via
`run-acceptance-tests.sh` **without** a URL, which seeds `SEED_TEST_ONLY=true`.

⇒ The same Playwright acceptance suite runs against two different data sets. A
`testOnly`-dependent test (e.g. print's `/view/test-image-archive` spread-mode
coverage) passes in CI and in `/verify-pr` but would fail `/dispatch-qa`'s Step 3c
smoke check.

PR #646 correctly makes the **walkthrough** public-data-only — and hardens that
seeding into policy — but does not reconcile Step 3c. **Recorded as non-blocking
follow-up** (per QA decision): a separate change should either route Step 3c through
the no-URL `SEED_TEST_ONLY=true` path, or drop the acceptance smoke check from
`/dispatch-qa` since CI's `acceptance` job already covers it.

This finding does **not** block PR #646 — QA passes clean.

PR #666, opened 2026-05-19
dispatch: extract worktree resolution into dispatch-resolve-worktree

natb1:
## Summary

`/dispatch` Step 3 ran ~35 lines of SKILL.md prose the agent re-executed every
invocation: parse `git worktree list --porcelain`, match the `<issue>-<slug>`
branch convention, generate a sanitized branch name, and branch four ways. All
of it is deterministic string/list processing — and the branch-slug rules were
effectively duplicated against the `WorktreeCreate` hook regex.

This extracts that logic into `dispatch-resolve-worktree <issue> <explicit|queue>`,
which prints exactly one decision line:

- `here` — current branch already is `<issue>-*`; no `EnterWorktree` needed
- `enter <path>` — re-use an existing `<issue>-*` worktree
- `create <branch>` — no worktree; `<branch>` is the sanitized name
- `conflict <path>` — a queue-selected target already has a worktree

The `here` check is mode-independent and fires before the worktree scan. An
existing `<issue>-*` worktree yields `enter` in `explicit` mode and `conflict`
in `queue` mode. Generated branch names match the `WorktreeCreate` hook form
`^[0-9]+-[a-z0-9]+(-[a-z0-9]+)*$`, full name ≤ 32 chars.

SKILL.md Step 3 collapses to: run the script, act on its single line. The
`EnterWorktree` call, the `sync-issue-context` re-syncs, the
`tmp/dispatch-worktree` marker, and the `restore-dispatch-skill.sh` recovery
paragraph stay with the agent.

## Changes

- **New** `.claude/skills/dispatch/scripts/dispatch-resolve-worktree`.
- `test-dispatch-scripts.sh` gains a `dispatch-resolve-worktree` section
  covering each output case, sanitization, truncation, `here` precedence, and
  argument validation (70 → 86 tests, all passing).
- `dispatch/SKILL.md` Step 3 rewritten to call the script.

## Testing

- `bash .claude/skills/dispatch/scripts/test-dispatch-scripts.sh` → 86/86 pass.
- Smoke from this worktree: `dispatch-resolve-worktree 656 explicit` → `here`.

Closes #656.

### DIFF
--- a/.claude/skills/dispatch/SKILL.md
+++ b/.claude/skills/dispatch/SKILL.md
@@ -101,41 +101,42 @@ If a named target issue is **closed**, report it and **stop**.
 
 ## 3. Resolve the Worktree
 
-Resolve or create the final target's worktree via `EnterWorktree`, matching
-`pr-workflow` Sections 2–3. `EnterWorktree` accepts exactly one of `path` (switch to
-an existing worktree) or `name` (create a new one; fires the `WorktreeCreate` hook).
+Run the worktree-resolution script, matching `pr-workflow` Sections 2–3. Pass
+`explicit` when the target was named by an explicit `/dispatch` argument,
+otherwise `queue`:
 
-- **Already in the target's worktree** (current branch starts with `<issue>-`) → no
+```bash
+.claude/skills/dispatch/scripts/dispatch-resolve-worktree <N> <explicit|queue>
+```
+
+It prints exactly one decision line — act on it. `EnterWorktree` accepts exactly
+one of `path` (switch to an existing worktree) or `name` (create a new one).
+
+- **`here`** → the current branch already is the target's worktree; no
   `EnterWorktree` needed. Re-sync issue context:
   ```bash
   .claude/skills/ref-pr-workflow/scripts/sync-issue-context <N>
   ```
-  (`dangerouslyDisableSandbox: true` — `sync-issue-context` calls `gh`.) Then go
-  straight to creating the marker below.
-- **An existing worktree matches** `<issue>-*` (parse `git worktree list --porcelain`
-  as blank-line-delimited records):
-  - **Named by an explicit `/dispatch` argument** (recycle-after-completion case) →
-    `EnterWorktree` with `path:` set to that path. After entering, re-sync issue
-    context from the worktree:
-    ```bash
-    .claude/skills/ref-pr-workflow/scripts/sync-issue-context <N>
-    ```
-    (`dangerouslyDisableSandbox: true` — `sync-issue-context` calls `gh`.)
-  - **Queue-selected (no argument)** → another session already owns this worktree.
-    The queue scan skips worktree'd issues, so this arises only when Step 2
-    leaf-tracing retargets to a blocker or sub-issue that has one. Report the
-    conflict (name the worktree path and issue `<N>`) and **stop**; do not
-    `EnterWorktree`.
-- **No existing worktree** → generate a sanitized branch name `<issue>-<slug>`:
-  lowercase the issue title, replace non-alphanumeric runs with `-`, collapse repeated
-  `-`, strip leading/trailing `-`, and truncate so the full branch name is ≤ 32
-  characters. `EnterWorktree` with `name:` set to that branch name.
+  (`dangerouslyDisableSandbox: true` — `sync-issue-context` calls `gh`.)
+- **`enter <path>`** → re-use an existing `<issue>-*` worktree (the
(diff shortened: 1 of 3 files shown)


natb1:
## QA pass — issue #656

Single user-acceptance QA pass (non-browser path — `/dispatch` skill/script
change). The walkthrough exercised `dispatch-resolve-worktree` against **real**
git state, distinct from the mocked unit tests.

**Result: 5 PASS / 0 FAIL / 0 SKIP — clean pass.**

| # | Item | Result |
|---|---|---|
| 1 | `here` decision from inside the target worktree | PASS |
| 2 | `enter` vs `conflict` for the same worktree'd target | PASS |
| 3 | `create` decision — branch name conforms | PASS |
| 4 | Argument validation — clear error, no fallback | PASS |
| 5 | SKILL.md Step 3 calls the script, handles all four outputs | PASS |

Notes:

- **Item 1** — `dispatch-resolve-worktree 656 explicit` and `... 656 queue` both
  print `here`; the check is mode-independent and fires before the worktree scan.
- **Item 2** — for issue 660 (which has a worktree), `explicit` → `enter <path>`
  and `queue` → `conflict <same path>`: same target, the mode argument flips the
  decision (AC #3).
- **Item 3** — `dispatch-resolve-worktree 651 explicit` →
  `create 651-dispatch-monitor-in-progress`; the branch is exactly 32 chars and
  matches the `WorktreeCreate` form `^[0-9]+-[a-z0-9]+(-[a-z0-9]+)*$` (AC #2).
- **Item 4** — no args, `656 bogus`, and `abc explicit` each exit 1 with the
  usage error on stderr and no decision line on stdout.
- **Item 5** — SKILL.md Step 3 invokes the script and has a handling branch for
  each of `here`/`enter`/`create`/`conflict`; the former inline
  `git worktree list --porcelain` parsing prose is gone (AC #4).

AC #5 (test suite covers each output case; full suite passes) was verified by the
pre-QA `origin/main` merge — `test-dispatch-scripts.sh` is green at 86/86.

natb1:
## /review-fix: review phase complete

Ran the generic `/review` against this PR — extraction of `/dispatch` Step 3
worktree resolution into `dispatch-resolve-worktree`.

### Recommendation: approve

No actionable findings — no code changes applied this phase.

### What the review checked

- **Correctness** — prefix matching (`<N>-*`) is unambiguous (the `-` separator
  rules out `4` matching `42-foo`); the `here` check fires before the worktree
  scan, mode-independent; slug truncation can leave a trailing `-` and the
  `${slug%-}` strip handles it (sed had already collapsed runs, so at most
  one); multi-byte titles are flattened to ASCII before truncation so bash's
  `${var:0:N}` byte/char distinction doesn't matter.
- **Conventions** — `set -euo pipefail`, quoted expansions, process-sub
  while-loop (parent-shell scope), `jq -er` to fail on null title, `printf '%s'`
  for slug input. Errors-over-fallbacks honored throughout (bad args, missing
  title, empty post-sanitize slug all exit non-zero with a labeled message).
- **Hook compatibility** — generated branch names match the `WorktreeCreate`
  regex `^[0-9]+-[a-z0-9]+(-[a-z0-9]+)*$` with full name ≤ 32 chars. Test 6
  asserts both length and regex.
- **Security** — slug is restricted to `[a-z0-9-]` before interpolation; no
  injection surface even on a hostile `gh` response.
- **Performance** — `gh issue view` is gated behind the worktree scan, so the
  network call only fires on the `create` path.
- **Test coverage** — 86/86 pass; 16 new tests cover all four output cases,
  the explicit-vs-queue distinction (acceptance criterion 3), sanitization,
  truncation against the hook regex (acceptance criterion 2), `here`
  precedence, non-matching worktree fallthrough, all five argument-validation
  paths, and the empty-slug failure mode.

### Informational notes (not fixed)

1. **Detached-HEAD worktrees aren't tracked** — `git worktree list --porcelain`
   emits a `detached` line instead of `branch <ref>` for those, and the script
   only inspects `branch` lines. Unreachable in this workflow; worktrees stay
   on their branches.
2. **`gh issue view` failures don't surface the underlying error** — the
   script prints `error: failed to fetch title for issue <N>` and drops the
   `gh`/`jq` diagnostic. The user can re-run `gh issue view <N>` to debug.
   Keeping the error to one line is more usable for the common case.
3. **Porcelain record order is assumed** (`worktree`, `HEAD`, `branch`) — this
   is documented and stable in git; no defensive guard needed.
4. **No `gh` call on `here`/`enter`/`conflict` paths** — already optimized;
   noting as a positive.

Applying `dispatch:reviewed`.

PR #675, opened 2026-05-19
firestoreutil: fix test typecheck errors and add CI ratchet (#668)

### LINKED ISSUE #668: firestoreutil: fix typecheck errors in namespace.test.ts and seed.test.ts
## Problem

`npx tsc --noEmit --project firestoreutil` reports **12 typecheck errors** in
test files on `origin/main`:

- `firestoreutil/test/namespace.test.ts` — 1 error (TS6133, unused import)
- `firestoreutil/test/seed.test.ts` — 11 errors (1× TS6133, 10× TS2322)

These errors do not break CI today because `vitest` runs the compiled JS without
typechecking and lint does not catch type errors. They surfaced during the
`/review` phase of #598 phase 1 (PR #620), which only moved `issue-state` and
did not touch these test files.

## Root cause

1. **Unused imports / params (TS6133, 2 sites):**
   - `namespace.test.ts:2` imports `type Namespace` but never references it.
   - `seed.test.ts:18` declares a `path: string` parameter in a mock factory
     whose body never reads it.

2. **Branded type assignment (TS2322, 10 sites in `seed.test.ts`):**
   `SeedSpec.namespace` is typed as the branded
   `Namespace = string & { readonly __brand: unique symbol }`, but the tests
   build `SeedSpec` literals with plain string values like
   `namespace: "app/emulator"`. Wrap each literal with `validateNamespace(...)`.

## Acceptance criteria

- [ ] `npx tsc --noEmit --project firestoreutil` exits clean (0 errors).
- [ ] `npx vitest run --root firestoreutil` continues to pass — no regressions.
- [ ] Only `firestoreutil/test/namespace.test.ts` and
      `firestoreutil/test/seed.test.ts` are modified. No `src/` changes.
- [ ] A CI guard runs `tsc --noEmit` per workspace package so future
      type-error drift fails the build (extend the unit-test or lint workflow,
      or add a dedicated typecheck job).

## Effort / Impact

- Effort: S (mechanical fix, 12 sites in 2 files; CI guard wiring may be S–M).
- Impact: S (test-only cleanup; restores per-package `tsc --noEmit`
  cleanliness and prevents regression).

natb1:
Closes #668

Two units:

1. **`firestoreutil` test typecheck fix.** Wraps `SeedSpec.namespace` literals
   in `firestoreutil/test/seed.test.ts` with the `validateNamespace(...)` brand
   constructor (the brand constructor for
   `Namespace = string & { readonly __brand: unique symbol }`), and drops two
   unused identifiers — the `type Namespace` import in `namespace.test.ts` and
   an unused `path: string` parameter in `mockDoc`'s vi.fn factory. Tests-only
   change; `vitest run` behavior unchanged. Clears the 19 errors
   `npx tsc --noEmit --project firestoreutil` was reporting (12 reported in
   the issue; 7 more accrued before this fix landed).

2. **Implicit typecheck CI ratchet.** Adds
   `.claude/skills/ref-pr-workflow/scripts/run-typecheck.sh` and wires a new
   `typecheck` job in `.github/workflows/unit-tests.yml` parallel to
   `unit-tests` and `lint`. The guard runs `npx tsc --noEmit --project <ws>`
   on each changed workspace's HEAD version, but only after confirming
   `origin/main`'s version of that workspace typechecks clean. Workspaces
   already dirty on `origin/main` are skipped with a one-line notice. New
   workspaces (absent from `origin/main`) must start clean. The protected
   set is self-maintaining — no allowlist, no script-marker, no manual
   opt-in. Includes a `test-run-typecheck.sh` covering 5 scenarios
   (clean/clean, clean/dirty, dirty/dirty, working-tree-dirty guard,
   cleanup invariant).

The CI guard satisfies #668's acceptance criterion that "a CI guard runs
`tsc --noEmit` per workspace so future type-error drift fails the build,"
without requiring this PR to fix the 419 typecheck errors across 8 other
currently-dirty workspaces (which is far outside the issue's S–M estimate).
Those workspaces auto-promote to the protected set the moment any future PR
makes them clean on `origin/main`.

One-PR lag note: because this PR is itself fixing `firestoreutil`,
`origin/main` for `firestoreutil` is still dirty when CI runs on this branch —
so the guard skips `firestoreutil` here. Once this PR merges, the next PR's
CI run will see `firestoreutil` clean on `origin/main` and enforce it.


natb1:
## Dispatch QA — clean pass

Backend / tooling PR — non-browser walkthrough.

**Items walked:** 5
**PASS:** 5 &nbsp; **FAIL:** 0 &nbsp; **SKIP:** 0

| # | Item | Outcome |
|---|------|---------|
| 1 | `npx tsc --noEmit --project firestoreutil` — clean | PASS (exit 0, acceptance criterion 1) |
| 2 | `npx vitest run --root firestoreutil` — no regressions | PASS (60/60 tests, acceptance criterion 2) |
| 3 | `test-run-typecheck.sh` — script's own self-tests | PASS (16/16 assertions, 5 scenarios) |
| 4 | `run-typecheck.sh` end-to-end on this branch | PASS (exit 0) |
| 5 | CI `typecheck` job wired and green on PR #675 | PASS (all 12 status checks SUCCESS) |

**Item 4 note.** `firestoreutil` shows as `skipping — origin/main has pre-existing typecheck errors`, alongside `authutil`, `blog`, `firebaseutil`. This is the script's expected dirty/dirty behavior (covered by self-test #3): the baseline still has the errors this PR fixes, so the script declines to compare. Once this PR merges and `origin/main` baselines clean, the workspace will be guarded by the typecheck job against future regressions. Direct `tsc --noEmit` on the HEAD content (item 1) confirms the fix itself is in place.

**Bugs found:** none.

natb1:
## Review — fixes applied

Reviewed the diff against current `main`. The PR is in good shape; one
boilerplate-removal fix applied, the rest are informational.

### Applied (1)

- **`unit-tests.yml`: dropped the unused `Detect changed file categories` step
  from the new `typecheck` job** (commit `cba4c9f`). The step ran
  `detect-changes.sh` and set `id: changes`, but no subsequent step in that job
  consumed `steps.changes.outputs.*` — typecheck has no nix / rules /
  playwright / go dependencies, only node. Removing it trims a few seconds
  off every CI run and removes confusing dead boilerplate copied from the
  `unit-tests` / `lint` jobs.

### Informational (not changed)

- **`unset 'DIRTY_APPS[rules-test]'` runs unconditionally** in
  `run-typecheck.sh`. `unset` on a missing array key is a no-op, and this
  matches the existing `run-unit-tests.sh` convention. Leaving as-is.
- **Cleanup trap silently swallows `git checkout HEAD --` errors** (`2>/dev/null
  || true`). Defensible — the trap must not mask the real script exit reason —
  and the explicit immediate restore before the trap reduces the window. Could
  optionally log to stderr, but the current form is fine.
- **`git fetch ... 2>/dev/null || ... || true` falls back to "new workspace"
  branch if offline.** In CI `origin/main` is always pre-fetched via
  `fetch-depth: 0`; locally the user has it. The script's note already calls
  this out as idempotent.
- **`--app <ws>` explicit-mode codepath isn't directly tested.** Shares
  parsing + main loop with the auto-detect path that is tested. Low risk.
- **`pass_suffix` for the "(new workspace)" annotation is slightly clever.**
  Two distinct echo branches would also work, but the current form is concise.

### Strengths worth noting

- `test-run-typecheck.sh` is thorough — ephemeral repos + stub `npx` for
  hermetic tests; covers clean/clean, regression, dirty/dirty skip,
  working-tree guard, and cleanup invariants on both pass and fail paths.
- Self-maintaining ratchet design — no allowlist file to drift; workspaces
  auto-promote on first clean baseline.
- Defensive layering: pre-flight working-tree-dirty guard, `cleanup` trap
  registered on `EXIT INT TERM` with original `rc` captured at entry,
  explicit immediate-restore after each per-workspace baseline check.
- Fix wraps `SeedSpec.namespace` with the existing `validateNamespace(...)`
  brand constructor rather than `as Namespace` casts — idiomatic.

natb1:
## Security review — no findings

Ran `/security-review` against the PR diff. The reviewer examined the new shell script (`run-typecheck.sh`), its test harness, the added GitHub Actions `typecheck` job, and the test-only `.ts` edits.

Disposition summary:

- **Not fixed (no action needed)** — no findings of any severity to fix.

Per-surface notes from the review:

- `run-typecheck.sh` — inputs come from `--app` CLI args and `get-changed-apps.sh` output (workspace directory names). All expansions are quoted; no `eval`, no unquoted command substitution, no path-traversal sink. `tsc --project` and `git checkout HEAD -- <path>` treat the value as a path argument, not as code. Per project rules, shell command injection is not flagged absent a concrete attack path with untrusted input; inputs here are repo-internal workspace names.
- `.github/workflows/unit-tests.yml` (typecheck job) — no `pull_request_target`, no use of `${{ github.event.* }}` interpolation into a `run:` block, pinned action SHAs, default permissions. No injection sink.
- `test-run-typecheck.sh` and `firestoreutil/test/*.ts` — test-only files, excluded from security findings per project rules.
- No secrets, crypto, auth, deserialization, or data-exposure surface is touched by this PR.

Issue #684, opened 2026-05-20
dispatch: serialize concurrent invocations with a process-level lock


natb1:
## Context

`/dispatch` has long-running, state-mutating side effects: fetching `origin/main`, creating and (per the issue this blocks) removing worktrees, calling `gh` for label transitions and PR ready-flips. Per-target guards (worktree existence check from #661, `dispatch:*` label state machine) assume **serial** execution within a single tick. Two `/dispatch` invocations running concurrently — `/loop /dispatch` interleaved with a manual `/dispatch <N>`, or two loops in two sessions — can race on:

- target selection: both pick the same task before either claims it.
- worktree adoption (after the sweep issue): two ticks adopt the same orphan.
- worktree creation: both EnterWorktree the same `<N>-*` branch.
- label transitions: two phases applying conflicting `dispatch:*` labels.

## Proposed change

Acquire a non-blocking process lock at the top of `/dispatch`, held for the duration of the tick:

```bash
exec 9>"$PROJECT_ROOT/tmp/dispatch.lock"
flock -n 9 || { echo "another /dispatch is running — stopping"; exit 0; }
```

Released automatically on exit (fd closes). Non-blocking — a contending tick fails fast and clean rather than stacking; the next `/loop` iteration or operator re-runs.

Wrap acquisition + the success-message + the orderly stop in a small `dispatch-acquire-lock` script invoked from a new SKILL.md Step 0, consistent with the "thin SKILL.md, thick script" pattern.

## Acceptance criteria

- [ ] `/dispatch` SKILL.md Step 0 acquires `tmp/dispatch.lock` non-blocking via `dispatch-acquire-lock` (or equivalent script entry point). On failure, exits 0 with a clear "another /dispatch is running" message and no other side effects.
- [ ] The lock is held across all subsequent steps in the tick and released automatically on normal or abnormal exit.
- [ ] `tmp/dispatch.lock` is git-ignored (verify — `tmp/` already ignored).
- [ ] Test in `test-dispatch-scripts.sh` demonstrates that two parallel `flock -n` acquisitions on the same lock file yield exactly one success.
- [ ] No regression to single-tick behavior: with no contention, the lock acquisition is invisible to the rest of the workflow.

## Dependencies

Blocked by the dispatch worktree-sweep issue. The sweep's "adopt-and-resume" path is one concrete contention point the lock protects; sequencing the lock work second lets its acceptance criteria explicitly cover sweep adoption.

PR #670, opened 2026-05-19
budget: use schemeTableau10 in sankey chart for sibling consistency

natb1:
Closes #669

PR #626, opened 2026-05-02
Split style and htmlutil packages (#606)

natb1:
Closes #606

## Summary

- Extracts `@commons-systems/criticalcssutil` from `htmlutil` (owns the critters-based critical-CSS inliner used by `landing` and `fellspiral` build scripts).
- Extracts `@commons-systems/components` from `style` (owns runtime Lit components: hero, hero-render, panel-toggle, scroll-indicator, nav, autocomplete) using a flat layout.
- `htmlutil` keeps only `escapeHtml`.
- `style` becomes a CSS-only package (design tokens + component CSS).

## Test plan

- [x] `npx tsc --noEmit` passes for every affected package
- [x] `npm test --prefix criticalcssutil`, `htmlutil`, `components` pass
- [x] App builds succeed: `landing`, `fellspiral`, `budget`, `print`

PR #686, opened 2026-05-20
dispatch: sweep merged worktrees and adopt orphaned ones

natb1:
## Summary

`/dispatch` queue-scan mode skipped any target whose `<N>-*` worktree existed (the #661 guard). The guard could not distinguish **active** worktrees (a live `claude` process with cwd inside) from **orphaned** ones (no live process — in-progress work whose owning session exited without `WorktreeRemove` firing). Orphans silently accumulated. Separately, worktrees whose branches squash-merged into `main` also lingered because the `WorktreeRemove` hook fires inconsistently.

This PR adds a sweep step that runs in `/dispatch` queue-scan mode and:

1. Removes merged-and-in-sync worktrees, using `gh pr list --state merged` as the authoritative source (squash-merge creates a new SHA so `git merge-base --is-ancestor` is not used).
2. Classifies the rest as active vs orphaned by walking `/proc/<pid>/{comm,cwd}` — no `sessions.json` consulted (known ~50% stale).
3. Adopts the oldest orphan by HEAD commit time, emitting `worktree <N> <branch>` on stdout — the same shape `dispatch-select-target` already emits for current-worktree mode, so SKILL.md routes it on an existing path.
4. Halts with `cleanup-unknown:<path>` on stderr when a worktree has neither an open PR nor an inferable issue number — SKILL.md prompts the user before any destructive operation on potentially-unmerged code.

### Files

- `lib-worktree-in-sync.sh` — new sourceable helper exposing `worktree_in_sync <path> [log]`, extracted from `worktree-remove.sh` so the sweep and the hook share one definition of "safe to remove."
- `dispatch-sweep` — new script. Default mode does the full sweep; `--cleanup-unknown <path>` is the explicit re-entry that removes a single unknown worktree after user confirmation. Env vars `DISPATCH_SWEEP_PROC_ROOT`, `DISPATCH_SWEEP_LOG_FILE`, `DISPATCH_SWEEP_NOW` enable testability.
- `test-dispatch-scripts.sh` — 6 new tests (22 assertions) covering merged classification, active-vs-orphaned via synthetic `/proc`, oldest-orphan tiebreaker, issue-number inference (both paths), and `--cleanup-unknown`.
- `SKILL.md` — one new step in the queue-scan path of Step 1 routing the three sweep exit modes; all logic stays in the script.
- `worktree-remove.sh` — now sources the shared helper; behavior unchanged.

### Verification

- `bash .claude/skills/dispatch/scripts/test-dispatch-scripts.sh` — 109/109 pass.
- `bash .claude/hooks/test-worktree-remove.sh` — 53/53 pass.
- Live sweep run cleaned up two newly-merged worktrees (#606, #669) on the verification machine and correctly classified 8 active sessions; `tmp/dispatch-sweep.log` recorded every decision.

### Out of scope

Concurrency safety against parallel `/dispatch` ticks racing on the sweep — deferred to the follow-up issue that blocks on this one.

Closes #683

## Test plan
- [ ] CI passes on the branch.
- [ ] After merge, manually run `/dispatch` from the main worktree and confirm the sweep step fires before target selection.
- [ ] After merge, leave a stale orphaned worktree on disk and confirm the next `/dispatch` tick adopts it.


natb1:
## /dispatch-qa results — non-browser walkthrough

Coverage scope: bash scripts under `.claude/` only (`dispatch-sweep`,
`lib-worktree-in-sync.sh`, `worktree-remove.sh` refactor, `test-dispatch-scripts.sh`
additions). No browser surface — every plan item is machine-verifiable.

**Items walked: 7 · PASS: 7 · FAIL: 0 · SKIP: 0**

| # | Item | Result |
|---|------|--------|
| 1 | Unit test suite (`test-dispatch-scripts.sh`) — all 6 new dispatch-sweep tests pass | PASS — 109/109 |
| 2 | dispatch-sweep rejects bogus flag and missing-path `--cleanup-unknown` | PASS — exit 2, clear ERROR on stderr |
| 3 | `--cleanup-unknown` refuses paths outside `worktrees/` and refuses `main` | PASS — exit 2, no `git worktree remove` runs |
| 4 | Real-world full sweep against the live worktrees tree | PASS — exit 0, this worktree correctly ACTIVE; historical log shows real REMOVE_MERGED entries for PRs #626 and #670 |
| 5 | Shared `worktree_in_sync` helper sourceable and accurate | PASS — returns 0 on clean+pushed worktree |
| 6 | `WorktreeRemove` hook now sources the extracted helper, no inline duplicate | PASS — line 21 of `worktree-remove.sh`, no `rev-list`/`status --porcelain` left inline |
| 7 | `SKILL.md` routes all three sweep exit modes | PASS — empty stdout, `worktree <N> <branch>` adoption, and `cleanup-unknown:<path>` with `AskUserQuestion` loop |

### Note (informational, not a FAIL)

The issue body's AC reads: "running **after** the #660 health gate and **before**
the existing target-selection ladder." The implementation runs the sweep
**before** `dispatch-select-target` (and therefore before the health gate, since
the gate lives inside that script).

Functionally this is defensible: adoption is treated as `explicit`, which
already bypasses the health gate per SKILL.md's stated rule that
current-worktree continuation bypasses the gate. Merged-worktree cleanup is
non-destructive regardless of main's state. End-user behavior is correct; only
the literal wording of the AC differs from the placement.

Worth a flag in case strict ordering is preferred.

🤖 [/dispatch-qa](https://github.com/natb1/commons.systems/blob/main/.claude/skills/dispatch-qa/SKILL.md)

natb1:
## /review summary

Ran `/review` on this PR. The design is sound — the shared in-sync helper, `gh pr list` as the merged-PR oracle (avoiding the squash-merge SHA pitfall), safety-fenced `--cleanup-unknown`, and best-effort degradation when `gh` is unavailable all check out. Tests stay green (123/123 dispatch + 53/53 worktree-remove) after the merge with current `origin/main`.

### Fixes applied in this commit

1. **SKILL.md — handle unexpected `dispatch-sweep` exits.** The routing block only described `Exit 0 + empty stdout`, `Exit 0 + worktree adoption`, and `Non-zero + cleanup-unknown stderr`. Added a fourth bullet: any other non-zero exit (e.g. `exit 2` for invalid invocation, transient `gh` / `git` failure) logs the stderr contents and falls through to `dispatch-select-target` as defense-in-depth — a malformed sweep call should not stall the workflow.

2. **`dispatch-sweep` — clearer `BRANCH_DELETE_FAILED` log message.** The previous parenthetical "(worktree already removed)" misread the situation: by the time `branch -D` runs, *this* call removed the worktree (the enclosing `git worktree remove` on the line above succeeded). Reworded to "(worktree removed; branch retained)" so the log reflects what actually happened.

3. **`lib-worktree-in-sync.sh` — document the shell-option side effect.** Sourcing the helper sets `-u` and `-o pipefail` in the caller shell. Both current callers (`worktree-remove.sh`, `dispatch-sweep`) already use those options, so this is currently invisible — but a sourceable helper that silently changes `set` flags can surprise future callers. Added a "Side effect" paragraph to the header so the constraint is explicit. (No behavior change.)

4. **Tests — `--cleanup-unknown` safety-guard coverage.** Added three new tests covering the destructive-operation perimeter: rejecting a path outside `$WORKTREES_ROOT`, refusing to remove `main`, and rejecting an invocation with no path argument. Each asserts `exit 2` and a specific stderr message. These are the highest-risk paths in the script and now have explicit assertions.

5. **SKILL.md — use `tmp/` instead of `$TMPDIR` for sweep stderr capture.** Changed `2>$TMPDIR/dispatch-sweep-stderr` to `2>tmp/dispatch-sweep-stderr` for consistency with `tmp/dispatch-sweep.log` (the sweep's own log) and the broader convention. `tmp/` is git-ignored and project-rooted.

6. **`dispatch-sweep` — log when `$PROC_ROOT` is not a directory.** Previously a missing `$PROC_ROOT` silently fell through the `/proc` walk, leaving every surviving worktree classified orphaned. A misconfigured `DISPATCH_SWEEP_PROC_ROOT` is now observable in the sweep log as `PROC_ROOT_MISSING`.

### Out-of-scope findings (not implemented here)

- **Concurrency safety against parallel `/dispatch` ticks** — already filed as #684 ("dispatch: serialize concurrent invocations with a process-level lock"), explicitly blocked-by #683 in its body. The PR body deferred this; no action needed.

### Unfixed findings (informational)

None — all the actionable findings from `/review` were implemented above.

### Verification

- 123/123 dispatch script tests pass (109 existing + 14 new assertions from the three safety-guard tests).
- 53/53 `worktree-remove` hook tests pass (the helper is now shared but the hook's behavior is unchanged).
- Branch is up to date with `origin/main` after the second merge in this `/review-fix` pass.

natb1:
## Security review

Ran `/security-review` against `origin/main` (merged in at `3f32c8d`).

**Findings: none.**

Files reviewed for security implications:
- `.claude/skills/dispatch/scripts/dispatch-sweep` — new worktree reconciliation script
- `.claude/skills/dispatch/scripts/lib-worktree-in-sync.sh` — shared in-sync helper
- `.claude/hooks/worktree-remove.sh` — now sources the helper
- `.claude/skills/dispatch/scripts/dispatch-select-target` — adds `--health-only` mode

### Defenses verified

**`--cleanup-unknown <path>` (the only path that can destroy unmerged work)** at `dispatch-sweep:72-111` enforces three independent guards before `git worktree remove --force`:

1. `realpath -m` canonicalization, then `dirname == WORKTREES_ROOT` (defeats traversal and symlink escapes — `realpath` resolves before validation).
2. `basename != "main"`.
3. Path must appear in `git worktree list --porcelain` (unregistered paths exit 0 with a log line).

All arguments are double-quoted.

**Branch names from `gh pr list` / `git worktree list`** flow into `git branch -D "$wt_branch"` and bash array keys. Git refname rules forbid leading `-`, so option-injection is impossible; quoting prevents shell-injection.

**`/proc` walk** validates PID against `^[0-9]+$`, reads `comm`/`cwd` only as data for pattern matches (`[[ comm == .claude* ]]`, `case "$cwd/" in "$wt_path/"*)`), and never `eval`/`exec`s any read value.

**`gh` JSON** is parsed via `jq -r '... | @tsv'`, consumed with tab-delimited `read`; values flow only as quoted args to `git` or as array keys/stdout strings.

**`lib-worktree-in-sync.sh`** receives the worktree path only as a quoted arg to `git -C "$path"` (no shell evaluation) and `return`s rather than `exit`s, so sourcing it from `worktree-remove.sh` does not expand attack surface.

### Non-security note (not a finding)

`case "$cwd/" in "$wt_path/"*)` can be confused by glob metacharacters in a worktree path. This is a classification-correctness concern, not exploitable for code execution or unauthorized access.

— `/security-review-fix` (`/dispatch` security phase)

PR #711, opened 2026-05-21
dispatch: serialize concurrent invocations with a process lock

### LINKED ISSUE #684: dispatch: serialize concurrent invocations with a process-level lock

natb1:
Closes #684

`/dispatch` mutates shared state — fetches `origin/main`, creates/removes
worktrees, flips `gh` labels and PR draft status. Its per-target guards assume
serial execution within a tick, so two concurrent `/dispatch` invocations can
race on target selection, orphan-worktree adoption, worktree creation, and
`dispatch:*` label transitions.

An fd-based `flock` does not work here: `/dispatch` runs as many separate `Bash`
tool calls, each a fresh shell, so an fd `flock` releases the instant its call's
shell exits. This adds a PID-stamped lock file keyed on the Claude **session**
process, whose lifetime spans the whole tick.

## Changes

- **New `dispatch-acquire-lock` script.** Resolves the shared project-root lock
  file `tmp/dispatch.lock`, identifies this `/dispatch`'s session PID by walking
  process ancestry to the nearest `.claude` process, and under a brief blocking
  `flock` decides: a recorded, still-live `.claude` PID different from ours →
  print `busy`; otherwise (no record, a stale/dead/recycled PID, or our own
  session re-entering) → record our PID and print `acquired`. Release is
  implicit and self-healing — a dead session's PID stops passing the liveness
  check, so a crashed tick never wedges the queue.
- **SKILL.md Step 0.** A new `## 0. Acquire the Dispatch Lock` section runs the
  script as the first action, before the health gate, sweep, selection, and
  worktree resolution. On `busy` the skill stops cleanly with no side effects.
- **Tests.** Six new cases in `test-dispatch-scripts.sh`: first acquisition,
  two parallel sessions yielding exactly one `acquired`/one `busy`, dead-PID
  stale lock, recycled-non-`.claude`-PID stale lock, same-session re-entry, and
  misconfiguration. Full suite: 137/137 passing.

No `.gitignore` change needed — the project-root `tmp/` is outside any worktree
and `tmp/dispatch.lock` is already ignored.


natb1:
## Dispatch QA — clean pass ✅

User-acceptance QA for the `/dispatch` Step 0 process-level lock. Non-browser
path (SKILL.md + shell-script change).

**Scope note:** the 6 unit tests in `test-dispatch-scripts.sh` run in CI
(`unit-tests.yml`) and are green — this pass did not re-run them. It instead
exercised the **real-environment paths the CI tests deliberately stub**: the
real `/proc` ancestry walk, the real `.claude`-process liveness check, and real
project-root path resolution (CI always overrides `DISPATCH_LOCK_SESSION_PID`,
`DISPATCH_LOCK_PROC_ROOT`, `DISPATCH_LOCK_FILE`).

### Items walked (6)

| # | Item | Result |
|---|------|--------|
| 1 | SKILL.md Step 0 exists, correctly ordered, routes `acquired`/`busy` | PASS |
| 2 | Real-environment acquisition (real `/proc` walk → session PID, `acquired`) | PASS |
| 3 | Lock file at shared project root `tmp/dispatch.lock`, not per-worktree | PASS |
| 4 | Concurrent `/dispatch` from a different live session → `busy` | PASS |
| 5 | Same-session re-entry and stale (dead-PID) lock reclaim both proceed | PASS |
| 6 | Lock file is git-ignored — never committed | PASS |

**6 PASS / 0 FAIL / 0 SKIP**

### Notes

- Real `dispatch-acquire-lock` run resolved session PID `3989201` (`.claude-unwrapp`)
  via the real `/proc` ancestry walk and wrote
  `/home/n8/natb1/commons.systems/tmp/dispatch.lock` — the shared project root
  (parent of `.bare`), visible to every worktree.
- A simulated concurrent session got `busy` (exit 0) with the lock left
  unchanged; a stale dead-PID lock was correctly reclaimed.
- No bugs found.

natb1:
## Simplify phase

Three review agents (reuse, quality, efficiency) reviewed the changes on this branch.

### Findings that were fixed

Both applied in commit 3748eb5 to `.claude/skills/dispatch/scripts/dispatch-acquire-lock`:

- **Dead `[[ -f "$LOCK_FILE" ]]` guard / precedence ambiguity.** The Step 4
  defensive read was guarded by `[[ -f "$LOCK_FILE" ]]`, but the Step 3 `>>`
  open already creates the lock file, so the guard was always true. Combined
  with the trailing `|| true` it also produced an operator-precedence
  ambiguity — `( [[ -f ]] && read ) || true`. Dropped the guard.
- **In-file duplication of the `.claude`-comm check.** The `/proc` ancestry
  walk re-inlined the same "read `<pid>/comm`, test `.claude*` prefix" logic
  that the `is_live_claude` helper already encapsulates. The walk now calls
  `is_live_claude` directly.

`test-dispatch-scripts.sh` passes 142/142 after both changes.

### Informational findings

No action — surfaced for reference:

- **Project-root resolution idiom duplicated 4-way.** The
  `git rev-parse --git-common-dir` → `dirname` idiom now appears in
  `dispatch-acquire-lock`, `dispatch-sweep`, `worktree-create.sh`, and
  `worktree-remove.sh`. Pre-existing pattern; this script joins as the 4th
  copy. A real cleanup, but out of scope here and not introduced by this PR.
- **Test 6 assumes `$TMPDIR` is outside any git repo.** The non-git
  misconfiguration test `cd`s into a fresh `mktemp -d` and expects exit 2. If
  `$TMPDIR` ever sat inside a git repo, `git rev-parse` would succeed and the
  test would fail. True in the sandbox/CI environment, so it passes — but the
  assumption is implicit.

### Disregarded findings

- **"Nested conditionals 3+ deep" in the ancestry walk** — the walk's outer
  `while` contains one inner `while` (status-file scan); two levels, not over
  the 3+ threshold.
- The `flock` / `exec {lockfd}>>` usage, the `set -- $line` PPid word-split,
  and the `mkdir -p` before the redirect were flagged as candidates — all
  verified correct on inspection.

### Deferred findings

- **Shared `.claude`-process `/proc` detection.** `dispatch-acquire-lock` and
  the sibling `dispatch-sweep` independently hand-roll the same `PROC_ROOT`
  override plus comm-prefix check. Drift between the two detectors would be a
  real correctness bug. Extracting a shared `lib-claude-proc.sh` also touches
  `dispatch-sweep` (outside this PR's scope) — tracked in #714.

natb1:
## Review — `review` phase

Reviewed the process-lock change: the new `dispatch-acquire-lock` script, the
SKILL.md Step 0, and the lock test cases.

**Quality is strong.** Thorough header doc explaining why an fd-`flock` cannot
work and how the session-PID keying recovers; `set -uo pipefail` (not `-e`)
matching sibling `dispatch-sweep`; env overrides for testability; a correct
blocking-`flock` mutex wrapping an atomic read-check-write; `is_live_claude`
reused; idempotent re-entry and self-healing release both tested.

### Fixed

- **The `/proc` ancestry walk had no test coverage.** All six lock tests set
  `DISPATCH_LOCK_SESSION_PID`, bypassing the walk — yet the walk (with its
  `/proc/<pid>/status` PPid parsing and word-split) is the production mechanism
  the acceptance criteria name as "nearest ancestor whose `comm` starts with
  `.claude`". Added **Test 7**: it leaves `DISPATCH_LOCK_SESSION_PID` unset and
  builds a synthetic `/proc` tree so the script's `$PPID` is a non-`.claude`
  shell whose parent is a synthetic `.claude` session, then asserts the lock
  records the climbed-to ancestor PID rather than `$PPID`. New `lock_proc_status`
  test helper writes the synthetic `status` files. Two stale comments in
  `dispatch-acquire-lock` that claimed the walk is untested were corrected.
  Suite: 145/145 passing.

### Not fixed (informational)

- **SKILL.md Step 0 routes only `acquired` / `busy`.** The script's exit-2
  misconfiguration path (`not in a git repo and DISPATCH_LOCK_FILE is unset`)
  prints an error to stderr with empty stdout, which Step 0 does not route. Left
  as-is: `/dispatch` always runs inside a git worktree, so exit 2 is unreachable
  in practice.
- **PR body says "137/137 passing".** Stale after the `origin/main` merge (now
  145/145). Not a code change, and the count re-stales with each merge — left
  as-is.

natb1:

Ran `/security-review` against the branch (merged with current `origin/main` at `38f247a`).

**Result: no security vulnerabilities found.** No HIGH or MEDIUM findings; no fixes applied.

### Scope reviewed

The PR adds the `dispatch-acquire-lock` script, a `Step 0` in `dispatch/SKILL.md`,
and unit tests in `test-dispatch-scripts.sh`.

### Findings and dispositions

| Finding | Disposition |
|---|---|
| — | No findings. |

### Analysis notes

Every external input was traced to its sink:

- **PID values** read from the lock file and `/proc/<pid>/status` flow only into
  `[[ ]]` string comparisons and quoted path components (`$PROC_ROOT/$pid/comm`) —
  never `eval` or a shell. Malformed content fails the `[[ -r ]]` test rather than
  executing.
- **`set -- $line`** word-splits only the kernel-generated, numeric `PPid:` field,
  immediately validated by `^[0-9]+$`.
- **`LOCK_FILE`** is derived from `git rev-parse --git-common-dir` (trusted) or the
  test-only `DISPATCH_LOCK_FILE` override; used quoted throughout.
- The **`comm`-prefix PID-recycling limitation** is a fail-safe correctness
  tradeoff (it can only make a tick stop early), explicitly accepted in the issue
  spec — not a security issue.

The script matches the existing security posture of sibling scripts `dispatch-sweep`
and `lib.sh`.

PR #712, opened 2026-05-21
dispatch: add dispatch-find-pr and forbid ad-hoc phase inference

natb1:
Closes #673

`/dispatch` Step 4 names `dispatch-phase` as the canonical phase-derivation
script, but Steps 2 and 6 gate on PR existence without saying how to check it.
A recent `/dispatch #669` checked PR existence with an ad-hoc
`gh pr list --search "669 in:title"`; PR #670 existed on branch `669-…` but its
title had no "669", so the search returned empty and the phase was misderived
as `implement` instead of `qa`.

## Changes

- **New `dispatch-find-pr` script** — the canonical PR-existence helper. It
  correlates a PR to an issue by the `<issue>-` branch-name prefix, the same
  logic `dispatch-phase` uses, and honors `DISPATCH_PR_LIST`.
- **SKILL.md Step 2** — leaf-tracing skip condition now calls `dispatch-find-pr`
  instead of describing the check in prose.
- **SKILL.md Step 4** — explicit rule forbidding hand-rolled `gh` queries for
  phase derivation or PR-existence checks (names the title-search miss).
- **SKILL.md Step 6** — relevance-review gate now calls `dispatch-find-pr`.
- **test-dispatch-scripts.sh** — four `dispatch-find-pr` cases: matching
  branch prefix + matching title, matching prefix + non-matching title (the
  #670 regression case), no PR, and `DISPATCH_PR_LIST` override.

Test suite: 128/128 passing.


natb1:
## Dispatch QA — clean pass ✅

User-acceptance QA for #673 (forbid inferring phase from ad-hoc `gh` queries).
Non-browser path — single QA pass against the SKILL.md prose edits and live
`dispatch-find-pr` behavior.

**Items walked: 5 — 5 PASS / 0 FAIL / 0 SKIP.**

| # | Item | Result |
|---|------|--------|
| 1 | Step 2 leaf-tracing skip invokes `dispatch-find-pr <N>` and forbids title-search/ad-hoc `gh` queries | PASS |
| 2 | Step 4 forbids hand-rolled `gh` phase inference; names `dispatch-phase` + `dispatch-select-target` as the sole phase-derivation paths; flags `gh pr list --search "<N> in:title"` as wrong | PASS |
| 3 | Step 6 relevance-review skip runs `dispatch-find-pr <N>` (PR number → skip review; nothing → `ref-ready` check) | PASS |
| 4 | `dispatch-find-pr` live correlation: `673` → `712` (exit 0); no-PR issue `1` → empty (exit 0) | PASS |
| 5 | `dispatch-find-pr` input validation: missing / non-numeric arg → `usage:` error, exit 1 | PASS |

No bugs found.

Fixture-level `dispatch-find-pr` coverage (matching vs non-matching title — the
#669/#670 regression — no PR, and `DISPATCH_PR_LIST` override) lives in
`test-dispatch-scripts.sh`, which CI runs via `unit-tests.yml`; this PR's CI is
green, so the fixture suite was not re-run here.

natb1:
## `/simplify` report

The `simplify` phase of the dispatch workflow reviewed this PR's changes for
reuse, quality, and efficiency.

### Findings that were fixed

Fix commit `cf97cd3`.

- **`dispatch-find-pr` over-fetched `gh pr list` fields.** The self-issued
  `gh pr list` requested `number,headRefName,isDraft,statusCheckRollup,labels`,
  but the script reads only `number` and `headRefName`. `statusCheckRollup` in
  particular expands every check-run for every open PR — fetched and discarded.
  Trimmed the self-fetch to `--json number,headRefName` and corrected the
  `DISPATCH_PR_LIST` doc comment to state the real minimum field requirement (a
  superset list — e.g. the one `dispatch-phase` needs — is still accepted).
- **Missing issue-prefix disambiguation test.** `dispatch-find-pr` relies on the
  trailing `-` in `startswith("<num>-")` to keep issue `6` from matching branch
  `60-foo`. `dispatch-phase` tests this collision (test #12) but the
  `dispatch-find-pr` block did not. Added the parallel test (issue `6` vs branch
  `60-foo` → empty) plus the matching `gh` stub case.

### Informational findings

- **`dispatch-find-pr` mirrors `dispatch-phase`'s PR-list fetch block and
  branch-prefix `jq` filter.** This duplication is intentional — issue #673
  specified the implementation should "mirror `dispatch-phase`'s correlation
  logic." Extracting a shared `lib.sh` helper would restructure `dispatch-phase`
  (untouched by this PR) and the test harness's script-copying — a worthwhile
  refactor, but broader than this PR and better evaluated on its own. After the
  fix above, the two self-fetches request different field sets, so the remaining
  shared surface is just the `if DISPATCH_PR_LIST … else gh pr list … fi` shape
  and the one-line `jq` filter.
- **A `/dispatch` implement-path run now issues 3 `gh pr list` calls** (Step 2
  and Step 6 `dispatch-find-pr`, plus Step 4 `dispatch-phase`) where the pre-PR
  flow issued 1. The scripts self-fetch by design so the operator need not
  hand-manage a PR list; `/dispatch` is operator-paced, not a hot path, so this
  is acceptable. Step 6's `dispatch-find-pr` is additionally always "no PR" when
  reached — Step 6 runs only after Step 4 derives `implement`, which means no PR
  exists — so it is harmless defense-in-depth, and is explicitly required by the
  issue's acceptance criteria.

### Disregarded findings

- **"`dispatch-find-pr` is over-commented."** Its comment density matches the
  established sibling convention in `dispatch-phase` (an 18-line header, per-`jq`
  comments). The header note that this is the canonical PR-existence check and
  must not be replaced by title search is intentional WHY context that
  reinforces #673's whole purpose.
- **"Test fixtures hand-roll PR JSON instead of using the `make_pr` helper; the
  `title` field is dead data."** The hand-rolled fixtures faithfully encode the
  acceptance criteria's matching-title / non-matching-title pair (the #670
  regression case). `make_pr` cannot carry a `title` field, and would instead
  inject `isDraft`/`labels`/`statusCheckRollup` — three fields `dispatch-find-pr`
  never reads. The `title` field is intentional self-documentation and is
  explicitly disclaimed in the test comment.
- **"Use `lib.sh`'s `resolve_issue_number` for argument validation."**
  `resolve_issue_number` adds current-branch fallback that `dispatch-find-pr`
  deliberately does not want, and `dispatch-phase` — the script #673 directs
  `dispatch-find-pr` to mirror — does not use it either. The inline `^[0-9]+$`
  check matches `dispatch-phase`'s own issue-number discriminator regex.

natb1:

Ran `/review` against current `origin/main` (`c37e554`). Test suite: **134/134 passing**.

**Outcome: no actionable code changes.** The implementation is correct, consistent with `dispatch-phase`, well-tested, and matches #673's acceptance criteria.

### What was checked

- `dispatch-find-pr`'s correlation (`map(select(.headRefName | startswith("<N>-"))) | .[0].number`) is byte-for-byte the same logic `dispatch-phase` uses — single source of truth preserved.
- Robust shell: `set -euo pipefail`, all-digits argument validation, `DISPATCH_PR_LIST` honored with a self-fetch fallback, `gh` failure surfaces as a non-zero exit (no silent empty result).
- `.[0].number // empty` yields clean empty output on no-match — no crash.
- Five test cases including the #670 regression (matching branch prefix, non-matching title) and prefix-collision (issue `6` vs branch `60-foo`).
- No security concerns — `$num` is validated all-digits before reaching `jq`.

### Findings — all informational, none fixed

1. **Step 6's `dispatch-find-pr` call is unreachable-by-design.** Step 6 runs only for the `implement` phase, which `dispatch-phase` returns precisely when no PR matches the `<N>-` prefix; `dispatch-find-pr` uses identical correlation, so it can never print a number when Step 6 is reached. Kept as intentional defense-in-depth and because it's an explicit acceptance criterion of #673.
2. **PR body is stale** — says "four cases / 128/128 passing"; actual is five cases / 134/134 (the `origin/main` merge added tests). Not a code artifact, left as-is.
3. **Step 2 skip-condition wording** mentions the `pr <num> <branch> <phase>` queue-result case, which never actually enters Step 2 (scoped to "open issue with no PR"). Wording carried over from the prose it replaced; harmless, rewording would be scope creep.

natb1:

Ran `/security-review` against the branch after merging `origin/main`.

**Result: no security vulnerabilities identified — no fixes applied.**

The PR adds `dispatch-find-pr` (a small bash helper), prose changes to
`SKILL.md`, and test-harness additions. Findings and dispositions:

| Finding | Disposition |
|---|---|
| _(none)_ | No actionable vulnerabilities found |

Review notes:

- **Input validation** — the `<issue-num>` argument is strictly validated
  against `^[0-9]+$` before any use; the script exits on failure.
- **No injection path** — the validated numeric value reaches `jq` only
  through its safe `--arg` parameterization (a JSON string variable, never
  filter code). No shell-injection or `jq`-filter-injection path exists.
- **`DISPATCH_PR_LIST` handling** — interpolated only as `printf '%s'` input
  piped into `jq`'s strict JSON parser; malformed content causes a clean parse
  failure under `set -e`, not code execution. Environment variables are
  trusted values.
- **`gh pr list` invocation** — fixed command string with no interpolation of
  attacker-controlled data; branch names flow only through `jq`'s JSON parser
  via a pure `startswith` comparison.
- **No privilege boundary crossed** — the output (a PR number or empty)
  selects a workflow path, not an authorization or access decision.

The script follows the established secure pattern of its sibling
`dispatch-phase`: strict regex input validation, `jq --arg` parameterization,
and `printf '%s' | jq` for JSON handling.

PR #713, opened 2026-05-21
budget: decouple scroll-load table re-filter from chart serialization (#578)

### LINKED ISSUE #578: Flaky CI: budget e2e 'budget filter applies to scroll-loaded rows' on mobile
## Summary

The mobile-only Playwright test `budget/e2e/home-scroll.spec.ts:197` ("budget filter applies to scroll-loaded rows") intermittently fails on CI. Observed once on PR #573 (run 25177958803, job 73815145998); passed on retry without any code changes (rerun 73984728240).

```
[mobile] e2e/home-scroll.spec.ts:197 — budget filter applies to scroll-loaded rows
expect(b).toBe("Groceries");
Expected: "Groceries"
Received: ""
```

The test sets `#sankey-budget-filter` to "Groceries", scrolls to load older transactions, then asserts every non-`display:none` row has `dataset.budgetName === "Groceries"`. The failing assertion saw a row with `data-budget-name=""` and no `display:none` style.

## Hypothesis

CI-mobile timing race. A scroll-loaded row appearing with empty `data-budget-name` and no inline hide style requires both:

1. `buildRowParts` produced an unbudgeted row (`txn.budget === null` → empty attribute). See `budget/src/pages/home.ts:58-65`.
2. `filterTable()` did not run after the row was inserted.

The synchronous flow at `budget/src/pages/home-hydrate.ts:316-319` should call `filterTable` immediately after insertion via `notifyChart` → `TRANSACTIONS_APPENDED_EVENT` → listener at `budget/src/pages/home-chart.ts:520`. The strongest race candidate is the silent try/catch at `home-hydrate.ts:282-285` swallowing a `serializeChartTransactions` exception, but that path requires a budget-id mismatch that local 464-test runs do not reproduce.

A latent bug is also possible: `serializeChartTransactions` filters to `t.normalizedId === null || t.normalizedPrimary` (`home.ts:171-181`), so it processes a different subset than `buildRowParts`. With the right seed, a normalized member transaction's budget id could be unique to that member and absent from the primary — `serializeChartTransactions` would then throw on a budget the row-render path didn't surface, the event would never dispatch, and `filterTable` would never apply to the new rows.

## When it recurs — diagnostic plan

Add temporary instrumentation, push a debug commit, capture CI logs, then revert:

1. In `budget/src/pages/home-hydrate.ts` `notifyChart` (around line 278-286), log:
   - count and ids of transactions being serialized,
   - whether `serializeChartTransactions` throws and the message,
   - whether the event is dispatched.
2. In `budget/src/pages/home-chart.ts` `filterTable` (around line 470) and the `TRANSACTIONS_APPENDED_EVENT` listener (line 520-538), log:
   - whether the listener fires,
   - the value of `currentBudgetFilter` when filterTable runs,
   - per-row `dataset.budgetName` and computed `visible`.
3. Run CI; capture logs via `gh run view --log`.

Interpretation:
- **Listener never fires** → `serializeChartTransactions` threw. Fix the asymmetry between `buildRowParts` and `serializeChartTransactions`, or make the chart serializer tolerant of unknown budgets (return `budgetName: null` instead of throwing).
- **Listener fires, filterTable runs, but row appended after filterTable** → race. Fix by calling `filterTable` directly after `insertAdjacentHTML` instead of via the event indirection, or queue a microtask.
- **Listener fires, filterTable hides the row, but later something un-hides it** → broader bug, escalate.

## Mitigation candidates (without diagnosis)

If this becomes noisy enough to act on without root-cause diagnosis, add a polling sentinel to the test's scroll-load expectation (e.g. `await waitForFilterApplied(page, "Groceries")`) that retries until the filter has been applied or 1s passes. Turns the latent flake into a soft retry rather than a hard fail.

## Context

- First observed on PR #573 (schema-consolidation work, branch `563-budget-consolidate-five-per`), commit `0f86512`.
- The PR diff did not touch `budget/src/pages/home.ts`, `home-hydrate.ts`, or `home-chart.ts`.
- Local `run-acceptance-tests.sh budget` passes 464/464 across all three Playwright projects (mobile, tablet, desktop).
- See sibling test at `budget/e2e/home-scroll.spec.ts:155` (category filter) — same mechanism, no observed failures, possibly because it runs first.

## Critical files

- `budget/src/pages/home.ts:58-65, 164-181, 208-240`
- `budget/src/pages/home-hydrate.ts:185, 273-276, 278-286, 289-349`
- `budget/src/pages/home-chart.ts:470-498, 520-538`
- `budget/e2e/home-scroll.spec.ts:8-66, 197-222`

natb1:
Closes #578


natb1:
## Dispatch QA — issue #578

User-acceptance QA pass on the budget app (`/transactions`), public seed data.

**Result: clean pass — 4 PASS, 0 FAIL, 0 SKIP.**

### Pre-QA acceptance check

`run-acceptance-tests.sh budget` (URL mode) — **446 passed, 0 failed, 1 skipped**.
All `home-scroll.spec.ts` tests pass, including *"budget filter applies to
scroll-loaded rows"* (the flake this issue targets).

> The first run reported 3 `prerender.spec.ts` failures (`ENOENT` on
> `budget/dist/index.html`). URL mode runs Playwright against the Vite dev
> server and never builds `dist/` — a QA-harness limitation, unrelated to this
> PR. After `npm run build --prefix budget` produced `dist/`, the re-run was
> fully green.

### Walkthrough

| # | Item | Result |
|---|------|--------|
| 1 | Budget home page loads with seed transactions (118 rows, table + heading + seed notice) | PASS |
| 2 | Category Sankey chart renders (`#category-sankey` with one `<svg>`) | PASS |
| 3 | Budget filter narrows the table — `Food` → 59 visible rows, all `data-budget-name="Food"`; chart re-renders | PASS |
| 4 | Clearing the filter restores all rows — back to 103 visible (Food 59, Housing 23, Vacation 16, unbudgeted 5) | PASS |

No console errors; no 4xx/5xx network failures.

### Scroll-coverage note

The public Firestore seed loads every transaction at once with no scroll
sentinel, so scroll-pagination — the core of #578 — is not reachable from the
seed-data walkthrough. It is exercised with uploaded IDB data and is covered by
the pre-QA acceptance check (`home-scroll.spec.ts`) and the new smoke tests in
`home-scroll.smoke.test.ts`.

Walkthrough GIF: `dispatch-qa-walkthrough-578.gif` (5 frames, 1232×957).

natb1:
## Simplify phase

### Findings that were fixed

Commit `d19dbad`:

- **Duplicated `FilterTableOptions` literal** — the same five-field filter-state
  object was constructed verbatim in `update()` and in the new scroll-load
  listener (the second copy was introduced by this PR). Extracted a
  `currentFilterOpts()` closure in `home-chart.ts`; both call sites now use it.
- **Excessive comment density in the scroll-load listener** — three multi-line
  comments narrated *what* the code does and partly duplicated each other.
  Trimmed the chart-update-block comment to the non-obvious *why* (`render()` is
  called directly instead of `update()`), and replaced the four-line
  table-re-filter-block comment with a one-liner pointing to the listener header.
- **Overlapping comments in `notifyChart`** (`home-hydrate.ts`) — the catch-block
  and dispatch comments both restated "the dispatch must still fire."
  De-duplicated so each explains a distinct point.

### Informational findings

No action recommended — surfaced for reference:

- The `setTimeout(() => { throw error; }, 0)` deferred-rethrow appears 4× in
  `home-chart.ts` (2 pre-existing, 2 added here). It is an established local
  idiom; the file-wide inconsistency with `errorutil`'s `deferProgrammerError`
  is pre-existing and out of scope.
- When the chart container is disconnected, the chart-update block is skipped
  and `allTxns` no longer tracks scroll-loaded batches. This is a deliberate
  trade-off — a disconnected chart is invisible — and is correct as written.
- `filterTable` re-evaluates visibility for *every* row in the table on each
  scroll-load event, making a full scroll-through O(rows²). This is pre-existing
  (the old `update()` did the same) and not worsened by this PR; a targeted fix
  would require handles from each inserted batch back to its rows.

natb1:

Reviewed against current `origin/main`. The fix is correct and well-targeted:
the synchronous chain `renderTransactionRows → notifyChart → dispatchEvent →
listener → filterTable` has no async race, so the only ways `filterTable` was
skipped were the event not dispatching or the listener throwing first — and all
three changes (`resolveBudgetName` no longer throws, `notifyChart` dispatches
outside the catch, the listener splits into two independent blocks) close those
paths. Tests cover the listener's two-block independence and `resolveBudgetName`
degradation; `afterEach` clears `document.body`, so the appended
`#transactions-table` divs don't leak.

### Fixed

- **Misleading `resolveBudgetName` docstring.** The docstring's first sentence
  called an unknown budget id a legitimate scenario ("a budget added after the
  hydration-time map was built"), then its last sentence called the identical
  condition "genuine data corruption" — the table-path throw fires on both, so
  the corruption framing was misleading. Reworded to state the accurate
  rationale: the editable table path stays strict because silently degrading an
  unknown id to unbudgeted would misrepresent a transaction the user edits
  (commit b6c990f).

### Not fixed — for whoever closes #578

- **Acceptance criterion 2 is satisfied for the chart path only.** The criterion
  asks for a scroll-time missing budget id to be handled "rather than throwing."
  This PR does that for the chart (`resolveBudgetName` → `null`), but the table
  path (`buildRowParts`) still throws `DataIntegrityError`, which the scroll
  handler surfaces as "Data error — please re-upload your file." The author
  documented the strict table path as deliberate, and changing it (degrade vs.
  refresh-the-map) is a separate design decision out of scope here — but the
  criterion's wording ("rather than throwing") should be confirmed as satisfied
  before closing #578.

- **(nit) `notifyChart`'s dispatch-outside-catch has no direct unit test.** Low
  value: `resolveBudgetName` no longer throws, so the `notifyChart` catch is
  purely defensive; the realistic path is covered by the smoke test and the
  listener independence by `home-chart.test.ts`. Skipped.

natb1:
## Security review — `dispatch:security` phase

Ran `/security-review` against the branch with `origin/main` merged (HEAD `62de74a`).

**Result: no security vulnerabilities identified. No findings to fix.**

The PR restructures error-handling control flow (decoupling the table re-filter
from chart serialization) and changes `resolveBudgetName` from throwing on an
unknown budget ID to degrading it to `null`. Data-flow analysis confirmed no new
attack surface:

- `resolveBudgetName` (changed) feeds only `serializeChartTransactions`, whose
  output is consumed as `CustomEvent.detail` and by chart aggregation logic —
  never interpolated into HTML.
- Scroll-loaded row HTML is produced by `buildRowParts` / `renderTransactionRows`,
  unchanged by this PR. Those paths keep their own strict `budgetIdToName.get`
  lookup and still pass every interpolated value through `escapeHtml`. The
  `null`-degradation change does not reach the rendering path.
- `filterTable` reads `row.dataset.*` and writes `row.style.display` only — no
  HTML injection sink.
- `assertChartTransactions` still runs as a runtime type guard on the event
  `detail` before the chart consumes it.
- `container.textContent` assignments in the new listener use only static string
  literals.
- The `unbudgetedOnly` chart filter treating a degraded `null` as "unbudgeted"
  is a cosmetic correctness issue in a client-side derived view, not a security
  boundary — authorization is enforced by Firestore rules.

No injection point, no authentication/authorization bypass, and no sensitive
data exposure introduced.

PR #723, opened 2026-05-21
dispatch: scope the process lock to selection and queue waiters

natb1:
Closes #719

`/dispatch` Step 0 acquired a process-level lock and held it for the **entire
tick** — Step 0 through the phase skill. The lock guards only *target selection
and worktree resolution* (two ticks racing to pick the same task or adopt the
same worktree); once a session owns an issue worktree, that worktree is its
claim and per-issue phase work is already serialized by worktree ownership.
Holding the lock through the multi-minute phase skill needlessly serialized that
work, and contention *stopped* the losing tick instead of queueing it.

This turns the lock into a short-lived **selection queue**.

## Unit 1 — `dispatch-acquire-lock`: `--wait` and `--release` modes

- `--wait` polls the lock (re-running the atomic read-check-write each
  iteration) until acquired or `DISPATCH_LOCK_WAIT_TIMEOUT` (default 300s)
  elapses, instead of returning `busy` immediately. On a timeout it prints
  `busy` and writes a wait-duration + holding-PID diagnostic to stderr. Poll gap
  is `DISPATCH_LOCK_WAIT_INTERVAL` (default 5s).
- `--release` clears the recorded PID from the lock file only when it is the
  current session's own — printing `released`, else `noop`. Truncates rather
  than `rm` so a concurrent acquire's open fd is never orphaned.
- The flock + read-check-write is refactored into a `try_acquire` function using
  a command-substitution subshell with a fixed numeric fd, so the flock is
  released between polls. The no-arg `acquire` path routes through the same
  function — behavior is byte-identical to before.
- Tests 8-13 cover the new modes; the 7 original lock tests are unchanged. Full
  suite: 163/163.

## Unit 2 — `SKILL.md`: scope the lock to Steps 0-4

- Step 0 calls `dispatch-acquire-lock --wait`, so a `/dispatch` that loses the
  selection race queues behind the holder rather than bailing.
- A new "Releasing the lock" subsection states the lock covers **Steps 0-4
  only**. The lock is released via `--release` at the Step 4→5 proceed boundary
  (after the `tmp/dispatch-worktree` marker) and before every Step 1-4 stop
  path. Steps 5-7 — the phase skill and relevance review — run lock-free.

## Out of scope

The currently-wedged production `tmp/dispatch.lock` (a stale PID recorded by a
pre-#719 tick) needs a one-time manual clear before a normal `/dispatch` loop
runs clean — this PR fixes future ticks, not the existing stale record.

### DIFF
--- a/.claude/skills/dispatch/SKILL.md
+++ b/.claude/skills/dispatch/SKILL.md
@@ -27,23 +27,60 @@ resolution. Runs unconditionally, whether or not an issue-number argument was
 given.
 
 ```bash
-LOCK=$(.claude/skills/dispatch/scripts/dispatch-acquire-lock)
+LOCK=$(.claude/skills/dispatch/scripts/dispatch-acquire-lock --wait)
 ```
 
-Requires `dangerouslyDisableSandbox: true` — the script writes
-`$PROJECT_ROOT/tmp/dispatch.lock`, which is outside the sandbox write-allowlist
-(same reason `dispatch-sweep` runs that way; see `.claude/rules/sandbox.md`).
+Run this Bash call with **both** `dangerouslyDisableSandbox: true` and an
+elevated `timeout: 600000` (ms):
+
+- `dangerouslyDisableSandbox: true` — the script writes
+  `$PROJECT_ROOT/tmp/dispatch.lock`, which is outside the sandbox write-allowlist
+  (same reason `dispatch-sweep` runs that way; see `.claude/rules/sandbox.md`).
+- `timeout: 600000` — `--wait` blocks on contention until it acquires the lock
+  or `DISPATCH_LOCK_WAIT_TIMEOUT` (default 300 s) elapses, which exceeds the
+  default Bash-call timeout.
 
 Route on `$LOCK`:
 
 - **`acquired`** → this `/dispatch` holds the lock; proceed to Step 1.
-- **`busy`** → another `/dispatch` is active; report "another /dispatch is
-  running — stopping" and **stop** — run no sync, no health gate, no sweep, no
-  selection, and no phase skill.
+- **`busy`** → the wait timeout elapsed without acquiring — a wedged selection
+  in another `/dispatch`. The script's **stderr** carries a one-line diagnostic
+  naming the wait duration and the holding PID; report those and **stop** — run
+  no sync, no health gate, no sweep, no selection, and no phase skill.
 
-The lock is session-scoped and self-healing: when a tick's session ends —
-normally, by crash, or by kill — its PID dies and the next invocation detects
-the stale record and proceeds. A crashed tick never wedges the queue. Same-session
+### Releasing the lock
+
+The lock covers **Steps 0-4 only** — target selection and worktree resolution.
+Steps 5-7 (the phase skill and the pre-implementation relevance review) run with
+the lock **released**.
+
+Release the lock by running:
+
+```bash
+.claude/skills/dispatch/scripts/dispatch-acquire-lock --release
+```
+
+This needs `dangerouslyDisableSandbox: true` (same reason as Step 0); no
+elevated timeout is needed — `--release` returns immediately. It prints
+`released` or `noop`; both are fine — it is a no-op when the lock is already
+released or held by another session, so the skill does not branch on its output.
+
+Release happens at exactly two kinds of point:
+
+- **Proceed path** — as the final action of Step 4, after the
+  `tmp/dispatch-worktree` marker is written, before Step 5.
+- **Every Step 1-4 stop path** — immediately before reporting the stop reason
+  and stopping.
+
+Releasing after Step 4 is safe because the session then owns a worktree (or has
+stopped), and the selection scan skips worktree'd targets — no other tick can
+select the same issue. Later steps cross-reference *Releasing the lock* rather
+than repeating this command.
+
+The lock is scoped to selection and self-healing. The recorded PID is
+session-keyed: if a tick dies before its explicit release, the next tick detects
+the stale record and proceeds, and a `--wait` waiter re-checks holder liveness
+every poll, so it reclaims a dead holder's lock automatically. Same-session
 re-entry (e.g. after a context clear that re-invokes `/dispatch`) re-acquires
 cleanly because the recorded PID matches the re-entering session's own PID.
 
@@ -59,8 +96,9 @@ merge and push are no-ops when local main already equals `origin/main`.
 
 - If the working tree is dirty, stash before invoking — `/commit-merge-push` would
   otherwise commit pending changes directly to `main`.
-- If `/commit-merge-push` reports a merge conflict or push rejection, **stop** and
-  surface the error — do not proceed to target selection.
+- If `/commit-merge-push` reports a merge conflict or push rejection, release the
+  lock (see *Releasing the lock*), then **stop** and surface the error — do not
+  proceed to target selection.
 
 ## 2. Select the Target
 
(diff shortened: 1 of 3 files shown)


natb1:
## Dispatch QA — clean pass ✅

User-acceptance QA for #719 (scope the dispatch lock to target selection; add
`--wait` and `--release` to `dispatch-acquire-lock`). Non-browser path — only
`.claude/skills/dispatch/` files changed, so the walkthrough ran as shell
behavioral checks plus a SKILL.md inspection.

**Items walked: 8 — PASS 8, FAIL 0, SKIP 0.**

| # | Item | Result |
|---|------|--------|
| 1 | Single-shot acquire unchanged (`acquired` / `busy`, exit 0) | PASS |
| 2 | `--wait` acquires immediately when uncontended (free lock + own-session re-entry, `WAIT_TIMEOUT=0`) | PASS |
| 3 | `--wait` against a live foreign holder times out → `busy` + stderr diagnostic, lock untouched | PASS |
| 4 | `--wait` reclaims a holder that goes stale mid-wait → `acquired`, records own PID | PASS |
| 5 | `--release` clears an own-session lock (`released`); no-op against a foreign lock (`noop`, file unchanged) | PASS |
| 6 | Unknown argument → exit 2, error to stderr, no lock file touched | PASS |
| 7 | SKILL.md Step 0 invokes `--wait`; lock release wired at every Step 1-4 stop path and the Step 4→5 boundary | PASS |
| 8 | Full `test-dispatch-scripts.sh` suite: **163/163 passed**, including new `--wait`/`--release`/unknown-argument cases | PASS |

No bugs found. All acceptance criteria for #719 are exercised and pass.

natb1:
## Simplify phase

Ran `/simplify` against the branch diff with three parallel reviewers (code
reuse, code quality, efficiency). The full `test-dispatch-scripts.sh` suite
passes (163/163) after the fixes.

### Findings that were fixed

Fixed in 794d161 (`.claude/skills/dispatch/scripts/dispatch-acquire-lock`):

- **Silent sentinel-decode fallback.** The `try_acquire` and `--release` `case`
  statements had `*)` catch-all branches that silently folded any unexpected
  flock-subshell output into "contended" / "noop". Per `.claude/rules/code-style.md`
  ("prefer clear errors over defensive fallbacks"), both now exit 2 with a loud
  internal-error message naming the unexpected output; the exit-code header doc
  was updated to cover the new exit path.
- **Comment referencing the task.** A mode-dispatch comment described the
  `acquire` path as "byte-identical to the pre-#719 behavior" — reworded to
  describe the behavior itself ("Single-shot: one read-check-write attempt, no
  polling"), since an inline source comment naming an issue number ages poorly.

### Informational findings

Surfaced for reference; no action recommended:

- **`/proc` "live .claude" detection is duplicated** between `dispatch-acquire-lock`'s
  `is_live_claude` and `dispatch-sweep`'s comm check. This is pre-existing — it
  predates this PR — and is already tracked by #714, which is chartered to
  extract it into `lib-claude-proc.sh`. Left untouched here so #714 owns the
  extraction.
- **The `--wait` timeout is a floor, not a ceiling** — worst-case wall time is
  `DISPATCH_LOCK_WAIT_TIMEOUT + DISPATCH_LOCK_WAIT_INTERVAL + one try_acquire`.
  With the 300 s default this stays well under the 600 s harness Bash-call cap,
  and SKILL.md already documents the cap relationship.
- **`WAIT_TIMEOUT=0` against a live foreign holder is not directly tested.** All
  four acceptance-criteria test cases are covered; Test 8 exercises
  `WAIT_TIMEOUT=0` (own-session) and Test 9 the timeout-against-foreign-holder
  path. The `WAIT_TIMEOUT=0` + foreign-holder combination is a marginal extra
  edge of the existing coverage.

### Disregarded findings

- **`try_acquire` / `--release` share a 2-line `exec 9>>; flock 9` subshell
  preamble.** Extracting it would add an indirection layer and another
  subshell-scoping caveat; the divergent business logic (record-PID vs truncate)
  is correctly *not* duplicated. Two lines of fd-open boilerplate are the
  acceptable kind of repetition.
- **`CONTENDED_PID` is a global set as a side effect of `try_acquire`.** This is
  idiomatic bash for a function that must return both an exit status and a
  datum; the global is named clearly and documented in the function's contract
  comment. Capture-and-reparse would add complexity without benefit.
- **Test 10 captures stderr (`2>&1`) where sibling lock tests use `2>/dev/null`.**
  Deliberate: Test 10 runs `--wait` backgrounded, and `2>&1` surfaces
  diagnostics from a run whose stderr would otherwise be invisible if the
  assertion failed.
- **The arg-parse `exit 2` branches for `$1` vs `$2` look near-duplicate.** They
  are two one-line error exits with distinct messages (unknown first argument
  vs. an extra second argument), not duplicated business logic; the current
  form is clear.

natb1:
## Fixed

_None._ — no in-scope correctness bug. The `try_acquire` flock semantics, the
`--release` truncate-not-`rm` discipline, and the SKILL.md release matrix
(every defined Step 1-4 stop path + the Step 4 proceed boundary) are all
correct.

## Informational

- `--wait` is a **bounded** wait, not an unbounded queue: after
  `DISPATCH_LOCK_WAIT_TIMEOUT` (default 300s) the loser prints `busy` and the
  SKILL.md `busy` route still **stops** the tick. The code and SKILL.md
  document this honestly; noting only that the PR's "queue waiters" framing is
  a 300s-bounded queue.
- Step 0 correctness depends on the model setting `timeout: 600000` on the
  `--wait` Bash call — a wait needing >120s would otherwise be killed by the
  default Bash-call timeout. The SKILL.md instructs this explicitly; no
  code-level fallback is possible since the script cannot control the harness
  timeout.

## Dismissed

- `--release` on an empty/unheld lock file (`noop`) and a trailing extra
  argument (`$2` rejection) are not directly unit-tested — both trivially
  correct, low value.
- "The phase skill in Step 5 onward runs lock-free" names Step 5 where the
  phase skill itself is Step 6 — minor imprecision; "Step 5 onward" is the
  correct lock-free scope, so the statement is not wrong.

## Deferred

- The lock's holder-liveness check (`is_live_claude`) treats **any** process
  whose `comm` starts with `.claude` as a valid `/dispatch` lock holder —
  including `.claude-unwrapped --bg-spare` background-spare daemons, which are
  long-lived and are not `/dispatch` sessions. This is the root cause of the
  production deadlock observed while dispatching this PR: `tmp/dispatch.lock`
  was wedged on PID 527052, a live `--bg-spare` daemon, forcing every
  `/dispatch` to report `busy`. This PR's `--release` + Steps-0-4 scoping
  shrink the leak window for cleanly-completing ticks but leave the recurrence
  vector open (a tick whose session-PID resolves to / is recycled onto a spare
  daemon and then fails to release). Out of scope here — this PR is the
  lock-scoping change and is already past QA/simplify; hardening PID
  identification warrants its own QA pass. → #724

natb1:
## 🔒 Security review

Ran `/security-review` against the full branch diff — 3 files: `dispatch-acquire-lock` (logic), `SKILL.md` (docs), `test-dispatch-scripts.sh` (tests).

**Findings: none.** No HIGH, MEDIUM, or LOW security issues; no fixes were required.

The new code in `dispatch-acquire-lock` — the `--wait` poll loop, the `--release` truncation path, and the `try_acquire` sentinel decoding — was reviewed for injection, path traversal, and privilege-boundary issues:

- The `recorded` PID (the only input that is neither an env var nor argv) is read with `IFS= read -r` and used only in quoted `[[ ]]` comparisons and as a quoted argument to `is_live_claude` — no word-splitting, glob expansion, `eval`, or command substitution touches it.
- `--release` truncates only `$LOCK_FILE`, a path resolved from a trusted env var or `git rev-parse`; the truncation is `flock`-guarded and gated on `recorded == SESSION_PID`, so it only ever clears the script's own lock.
- `SESSION_PID`, `WAIT_TIMEOUT`, and `WAIT_INTERVAL` come from `$PPID` / the `/proc` walk / trusted env vars.

The lock is a cooperative advisory mutex between same-user `/dispatch` ticks — not an authentication, authorization, or trust boundary. The hardening commit (unexpected subshell sentinel → `exit 2`) is a strict robustness improvement with no security downside.

PR #734, opened 2026-05-21
dispatch: harden lock holder identification against --bg-spare daemons

natb1:
Closes #724

### DIFF
--- a/.claude/skills/dispatch/scripts/dispatch-acquire-lock
+++ b/.claude/skills/dispatch/scripts/dispatch-acquire-lock
@@ -9,8 +9,9 @@
 # An fd-based `flock` cannot serialize /dispatch: the skill runs as many
 # separate Bash tool calls, each a fresh shell, so an fd flock would release
 # the instant its Bash call exits. The lock therefore keys on the Claude
-# *session* process — the nearest ancestor whose comm starts with ".claude" —
-# whose lifetime spans the whole tick. The recorded PID is the lock holder.
+# *session* process — the nearest ancestor whose comm starts with ".claude"
+# and is not a background daemon (e.g. a `--bg-spare` spare) — whose lifetime
+# spans the whole tick. The recorded PID is the lock holder.
 #
 # Release is implicit and self-healing: when the session dies its PID stops
 # being a live ".claude" process, so the next tick's read-check-write reclaims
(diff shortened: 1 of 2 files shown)


natb1:
## Dispatch QA — clean pass ✅

User-acceptance QA for #724: hardening `dispatch-acquire-lock` so a
`--bg-spare` background daemon's PID is never treated as a live lock holder.

Pure shell-script change → **non-browser path**. The walkthrough ran against
the host's **real `/proc`** and the **real `dispatch-acquire-lock` script** —
coverage the synthetic-`/proc` unit tests (which CI runs) cannot provide.

### Items walked — PASS 3 / FAIL 0 / SKIP 0

| # | Item | Result |
|---|------|--------|
| 1 | A real `--bg-spare` daemon is indistinguishable from a session by `comm` alone | PASS |
| 2 | Production wedge reproduced and fixed — a lock recording a live `--bg-spare` PID is reclaimable | PASS |
| 3 | Regression — a live, genuine `.claude` process is still honored as a holder | PASS |

**Item 1** — live `--bg-spare` daemon PID 553930 reports `comm` = `.claude-unwrapp`,
byte-identical to a real `/dispatch` session's comm; only its NUL-separated
`cmdline` (carrying `--bg-spare`) distinguishes it. This is exactly the
ambiguity the old comm-only check could not resolve.

**Item 2** — a scratch lock file recording the live `--bg-spare` daemon PID
553930 was handed to the real `dispatch-acquire-lock`: it printed `acquired`
(exit 0) and reclaimed the lock. The production deadlock — every `/dispatch`
reporting `busy` forever against a wedged `--bg-spare` PID — is resolved.

**Item 3** — a scratch lock file recording live PID 553867
(`.claude-unwrapped agents`, no `--bg-spare`) still produced `busy` with the
lock untouched. The new cmdline-marker check rejects only `--bg-spare`, never
a genuine `.claude` session.

### Coverage note

The issue's second acceptance criterion — the session-PID ancestry walk skips
a `--bg-spare` ancestor — cannot be reproduced against real `/proc` (it needs a
`--bg-spare` daemon as an actual ancestor of the script). It is covered by
synthetic-`/proc` **Test 15**, which CI runs. The walk uses the same
`is_live_claude` predicate hardened in this PR.

### Sanity check

`test-dispatch-scripts.sh` — **169/169 passed**, including Test 14 (bg-spare
holder reclaimable) and Test 15 (ancestry walk skips a spare ancestor).

No bugs found.

natb1:
## `/simplify` — simplify phase

Ran the three-agent simplify review (reuse, quality, efficiency) over the
branch diff (`dispatch-acquire-lock` + `test-dispatch-scripts.sh`). The change
is clean — **no code fixes applied**.

- **Reuse:** No duplication. No `lib-claude-proc.sh` exists yet, so the
  cmdline-marker logic is genuinely new; the `--bg-spare` exclusion is not
  reimplemented elsewhere. The pre-existing `/proc/<pid>/comm` read shared
  between `dispatch-acquire-lock` and `dispatch-sweep` predates this branch and
  is the consolidation #714 tracks — out of scope here.
- **Efficiency:** The added `cmdline` read in `is_live_claude` is gated behind
  the `comm` short-circuit and costs one `stat` + one `tr` fork per call —
  negligible for a lock script invoked a handful of times per tick.
- **Quality:** No redundant state, parameter sprawl, or deep nesting. The
  explicit `return 0`/`return 1` ladder and `tr` command-substitution edge
  cases (empty cmdline, unreadable file) are all handled correctly.

### Informational findings

- **`dispatch-acquire-lock:104-105`** — the inline comment above the `cmdline`
  read partly restates the "unreadable cmdline → comm-only verdict" rationale
  already given in the `is_live_claude` header (lines 96-97). Left as-is: the
  comment's facts (the NUL-separated `/proc` cmdline format; the deliberately
  surprising `|| return 0` on a failed read) are both non-obvious, and the
  inline placement aids local comprehension at the point of use. Trimming it
  is a wash, not a clear improvement — recorded for human reference only.

natb1:
## Fixed

_None._ The review found no in-scope code changes to apply — the implementation is correct and the full `test-dispatch-scripts.sh` suite passes (169/169).

## Informational

- `is_live_claude` now means "is a live Claude *session* process" — it rejects `--bg-spare` daemons — while the function name is unchanged. The updated header comment documents the session semantics, so a rename would be cosmetic churn and is unnecessary.
- `dispatch-sweep` (line 241) classifies a worktree as active via the same `.claude*` comm-only prefix test, with no cmdline check; a `--bg-spare` daemon whose cwd lies inside a worktree would be mis-classified as an active session. This is out of scope for #724 (scoped to the lock script). #714's planned `lib-claude-proc.sh` extraction is the natural place to carry this `--bg-spare` exclusion across to the sweep — landing #724 first means #714 rebases onto the hardened detection.

## Dismissed

_None._

## Deferred

_None._ The `dispatch-sweep` parallel-detection observation is already covered by the planned consolidation in #714; filing a separate issue would duplicate it.

natb1:

`/security-review` ran against the branch diff (`dispatch-acquire-lock` +
`test-dispatch-scripts.sh`). **No vulnerabilities found** — no fixes applied.

### Findings

|---|---|---|
| _(none)_ | — | — |

The security review reported no HIGH, MEDIUM, or LOW findings.

### Why the change is clean

This PR hardens a local concurrency-control lock primitive — it adds a
read-only `/proc/<pid>/cmdline` inspection and a bash glob-pattern substring
match to exclude `--bg-spare` background daemons from lock-holder identification.

- **No injection.** `/proc/<pid>/cmdline` content is piped through
  `tr '\0' ' '` and matched only with `[[ "$cmdline" == *"$marker"* ]]` — a pure
  string comparison. No `eval`, no command substitution on the content, no path
  interpolation. Untrusted process cmdline data cannot inject shell commands.
- **No new path-traversal vector.** The recorded PID flows into the `/proc`
  path, but `is_live_claude` performs only a read-only `-r` test plus a string
  match. The recorded PID already flowed into `comm_file` before this PR; the
  worst case is a wrong busy/acquired verdict, not disclosure or code execution.
- **No privilege boundary crossed.** All `/dispatch` ticks run as the same local
  user; there is no network surface and no untrusted-user threat model.
- **Conservative fallback.** An unreadable cmdline file falls back to the
  comm-only verdict, preserving prior behavior.
- **Test changes** (`lock_proc_cmdline` helper, Tests 14–15) write synthetic
  files under the test-only `$DISPATCH_LOCK_PROC_ROOT` — no security relevance.

PR #737, opened 2026-05-21
dispatch: accept a PR number, resolving it to the issue it closes

natb1:
Closes #736

`/dispatch` documents its optional argument as an issue number, but every
downstream `dispatch-*` script keys on the `<issue>-<slug>` branch convention.
Passing a PR number silently misbehaved — `dispatch-find-pr` found no branch,
`dispatch-phase` derived `implement`, and `dispatch-resolve-worktree` created a
fresh empty worktree on a wrong `<pr#>-…` branch.

This adds a `dispatch-resolve-arg` helper that normalizes the `/dispatch`
argument before any downstream script runs:

- An issue number passes through unchanged.
- A PR number resolves to the single issue it closes, via GitHub's `Closes #N`
  closing-issue references (`gh pr view --json closingIssuesReferences`).
- Issue vs PR is discriminated by the `pull_request` key on the REST issues
  endpoint (which returns PRs too).
- A PR closing zero issues, a PR closing multiple issues, or a number that is
  neither each stop `/dispatch` with a clear error and create no worktree.

`SKILL.md` Step 2's argument branch is wired to the new script; the no-argument
queue path is untouched. `test-dispatch-scripts.sh` gains gh-stub arms and a
7-case section covering issue pass-through, leading-`#` tolerance, PR→issue
resolution, and the zero/multiple/not-found/non-numeric error paths.

### DIFF
--- a/.claude/skills/dispatch/SKILL.md
+++ b/.claude/skills/dispatch/SKILL.md
@@ -9,8 +9,9 @@ Selects the single most pressing task, resolves its worktree, derives the curren
 workflow phase from PR/issue status, and dispatches **exactly one phase skill** —
 then stops. Re-invoke `/dispatch` (or `/loop /dispatch`) to advance to the next phase.
 
-`/dispatch` takes an **optional issue-number argument** (leading `#` optional). With
-an argument, it targets that issue and skips the queue scan.
+`/dispatch` takes an **optional issue-or-PR-number argument** (leading `#`
+optional). With an argument, it targets that issue and skips the queue scan; a
+PR number is resolved to the issue that PR closes (see Step 2).
 
 Run `/dispatch` from the **main worktree**, or from inside an issue worktree to
 continue that issue. Step 4 switches into the target's worktree via `EnterWorktree`;
@@ -102,8 +103,26 @@ merge and push are no-ops when local main already equals `origin/main`.
 
 ## 2. Select the Target
 
-- **Issue argument given** → strip any leading `#`; that issue is the target.
-  Skip the queue scan.
+- **Issue or PR argument given** → strip any leading `#`, then normalize the
+  number to a target issue via the resolver script (`dangerouslyDisableSandbox:
+  true` — it calls `gh`):
+
+  ```bash
+  TARGET=$(.claude/skills/dispatch/scripts/dispatch-resolve-arg <arg>)
+  ```
+
+  An issue number passes through unchanged; a PR number resolves to the single
+  issue that PR closes (GitHub's `Closes #N` closing-issue references). Route on
+  the exit code:
+
+  - **Exit 0** → `$TARGET` is the target issue number; that issue is the target.
+    Skip the queue scan.
+  - **Non-zero exit** → release the lock (see *Releasing the lock*), report the
+    script's stderr message, and **stop**; create no worktree. This covers a PR
+    that closes no issue, a PR that closes more than one issue, and an argument
+    that is neither an issue nor a PR — consistent with the other Step 1-4 stop
+    paths.
+
 - **No argument** → run the `origin/main` CI health gate first, then the
   worktree sweep, then target selection. Both gh-calling scripts need
   `dangerouslyDisableSandbox: true`.
--- a/.claude/skills/dispatch/SKILL.md
+++ b/.claude/skills/dispatch/SKILL.md
@@ -24,8 +24,8 @@ Run `gh` commands (`gh label create`, `gh pr edit`, and the scripts that invoke
 
 Run this as the **very first action** — before the `origin/main` sync, the
 `origin/main` health gate, the worktree sweep, target selection, and worktree
-resolution. Runs unconditionally, whether or not an issue-number argument was
-given.
+resolution. Runs unconditionally, whether or not an issue-or-PR-number argument
+was given.
 
 ```bash
 LOCK=$(.claude/skills/dispatch/scripts/dispatch-acquire-lock --wait)
--- a/.claude/skills/dispatch/scripts/dispatch-resolve-arg
+++ b/.claude/skills/dispatch/scripts/dispatch-resolve-arg
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+# Resolve a /dispatch argument (issue or PR number) to an issue number.
+# Usage: dispatch-resolve-arg <number>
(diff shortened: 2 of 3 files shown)


natb1:

**Issue #736** — `/dispatch` accepts a PR-number argument, resolving it to the
issue the PR closes.

Non-browser PR (changes under `.claude/skills/dispatch/`). The stubbed unit
tests in `test-dispatch-scripts.sh` are already run by CI, so this QA pass
exercised the new `dispatch-resolve-arg` helper against the **live GitHub API**
to confirm the real `gh` response shapes match what the script expects.

### Items walked (5)

| # | Item | Result |
|---|------|--------|
| 1 | Issue number passes through unchanged (`resolve-arg 736` → `736`, exit 0) | PASS |
| 2 | Leading `#` is accepted (`resolve-arg '#736'` → `736`, exit 0) | PASS |
| 3 | PR number resolves to the issue it closes (`resolve-arg 737` → `736`, exit 0) | PASS |
| 4 | Unknown number → clear error, exit 2 (`resolve-arg 999999`; stderr names `#999999` as not found) | PASS |
| 5 | Non-numeric argument → clear usage error, exit 1 (`resolve-arg abc`; stderr shows the bad argument) | PASS |

**Counts:** PASS 5 · FAIL 0 · SKIP 0

No bugs found.

### Coverage note

The *PR-closes-zero-issues* (exit 3) and *PR-closes-multiple-issues* (exit 4)
error paths require manufactured GitHub state to reproduce against live data.
They are covered by `test-dispatch-scripts.sh` cases 4 and 5, which CI runs on
every push — not re-walked here.

natb1:
## `/simplify-fix` report — `simplify` phase

### Findings that were fixed (commit c18d119)

- **Awkward `if ...; then :; else` control flow** — the `gh api` call used an
  empty `:` true-branch with the error handling in the `else`; rewritten as a
  negated `if !` guard, matching the sibling `dispatch-*` scripts.
- **`gh api` errors silently misreported as "not found"** — `2>/dev/null`
  swallowed *every* `gh api` failure and the script treated any non-zero exit as
  a 404, so an auth or network failure was reported as "#N is neither an issue
  nor a PR." It now captures stderr (`2>&1`), reports a genuine 404 as exit 2 and
  any other failure as a distinct exit-1 error — the discriminating idiom sibling
  `issue-parent` already uses, and consistent with `.claude/rules/code-style.md`
  ("prefer clear errors over defensive fallbacks").
- **`echo "$var" | jq` → `printf '%s' "$var" | jq`** — aligned with the
  `printf '%s'` pipe convention every sibling `dispatch-*` script uses;
  `printf '%s'` avoids `echo`'s leading-dash/backslash edge cases.
- **Multi-issue error: `jq | tr | sed` → single `jq ... | join(" ")`** — the rare
  "PR closes multiple issues" branch formatted the issue list with three piped
  tools; collapsed into one `jq` expression.
- **Inline WHAT-narrating comments removed** — comments restating what adjacent
  code plainly does (`# Strip a single leading '#'`, `# Validate: must be a
  positive integer`, etc.) and a three-line block that duplicated the header's
  discrimination rationale; the header retains the non-obvious WHY.
- **Tests** — added a missing-argument case (exit 1) and a non-404 `gh` failure
  case (exit 1, not misreported as exit 2), with a stub `.err` fixture branch to
  model a non-404 failure. All 184 dispatch-script tests pass.

### Disregarded findings

- **"Source `lib.sh` / reuse `resolve_issue_number`"** — `resolve_issue_number`
  treats an empty argument as "derive from the current branch name", which would
  be wrong here (a missing argument must be a usage error), and it does not strip
  a leading `#`. Every `dispatch-*` sibling script hand-rolls its own
  integer-validation regex and does not source `lib.sh` — only the smaller
  `issue-*` scripts do. The inline validation is consistent with the script's
  true siblings.
- **"`count` from `jq` could be non-numeric, causing a `[[ -eq ]]` syntax
  error"** — unreachable under `set -euo pipefail`: a failed `gh pr view` or `jq`
  in a command substitution aborts the script before `count` is compared.
  Matches sibling `dispatch-phase`'s identical pattern.

natb1:
## Fixed
- SKILL.md Step 0 wording aligned with the PR's new terminology — "an issue-number argument" → "an issue-or-PR-number argument", matching the updated intro and Step 2: 10448e0

## Informational
- `dispatch-resolve-arg` relies on `set -e` to propagate a `gh pr view` failure rather than wrapping it with a custom message (unlike the `gh api` 404/non-404 handler). This is correct as-is: the number was just confirmed to resolve via the issues API one line earlier, so any `gh pr view` failure is purely transient, and `set -e` surfaces gh's own error — a clear error, not a silent fallback.
- The `pull_request` discriminator's `jq -e 'has("pull_request")' 2>/dev/null` swallows a jq parse error and falls through to the "issue" branch. Misclassification would require `gh api` to return success with malformed JSON — near-impossible — so this is an observation only.

## Dismissed
- SKILL.md Step 2 instructs the reader to "strip any leading `#`" before calling the resolver — harmless belt-and-suspenders redundancy, since `dispatch-resolve-arg` strips `#` itself (`ARG="${ARG#\#}"`) and test case 2 exercises that path.

## Deferred
_None._

natb1:
## Security review

Ran `/security-review` against this branch's changes (`dispatch-resolve-arg`, the
`SKILL.md` Step 2 wiring, and the `test-dispatch-scripts.sh` cases).

**No actionable findings — nothing to fix.**

The only user-controlled input is the numeric `/dispatch` argument. It is stripped
of a leading `#` and validated against the anchored regex `^[1-9][0-9]*$` before
any use, so `$N` is digits-only by the time it reaches the `gh`/`jq` calls:

- `gh api "repos/{owner}/{repo}/issues/$N"` and `gh pr view "$N" ...` — both
  double-quoted single arguments; a digits-only `$N` cannot inject arguments or
  commands.
- All `jq` programs are fixed string literals — no interpolation of user or
  PR-supplied input; `closing_json` is parsed, never `eval`'d.
- The script issues only read-only `gh` API calls, creates no files, and runs no
  `eval` — no new privilege or data-exposure surface.

`test-dispatch-scripts.sh` is test-only infrastructure with controlled fixtures
and never runs in production.

|---|---|---|
| _(none)_ | — | No vulnerabilities identified |

PR #744, opened 2026-05-21
budget: replace manual hex parsing in sankey categoryColor with d3-interpolate

natb1:
Closes #681

### DIFF
--- a/budget/package.json
+++ b/budget/package.json
@@ -24,6 +24,7 @@
     "@commons-systems/style": "*",
     "@observablehq/plot": "^0.6.17",
     "d3-hierarchy": "^3.1.2",
+    "d3-interpolate": "^3.0.1",
     "d3-scale": "^4.0.0",
     "d3-scale-chromatic": "^3.1.0",
     "d3-shape": "^3.2.0",
@@ -32,6 +33,7 @@
   },
   "devDependencies": {
     "@types/d3-hierarchy": "^3.1.7",
+    "@types/d3-interpolate": "^3.0.4",
     "@types/d3-scale": "^4.0.9",
     "@types/d3-scale-chromatic": "^3.1.0",
     "@types/d3-shape": "^3.1.8",
--- a/budget/src/pages/sankey-render.ts
+++ b/budget/src/pages/sankey-render.ts
@@ -1,4 +1,5 @@
 import { hierarchy, tree, type HierarchyNode } from "d3-hierarchy";
+import { interpolateRgb } from "d3-interpolate";
 import { schemeTableau10 } from "d3-scale-chromatic";
 import { formatCurrency } from "../format.js";
 import type { CategoryNode } from "./category-node.js";
@@ -20,14 +21,8 @@ const NODE_SCALE = 0.6;
 function categoryColor(topLevelIndex: number, depth: number): string {
   const base = schemeTableau10[topLevelIndex % schemeTableau10.length];
   if (depth <= 1) return base;
-  const r = parseInt(base.slice(1, 3), 16);
-  const g = parseInt(base.slice(3, 5), 16);
-  const b = parseInt(base.slice(5, 7), 16);
   const factor = Math.min(0.3 * (depth - 1), 0.6);
-  const lr = Math.round(r + (255 - r) * factor);
-  const lg = Math.round(g + (255 - g) * factor);
-  const lb = Math.round(b + (255 - b) * factor);
-  return `#${lr.toString(16).padStart(2, "0")}${lg.toString(16).padStart(2, "0")}${lb.toString(16).padStart(2, "0")}`;
+  return interpolateRgb(base, "#fff")(factor);
 }
 
 function tooltipText(data: CategoryNode, rootValue: number): string {
--- a/budget/src/pages/sankey-render.ts
+++ b/budget/src/pages/sankey-render.ts
@@ -18,11 +18,13 @@ const SVG_NS = "http://www.w3.org/2000/svg";
 const OVERLAP_GAP = 8;
 const NODE_SCALE = 0.6;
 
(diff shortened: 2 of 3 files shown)


natb1:
## Dispatch QA — issue #681

**Result: clean pass** — 4 PASS / 0 FAIL / 0 SKIP.

Walkthrough ran in **non-browser mode**: the Chrome `navigate` permission was
denied, so per the skill the QA pass fell through to the non-browser path. For
this PR that is a stronger verification — the change is a pure refactor of
`categoryColor`, so equivalence can be proven exactly rather than spot-checked
visually.

### Items walked

| # | Item | Result |
|---|------|--------|
| 1 | Sankey chart renders with category colors | PASS |
| 2 | Top-level category nodes use base palette colors | PASS |
| 3 | Sub-category nodes use lightened colors | PASS |
| 4 | Visual appearance of the sankey chart | PASS (by equivalence proof) |

### How each was verified

- **Item 1** — pre-QA acceptance suite (446 passed, 1 skipped). Sankey-specific
  tests passed, including `sankey chart has node elements`, `sankey chart renders
  with seed data`, `sankey controls present`, and `clicking a Sankey node text
  label sets the category filter` — all exercising `sankey-render.ts` against real
  seed data.
- **Items 2–4** — an equivalence check compared the new `interpolateRgb`-based
  `categoryColor` against the old hand-rolled hex lerp across **all 128
  `(topLevelIndex 0–15, depth 1–8)` combinations**. Every combination produced
  **RGB-identical** output. The depth-1 base-palette path and the `depth >= 3`
  factor clamp at `0.6` are both preserved. The rendered chart is provably
  pixel-identical to the pre-refactor baseline.

The output *string* format changed for lightened depths (`#rrggbb` →
`rgb(r, g, b)`), but the RGB *values* are byte-identical and both are valid SVG
`fill` values.

### Bugs found

None.

### Note on the pre-QA acceptance run

The first acceptance run failed 3 `prerender.spec.ts` tests because
`run-acceptance-tests.sh` external-URL mode runs the QA dev server and never
builds `budget/dist/`, which that spec reads from disk. Building `dist/` and
re-running produced **446 passed, 1 skipped**. This is a harness gap unrelated to
this PR — CI's `acceptance` job builds `dist/` and passed.

natb1:
## `/simplify` review

Reviewed the changed files (`budget/src/pages/sankey-render.ts`, `budget/package.json`, `package-lock.json`) for reuse, quality, and efficiency.

### Findings that were fixed

- **Extracted the `"#fff"` lightening-target literal into a named constant.** `categoryColor` now interpolates toward a module-scoped `LIGHTEN_TARGET` constant instead of a bare `"#fff"` string, matching the existing `OVERLAP_GAP` / `NODE_SCALE` constants in this file and the `NOT_BUDGETED_COLOR` precedent in `budgets-pie-chart.ts`. Fixed in `8f23a41`.

### Informational findings

No action recommended — surfaced for reference:

- **`0.3` / `0.6` magic numbers in the `factor` calculation** are pre-existing — the diff leaves the `factor` line untouched. Not a regression introduced by this PR.
- **Output format change:** `interpolateRgb(base, LIGHTEN_TARGET)(factor)` returns an `rgb(r, g, b)` string rather than `#rrggbb`. Verified safe — the return value is consumed only as an SVG `fill` attribute, which accepts both formats. `budget/test/pages/sankey-render.test.ts` makes no color assertions and passes unchanged.
- **Per-call interpolator allocation:** `interpolateRgb(...)` builds a fresh interpolator on every `categoryColor` call. `categoryColor` runs roughly twice per visible node on an interaction-triggered render path (not an animation loop) dominated by DOM mutation, so the allocation is negligible. Memoizing per `base` color (only 10 distinct values) would add a module-level `Map` and lookup logic, working against this PR's goal of removing hand-rolled code — not worth it.
- **Reuse check:** no existing color-lightening helper exists in the budget app or any shared package; `interpolateRgb` is the correct reuse choice. No other file in the budget app does manual hex/color math.

natb1:
## Fixed

_None._ — the change is correct as-is: `interpolateRgb` defaults to gamma 1, a per-channel linear lerp `c + (255 − c)·factor` that is byte-for-byte the old hex math, and d3-color rounds channels on output exactly as the old `Math.round` did.

## Informational

- `categoryColor` now returns two color-string formats — `#rrggbb` hex for `depth <= 1` (the early `return base`) and `rgb(r, g, b)` for `depth > 1` (`interpolateRgb`'s returned string). Both are valid SVG `fill` values and render identically; the only call sites (`sankey-render.ts:171`, `:196`) pass the result straight to `setAttribute("fill", …)` and never re-parse it, so there is no functional impact.

## Dismissed

_None._

## Deferred

_None._ — adding a direct unit test for the private (non-exported) `categoryColor` helper would require exporting it and contradicts the issue's "test file passes unchanged" criterion; not worth a tracking issue for a 4-line pure function.

natb1:
## Security review

Ran `/security-review` against the branch (merged current `origin/main` first).

**No security vulnerabilities found.**

The change replaces manual hex string parsing in `categoryColor()`
(`budget/src/pages/sankey-render.ts`) with d3-interpolate's `interpolateRgb`,
and declares `d3-interpolate` / `@types/d3-interpolate` as direct dependencies.

`categoryColor()` operates entirely on fixed, non-attacker-controlled inputs —
numeric indices and depths from the d3 layout, and hardcoded color constants
(`schemeTableau10` palette literals and `"#fff"`). Its output is consumed via
`setAttribute("fill", ...)` on SVG elements, which sets an inert attribute value
and is not an HTML/script injection sink. The newly declared packages were
already present as transitive dependencies via `d3-scale-chromatic`, so no new
code enters the dependency tree.

No HIGH or MEDIUM severity findings — no fixes applied.

PR #745, opened 2026-05-21
ci: extend go-tests to cover all Go modules

natb1:
Closes #514

## Summary

The `go-tests` CI job in `unit-tests.yml` and the `go=true` detection in
`detect-changes.sh` were both hardcoded to the `budget-etl` module. Changes to
the repo's other two Go modules — `scaffolding/firebase` and `productivity-tui`,
each with its own `go.mod` and passing test suite — never ran `go test` in CI
and could break silently.

## Changes

- **`detect-changes.sh`** — discovers Go module roots from `go.mod` locations
  and sets `go=true` when a changed path is prefixed by any module root, not
  only `budget-etl/`.
- **`unit-tests.yml`** — the `go-tests` job sets up a single Go toolchain
  (`go-version: stable`, satisfying all modules' `go` directives) and runs
  `go test ./...` in every discovered `go.mod` module, printing a per-module
  header and failing the job if any module fails.

Modules are discovered from `go.mod` locations, so a future Go module needs no
CI edit.

## Verification

All three modules' `go test ./...` pass locally:
- `budget-etl` — green
- `scaffolding/firebase` — green
- `productivity-tui` — green

### DIFF
--- a/.claude/skills/dispatch/scripts/detect-changes.sh
+++ b/.claude/skills/dispatch/scripts/detect-changes.sh
@@ -32,13 +32,10 @@ fi
 if echo "$CHANGED" | grep -qE '^(firestore\.rules$|storage\.rules$|rules-test/|\.claude/skills/dispatch/scripts/|firebase\.json$|package\.json$|package-lock\.json$)'; then
   echo "rules=true" >> "$GITHUB_OUTPUT"
 fi
-# go-tests needs the Go toolchain. Set go=true when any file under a Go module
-# changed. Module roots are discovered from go.mod locations, so a new Go
-# module needs no edit here.
-GO_MODULE_PREFIXES=$(
-  find . -name go.mod -not -path './node_modules/*' \
-    -exec dirname {} \; | sed 's|^\./||; s|$|/|'
-)
+# go-tests needs the Go toolchain. Set go=true when a changed file is under a
+# discovered Go module. list-go-modules.sh discovers module roots from go.mod
+# locations, so a new Go module needs no edit here.
+GO_MODULE_PREFIXES=$("$(dirname "$0")/list-go-modules.sh" | sed 's|$|/|')
 if [ -n "$GO_MODULE_PREFIXES" ]; then
   GO_REGEX=$(printf '%s\n' "$GO_MODULE_PREFIXES" | paste -sd'|' -)
   if echo "$CHANGED" | grep -qE "^($GO_REGEX)"; then
--- a/.claude/skills/dispatch/scripts/detect-changes.sh
+++ b/.claude/skills/dispatch/scripts/detect-changes.sh
@@ -32,8 +32,16 @@ fi
 if echo "$CHANGED" | grep -qE '^(firestore\.rules$|storage\.rules$|rules-test/|\.claude/skills/dispatch/scripts/|firebase\.json$|package\.json$|package-lock\.json$)'; then
   echo "rules=true" >> "$GITHUB_OUTPUT"
 fi
-# go-tests needs the Go toolchain. The budget-etl module is not a package.json
-# workspace, so set go=true when anything in budget-etl/ changed.
-if echo "$CHANGED" | grep -qE '^budget-etl/'; then
-  echo "go=true" >> "$GITHUB_OUTPUT"
+# go-tests needs the Go toolchain. Set go=true when any file under a Go module
+# changed. Module roots are discovered from go.mod locations, so a new Go
+# module needs no edit here.
+GO_MODULE_PREFIXES=$(
+  find . -name go.mod -not -path './node_modules/*' \
+    -exec dirname {} \; | sed 's|^\./||; s|$|/|'
+)
+if [ -n "$GO_MODULE_PREFIXES" ]; then
+  GO_REGEX=$(printf '%s\n' "$GO_MODULE_PREFIXES" | paste -sd'|' -)
+  if echo "$CHANGED" | grep -qE "^($GO_REGEX)"; then
+    echo "go=true" >> "$GITHUB_OUTPUT"
+  fi
 fi
--- a/.claude/skills/dispatch/scripts/list-go-modules.sh
+++ b/.claude/skills/dispatch/scripts/list-go-modules.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Print each Go module's directory (one per go.mod found in the repo),
+# relative to the repo root, sorted. The go-tests CI job and detect-changes.sh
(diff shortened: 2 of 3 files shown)


natb1:
## Dispatch QA — clean pass ✅

Single user-acceptance QA pass, non-browser path (CI/tooling change only).

**Items walked (4):**

1. **`detect-changes.sh` discovers all Go modules** — PASS. The `find`-based
   discovery locates `budget-etl`, `productivity-tui`, and `scaffolding/firebase`
   and builds the alternation regex `productivity-tui/|budget-etl/|scaffolding/firebase/`.
   No module list is hardcoded.
2. **`detect-changes.sh` sets `go=true` for a change under any module** — PASS.
   `go=true` for changed files under each of the three modules (and multi-line
   diffs containing a Go file); absent for frontend, root-doc, and non-Go
   `scaffolding/` subdir changes.
3. **`unit-tests.yml` runs `go test ./...` for every module** — PASS. The loop
   ran `go test ./...` in all three module directories; every suite passed and
   the loop exited 0.
4. **CI fails when any module's `go test` fails** — PASS. With a deliberately
   failing test injected into `productivity-tui/internal/ratelimits`, the loop
   continued to `scaffolding/firebase` (no short-circuit) and exited 1. The
   injected file was removed afterward.

**Results:** PASS 4 / FAIL 0 / SKIP 0

**Bugs found:** none

All four acceptance criteria for #514 are met — the `go-tests` job and
`detect-changes.sh` discover Go modules from `go.mod` locations, so a future
module needs no CI edit.

natb1:
## Simplify phase

`/simplify` reviewed this PR's changes to `detect-changes.sh` and `unit-tests.yml` for reuse, quality, and efficiency.

### Findings that were fixed

Fixed in `c4fc450`:

- **Duplicated Go-module discovery.** The `find . -name go.mod -not -path './node_modules/*'` discovery was written verbatim in both `detect-changes.sh` and the `go-tests` job's run block — the change gate and the test runner could silently drift apart if one was edited without the other. Extracted into a shared `list-go-modules.sh` that both consume, so the module set has a single definition.
- **`Set up Go` step name.** The step had lost its conditional suffix. Restored it to `Set up Go (if Go files changed)` to match the sibling `Install nix (if nix files changed)` / `Set up Java (if firestore.rules changed)` steps.

### Informational findings

- `detect-changes.sh` builds a `grep -E` alternation from the discovered module paths without escaping regex metacharacters. No current bug — the repo's module directories (`budget-etl`, `productivity-tui`, `scaffolding/firebase`) contain no metacharacters, and the trailing-`/` suffix correctly anchors prefix matching. A future module under a path containing `.` or `+` would match too loosely; noted for awareness.
- `detect-changes.sh` and the `go-tests` loop both silently do nothing if module discovery returns an empty list (e.g. a broken checkout) rather than erroring. This is consistent with the file's sibling `nix`/`playwright`/`rules` detection blocks, and an empty module set on a real checkout would surface as a failure elsewhere first.

### Disregarded findings

- A review suggestion to replace `go-version: stable` with `go-version-file: <module>/go.mod` was rejected. Pointing `go-version-file` at a specific module reintroduces a hardcoded module reference, which contradicts issue #514's acceptance criterion that adding a future Go module needs no CI edit. `stable` satisfies every module's `go` directive (current minimums 1.23.3 / 1.24.0) and stays maintenance-free.

### Deferred findings

- `detect-changes.sh` has no automated test coverage — the whole script, not just the new Go discovery, is untested, while `test-dispatch-scripts.sh` covers its sibling dispatch scripts. Building a test harness for it is a separate unit of work. Filed as #749 (blocked by #514).

natb1:
## Fixed

- `go-tests` job silently passed when Go-module discovery failed or returned nothing — the `Run Go tests` step fed `list-go-modules.sh` via process substitution (`< <(...)`), so a discovery failure or empty result ran the loop zero times and exited green having tested nothing. The step now captures the module list into a variable (a discovery failure aborts under `set -e`) and fails explicitly with a `::error::` annotation on an empty result: 37d3bb1

## Informational

- `unit-tests.yml` switches the toolchain from `go-version-file: budget-etl/go.mod` to `go-version: stable`. This is a deliberate, documented choice for a single shared toolchain, but it un-pins the Go version: CI now tracks the latest stable Go, which can surface a Go-version-specific breakage unrelated to a change and differs from a contributor's local Go. Acceptable for this repo's three modules (`go` directives 1.23.3–1.24.0, all satisfied by `stable`); noted as a tradeoff.
- `list-go-modules.sh` filters `node_modules` with `-not -path './node_modules/*'`, which excludes matches but does not prune traversal — `find` still descends into `node_modules`. Harmless here (the `go-tests` job installs no `node_modules`, and any nested `go.mod` would still be filtered out); `-prune` would be marginally faster for local runs.
- Neither `list-go-modules.sh` nor the new Go-detection branch in `detect-changes.sh` has a unit test. The existing `test-dispatch-scripts.sh` suite scopes to the `dispatch-*` decision scripts, and `detect-changes.sh` was already untested, so this matches the current convention rather than introducing a coverage gap.

## Dismissed

- `list-go-modules.sh` sorts without `LC_ALL=C` — not a defect: neither consumer (the CI test loop, the `paste`-joined detect-changes regex) depends on a specific collation order, so locale variance is immaterial.

## Deferred

_None._

natb1:
## Security review — `dispatch:security-reviewed`

Ran `/security-review` against the branch (3 files: `detect-changes.sh`,
`list-go-modules.sh`, `unit-tests.yml`).

**No security findings.** No actionable vulnerabilities, no fixes applied.

Summary of the analysis:

- **Trigger context is benign.** `unit-tests.yml` runs on `push` (not
  `pull_request_target`) with a read-only `GITHUB_TOKEN`. It runs only on
  in-repo branches; fork PRs cannot push to it, so untrusted external input
  never reaches this workflow.
- **No command injection in the discovered-path data flow.** A `go.mod` at an
  attacker-controlled path feeds `dir` into `cd "$dir" && go test ./...`.
  `dir` is a double-quoted argument to `cd` — never passed to `eval` or an
  unquoted expansion — so metacharacters are treated as a literal path.
- **Regex interpolation in `detect-changes.sh`** builds `GO_REGEX` from module
  paths for `grep -qE`. Metacharacters in a directory name could only flip the
  `go=true`/`go=false` gate — no privilege escalation or code execution.
- **`go test ./...` across all modules adds no new code-execution surface.**
  Running `go test` on a contributor branch already executes arbitrary test
  code; generalizing from one hardcoded module to all discovered modules runs
  the same code on the same ephemeral, read-only-token runner. No new secret
  access, no expanded permissions, no new trust boundary crossed.

The PR is a faithful generalization of existing CI behavior and introduces no
exploitable security risk.

PR #750, opened 2026-05-21
url-encode group IDs and path segments in upload-media-core (#697)

natb1:
Closes #697

`core::lookup_group_members` interpolated the operator-supplied `--group <id>`
argument straight into a Firestore REST URL. A `/`, `?`, or `#` in the value
escaped its document-path segment — at best a confusing 404, at worst a request
landing on an unintended document. `core::create_firestore_doc` had the same
latent exposure for its path arguments.

## Changes

- Add two helpers in `scaffolding/scripts/upload-media-core.sh`:
  - `core::url_encode_segment` — percent-encodes a whole value as one path
    segment, so an internal `/` becomes `%2F` (`foo/bar` → `foo%2Fbar`). Used
    for the operator-supplied group ID.
  - `core::url_encode_path` — percent-encodes each `/`-delimited segment,
    preserving the structural separators. Used for the multi-segment constants
    `project`, `groups_path`, `collection_path`.
- `core::lookup_group_members` builds its URL from encoded locals; both error
  messages still quote the original unencoded group ID so output matches what
  the operator typed.
- `core::create_firestore_doc` builds its URL from encoded locals.
- `print/scripts/test-upload-media.sh`: the `curl` stub now logs every
  `https://` URL to `curl-urls.log`; new Test 13 runs `--group 'has/slash'` and
  asserts the URL contains `groups/has%2Fslash` and no raw `groups/has/slash`.
  Test 14 runs the same group ID against a failing lookup and asserts the
  error message quotes the original unencoded `has/slash`.

## Verification

- `bash print/scripts/test-upload-media.sh` — 38/38 passed, 14 test blocks.
- `bash audio/scripts/test-upload-media.sh` — 43/43 passed, no regression.
- `bash -n scaffolding/scripts/upload-media-core.sh` — clean.

### DIFF
--- a/print/scripts/test-upload-media.sh
+++ b/print/scripts/test-upload-media.sh
@@ -77,6 +77,7 @@ for ((i=0; i<${#ARGS[@]}; i++)); do
     -d) HAS_DATA=true; echo "${ARGS[$((i+1))]}" > "$STUB_DIR/curl-body.json"; i=$((i+1)) ;;
     -o) RESP_FILE="${ARGS[$((i+1))]}"; i=$((i+1)) ;;
     --config) i=$((i+1)) ;;
+    https://*) echo "${ARGS[$i]}" >> "$STUB_DIR/curl-urls.log" ;;
   esac
 done
 
@@ -279,6 +280,14 @@ stderr=$(bash "$UPLOAD_SCRIPT" "$TMPDIR_TEST/stub/test-file.cbz" "Title" "epub" 
 assert_eq "exits 1" "1" "$exit_code"
 assert_contains "stderr mentions HTTP code" "HTTP 403" "$stderr"
 assert_contains "stderr mentions cleanup" "gsutil rm" "$stderr"
+teardown
+
+echo "Test 13: --group with slash -> group ID is percent-encoded in Firestore URL"
+setup
+output=$(bash "$UPLOAD_SCRIPT" "$TMPDIR_TEST/stub/test-file.cbz" "Private Item" "epub" --group 'has/slash' 2>&1)
+curl_urls=$(cat "$TMPDIR_TEST/stub/curl-urls.log")
+assert_contains "URL contains percent-encoded group ID" "groups/has%2Fslash" "$curl_urls"
+assert_not_contains "URL does not contain raw slash in group ID" "groups/has/slash" "$curl_urls"
 teardown
 
 echo ""
--- a/print/scripts/test-upload-media.sh
+++ b/print/scripts/test-upload-media.sh
@@ -70,12 +70,14 @@ STUB
 STUB_DIR="$(cd "$(dirname "$0")/.." && pwd)/stub"
 RESP_FILE=""
 HAS_DATA=false
-# Parse args to detect -d (POST) vs GET, and find -o target
+# Parse args: detect -d (POST) vs GET, find -o target, log request URLs.
+# Value-taking flags skip their value so it can't be mistaken for a URL.
 ARGS=("$@")
 for ((i=0; i<${#ARGS[@]}; i++)); do
   case "${ARGS[$i]}" in
     -d) HAS_DATA=true; echo "${ARGS[$((i+1))]}" > "$STUB_DIR/curl-body.json"; i=$((i+1)) ;;
     -o) RESP_FILE="${ARGS[$((i+1))]}"; i=$((i+1)) ;;
+    -w) i=$((i+1)) ;;
     --config) i=$((i+1)) ;;
     https://*) echo "${ARGS[$i]}" >> "$STUB_DIR/curl-urls.log" ;;
   esac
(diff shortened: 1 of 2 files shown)


natb1:
## Dispatch QA — clean pass ✅

User-acceptance QA for #697 (URL-encode group IDs in the shared upload-media core). Scripts-only PR, so the non-browser walkthrough applies.

**Items walked: 4 — 4 PASS / 0 FAIL / 0 SKIP**

| # | Item | Result |
|---|------|--------|
| 1 | `print/scripts/test-upload-media.sh` passes incl. new Test 13 | ✅ PASS — 36/36 |
| 2 | Group ID with reserved char percent-encoded in Firestore URL | ✅ PASS |
| 3 | Error message quotes the original (unencoded) group ID | ✅ PASS |
| 4 | `audio/scripts/test-upload-media.sh` regression (shared core) | ✅ PASS — 43/43 |

**Detail**

- **Item 2** — Test 13 asserts the stubbed `curl` receives `groups/has%2Fslash` and *not* a raw `groups/has/slash`: the `/` stays inside one document-path segment.
- **Item 3** — A targeted 404 run with `--group 'has/slash'` produced stderr `error: group 'has/slash' not found (HTTP 404)` while the recorded URL was `.../documents/print/prod/groups/has%2Fslash` — the operator sees the slash they typed; the wire request is encoded.

**Coverage limitation**

- The acceptance criteria's manual smoke test against a dev bucket is out of scope for automated QA — it needs real Firestore/GCS credentials. The stubbed-`curl` path (items 2–3) exercises the encoding logic without live infrastructure; live verification is deferred to manual operator verification.

No bugs found.

natb1:
## Simplify phase

`/simplify` reviewed the diff for #750 across reuse, quality, and efficiency.
The two new encoding helpers, their comments, quoting, and `local` ordering were
all confirmed correct as written, and `jq` is a pre-existing dependency — the
items below are the actionable findings.

### Findings that were fixed

Fixed in `679b6a5`:

- **Stale curl-stub comment + unguarded `-w` value.** The PR added a `https://*`
  arm to the curl stub's argument loop but left the `# Parse args...` comment
  stale, and the loop did not consume `-w`'s value — leaving the new URL-capture
  arm exposed to a flag-value false-match. Updated the comment and added a
  `-w) i=$((i+1)) ;;` arm so value-taking flags skip their value.
- **Test 14: error-message coverage.** #697's acceptance criterion "error
  messages still display the original (unencoded) group ID" was unverified —
  Test 13 exercises only the success path. Added Test 14: a slash-containing
  group ID against a failing lookup, asserting the error message quotes the raw
  `has/slash`. `print/scripts/test-upload-media.sh` is now 38/38 across 14 test
  blocks; the PR body's verification line was updated to match.

### Informational findings

- `core::lookup_group_members` and `core::create_firestore_doc` encode `project`
  with `core::url_encode_path` (the slash-preserving variant). A GCP project ID
  is always a single segment, so `core::url_encode_segment` would be marginally
  more precise — but the two are behavior-identical for a slash-free value, and
  applying `url_encode_path` uniformly to all three app-controlled constants
  (`project`, `groups_path`, `collection_path`) is a defensible, consistent
  choice. No change made.

### Deferred findings

- **#754** — `print/scripts/attach-markdown.sh:55` builds a Firestore document
  URL by raw interpolation of the operator-supplied `DOC_ID`, the same bug class
  #697 fixes for `group_id`. That script does not source the upload-media core,
  so it is out of scope for #750. Filed as #754, blocked by #697 (the fix should
  reuse the `core::url_encode_*` helpers this PR adds).

natb1:
## Fixed

_None._

## Informational

- The two-helper split is well-matched to the use cases: `core::url_encode_segment` percent-encodes a whole value (so an internal `/` in a document ID becomes `%2F`), while `core::url_encode_path` encodes each `/`-delimited segment so multi-segment constants keep their structure. The `(default)` database literal and the URL skeleton are correctly left outside encoding — routing the whole URL through `@uri` would have broken `(default)`.
- The shared core in `scaffolding/scripts/upload-media-core.sh` is exercised by both `audio` and `print`, but only `print/scripts/test-upload-media.sh` gained the URL-encoding regression tests (Tests 13/14). This matches issue #697's acceptance criteria, and the encoding logic under test is identical for both apps, so coverage is adequate — noting only for awareness.
- Verified locally: `print` suite 38/38, `audio` suite 43/43, `bash -n` clean on `upload-media-core.sh`.

## Dismissed

- The new `-w) i=$((i+1)) ;;` case in the `curl` stub is defensively correct but not strictly required — the `-w` value (`%{http_code}`) never starts with `https://`, so it could not be mis-logged as a request URL. It is harmless and keeps the stub's value-taking-flag parsing consistent; no change needed.

## Deferred

_None._

natb1:
## Security review — `dispatch:security-reviewed`

Ran `/security-review` against the branch after merging current `origin/main`.

**No security vulnerabilities found.**

This PR is a security improvement: it URL-encodes group IDs and path segments
before interpolating them into Firestore REST URLs, closing a path-segment-escape
class of bug.

### Findings and dispositions

|---|---|---|
| _(none)_ | — | No findings — review surfaced no actionable vulnerabilities. |

### Notes

- The only externally-influenced input, `group_id` (the operator-supplied
  `--group <id>` argument), is now confined to a single URL path segment via
  `jq '$s | @uri'`, which percent-encodes all RFC 3986 reserved characters
  (`/`, `?`, `#`, …). This prevents a malformed or malicious group ID from
  escaping its document-path segment or contaminating the query string.
- `project`, `groups_path`, and `collection_path` are hardcoded app constants;
  encoding them is harmless defense-in-depth.
- Error messages still quote the raw unencoded `group_id` (to stderr, not a
  URL) — correct and not an injection vector.
- The `print/scripts/test-upload-media.sh` changes are test-only and not
  reachable as an injection vector.

No fixes were required, so no follow-up commit was made.

Issue #556, opened 2026-04-17
budget: cleared-state auto-suggest from statement-item signal


natb1:
Parent: #551

## Context

During balance-matching reconciliation, the user toggles a "cleared" flag on
each leg the bank has also reported. Legs imported by the ETL already carry a
`statementItemId` (set when the leg's originating transaction was linked to an
OFX line). These legs are, by definition, reported by the bank and are strong
candidates for being marked cleared.

Rather than auto-check them and silently advance the reconciliation, the UI
surfaces them as suggestions the user explicitly confirms. This keeps the user
in control (so import noise or ambiguity stays visible) while saving them from
checking every box by hand.

## What changes

**Hint generator — `budget/src/reconcile-hints.ts`**
Rename and repurpose `budget/src/reconciliation.ts`. Export
`suggestClearedLegs(legs: JournalLeg[], statementItems: StatementItem[], toleranceDays: number): Set<JournalLegId>`.
The existing greedy match algorithm becomes the signal source: it identifies
legs whose `statementItemId` links to a bank-reported line, plus legs whose
amount+date proximity puts them close enough to a statement item to be
confident.

**UI integration**
- Suggested-cleared legs render with a visual highlight (e.g., a "Bank reported"
  badge) and a pre-suggested checkbox state that is visually distinguishable
  from a user-confirmed cleared checkbox (e.g., dotted border) until the user
  clicks to confirm.
- A "Confirm all suggestions" button bulk-promotes the full suggested set into
  true cleared state.
- Uncleared legs without a hint stay neutral.

## Builds on

- #553 (balance-matching UX) — blocker (cleared UI is the integration
  surface)

## Acceptance criteria

- [ ] `suggestClearedLegs` returns the set of legs that should be visually
  suggested as cleared
- [ ] Suggested-cleared legs render with a "Bank reported" badge and a
  distinguishable unconfirmed-but-suggested checkbox state
- [ ] "Confirm all suggestions" button promotes the full suggested set to
  user-confirmed cleared
- [ ] User can override a suggestion by toggling the checkbox to uncleared
- [ ] Unit tests cover: explicit `statementItemId` link → suggested;
  amount+date match within tolerance → suggested; mismatch beyond tolerance →
  not suggested

Issue #553, opened 2026-04-17
budget: balance-matching reconciliation UX


natb1:
Parent: #551

## Context

The current `/accounts/reconcile` view shows a three-column match-pairing layout
(matched, unmatched-items, unmatched-transactions). During QA of #455, it became
clear that the three-column view does not answer the question the user is
actually asking: "does my view of this account agree with the bank's view of
this account?" The user has to scan three columns and reason column-to-column to
tell.

Balance-matching reconciliation answers that question directly. The user enters
the bank's statement-ending-balance, toggles a per-entry "cleared" flag on each
entry the bank has also reported, and the app shows a live comparison of
cleared-total vs. bank-total. On match, the cleared entries lock as reconciled
and a `reconciliation-events` record is written.

This issue delivers the happy-path (match) flow on top of the double-entry data
model. Mismatch UX and cleared-state auto-suggest are tracked separately.

## What changes

**Primary panel — journal-leg list for the selected account+period**
Replace the three-column layout with a single list of `JournalLeg` rows for the
selected account. Columns: cleared checkbox, date, description (from the entry),
amount (signed per the account's type — debit positive for assets/expenses,
credit positive for liabilities/equity/income), running balance.

**Header/sidebar — live totals and reconcile button**
Show cleared balance (sum of signed amounts of legs where `cleared == true`),
statement-ending balance (from the most recent `Statement` record for this
account+period, if available), difference (cleared − statement-ending), and a
Reconcile button.

**Reconcile dialog — match path**
Clicking Reconcile opens a modal. The initial value is the statement-ending
balance (if known). On submit:

- If the user-entered bank balance equals the cleared balance (cent-tolerant):
  for each cleared leg, write `reconciledAt` and `reconciledEventId`; write a
  `reconciliation-events` document with `bankBalance`, `clearedBalance`,
  `adjustment: 0`, `reconciledBy`, `reconciledAt`, `legIds`. Reconciled legs
  render disabled (cannot uncheck), and the reconciliation surfaces in a "Past
  reconciliations" section.
- If mismatch: the dialog surfaces the difference and keeps the dialog open.
  Mismatch-resolution UX is tracked in the sibling issue — for this issue, the
  dialog only shows the difference.

**Remove**
Three-column CSS classes (`.reconcile-columns`, `.reconcile-column-*`,
`.reconcile-match-*`) and their render paths. Aging indicator
(`.reconcile-aging`) remains, now applied to uncleared legs older than 30 days.

## Builds on

[...]

PR #752, opened 2026-05-21
dispatch: propagate gh API failures in dispatch-trace-leaf as hard errors

natb1:
## Summary

Fixes a bug in `dispatch-trace-leaf` where a transient `gh` API failure during
blocker/sub-issue resolution was swallowed, causing `/dispatch` to mis-select a
blocked issue as startable.

`open_children()` resolved blockers via `issue-blocking` and sub-issues via
`issue-sub-issues`, both of which exit non-zero **only** on a real `gh` failure
(empty output on success means genuinely no children). The old `|| ...=""`
swallowed that exit, collapsing "API call failed" into "this issue has no
children" — so a blocked issue looked like a childless startable leaf. This is
how `/dispatch` mis-selected #556 (recorded `blocked_by` open issue #553).

## Changes

All in `.claude/skills/dispatch/scripts/dispatch-trace-leaf`:

- `open_children` propagates an `issue-blocking` / `issue-sub-issues` failure
  (`|| return 1`), drops `2>/dev/null` so the diagnostic surfaces, and adds an
  explicit `return 0` so a genuine no-children result is not misread as a
  failure once callers check `$?`.
- `find_leaf` captures `open_children`'s exit status explicitly (`set -e` is
  disabled inside an `if`-condition call) and returns internal code `3` on a
  hard error, re-propagated through every recursive frame.
- The top level maps code `3` to `exit 1` with a clear stderr message, in both
  `queue` and `explicit` mode (an explicit-mode failure no longer falls through
  to the historical print-N cycle fallback).

Regression tests in `test-dispatch-scripts.sh` cover an `issue-blocking`
failure, an `issue-sub-issues` failure, and an `issue-blocking` failure in
`explicit` mode, via fake-`gh` failure injection.

## Verification

- `test-dispatch-scripts.sh`: 181/181 pass. The 3 new tests fail against the
  pre-fix script, confirming they catch the bug.
- `dispatch-trace-leaf 556 queue` still exits 2 ("all open leaves
  worktree-conflicted") on the no-failure path — behavior unchanged.

Closes #751

### DIFF
--- a/.claude/skills/dispatch/scripts/dispatch-trace-leaf
+++ b/.claude/skills/dispatch/scripts/dispatch-trace-leaf
@@ -138,7 +138,7 @@ find_leaf() {
   # Descend into children lowest-numbered first; return the first leaf found.
   # In queue mode, skip any child that already has a worktree — if every child
   # is skipped the loop finds nothing and we bubble up.
-  local kid leaf rc
+  local kid leaf
   while IFS= read -r kid; do
     [[ -z "$kid" ]] && continue
     [[ "$MODE" == "queue" && -n "${CONFLICTED[$kid]+x}" ]] && continue
--- a/.claude/skills/dispatch/scripts/dispatch-trace-leaf
+++ b/.claude/skills/dispatch/scripts/dispatch-trace-leaf
@@ -28,7 +28,8 @@
 #
 # Exit codes:
 #   0 — a leaf was found and printed to stdout
-#   1 — usage error (bad issue number or missing/invalid mode)
+#   1 — usage error, or a sibling gh script (issue-blocking / issue-sub-issues)
+#       failed — a hard failure the caller must not treat as "no work"
 #   2 — queue mode only: every reachable open leaf is worktree-conflicted
 set -euo pipefail
 
@@ -66,18 +67,21 @@ open_children() {
   local n="$1"
   local children=()
 
-  # Gather open blockers.
+  # Gather open blockers. A non-zero exit from issue-blocking is a real gh API
+  # failure (empty output on success means genuinely no blockers) — propagate
+  # it as a hard error instead of collapsing it into "this issue has no
+  # children". No 2>/dev/null: let issue-blocking's stderr diagnostic surface.
   local blocker_json
-  blocker_json=$("$SCRIPT_DIR/issue-blocking" "$n" 2>/dev/null) || blocker_json=""
+  blocker_json=$("$SCRIPT_DIR/issue-blocking" "$n") || return 1
   if [[ -n "$blocker_json" ]]; then
     while IFS= read -r num; do
       [[ -n "$num" ]] && children+=("$num")
     done < <(printf '%s' "$blocker_json" | jq -r 'select(.state == "OPEN" or .state == "open") | .number' 2>/dev/null || true)
   fi
 
-  # Gather open sub-issues.
+  # Gather open sub-issues. Same hard-error contract as the blocker lookup.
   local sub_json
-  sub_json=$("$SCRIPT_DIR/issue-sub-issues" "$n" 2>/dev/null) || sub_json=""
+  sub_json=$("$SCRIPT_DIR/issue-sub-issues" "$n") || return 1
   if [[ -n "$sub_json" ]]; then
     while IFS= read -r num; do
       [[ -n "$num" ]] && children+=("$num")
@@ -88,6 +92,10 @@ open_children() {
   if [[ ${#children[@]} -gt 0 ]]; then
     printf '%s\n' "${children[@]}" | sort -un
   fi
+  # Explicit success: the function would otherwise fall off the end returning
+  # the [[ ]] test's exit status (1 when there are no children), which a
+  # caller checking $? cannot distinguish from a sibling-script failure.
+  return 0
 }
 
 # Depth-first search for the lowest-numbered reachable open leaf.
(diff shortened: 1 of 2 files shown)


natb1:
## Dispatch QA — clean pass ✅

User-acceptance QA for #751 (`dispatch-trace-leaf` swallows transient `gh` failures and mis-selects blocked issues).

**Path:** non-browser (scripts-only PR — `dispatch-trace-leaf` + `test-dispatch-scripts.sh`).

CI's `hook-tests` job already runs the stub-based unit suite (including the 3 new regression tests), so this walkthrough uses **real `gh` integration checks** that don't duplicate CI — driving the actual `dispatch-trace-leaf` → `issue-blocking` → `gh_api_array` chain. A nonexistent issue number produces a genuine HTTP 404, faithfully modeling the transient `gh` rate-limit/network blip the issue describes.

**Result: 3 PASS / 0 FAIL / 0 SKIP**

| # | Item | Result |
|---|------|--------|
| 1 | Real `gh` failure → hard abort, visible operator diagnostic (queue mode) | PASS |
| 2 | Real `gh` failure → `explicit` mode also exits 1, not the print-N cycle fallback | PASS |
| 3 | No-regression — genuine childless leaf still emitted (happy path, real `gh`) | PASS |

**Item 1** — `dispatch-trace-leaf 999999999 queue` exits `1` with empty stdout (no issue mis-emitted as startable). Stderr carries *both* diagnostics — the fix's central win, since the old `2>/dev/null` swallowed them:
```
error: gh api call failed for /repos/{owner}/{repo}/issues/999999999/dependencies/blocked_by: gh: Not Found (HTTP 404)
error: dispatch-trace-leaf 999999999: a blocker/sub-issue lookup failed (gh API error); aborting rather than treating 999999999 as startable
```

**Item 2** — `dispatch-trace-leaf 999999999 explicit` exits `1` with empty stdout — explicit mode aborts on the hard error rather than falling through to the historical "print N" cycle fallback.

**Item 3** — `dispatch-trace-leaf 751 explicit` exits `0` and prints `751` — the new `return 0` in `open_children` and the explicit status capture leave the genuine no-children path unchanged.

No browser component; no GIF. No bugs found.

natb1:
## Simplify phase

Ran `/simplify` over the PR diff with three parallel review agents (reuse,
quality, efficiency). Reuse and efficiency found the change clean — it routes
through the existing sibling scripts rather than re-implementing `gh` calls,
matches `sync-issue-context`'s hard-fail pattern, and adds no per-node cost (the
failure path now short-circuits the remaining `gh` calls instead of continuing
the walk). Quality found three small cleanups, all fixed.

### Findings that were fixed

Fix commit: 7995d10

- **Redundant `local` re-declaration** — `find_leaf` declared `local kids rc`
  and then later `local kid leaf rc`; the second `rc` just re-declared the same
  function-scope local. Dropped it (`local kid leaf`).
- **Change-narration comment (test 13)** — the test header carried a "Without
  the fix, the swallowed failure makes 800 look like…" sentence that narrated
  the bug being fixed and duplicated the test's own body comment. Trimmed to the
  one-line assertion, matching tests 1–12's house style.
- **Editorializing comment (test 15)** — the test header's "Guards the subtlest
  regression: …" sentence editorialized and restated the first sentence. Trimmed
  to the one-line assertion.

### Informational findings

- `dispatch-trace-leaf` has two comments — at the `open_children` call and at
  the recursive `find_leaf` call — that both note `set -e` is disabled inside an
  `if`-condition. The lower one was left as-is: it also carries the distinct
  point that `rc 3` must be distinguishable from `rc 1`. No action recommended.

All 193 `test-dispatch-scripts.sh` tests pass after the cleanup.

natb1:
## Review

Reviewed the `dispatch-trace-leaf` hard-error propagation change. The core logic is
correct: `open_children` propagates a sibling-script failure (`|| return 1`) with an
explicit `return 0` for the genuine no-children case, `find_leaf` captures the status
explicitly and returns internal code `3`, and the top level maps `3` to `exit 1` in
both modes. Verified the full suite passes (195/195) and the sibling-script contract
the fake stubs model — `issue-blocking` / `issue-sub-issues` exit non-zero only on a
real `gh` failure (via `gh_api_array`), empty output on a genuine no-children result.

## Fixed
- Recursion re-propagation of the `rc`-3 hard error had no test coverage — tests 13–15 all inject the failure at the *root* issue, so `find_leaf` returns `3` from its first frame and the descent-loop branch `[[ "$rc" -eq 3 ]] && return 3` is never exercised. Added test 16: a `gh` failure one level deep (sub-issue `801`'s blocker lookup), confirming the hard error is carried up through a recursive frame rather than masked as "no leaf in this subtree". Mutation-checked — the test fails if the re-propagation line is removed: c0d3f87

## Informational
- The comment "`set -e` is disabled here (`find_leaf` is always called from an `if` condition)" is slightly imprecise: the recursive call `leaf=$(find_leaf "$kid")` is a plain assignment, not an `if` condition. The conclusion still holds — the errexit-suppressed context established by the top-level `if LEAF=$(find_leaf ...)` is inherited through every command-substitution fork into all recursive frames and into `open_children`, so the explicit status capture is correct. No code change needed; noted in case a future refactor of the top-level `if` is contemplated, which would silently break the suppression.

## Dismissed
_None._

## Deferred
_None._

natb1:

The dispatch workflow's `security` phase ran the generic `/security-review`
against this branch's changes (`origin/main` merged first; reviewed at
`c0d3f87`).

**Result: no security vulnerabilities found.**

### Findings and dispositions

No findings. The diff modifies two bash scripts:

- `dispatch-trace-leaf` — converts swallowed sibling-script failures
  (`2>/dev/null` + `|| ...=""`) into hard-error propagation
  (`|| return 1`, exit code `3` → `exit 1`).
- `test-dispatch-scripts.sh` — adds fake-`gh` failure-injection tests (13–16).

The change introduces no new injection, path-traversal, auth, or
data-exposure surface. The only variable interpolated into the new stderr
diagnostic is `$ISSUE_NUM`, already validated against `^[1-9][0-9]*$`. The new
behavior is strictly more conservative: a transient `gh` API failure now aborts
selection instead of mis-classifying a blocked issue as startable.

No fixes were applied — there were no actionable findings.

Marking the PR ready; `security` is the workflow's terminal phase.

PR #762, opened 2026-05-21
dispatch: show the worktree's PR phase in the status line

natb1:
Closes #718

### DIFF
--- a/.claude/hooks/statusline.sh
+++ b/.claude/hooks/statusline.sh
@@ -4,6 +4,9 @@
 # productivity-tui rate-limits header has data to render.
 # Visible output: identical to the user's global statusLine (model | cwd | tokens).
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+DISPATCH_PHASE_TTL=60  # seconds; short TTL — phase changes happen minutes apart so 60s staleness is imperceptible, and it caps `gh pr list` at one call per minute per worktree.
+
 input=$(cat)
 
 # Side effect — write rate_limits.json. Discard stdout (its summary line) and
@@ -25,3 +28,33 @@ if [ "$usage" != "null" ]; then
 else
   printf "\033[36m%s\033[0m | \033[33m%s\033[0m" "$model" "$cwd"
 fi
+
+# Dispatch-phase segment — shows the worktree's /dispatch PR phase. The phase is
+# cached to a file with a TTL and refreshed in a detached background process, so
+# the status line never blocks on dispatch-phase's `gh pr list` network call.
+# `print_dispatch_segment ... || true` keeps a git/dispatch-phase/cache failure
+# from blanking the other segments.
+cwd_raw=$(echo "$input" | jq -r '.workspace.current_dir')
+print_dispatch_segment() {
+  local cwd_raw="$1"
+  local branch issue cache phase_script now mtime phase
+  branch=$(git -C "$cwd_raw" branch --show-current 2>/dev/null) || return 0
+  [[ "$branch" =~ ^([0-9]+)- ]] || return 0          # main / non-issue → nothing
+  issue="${BASH_REMATCH[1]}"
+  cache="$SCRIPT_DIR/../../tmp/dispatch-phase"
+  phase_script="$SCRIPT_DIR/../skills/dispatch/scripts/dispatch-phase"
+  now=$(date +%s)
+  mtime=$(stat -c %Y "$cache" 2>/dev/null || echo 0)
+  if (( now - mtime >= DISPATCH_PHASE_TTL )); then
+    mkdir -p "$(dirname "$cache")" 2>/dev/null
+    touch "$cache" 2>/dev/null                       # bump mtime → no stampede
+    setsid bash -c '
+      p=$("$1" "$2" 2>/dev/null) && [ -n "$p" ] && printf "%s" "$p" > "$3"
+    ' _ "$phase_script" "$issue" "$cache" >/dev/null 2>&1 </dev/null &
+  fi
+  phase=$(cat "$cache" 2>/dev/null) || return 0
+  [[ -n "$phase" ]] || return 0
+  printf " | \033[32m#%s %s\033[0m" "$issue" "$phase"
+}
+
+print_dispatch_segment "$cwd_raw" || true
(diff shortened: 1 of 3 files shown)


natb1:
## Dispatch QA — clean pass ✅

User-acceptance QA for #718 (dispatch-phase status line segment). Non-browser
walkthrough — the real `statusline.sh` hook was run against the real worktrees,
exercising the genuine `git` / `dispatch-phase` / file-cache path end-to-end
(the `test-statusline.sh` unit suite stubs those).

**Items walked: 5 — 5 PASS / 0 FAIL / 0 SKIP.**

| # | Item | Result |
|---|------|--------|
| 1 | Issue worktree shows the dispatch-phase segment | PASS — renders `… \| 17k/200k tokens (8%) \| #718 qa` |
| 2 | `main` worktree renders no dispatch segment | PASS — `model \| cwd \| tokens` only |
| 3 | Phase sourced from `dispatch-phase`, is a valid phase | PASS — hook render `qa` matches `dispatch-phase 718` |
| 4 | Status line renders fast — no blocking network call | PASS — cold run 0.041s; `gh pr list` runs detached |
| 5 | Existing `model \| cwd \| tokens` output + `rate_limits.json` side effect unchanged | PASS — diff is purely additive (0 lines removed) |

Graceful degradation on a transient `dispatch-phase`/`gh` failure is covered by
`test-statusline.sh` case 5 (unit suite, wired into `unit-tests.yml`), so it was
not re-walked here.

No bugs found.

natb1:
## Simplify phase

Reviewed the changed files (`statusline.sh`, `test-statusline.sh`,
`unit-tests.yml`) for reuse, quality, and efficiency.

### Findings that were fixed

All fixed in `a8bc60f`:

- **`statusline.sh`** — `.workspace.current_dir` was parsed by `jq` twice per
  render (once for the display `cwd`, once for `cwd_raw`). Extracted once and
  reused. The status line re-renders roughly every 300ms, so this drops one
  `jq` process per render.
- **`statusline.sh`** — branch issue-number regex tightened from `^([0-9]+)-`
  to `^([1-9][0-9]*)-`, matching `dispatch-phase`, `lib.sh`, and the rest of
  the dispatch tooling (rejects leading-zero issue numbers).
- **`statusline.sh`** — wrapped an over-long (~150-char) TTL comment and
  trimmed the segment comment that restated the code, keeping the non-blocking
  rationale and the `|| true` purpose.
- **`test-statusline.sh`** — `STUB_BRANCH`/`STUB_PHASE`/`STUB_FAIL` resets
  moved into `setup_case` (matching `test-worktree-remove.sh`'s `setup_root`),
  removing 6 ad-hoc per-case `unset` lines and the cross-case env-leak risk.
- **`test-statusline.sh`** — added a `trap ... EXIT` plus per-case `rm -rf` so
  the stub-git dir and per-case temp trees no longer leak, matching
  `test-dispatch-scripts.sh`'s cleanup.
- **`test-statusline.sh`** — `poll_cache` counter renamed `elapsed` → `iters`;
  it counts loop iterations, not seconds.

All 13 statusline tests still pass; both scripts pass `bash -n`.

### Informational findings

- `test-statusline.sh` defines its own assertion harness rather than sourcing
  `.claude/skills/dispatch/scripts/test-helpers.sh`. Consistent with the
  sibling `test-worktree-remove.sh`, and the async-specific assertions
  (`assert_poll`, `assert_cache_not_populated`) have no equivalent in
  `test-helpers.sh`. No change.
- `statusline.sh`'s `mkdir -p ... 2>/dev/null` silently swallows a failed
  cache-dir creation. This is intentional silent degradation, consistent with
  the file's `|| true` style and the requirement that a failure must not blank
  the other segments. No change.
- A few-millisecond window at the TTL boundary lets concurrent renders each
  spawn a `dispatch-phase` refresh. The `touch`-before-spawn debounce bounds
  this to a small handful of processes (not an unbounded herd), and the
  requirement explicitly allows a briefly stale phase. No change.

natb1:
## Review

`/review-fix` ran the generic `/review` against this PR. The dispatch-phase
status-line segment is well-built: a correct `touch`-before-spawn stampede
guard caps `dispatch-phase` at one `gh pr list` per 60s, graceful degradation
(`|| return 0`, trailing `|| true`) keeps a transient failure from blanking the
other segments, and the test suite (6 cases / 13 assertions) covers main,
fresh/stale/missing cache, background refresh, stub failure, and the no-token
render path. All 13 tests pass locally. The visible `model | cwd | tokens`
output and the `rate_limits.json` side effect are unchanged.

## Fixed

_None._

## Informational

- The background cache write `printf "%s" "$p" > "$cache"` truncates-then-writes
  rather than writing to a temp file and renaming. A status-line render whose
  synchronous `cat "$cache"` races the background write could drop the dispatch
  segment for a single frame. It self-heals on the next render, and the
  requirement explicitly tolerates a transient blank dispatch segment — no
  change needed.
- The cache file `tmp/dispatch-phase` is not keyed by issue number. If a
  worktree is recycled for a different issue, the first render after the branch
  change shows the new issue number beside the previous issue's cached phase
  until the background refresh lands (~1s). Negligible and self-healing;
  worktree recycling is rare.

## Dismissed

_None._

## Deferred

_None._

natb1:

Ran `/security-review` against the branch (merged with current `origin/main`).

**Result: no actionable security findings.**

The PR adds a cached dispatch-phase segment to `.claude/hooks/statusline.sh`, a
new test suite (`test-statusline.sh`), and a CI wiring step in
`unit-tests.yml`. All security-relevant data flows newly introduced were
traced:

- The `setsid bash -c '...'` background invocation uses a fixed, single-quoted
  script body with values passed as positional parameters (`"$1"`/`"$2"`/`"$3"`)
  referenced only inside double quotes — the injection-safe pattern. No
  attacker-controlled data is interpolated into shell code.
- `issue` is constrained to digits-only by the anchored regex `^([1-9][0-9]*)-`
  before any use.
- `git -C "$cwd_raw"` quotes the harness-supplied `current_dir`, preventing
  argument/command injection; a bad value fails harmlessly via `|| return 0`.
- `cache` and `phase_script` derive from `SCRIPT_DIR` (the hook's own path), not
  from input.
- `printf` calls use literal format strings with `%s` — no format-string
  injection from `issue` or `phase`.
- The new test file and the `unit-tests.yml` `run:` step are CI/test-only and
  introduce no production attack surface.

No HIGH or MEDIUM severity findings. No fixes were required; no code changed in
this phase.

PR #764, opened 2026-05-21
dispatch: local config files and project helpers for the JIT engine

natb1:
Sub-issue A of #726. Builds the foundation the JIT engine (#729) and due-date
prioritization (#730) both need: a way to read local, non-version-controlled
config, and helpers to add an issue to a GitHub project and read/write a
project item's Status.

- `dispatch-config-load <projects|jit>` — locates `<project-root>/dispatch.config/`
  (overridable via `DISPATCH_CONFIG_DIR`), validates the projects/jit schemas,
  prints `no-config` for an absent file, and reports clear errors on malformed
  config.
- `projects.example.json` / `jit.example.json` — committed schema examples with
  non-identifying placeholders.
- `dispatch-project-item-add` / `dispatch-project-status-read` /
  `dispatch-project-status-write` — GitHub-project helpers that resolve project
  metadata through the loader; the writer resolves field/option ids via
  `gh project field-list`.
- New `test-dispatch-scripts.sh` coverage with a mocked `gh` (208/208 pass).

`SKILL.md` is intentionally unchanged — this sub-issue ships scripts only.

Closes #727

### DIFF
--- a/.claude/skills/dispatch/scripts/dispatch-config-load
+++ b/.claude/skills/dispatch/scripts/dispatch-config-load
@@ -1,0 +1,228 @@
+#!/usr/bin/env bash
+# dispatch-config-load — load and validate a local dispatch config file.
+#
+# Reads non-version-controlled config from a project-root dispatch.config/
+# directory (outside every git worktree, shared across worktrees). Validates
+# the parsed JSON against the schema for the requested config type and prints
+# the normalized result to stdout.
+#
+# Usage:
+#   dispatch-config-load <projects|jit>
+#
+# Config files live in <project-root>/dispatch.config/. The project root is
+# the parent of git --git-common-dir (same idiom as dispatch-acquire-lock).
+#
+# Stdout protocol:
+#   <normalized JSON>   file present and valid — prints jq-normalized JSON.
+#   no-config           file absent — the feature is treated as inert.
+#
+# Exit codes:
+#   0  valid JSON printed, or file absent (no-config).
+#   1  file present but invalid JSON or fails schema validation.
+#   2  usage error or environment misconfiguration.
+#
+# Environment overrides for testability:
+#   DISPATCH_CONFIG_DIR   Override the config directory path. When set and
+#                         non-empty, the script does NOT require a git repo.
+#                         Default: <project-root>/dispatch.config/
+#
+# ---- Schema: projects.json --------------------------------------------------
+#
+# Top-level object with a required "projects" array. Each element is an object
+# with these required fields:
+#
+#   key             string (non-empty) — local identifier for the project.
+#   owner           string (non-empty) — GitHub org or user owning the project.
+#   number          number             — GitHub project number.
+#   statusField     string (non-empty) — name of the project's Status field.
+#   statusInProgress string (non-empty) — Status option name for "in progress".
+#   statusDone      string (non-empty) — Status option name for "done".
+#
+# Example (see projects.example.json):
+#   {
+#     "projects": [
+#       { "key": "example-project", "owner": "example-owner", "number": 1,
+#         "statusField": "Status", "statusInProgress": "In Progress",
+#         "statusDone": "Done" }
+#     ]
+#   }
+#
+# ---- Schema: jit.json -------------------------------------------------------
+#
+# Top-level object with a required "jits" array. Each element is an object
+# with these required fields:
+#
+#   key     string (non-empty) — unique identifier for the jit. Used as the
+#           label suffix: the engine stamps "jit:<key>" on created issues.
+#   repo    string (non-empty) — GitHub repo in "owner/repo" form.
+#   label   string (non-empty) — full label name, typically "jit:<key>".
+#   title   string (non-empty) — issue title.
+#   body    string (non-empty) — issue body.
+#   project string (non-empty) — "key" of an entry in projects.json.
+#
+# Optional timing fields (each must be a string if present):
+#
+#   remindAfterClose  duration string, e.g. "12h" — cadence jit: how long
+#                     after the last closed issue to create the next one.
+#   dueAfterClose     duration string, e.g. "24h" — cadence jit: due time
+#                     offset from the last closed issue's closedAt.
+#   dueAfterCreate    duration string — check-script jit: due time offset
+#                     from the issue's createdAt.
+#   debounce          duration string, e.g. "1h" — how often the jit engine
+#                     re-checks this jit (skips network calls within window).
+#   check             object with "script" (string) and optional "args" keys —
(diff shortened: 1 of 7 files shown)


natb1:

Non-browser path (dispatch scripts + config only). All items machine-verified.

**Items walked:** 5 &nbsp;·&nbsp; **PASS:** 5 &nbsp;·&nbsp; **FAIL:** 0 &nbsp;·&nbsp; **SKIP:** 0 — clean pass.

| # | Item | Result |
|---|------|--------|
| 1 | Committed example config files (`projects.example.json`, `jit.example.json`) are valid JSON with only non-identifying placeholders | PASS |
| 2 | Loader accepts the shipped example files against its documented schema (`dispatch-config-load projects`/`jit` → exit 0, normalized JSON) | PASS |
| 3 | Loader's real `git --git-common-dir` config-dir resolution works end-to-end — with no `dispatch.config/` present, prints `no-config` and exits 0 (a path the unit tests bypass) | PASS |
| 4 | Loader + project-helper behavior covered by `test-dispatch-scripts.sh` — **216/216 passed** | PASS |
| 5 | `.claude/skills/dispatch/SKILL.md` is unchanged by this sub-issue | PASS |

**Bugs found:** none.

**Out of scope (by design):** live invocation of the three project-helper scripts (`dispatch-project-item-add`, `-status-read`, `-status-write`). They call `gh` against a real GitHub project and the writer mutates project `Status` — running them live during QA would add a real issue to a real project. Per acceptance criterion #6, their behavior is verified through the mocked-`gh` test suite (item 4).

natb1:
## Simplify phase — review summary

Three review agents (reuse, quality, efficiency) examined the seven changed files. Findings below.

### Findings that were fixed

All applied in commit `f38c39a`:

- **`dispatch-project-item-add` — validate `gh`'s returned item id.** The script printed `jq -r '.id'` directly; a `gh` exiting 0 with empty or `.id`-less output would emit an empty line and exit 0, so a caller (the JIT engine, #729) would capture an empty item id as success. Now resolves `.id // empty` and exits with a clear error when absent — consistent with `code-style.md` (prefer clear errors over buried ones).
- **`dispatch-project-status-write` — guard the resolved `ITEM_ID`.** The two IDs resolved just above it (`FIELD_ID`, `OPTION_ID`) were guarded; `ITEM_ID` was not. Added the matching non-empty check.
- **`dispatch-project-status-read` — document the `--limit 5000` bound.** `gh project item-list` has no server-side filter for a specific issue URL, so the item is matched client-side. Added a comment explaining the cap and its silent-truncation failure mode (a project larger than the cap could wrongly report an issue as absent).
- **`dispatch-config-load` — removed dead code.** The `PARSE_ERR=""` variable was unconditionally overwritten and never read outside its `if`; dropped it and merged the two parse comments into one.

### Informational findings

Surfaced for reference; no action taken.

- **`dispatch-project-status-write` resolves the config twice.** It loads the config and resolves the project, then calls `dispatch-project-status-read`, which repeats both. A bounded waste (one extra file read + a few `jq` calls per call). The efficiency review rated this acceptable for a script that runs at most once per dispatch tick — fixing it would push pre-resolved project metadata through the reader's interface, adding parameter complexity for no real gain.
- **The project helpers do not use the `gh_api_array` helper from `lib.sh`.** Not a true duplication: `gh_api_array` wraps `gh api` calls returning a top-level JSON array, while these scripts use the higher-level `gh project …` subcommands returning objects. Noted only as a consistency observation for sub-issues B/C (#729, #730), which will add more `gh project` calls.

### Deferred findings

- **Project-root resolution is now triplicated.** The `git --git-common-dir` → `dirname` idiom in `dispatch-config-load` also appears verbatim in `dispatch-acquire-lock` and `dispatch-sweep`; this PR turned a 2× copy into a 3×. The clean fix — a `resolve_project_root()` helper in `lib.sh`, with all three call sites migrated — is a cross-cutting refactor of unrelated dispatch-lock infrastructure and is out of scope for this foundation sub-issue. Tracked in **#773** (blocked by #727).

natb1:
## Fixed

- `dispatch-config-load` did not cleanly reject config whose top-level JSON is not an object: a top-level array/number/string aborted under `set -e` with a raw `jq:` error and exit code 5 (the script documents only exit codes 0/1/2), and an empty/whitespace-only file passed as exit 0 with empty output. Both now exit 1 with a clear `error: <file>: ...` message; two new tests in `test-dispatch-scripts.sh` cover them: 22254d6

## Informational

- `dispatch-config-load`'s jit validator type-checks the optional timing fields (`remindAfterClose`, `dueAfterClose`, `dueAfterCreate`, `debounce`) as strings-if-present, but does not type-check the optional `check` object that the script's own schema comment documents. Likely intentional — the structure of `check` (`script` + `args`) is the JIT engine's concern (#729) — but the omission is inconsistent with the other optional-field checks.

## Dismissed

_None._

## Deferred

_None._

natb1:

Ran `/security-review` against the branch (5 commits ahead of `origin/main`, HEAD `22254d6`).

**Result: no security vulnerabilities found.** No fixes were applied.

### Findings and dispositions

The PR adds four bash scripts (`dispatch-config-load`, `dispatch-project-item-add`, `dispatch-project-status-read`, `dispatch-project-status-write`), two example JSON config files, and `test-dispatch-scripts.sh` additions. Data flow was traced from every input (CLI args, the `DISPATCH_CONFIG_DIR` env var, local JSON config contents, `gh` output) to every sensitive sink (file reads, path construction, `gh` subprocess calls).

- **Command injection — not present.** Every `gh` invocation passes interpolated values (`$NUMBER`, `$OWNER`, `$ISSUE_URL`, `$STATUS_VALUE`, `$ITEM_ID`, `$FIELD_ID`, `$OPTION_ID`, `$PROJECT_NODE_ID`) as individually double-quoted, discrete arguments. No `eval`, no `sh -c`, no backtick command-building, no unquoted expansion feeding a subprocess.
- **Path traversal — not present.** `dispatch-config-load` builds `CONFIG_FILE="$CONFIG_DIR/${CONFIG_TYPE}.json"`; `CONFIG_TYPE` is constrained to exactly `projects` or `jit` by a `case` statement that exits before the path is built, so it cannot carry `../`. `CONFIG_DIR` derives from a developer-controlled env var or `git --git-common-dir`, matching the established `dispatch-acquire-lock` pattern.
- **Code execution via JSON — not present.** Config content reaches only `jq` (as data, with static filter programs) and `gh` (as quoted args).
- **Secrets handling — none added.** The scripts carry no credentials; `gh` manages its own auth. The example JSON files contain only non-identifying placeholders.

The single unquoted word-split (`set -- $args`) appears only in the test `gh` stub — a mock fixture, not production code — so it is out of scope.

Issue #756, opened 2026-05-21
dispatch: self-close and baton-pass — the dispatch-spawn script and /dispatch Step 8


natb1:
## Context

`/dispatch` runs one phase per invocation and then stops — leaving the
background job idle, still owning its worktree, until a human closes it. There
is no mechanism for a job to hand the workflow forward and end itself.

This sub-issue delivers the baton-pass for the autonomous (no-input-needed)
path: a job that completes a phase spawns the next `/dispatch` job and
self-closes.

## Approach

**New `.claude/skills/dispatch/scripts/dispatch-spawn`** — the shared primitive
that starts the next `/dispatch` background job:

- prunes stale stopped `dispatch-*` agents (`claude rm`) by walking the
  on-disk ledger at `~/.claude/jobs/<id>/state.json`, filtering on a terminal
  `state` value (`stopped`, `done`, `failed`) and on the recorded `--name`
  starting with `dispatch-`, and passing each *directory basename* (the
  job-id) to `claude rm`. `claude agents --json` lists only LIVE sessions, so
  the registry is the wrong source for a stopped-agent reaper, and `claude rm`
  wants the job-id — not the registry's `.sessionId`;
- spawns `claude --bg --name dispatch-<short-id> "/dispatch"` from the main
  worktree, in the "auto mode" permission posture #725 specifies;
- is dedup-guarded — spawns nothing if another `dispatch-*` agent is already
  running, via the #739 liveness helper;
- verifies the new agent registered in `claude agents --json`; non-zero exit on
  failure.

Runs with `dangerouslyDisableSandbox` (the `claude` CLI reaches the local
supervisor daemon over a socket).

**New `.claude/skills/dispatch/scripts/dispatch-self-close`** — the self-close
primitive Step 9 calls. `claude stop <id>` expects the *job-id* (the basename
of `$CLAUDE_JOB_DIR`), not the conversation session UUID and not the
registry's `.sessionId`. The script extracts the basename and runs
`claude stop <job-id>`; when `$CLAUDE_JOB_DIR` is unset (an interactive
session, not a managed background job) it is a no-op with a stderr
diagnostic — the foreground-safe gate that prevents an interactive `/dispatch`
from stopping the user's live conversation. Runs with
`dangerouslyDisableSandbox` for the same socket reason.

**`/dispatch` SKILL.md — new Step 9 "Hand off and self-close"** — every "stop"
on the phase-completed and early-stop paths routes here:

[...]

PR #765, opened 2026-05-21
dispatch: run /dispatch daily at 9 AM via a home-manager systemd timer

natb1:
Closes #725

Adds the first `systemd.user` unit in the repo: a oneshot service that runs
`claude --bg --permission-mode auto /dispatch` and a paired timer that fires it
every 30 minutes, so the issue queue advances unattended while still parking in
`Needs input` when a phase needs a human decision.

- `nix/home/dispatch-timer.nix` — new module declaring the service + timer.
- `nix/home/default.nix` — imports the new module.

Design notes:

- `KillMode = "process"` keeps the detached `claude --bg` session alive after
  the oneshot unit deactivates — the background supervisor daemon owns the
  session lifetime, not the timer service.
- `Persistent = true` runs a tick missed while WSL was down on the next start.
- `--permission-mode auto` replicates the user's normal interactive posture, so
  a phase needing a decision parks in `Needs input` rather than auto-denying.
- PATH is set explicitly because a systemd user service does not inherit the
  login shell's PATH and the devShell/direnv environment is unavailable.

The module header documents the WSL prerequisites (`systemd=true`,
`loginctl enable-linger`) and the VM-lifecycle limitation.

The runtime acceptance criteria — timer cadence, spawned session survival,
`Needs input` parking, lock-based overlap exit — are all post-`home-manager
switch` checks verified in the QA phase, not at implementation time.


natb1:
## Dispatch QA — issue #725

Non-browser walkthrough (config-only PR: `nix/home/dispatch-timer.nix` + `nix/home/default.nix`).

**Items walked:** 3 of 5 — walkthrough stopped on the first FAIL.

| # | Item | Result |
|---|------|--------|
| 1 | Nix module evaluates and produces the systemd units | ✅ PASS |
| 2 | Timer fires on a ~30-minute cadence | ✅ PASS |
| 3 | Service `ExecStart` points to a runnable command | ❌ **FAIL** |
| 4 | Module header documents purpose / WSL prereqs / VM-lifecycle limitation | — not walked |
| 5 | `KillMode` keeps the detached session alive | — not walked |

**Counts:** PASS 2 / FAIL 1 / SKIP 0 / not walked 2

### Bug found

**Item 3 — `ExecStart` references a script that does not exist.**

The service's `ExecStart` is:

```
/home/n8/.../worktrees/main/.claude/skills/dispatch/scripts/dispatch-spawn
```

`dispatch-spawn` is not in the repo — `git log --all` for that path is empty. It is
introduced by #756 (the dispatch self-close / baton-pass epic #755), which is still
**open**. If this PR merges and `home-manager switch` runs, the oneshot service fails
on every 30-minute fire with `203/EXEC` (No such file or directory) — the timer never
actually runs `/dispatch`.

This follows the @natb1 comment on #725 ("`ExecStart` should invoke `dispatch-spawn`"),
but that comment assumes #756 has landed first ("with the baton-pass in place"). #725
carries no `blocked_by` dependency on #756, so nothing enforces that ordering.

Two further notes:

- The PR description says the service runs `claude --bg --permission-mode auto /dispatch`,
  but the code uses `dispatch-spawn` — the body and the code disagree.
- What was verifiable passed: the module evaluates, the timer cadence is correct
  (`*:0/30` → every 30 min), and the service `PATH` includes claude/gh/git/node/jq.

Resolution requires a decision on epic merge ordering (block #725 on #756, or make #725
self-contained with a raw `claude --bg` `ExecStart` and swap to `dispatch-spawn` as a
follow-up). QA is paused on that decision.

natb1:
## Dispatch QA — PR #765 (issue #725)

**Result: FAIL** — one bug found on the first walkthrough item. Non-browser
(infrastructure) QA path; single pass, stopped on the first FAIL.

### Items walked

- **1. Timer service `ExecStart` is a runnable command** — ❌ **FAIL** (see below)

Items **2–4** were incidentally confirmed OK during investigation: the module is
imported in `nix/home/default.nix`; `nix eval` of the home configuration
succeeds; `OnCalendar = "*:0/30"` gives a 30-minute cadence; the header comment
documents the module's purpose, the WSL prerequisites, and the VM-lifecycle
limitation.

Item **5** (post-`home-manager switch` runtime criteria — timer registration,
spawned-session survival, `Needs input` parking, lock-based overlap exit) was
**not reached** — it is moot until the bug below is resolved, since the service
cannot exec at all.

**PASS 3 (incidental) · FAIL 1 · SKIP 0 · NOT WALKED 1**

### Bug found — item 1

`nix/home/dispatch-timer.nix` sets:

```nix
ExecStart = "${mainWorktree}/.claude/skills/dispatch/scripts/dispatch-spawn";
```

**`dispatch-spawn` does not exist** — not on this branch, not on `origin/main`,
not in the main worktree. The only file in the repo that mentions
`dispatch-spawn` is this new module.

`dispatch-spawn` is created by **#756** ("dispatch: self-close and baton-pass —
the dispatch-spawn script and /dispatch Step 8"), which is still **open with no
PR**. Issue #725 is recorded as **`blocked_by #756`**.

Consequences:

- When the timer fires, the oneshot `dispatch.service` fails to exec (systemd
  `203/EXEC`) — the queue never advances. The acceptance criterion *"On each
  fire, the service runs `claude --bg "/dispatch"` from the main worktree"* is
  not met.
- **PR body / code mismatch:** the PR description says `ExecStart` runs
  `claude --bg --permission-mode auto /dispatch`, but the committed code (and
  its header comment) instead invoke `dispatch-spawn`.

### Resolution is a sequencing decision

This is not a defect fixable by editing PR #765 in isolation — it is a
sequencing question between two valid directions:

- **A — decouple and ship now:** point `ExecStart` directly at
  `claude --bg --permission-mode auto /dispatch` (matching the PR body and
  #725's own description, which handles concurrency via `/dispatch`'s Step 0
  lock), drop the `dispatch-spawn`/#756 references from the header comment, and
  remove the `blocked_by #756` dependency. PR #765 becomes self-contained and
  mergeable now; #756 later re-points the timer at `dispatch-spawn`.
- **B — hold for #756:** keep the `dispatch-spawn` target; the recorded
  `blocked_by #756` is correct. PR #765 waits until #756 lands and
  `dispatch-spawn` exists.

QA stopped here and is awaiting that decision before applying a fix. No
`dispatch:*` label was applied.

Issue #740, opened 2026-05-21
dispatch: replace dispatch-sweep /proc walk with claude agents --json


natb1:
## Context

`dispatch-sweep` Step 4 classifies each surviving worktree ACTIVE/ORPHANED by walking `$DISPATCH_SWEEP_PROC_ROOT` for `.claude*` processes and resolving their `cwd`. Brittle (comm truncation, symlink resolution, sandbox PID-namespace isolation) and yields only a bare PID for diagnostics.

Sub-issue of #738. Depends on the shared `claude agents --json` liveness helper (sub-issue 1).

## Proposed change

Replace the Step 4 `/proc` walk with the shared helper. For each surviving worktree, classify ACTIVE if a live session's `cwd` is inside it.

- Log `sessionId`, `name`, and `status` on the `ACTIVE` line instead of `pid`.
- Apply the helper's fail-safe contract: if the daemon cannot be queried, classify every surviving worktree ACTIVE (never adopt/clean on an unknown signal).
- Remove the inline `/proc` loop and `DISPATCH_SWEEP_PROC_ROOT`; update the synthetic-`/proc` sweep tests to drive the fake-`claude agents` override.

This is a detection-mechanism swap only. The sweep's downstream behavior — orphan adoption (resuming a potentially interrupted implementation task), merged-cleanup, and `--cleanup-unknown` — is unchanged. The one-orphan-per-tick adoption cadence also stays: once sub-issue 3 lands, liveness-aware tracing no longer mislabels un-adopted orphans.

## Acceptance criteria

- [ ] `dispatch-sweep` Step 4 classifies ACTIVE/ORPHANED via the shared `claude agents --json` helper; the `/proc` walk and `DISPATCH_SWEEP_PROC_ROOT` are removed.
- [ ] `ACTIVE` log lines include `sessionId`, `name`, and `status`.
- [ ] When the daemon cannot be queried, the sweep classifies all surviving worktrees ACTIVE and adopts/cleans nothing.
- [ ] `test-dispatch-scripts.sh` sweep tests are updated to the fake-`claude agents` override and pass.
- [ ] Orphan adoption, merged-cleanup, and `--cleanup-unknown` behavior are unchanged.

PR #768, opened 2026-05-21
dispatch: add shared claude agents --json session-liveness helper

natb1:
Closes #739

Sub-issue 1 of #738 — the foundation. Adds the shared session-liveness
helper only; sub-issues 2-4 later migrate the consuming scripts onto it.

## What

`lib-claude-agents.sh` — a sourceable bash helper backed by
`claude agents --json` (Claude Code >= 2.1.146), mirroring the
`lib-worktree-in-sync.sh` sibling. It exposes two functions:

- `claude_sessions_under <path>` — the low-level primitive. Emits one
  tab-separated line per live session under `<path>`
  (`sessionId<TAB>pid<TAB>status<TAB>name`), or signals **unknown**
  (return 1) for the failures it can detect: `claude` missing, non-zero
  exit, or output that is not a JSON array. An empty `[]` is a
  successful no-sessions result (return 0, no lines) — distinct from
  unknown.
- `worktree_has_live_session <path>` — the fail-safe predicate. Folds
  unknown into the occupied branch, so a caller can never green-light a
  second session on a worktree it could not query.

## Tests

A new `lib-claude-agents.sh` section in `test-dispatch-scripts.sh`
covers a live session, no session, an empty `[]`, a daemon-query
failure, a missing `claude` binary, non-array output, and
multi-session parsing — via a `CLAUDE_AGENTS_CMD` override that injects
a fake `claude` with no real daemon. Full suite: 200/200 passing.

## Docs

`sandbox.md` gains a `claude agents --json` section: the command
reaches the daemon over a Unix socket, so a sandboxed call returns an
empty `[]` indistinguishable from "no sessions" — callers must use
`dangerouslyDisableSandbox: true`.

## Scope

Per acceptance criterion #5, no consuming script (`dispatch-sweep`,
`dispatch-acquire-lock`, `dispatch-resolve-worktree`,
`dispatch-trace-leaf`) is modified.

### DIFF
--- a/.claude/rules/sandbox.md
+++ b/.claude/rules/sandbox.md
@@ -52,6 +52,19 @@ Sandboxed Bash calls run in an isolated network namespace. Servers started with
 Use `dangerouslyDisableSandbox: true` on Bash calls that check local server
 connectivity (e.g., `curl http://localhost:*`, `ss -tlnp`, readiness polls).
 
+## claude agents --json
+
+`claude agents --json` lists live Claude sessions. It reaches the local Claude
+daemon over a Unix socket, which the sandbox's network-namespace isolation
+blocks. A sandboxed call does not error — it returns an empty `[]`,
+indistinguishable from a genuine "no live sessions" result. A caller that
+trusts that `[]` would wrongly conclude a worktree is free.
+
+Use `dangerouslyDisableSandbox: true` on every Bash call that runs
+`claude agents --json` directly, or that runs a script sourcing
+`.claude/skills/dispatch/scripts/lib-claude-agents.sh` (whose
+`claude_sessions_under` / `worktree_has_live_session` helpers shell out to it).
+
 ## pass / GPG pinentry
 
 `pass show <path>` decrypts a GPG-encrypted secret store entry. GPG cannot
--- a/.claude/skills/dispatch/scripts/lib-claude-agents.sh
+++ b/.claude/skills/dispatch/scripts/lib-claude-agents.sh
@@ -59,31 +59,32 @@ if [[ -z "${_LIB_CLAUDE_AGENTS_LOADED:-}" ]]; then
       return 1
     fi
 
-    # Capture stdout and exit code separately — `local out=$(...)` would mask
-    # the exit code. 2>/dev/null drops daemon noise; only the exit code and a
-    # well-formed JSON array on stdout are trusted.
-    local out rc
-    out=$("${CLAUDE_AGENTS_CMD:-claude}" agents --json --cwd "$path" 2>/dev/null)
-    rc=$?
-    if [[ "$rc" -ne 0 ]]; then
-      # Non-zero exit: `claude` missing (127), the daemon unreachable, or any
-      # other failure. The session state cannot be determined — unknown.
+    # 2>/dev/null drops daemon noise; only the exit code and a well-formed
+    # JSON array on stdout are trusted. A non-zero exit — `claude` missing
+    # (127), the daemon unreachable, or any other failure — means the session
+    # state cannot be determined: unknown.
+    local out
+    if ! out=$("${CLAUDE_AGENTS_CMD:-claude}" agents --json --cwd "$path" 2>/dev/null); then
       return 1
     fi
 
-    # A zero exit must still yield a JSON array. Anything else — empty output,
-    # an object, malformed JSON — is unknown, not a definite "no sessions".
-    if ! jq -e 'type == "array"' >/dev/null 2>&1 <<<"$out"; then
+    # A zero exit with empty (or whitespace-only) output is unknown too — not a
+    # definite "no sessions".
+    if [[ -z "${out//[[:space:]]/}" ]]; then
       return 1
     fi
 
-    # Valid array (including `[]`): one TSV line per session. Capture first so
-    # a degenerate array (e.g. of non-objects) that makes jq fail extraction
-    # yields unknown (return 1) rather than partial output.
+    # One jq pass validates the JSON is an array and extracts the TSV. A
+    # non-array — object, scalar, or malformed JSON — hits `error`; a degenerate
+    # element that breaks extraction errors mid-pass; either way jq exits
+    # non-zero and the result is unknown. Capture first so partial output from
+    # a mid-stream error is discarded rather than emitted.
     local lines
-    lines=$(jq -r '.[] | [.sessionId, .pid, .status, .name] | @tsv' <<<"$out" 2>/dev/null)
-    rc=$?
-    if [[ "$rc" -ne 0 ]]; then
+    if ! lines=$(jq -r '
+      if type == "array"
+      then .[] | [.sessionId, .pid, .status, .name] | @tsv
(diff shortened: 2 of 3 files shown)


natb1:
## Dispatch QA — clean pass ✅

User-acceptance QA for the shared `claude agents --json` session-liveness helper (`lib-claude-agents.sh`). Non-browser path — the PR touches only `.claude/` scripts and docs.

The 7 unit tests in `test-dispatch-scripts.sh` already cover the helper against *fakes* (live / empty `[]` / unknown / missing-binary / non-array / multi-session). This QA pass covers what fakes cannot: integration with the **real** `claude agents --json` daemon, plus the user-facing sandbox doc.

**Items walked: 5 — PASS: 5, FAIL: 0, SKIP: 0**

| # | Item | Result |
|---|------|--------|
| 1 | Helper sources cleanly; exposes `claude_sessions_under` + `worktree_has_live_session` | PASS |
| 2 | `claude_sessions_under` parses live `claude agents --json --cwd` output (real daemon) | PASS |
| 3 | `worktree_has_live_session` reports this worktree as occupied | PASS |
| 4 | `worktree_has_live_session` reports a sessionless dir as free | PASS |
| 5 | `.claude/rules/sandbox.md` documents the `claude agents --json` constraint | PASS |

Item 2 ran against the actual daemon and correctly emitted the TSV line for this QA session itself (`sessionId e75165a4-…`, `status busy`) — confirming the helper parses the real output shape, not just hand-crafted fake JSON.

### Observation (not a bug)

The helper is a `#!/usr/bin/env bash` library. Sourced directly into the QA harness's interactive shell (zsh 5.9), it fails to resolve `claude` from a parameter-expansion command word inside a function and reports "unknown". Items 2–4 were therefore evaluated under **bash** — the helper's declared environment, matching its real consumers (the bash dispatch scripts) and `test-dispatch-scripts.sh`. Under bash all items pass. Even on the zsh-misuse path the helper degrades **fail-safe**: "unknown" folds to "occupied", never wrongly to "free" — consistent with the design contract.

No bugs found.

natb1:
## `/simplify` — `lib-claude-agents.sh` session-liveness helper

A `/simplify` pass over the PR's three changed files (`lib-claude-agents.sh`, the
`test-dispatch-scripts.sh` test section, and the `sandbox.md` doc section).

### Findings that were fixed

All fixed in commit a0316f6.

- **Capture-then-check-`$?` replaced with the `if ! var=$(...)` idiom.**
  `claude_sessions_under` captured the `claude` call and the `jq` extraction with
  `out=$(...); rc=$?; if [[ "$rc" -ne 0 ]]; then`. Both now use
  `if ! var=$(...); then` — the form the sibling `lib-worktree-in-sync.sh`
  already uses — dropping the `rc` variable and a stale comment claiming the
  capture form was needed to avoid masking the exit code.
- **Two `jq` invocations merged into one.** The helper ran `jq` twice on the same
  captured output: once for `type == "array"` validation, once for the `@tsv`
  extraction. They are now a single `jq` pass
  (`if type == "array" then .[] | ... | @tsv else error(...) end`), halving the
  per-call `jq` subprocess count. The old `jq -e` type check also implicitly
  rejected empty output as "unknown"; an explicit empty/whitespace-output guard
  now preserves that contract.
- **Added a regression test for zero-exit-with-empty-output.** The contract case
  "`claude` exits 0 but prints nothing → unknown" was uncovered by the prior
  tests, and is exactly the case the single-`jq` consolidation had to handle
  explicitly. Test 8 now locks it in. Full suite: 210/210 passing.

### Informational findings

- The `lib-claude-agents.sh` test section repeats the
  `if out=$(...); then rc=0; else rc=$?; fi` rc-capture idiom (and a parallel
  `live=occupied`/`free` idiom) across its tests. The repetition is exact and
  could fold into a one-line test helper, but it matches the inline-capture
  convention used elsewhere in `test-dispatch-scripts.sh` — no shared rc-capture
  helper exists — and keeps each test self-contained. Left as-is.
- The `type == "array"` validation parallels `gh_api_array` in `lib.sh`. Not
  extracted into a shared validator: `gh_api_array` is coupled to `gh api` and
  *exits* on a non-array, whereas `claude_sessions_under` runs an arbitrary
  command and must *return* "unknown" — a shared helper would cost more clarity
  than the ~2 shared lines save.
- The helper's per-worktree `--cwd` design means a `dispatch-sweep` over N
  worktrees spawns N `claude` subprocesses / socket round-trips. This is a design
  consideration for sub-issue #740 (which migrates `dispatch-sweep` onto the
  helper) and is already in that issue's scope — no separate tracking issue.

### Disregarded findings

- *Trim the helper's 4-line header "Sandbox:" paragraph to a one-line pointer to
  `.claude/rules/sandbox.md`.* The paragraph is already a compressed summary, not
  a verbatim duplicate, and carries the safety-critical WHY — a sandboxed `[]` is
  indistinguishable from a genuine "no sessions". Documenting that caveat inline
  where a maintainer of the helper will see it is consistent with how the sibling
  `lib-worktree-in-sync.sh` documents its own side effects. Not a real
  simplification.

natb1:
## Fixed
- Add a test verifying `claude_sessions_under` invokes `claude` with `agents --json --cwd <path>`: a4c56c1 — the existing fake `claude` ignored its argv, so a regression dropping `--cwd` (reporting every session globally instead of only those under `<path>`) would have passed the suite silently.

## Informational
- The header comment's "unknown" causes list (`claude` missing, non-zero exit, non-array output) omits a missing `jq` binary, which also yields return 1 via the `if ! lines=$(jq …)` guard. The behavior is correct and fail-safe; only the comment is slightly incomplete.

## Dismissed
_None._

## Deferred
_None._

natb1:

`/security-review` examined every change on this branch: the new
`.claude/skills/dispatch/scripts/lib-claude-agents.sh` helper, the
`.claude/rules/sandbox.md` documentation section, and the
`test-dispatch-scripts.sh` unit tests.

**No security vulnerabilities found — no fixes applied.**

### Findings and dispositions

|---|---|---|
| _(none)_ | — | The review identified no HIGH or MEDIUM vulnerabilities. |

### What was checked

- **Command injection** — `$path` flows into `claude agents --json --cwd "$path"`
  properly double-quoted as a distinct argv element; never `eval`'d or
  word-split. The `jq` filter is a fixed single-quoted literal with no
  interpolation. `CLAUDE_AGENTS_CMD` is a trusted env override.
- **Path traversal** — `$path` is passed verbatim to `claude --cwd`; the helper
  performs no filesystem reads or writes of its own.
- **Fail-safe logic** — non-zero exit, empty/whitespace-only output, and
  non-array/malformed JSON all resolve to "unknown"; `worktree_has_live_session`
  folds "unknown" into "occupied," the security-conservative direction. The
  capture-then-emit pattern prevents partial mid-stream `jq` output from being
  treated as valid sessions.
- **Data exposure** — `2>/dev/null` discards daemon stderr; the TSV output
  carries only `sessionId`, `pid`, `status`, `name` — operational metadata, not
  secrets.

The documentation file and the unit-test file are excluded from security-finding
scope per the review instructions.

PR #791, opened 2026-05-22
dispatch: skip a PR whose closing issue is blocked_by an open issue

natb1:
## Summary

`/dispatch`'s PR ladder (`security` → `review` → `code-review` → `verify` →
`qa`) selected draft PRs by phase without ever checking whether a closing issue
was `blocked_by` an open issue. Only the `help wanted` issue tier honored
blockers (via `dispatch-trace-leaf`). So once a PR existed, `/dispatch` advanced
it through every phase blind to its closing issue's blockers — as happened to
PR #765 (closes #725, which is `blocked_by` open #756).

## Changes

- **`lib.sh`** — new `count_open_blockers` helper: reuses `gh_api_array` against
  the `blocked_by` dependency endpoint and prints the count of **open** blockers.
  An API failure is surfaced as a clear error, never swallowed as "not blocked".
- **`dispatch-select-target`** — sources `lib.sh`; the PR-record loop skips a PR
  when any issue it closes is `blocked_by` an open issue. The skip runs before
  `FIRST_PR` is populated, so `--qa` mode inherits it. A closing issue blocked
  only by already-closed issues is not gated.
- **`SKILL.md`** — Step 2 documents the PR-ladder blocker skip; Step 3 adds the
  explicit-path counterpart: `/dispatch <N>` reports the blocker and stops
  whenever the named target is `blocked_by` an open issue, checked before leaf
  tracing and regardless of whether a PR exists.
- **`test-dispatch-scripts.sh`** — three new tests: open blocker → PR skipped;
  closed-only blockers → PR not skipped; `--qa` mode inherits the skip.

## Testing

`bash .claude/skills/dispatch/scripts/test-dispatch-scripts.sh` — 248/248 pass,
including the three new blocked-PR-skip tests.

Closes #786


natb1:
## dispatch-qa — user-acceptance QA (issue #786)

Single QA pass, non-browser path (scripts-only PR). **Clean pass — no bug found.**

**Items walked: 3 — PASS 3 / FAIL 0 / SKIP 0**

| # | Item | Result |
|---|------|--------|
| 1 | `count_open_blockers` reads the live GitHub `blocked_by` API correctly | PASS |
| 2 | Blocked-PR skip holds end-to-end for the real motivating case (#765 / #725 / #756) | PASS |
| 3 | SKILL.md documents the PR-ladder skip and the explicit-path blocked-target stop | PASS |

### 1 — `count_open_blockers` on the live API
`count_open_blockers 725` → `1` (issue #725 is `blocked_by` open #756); `count_open_blockers 786` → `0` (no blockers). Confirms the new helper's `jq` filter `select(.state == "open" or .state == "OPEN")` matches the **live** GitHub API response shape — the integration gap a stubbed unit test cannot cover.

### 2 — Blocked-PR skip, real motivating case
PR #765 is a draft with `closingIssuesReferences = [725]`; #725 is `blocked_by` open #756. Live `dispatch-select-target` → `issue 469` (not #765); live `dispatch-select-target --qa` → `empty` (not #765). Both scans ran the new blocker-skip code against the whole real queue and exited 0 — no `blocked_by` lookup failure.

PR #765's branch also has a local worktree, so today's live snapshot is doubly-skipped (the worktree skip preempts the blocker skip in ordering). The blocker skip in isolation is the operative guard for a blocked PR without a worktree — proven by unit tests 31–33 (CI-run) and by the verified inputs here: closing issues `[725]` + `count_open_blockers(725)=1` deterministically set `pr_blocked=true`.

### 3 — SKILL.md documentation
The PR-ladder description names "a PR whose closing issue is `blocked_by` an open issue is skipped". Step 3 documents the explicit-path counterpart: `/dispatch <N>` resolving to a target `blocked_by` an open issue releases the lock, reports the blocker, and stops — checked before leaf tracing and the PR-existence skip, applying whether or not a PR exists. Consistent with the adjacent closed-target stop rule.

All six acceptance criteria verified.

natb1:
## `/code-review-fix` — code-review phase

### Findings that were fixed

- **Blocker check ordering** — the per-closing-issue `count_open_blockers` `gh api` round-trip ran before `dispatch-phase` and the `done`/`waiting` continue, so PRs that would be discarded anyway still paid the cost. Moved the blocker check **after** the phase-derivation + `done`/`waiting` filter; `FIRST_PR` is still populated afterward so `--qa` still inherits the skip. Also collapsed the surrounding 7-line WHAT-block comment to one WHY-only line covering the placement rationale. Fixed in commit `aa5c9eb`. All 248 tests still pass.

### Informational findings

- **`dispatch-trace-leaf` does not use `count_open_blockers`** — `dispatch-trace-leaf`'s open-blocker resolution shells out to the `issue-blocking` script and filters JSON inline. It already sources `lib.sh` but uses a different abstraction. The contracts genuinely differ (full issue JSON for display vs. a count for skip logic), so the divergence is defensible; surfaced for awareness.
- **Inline worktree-list parser duplicates `lib.sh::list_worktree_records`** — `dispatch-select-target` manually walks `git worktree list --porcelain` rather than using the existing `list_worktree_records` / `split_worktree_record` helpers in `lib.sh`. Pre-existing pattern; not in scope of this PR.
- **`jq` subprocess per PR for closing-issue extraction** — `printf '%s' "$pr_closing" | jq -r '.[].number'` forks `jq` once per PR. Negligible (~5 ms each) vs the `gh api` round-trip; not worth changing.
- **`dispatch-select-target` vs `dispatch-trace-leaf` cross-script consistency** — both check `blocked_by` for open issues, but through different abstractions and with subtly different responsibilities. Documented for any future refactor that wants to unify them.

### Disregarded findings

- **`pr_blocked` bool is redundant** — false positive. `continue` inside the inner `while < <(...)` loop only continues that loop, not the outer PR loop, so the bool + `break` + outer-check pattern is the correct bash idiom.
- **SKILL.md explicit-path check not implemented in code** — false positive. The explicit-path blocker check IS implemented, as procedural instructions for the dispatching AI in Step 3 of `dispatch/SKILL.md` (run `issue-blocking <N>`, parse the result). The two paths use different mechanisms (procedural for the explicit path; direct API call in the queue loop) but reach the same outcome.
- **Drop `"OPEN"` from the case filter** — defense-in-depth dual-casing is consistent with `dispatch-trace-leaf`'s existing usage; removing it here would create an inconsistency across the dispatch scripts for a single-character savings.
- **Rename `count_open_blockers` to `has_open_blocker`** — the count form is more general; a future caller may need the actual count. The `> 0` boolean coercion at the call site is clear. Renaming adds churn for a stylistic preference, not an improvement.
- **Add a within-run cache for `count_open_blockers` lookups** — two open PRs closing the same issue is a rare edge case. The cache adds a `declare -A` and key-management for negligible gain.
- **Treat transient `gh api` failures as a soft skip instead of hard exit** — the project's `.claude/rules/code-style.md` explicitly prefers clear errors over defensive fallbacks. The current hard-exit behavior matches that rule.
- **`count_open_blockers` duplicates `issue-blocking`** — both hit the same REST endpoint but expose different data (count vs. full issue JSON). The duplication is by design; merging them would force one caller to pay for data it doesn't use.

### Deferred findings

_None — every actionable finding was either fixed in scope or judged not worth tracking as a follow-up._

natb1:
## Code review — `/code-review-fix`

### Findings that were fixed

- **Blocker-check ordering (efficiency)** — the per-PR `count_open_blockers` round-trip ran before the `done|waiting` phase filter, so PRs that would be discarded anyway still paid for the `gh api .../dependencies/blocked_by` call. Reordered so phase derivation + `done|waiting` filter run first; `FIRST_PR` is still populated afterward so `--qa` still inherits the skip. Fixed in `aa5c9eb`.
- **`dispatch-select-target` blocker-skip comment trimmed** — collapsed the 7-line preamble to a 3-line note covering only the placement rationale. Fixed in `aa5c9eb`.
- **`count_open_blockers` comment trimmed in `lib.sh`** — dropped the redundant restatement of `gh_api_array`'s own error contract. Fixed in `aa5c9eb`.
- **`SKILL.md` blocked-target guard tightened** — the new paragraph in Step 3 instructed the reader to apply the check "before leaf tracing and the PR-existence skip above" while sitting *after* those items, and narrated the PR-ladder counterpart instead of stating the rule. Compressed 8 lines → 4 and dropped the contradictory ordering language. Fixed in `3bd62c1`.

All 248 tests in `test-dispatch-scripts.sh` pass.

### Informational findings

- `count_open_blockers` accepts both `state == "open"` and `state == "OPEN"`. The REST `dependencies/blocked_by` endpoint returns lowercase; the dual-case match is defensive but matches the pattern already used in `dispatch-trace-leaf` for GraphQL-derived state values. Surfaced for awareness — no action.

### Disregarded findings

- **Reuse `issue-blocking` in the PR-skip loop instead of introducing `count_open_blockers`.** `issue-blocking` runs `gh issue view` per blocker (N+1 API calls) because human-facing callers need title/body/state. `count_open_blockers` filters in jq from one call because the PR-skip hot path only needs a bool. They intentionally solve different problems.
- **Fold `pr_closing | jq -r '.[].number'` into the same jq expression used for `pr_category`.** The current `while read … < <(jq …)` form matches the loop's idiom; the gain is purely cosmetic.
- **Refactor `pr_blocked=true; break` to `continue 2` or a helper function.** The flag/break pattern is the correct bash idiom inside a `while < <(...)` inner loop — `continue` only escapes the inner loop, not the outer PR loop. Refactor alternatives trade one local complexity for another with no readability gain.

### Deferred findings

- **#794** — `/dispatch` runs as a heartbeat (#725), so each tick now adds one serial `gh api` round-trip per actionable PR. The existing PR-list GraphQL query already pulls `closingIssuesReferences`; extending it to also pull the blocker state would collapse M+1 round-trips into 1. Filed as #794, `blocked_by` #786 (builds on the `count_open_blockers` helper this PR introduces).

natb1:
## `/review-fix` — review phase

### Fixed

- Multi-issue closure test gap — acceptance criterion "A PR closing multiple issues is skipped if any one is blocked" was uncovered by the three existing single-issue tests. Added test 34 in `test-dispatch-scripts.sh`: a PR closing two issues where only the SECOND is blocked is correctly skipped. Renumbered the following help-wanted-leaf-reachability tests 34-38 → 35-39. All 249 tests pass: 536eef3

### Informational

- `count_open_blockers` jq filter accepts both `"open"` and `"OPEN"` state values (`lib.sh:64`). The REST `/dependencies/blocked_by` endpoint returns lowercase per the test stubs; the `"OPEN"` branch is defensive against a state representation the API does not currently produce. The prior `/code-review-fix` pass already acknowledged this as Informational with the rationale that the dual-casing keeps the helper consistent with `dispatch-trace-leaf`'s existing usage. Surfaced again here for visibility; no action.

### Dismissed

_None._

### Deferred

- Explicit-path blocker check is LLM-procedural in `SKILL.md` Step 3 with no script enforcing it; extract to a `dispatch-check-blockers <N>` script that wraps `count_open_blockers` and exits with a recognizable code, mirroring the pattern used by `dispatch-resolve-arg` / `dispatch-find-pr` / `dispatch-resolve-worktree`. Builds on this PR's `count_open_blockers` helper. → #795

natb1:
## Security review

`dispatch:security-reviewed`

Both finding sources came back empty for this PR:

- **`/security-review`** (generic correctness pass over the diff): no findings.
  The new `count_open_blockers` helper interpolates `$issue_num` (sourced from
  `gh pr list --json closingIssuesReferences | jq -r '.[].number'` — a
  GitHub-managed integer field) into a `gh api` path argument and the result
  flows into an arithmetic `[[ -gt 0 ]]` context. No shell-eval sink and no
  attacker-influenced input on the path.
- **CodeQL code-scanning alerts** for `refs/pull/791/head`: none open.

No fixes applied.

PR #804, opened 2026-05-26
dispatch: key the dispatch lock on CLAUDE_CODE_SESSION_ID

natb1:
Closes #742

## Summary

Replaces the PID-keyed `tmp/dispatch.lock` mechanism with `CLAUDE_CODE_SESSION_ID`. Self-identity is now an env-var comparison; foreign-holder liveness is resolved via `claude agents --json` (the helper merged in #768). When the daemon cannot be queried, a recorded foreign holder is treated as live (fail safe — never steal a possibly-held lock).

The `/proc` ancestry walk, `is_live_claude`, `DISPATCH_LOCK_SESSION_PID`, `DISPATCH_LOCK_PROC_ROOT`, and the #724 `--bg-spare` cmdline exclusion are removed. A `--bg-spare` daemon is not an agent session and never appears in `claude agents --json`, so the wedge it caused is structurally eliminated.

`--wait` / `--release` / single-shot behavior, stdout protocol, and exit codes are unchanged.

## Acceptance criteria

- [x] `tmp/dispatch.lock` records `CLAUDE_CODE_SESSION_ID`; no `/proc` ancestry walk remains.
- [x] A foreign holder is reclaimed iff `claude agents --json` does not list its `sessionId`; same-session re-entry still re-acquires cleanly.
- [x] When the daemon cannot be queried, a recorded foreign holder is treated as live (lock not stolen).
- [x] `is_live_claude`, the ancestry walk, `DISPATCH_LOCK_PROC_ROOT`, and the #724 `--bg-spare` exclusion are removed.
- [x] `--wait` / `--release` / single-shot behavior and exit codes are unchanged.
- [x] `test-dispatch-scripts.sh` lock tests are updated to the session-id model and pass (291/291).

## Test plan

- [x] `bash .claude/skills/dispatch/scripts/test-dispatch-scripts.sh` — 291/291 pass
- [ ] CI green

### DIFF
--- a/.claude/skills/dispatch/SKILL.md
+++ b/.claude/skills/dispatch/SKILL.md
@@ -38,8 +38,10 @@ Run this Bash call with **both** `dangerouslyDisableSandbox: true` and an
 elevated `timeout: 600000` (ms):
 
 - `dangerouslyDisableSandbox: true` — the script writes
-  `$PROJECT_ROOT/tmp/dispatch.lock`, which is outside the sandbox write-allowlist
-  (same reason `dispatch-sweep` runs that way; see `.claude/rules/sandbox.md`).
+  `$PROJECT_ROOT/tmp/dispatch.lock`, which is outside the sandbox write-allowlist,
+  and queries `claude agents --json` over a local Unix socket for foreign-holder
+  liveness, which sandbox network-namespace isolation blocks (see
+  `.claude/rules/sandbox.md`).
 - `timeout: 600000` — `--wait` blocks on contention until it acquires the lock
   or `DISPATCH_LOCK_WAIT_TIMEOUT` (default 300 s) elapses, which exceeds the
   default Bash-call timeout.
@@ -49,7 +51,7 @@ Route on `$LOCK`:
 - **`acquired`** → this `/dispatch` holds the lock; proceed to Step 1.
 - **`busy`** → the wait timeout elapsed without acquiring — a wedged selection
   in another `/dispatch`. The script's **stderr** carries a one-line diagnostic
-  naming the wait duration and the holding PID; report those and **stop** — run
+  naming the wait duration and the holding sessionId; report those and **stop** — run
   no sync, no health gate, no sweep, no selection, and no phase skill.
 
 ### Releasing the lock
@@ -81,12 +83,14 @@ stopped), and the selection scan skips worktree'd targets — no other tick can
 select the same issue. Later steps cross-reference *Releasing the lock* rather
 than repeating this command.
 
-The lock is scoped to selection and self-healing. The recorded PID is
-session-keyed: if a tick dies before its explicit release, the next tick detects
-the stale record and proceeds, and a `--wait` waiter re-checks holder liveness
-every poll, so it reclaims a dead holder's lock automatically. Same-session
-re-entry (e.g. after a context clear that re-invokes `/dispatch`) re-acquires
-cleanly because the recorded PID matches the re-entering session's own PID.
+The lock is scoped to selection and self-healing. The recorded sessionId
+(`CLAUDE_CODE_SESSION_ID`) outlives any single Bash call within a tick: if a
+tick dies before its explicit release, the next tick detects that the recorded
+sessionId no longer appears in `claude agents --json` and reclaims the lock,
+and a `--wait` waiter re-checks holder liveness every poll, so a dead holder's
+lock is reclaimed automatically. Same-session re-entry (e.g. after a context
+clear that re-invokes `/dispatch`) re-acquires cleanly because the recorded
+sessionId matches the re-entering session's own `CLAUDE_CODE_SESSION_ID`.
 
 ## 1. Sync local main with `origin/main`
 
--- a/.claude/skills/dispatch/scripts/dispatch-acquire-lock
+++ b/.claude/skills/dispatch/scripts/dispatch-acquire-lock
@@ -4,37 +4,44 @@
 # /dispatch mutates shared state (fetches origin/main, creates/removes
 # worktrees, flips gh labels and PR draft status). Its per-target guards assume
 # serial execution within a tick; two concurrent invocations can race. This
-# script serializes them with a PID-stamped lock file.
+# script serializes them with a session-id-stamped lock file.
 #
 # An fd-based `flock` cannot serialize /dispatch: the skill runs as many
 # separate Bash tool calls, each a fresh shell, so an fd flock would release
 # the instant its Bash call exits. The lock therefore keys on the Claude
-# *session* process — the nearest ancestor whose comm starts with ".claude"
-# and is not a background daemon (e.g. a `--bg-spare` spare) — whose lifetime
-# spans the whole tick. The recorded PID is the lock holder.
+# *session* — identified by CLAUDE_CODE_SESSION_ID, a stable UUID exported into
+# every .claude session's environment — whose lifetime spans the whole tick.
+# The recorded sessionId is the lock holder.
 #
-# Release is implicit and self-healing: when the session dies its PID stops
-# being a live ".claude" process, so the next tick's read-check-write reclaims
-# the lock automatically. `--release` is an explicit optional fast-path that
-# the lock holder can call to release immediately — clearing the recorded PID
-# without waiting for self-healing to notice the session has moved on.
+# Foreign-holder liveness is resolved via `claude agents --json`, the daemon-
+# backed registry of live sessions. A recorded holder is "live" iff its
+# sessionId appears in the registry; otherwise it is reclaimable. The query is
+# fail-safe: any failure (binary missing, non-zero exit, empty output, non-
+# array JSON, jq failure) is treated as "live" so the lock is never stolen
+# from a possibly-held foreign holder when the daemon cannot be queried.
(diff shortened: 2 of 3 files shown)


natb1:
## Dispatch QA — clean pass

Non-browser walkthrough (scripts-only PR — `dispatch-acquire-lock` and its test suite).

### Items walked

| # | Item | Result |
|---|---|---|
| 1 | `test-dispatch-scripts.sh` — full lock-test suite (Tests 1–15 plus the rest of the dispatch tests) | PASS (291/291) |
| 2 | `tmp/dispatch.lock` records `$CLAUDE_CODE_SESSION_ID`, not a PID | PASS (UUID matches env exactly) |
| 3 | `/proc` walk, `is_live_claude`, `DISPATCH_LOCK_PROC_ROOT`, and `--bg-spare` exclusion fully removed from `dispatch-acquire-lock` | PASS (grep returns nothing) |
| 4 | `--release` round-trip: release truncates to empty, re-acquire writes back our sessionId | PASS |
| 5 | Only `dispatch-acquire-lock` and `test-dispatch-scripts.sh` are touched — no consumer scripts modified | PASS |

### Counts

- PASS: 5
- FAIL: 0
- SKIP: 0

### Notes

- The shared lock lives at the bare-repo project root (`/home/n8/natb1/commons.systems/tmp/dispatch.lock`), not the worktree's `tmp/` — this is the existing global-mutex behavior, unchanged by this PR.
- No GIF: non-browser path.

natb1:
## Code review (dispatch:code-reviewed)

Effort: `/code-review max` — 5 finder angles + 1-vote verify + gap sweep. 15 findings; classified and applied in commit 9ec57fc.

### Findings that were fixed

Fix commit: 9ec57fc

- **SKILL.md doc drift caused by this PR**: Step 0 sandbox justification now also cites the `claude agents --json` socket access; "holding PID" → "holding sessionId" in the busy-path description; the self-healing paragraph (lines 84–89) rewritten to reference sessionId / `CLAUDE_CODE_SESSION_ID` instead of PID.
- **dispatch-acquire-lock header comment**: now distinguishes the definite-zero case (`[]` → reclaim) from the opaque-failure case (whitespace-only, non-array JSON, jq error → fail safe to live), so a future maintainer cannot wrongly "fix" the `[]` branch to also fail safe and recreate the wedge.
- **is_live_session comment**: the prior "Mirrors the defensive pattern from lib-claude-agents.sh" wording was misleading — the helper uses return-1-for-unknown while is_live_session uses return-0-for-live (opposite return conventions). Replaced with an explicit description of the failure-mode coverage and the semantic difference (the helper asks "is anything running under this path?"; we ask "is THIS sessionId still alive?", with no `--cwd` filter).
- **fd 9 defensive close**: `is_live_session` now invokes `claude` with `9>&-`, closing the flock fd before exec. Today `claude agents --json` is a query-and-exit client and inheritance is harmless, but the cost of defending against any future helper that forks (which would inherit fd 9 and hold the flock past the subshell's exit, wedging the lock) is one redirection.
- **Test 4 missing assertion**: now also checks that the lock file is rewritten to the new sessionId after an empty-registry reclaim — Tests 3 and 14 had this assertion, Test 4 did not.
- **Tests 6a, 6b, 13 stderr assertions**: now match the specific error message (`"not in a git repo"`, `"CLAUDE_CODE_SESSION_ID is unset"`, `"unknown argument"`) instead of just "stderr non-empty". Without this, the tests would silently pass for the wrong reason if the guards in `dispatch-acquire-lock` were ever reordered.
- **Test 13 also sets `CLAUDE_CODE_SESSION_ID`**: removing the order-dependence on Step 0 (arg parse) running before Step 2 (session-id check).
- **New `lock_fake_claude_payload` helper + Tests 16/17/18**: covers `--release` on an absent lock file (→ `noop`), whitespace-only daemon stdout (→ foreign holder treated as live, lock not stolen), and non-array JSON stdout (`{...}` → same). These exercise fail-safe branches that previously had no test coverage.

Suite: 301/301 pass (was 291; +10 new assertions).

### Informational findings

Surfaced for human reference; no code change recommended.

- **is_live_session runs inside the flock-held subshell**, extending the critical section from microseconds (old /proc reads) to a daemon IPC round-trip. Under daemon load, concurrent `--wait` pollers serialize on the flock for the IPC duration. Architectural tradeoff of moving liveness off `/proc`; an alternative (release flock before the daemon call, then re-acquire and re-check) would reopen a TOCTOU window.
- **`printf '%s\n' "$SESSION_ID" >"$LOCK_FILE"` inside the try_acquire subshell** has no `set -e`; a filesystem-full or permission-denied write would silently print `OK` and the parent would decode `acquired`. This is the same pattern the script used before the migration — not introduced by this PR — and is intentional for the no-`set -e` style the script declares ("exits are handled explicitly").
- **The `--cwd` flag is absent from is_live_session's `claude agents --json` call**, while `lib-claude-agents.sh::claude_sessions_under` always passes it. This is correct (a sessionId is globally unique; we want the full registry, not a path filter) and is now documented in the rewritten `is_live_session` comment.

### Disregarded findings

False positives.

- **`lock_fake_claude_sessions` unquoted heredoc fragility when `$TMPDIR_TEST` contains spaces** — refuted. The heredoc body contains literal `"$TMPDIR_TEST/fake/payload.json"` with double-quotes; the quotes are written verbatim into the generated fake script, so a path with spaces produces `cat "/path with spaces/fake/payload.json"` which the fake's bash interprets correctly.

### Deferred findings

None.

natb1:
## Fixed
- Test for `is_live_session`'s jq parse-failure branch on malformed JSON (Test 19): 4a7c282
- Test for `--release` with `CLAUDE_CODE_SESSION_ID` unset (Test 6c, mirrors Test 6b's acquire-path coverage): 4a7c282

## Informational
- `is_live_session` (in `dispatch-acquire-lock`) and `claude_sessions_under` (in `lib-claude-agents.sh`) intentionally duplicate the opaque-failure enumeration — different scoping (global registry vs `--cwd`-filtered) and inverted return convention (LIVE vs UNKNOWN). Could promote to a shared `session_id_is_live` helper if a third caller appears; premature now.
- `jq` is an implicit hard dependency of `is_live_session`. Missing `jq` would make the script permanently report `busy` against any non-self foreign holder — the safe direction (never steal a lock).
- Hard dependency on Claude Code >= 2.1.146 for the daemon-backed registry. Already inherited from #738/#739; this PR adds no new constraints.
- `claude agents --json` daemon stalls could elongate a `--wait` poll (~60 daemon calls over the 300s default timeout). No change suggested; flag for monitoring.
- The `9>&-` close in `is_live_session` is forward-compatible defense against a future `claude agents --json` that spawns a background helper inheriting the flock fd.
- Code style and conventions consistent with sibling scripts: `set -uo pipefail`, explicit exit-code handling, no defensive fallbacks (unset `CLAUDE_CODE_SESSION_ID` exits 2 with a clear error rather than falling back to `$PPID`).
- SKILL.md Step 0 callout correctly updated — now explains both reasons for `dangerouslyDisableSandbox: true` (lock-file write outside the sandbox allowlist + the `claude agents --json` Unix-socket call).
- Test coverage matrix for the new mechanism is good: first acquisition, parallel acquisition, stale recorded sessionId, empty registry, same-session re-entry, missing-git-repo guard, missing-`CLAUDE_CODE_SESSION_ID` guard, `--wait` against live foreign, `--wait` reclaim mid-flight, `--release` own/foreign/absent, and the opaque-failure fail-safe set.
- Test 6b (missing-`CLAUDE_CODE_SESSION_ID` guard) is the right replacement for the removed ancestor-walk test — the new mechanism has no ancestor walk to exercise.
- No new attack surface. Replacing a PID with a UUID structurally eliminates kernel PID-recycling collisions; the `flock`-bounded read-check-write is unchanged.

## Dismissed
- Test 10 wording polish about the second `lock_fake_claude_sessions` call mid-`--wait` — nit; the existing comment already conveys that the fake reads `payload.json` at runtime, and the parent thread judged the added sentence not worth the noise.

## Deferred
_None._

natb1:
## /security-review-fix

Ran the generic `/security-review` against the working tree and fetched the open
CodeQL code-scanning alerts for this PR.

### /security-review findings

No security vulnerabilities identified. The review considered:

- **Command injection via `$recorded` from the lock file** — the recorded session
  id is passed only to `grep -qxF -- "$sid"` (fixed-string, exact-match, with the
  `--` separator preventing flag injection). No injection surface.
- **`CLAUDE_AGENTS_CMD` override** — documented test-only env hook; environment
  variables are trusted per this codebase's threat model.
- **Lock file path** — derived from `DISPATCH_LOCK_FILE` (env override) or
  `git rev-parse`; no user-controlled input reaches the path.
- **Liveness fail-safe** — daemon-query errors are treated as "live" so a
  possibly-held lock is never stolen. Behavior is covered by the new tests.
- **Session id format** — UUIDs are unguessable; comparisons use `==` and
  `grep -qxF`, not regex or `eval`.

### CodeQL code-scanning alerts

`gh api repos/{owner}/{repo}/code-scanning/alerts?state=open&ref=refs/pull/804/head`
returned an empty array. No open alerts on this PR's ref.

### Disposition

No code changes. Applying `dispatch:security-reviewed` and marking the PR ready.

PR #828, opened 2026-05-27
hooks: set session title to worktree basename on SessionStart

natb1:
## Summary

- Add `.claude/hooks/session-title.sh`: emits `{"hookSpecificOutput":{"sessionTitle":"<basename>"}}` where `<basename>` is `basename "$(git rev-parse --show-toplevel)"`. Exits 0 silently outside a git repo.
- Wire the hook into `SessionStart` under both the default matcher and `matcher: "clear"`, so the title applies on fresh startup, resume, and after `/clear`.

Concurrent Claude sessions across worktrees previously all showed the same generic title, making them indistinguishable in the Remote Control list, mobile app, and `claude --resume` picker. Claude Code 2.1.152 added the `hookSpecificOutput.sessionTitle` field; this PR uses it to label each session with its worktree basename (e.g. `825-set-claude-session-title-to`, `main`).

The hook is independent of the existing `productivity-tui/hooks/on-idle.sh` and `restore-dispatch-skill.sh` hooks — different output contracts, no overlap.

Closes #825

## Test plan

- [x] Hook prints expected JSON inside the worktree
- [x] Hook exits 0 with no stdout outside a git repo
- [x] `.claude/settings.json` is valid JSON
- [ ] Open a fresh session in `worktrees/main` and any issue worktree; verify each session's title reflects its basename (real-Claude check — runs post-merge)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #851, opened 2026-05-27
dispatch: greenfield router/worker split

natb1:
Closes #839

## Summary

Split monolithic `/dispatch` along its natural seam between *selection* and
*execution* into a router + worker pair. One session per worktree, plus a thin
router that selects work and spawns the per-worktree sessions.

- **`/dispatch`** (router; lives in `worktrees/main`): acquires the per-repo
  selection lock, syncs `origin/main`, runs JIT, selects a target, materializes
  the worktree (`git worktree add` + `direnv allow` + `sync-issue-context`),
  spawns `/dispatch-worker <N>` with `cwd=<worktree-path>`, then exits or
  self-closes. The router never enters a worktree.
- **`/dispatch-worker <N>`** (worker; lives in `worktrees/<N>-…`): verifies
  cwd, derives phase via `dispatch-phase`, runs exactly one phase skill, then
  spawns a fresh `/dispatch` router back in `worktrees/main` (or parks on
  `dispatch:office-hours`) and self-closes. The worker never selects another
  target.

The selection lock stays per-repo and serializes router selection only. The
execution path is per-worktree: N issues in flight = N concurrent workers,
each in its own worktree, advancing their own issues' phases without
contention. Per-worktree dedup is enforced at the spawn boundary by
`dispatch-spawn-worker`'s `claude_sessions_under <worktree-path>` query.

Resolves #825 for free: the worker is born with `cwd=<worktree-path>`, so
`.claude/hooks/session-title.sh` titles each worker as `<N>-…` instead of
`main`. No platform feature needed; the existing in-repo hook just works in
the new architecture.

## Migration (six dependency-ordered commits)

1. Introduce `/dispatch-worker` skill as the per-worktree execution half.
2. Add `dispatch-spawn-worker` primitive with per-worktree dedup + 8 tests.
3. Rewrite `/dispatch` as the router (700 → 492 lines; drop phase mapping,
   relevance review, 4-step handoff, `EnterWorktree`/`ExitWorktree`).
4. Drop `EnterWorktree`/`ExitWorktree` from the dispatch chain; rewrite
   `/dispatch-qa` Step 0 to operate in place; add ratchet test that fails if
   the strings reappear in any of 10 chain-skill SKILL.md files.
5. Rename `dispatch-spawn` → `dispatch-spawn-router` for call-site
   symmetry with `dispatch-spawn-worker`; rename env-var prefix.
6. Rescope lock documentation: per-repo selection lock (Steps 1–5) and
[...]

Issue #825, opened 2026-05-27
Set Claude session title to worktree name via SessionStart hook


natb1:
Claude Code 2.1.152 added a SessionStart hook option to set the session title via `hookSpecificOutput.sessionTitle`. The motivating goal is to label every Claude session with the basename of its worktree directory (e.g. `821-dispatch-clarify-in-phase-sk`, `main`) so concurrent sessions are distinguishable in `claude agents`, the Remote Control list, the mobile app, and `claude --resume`.

### Shipped — interactive case

The in-repo hook `.claude/hooks/session-title.sh`, wired into `SessionStart` (default and `clear` matchers), shipped in PR #828 (and follow-up `b6255c5` adding the `hookEventName` field). For any session whose **start cwd is a worktree** — i.e. interactive sessions launched inside a worktree — the title correctly reflects the worktree basename. The `implement-unit`/`plan-implement` post-condition scope from the original body also shipped (commit `4c9a1f8`).

### Not addressed — dispatch-chain case

For background `/dispatch` sessions, the in-repo hook is structurally insufficient: every `/dispatch` job today starts in `worktrees/main` (per `dispatch-spawn`'s contract), then later switches its cwd via `EnterWorktree` to the worktree it will actually work on. `SessionStart` fires at session start — long before `EnterWorktree` — so the hook always sees `main` and titles every dispatch-chain session `main`. The headline goal of distinguishing concurrent sessions in `claude agents` is therefore not met for the dispatch-chain, which is the dominant case in this repo.

The clean fix is the greenfield router/worker split in #839: `/dispatch` becomes a router that runs only in `worktrees/main` and spawns per-worktree workers (`/dispatch-worker <N>`) with their cwd set to the target worktree at spawn time. Each worker's `SessionStart` fires in its own worktree, so the existing `session-title.sh` hook produces the right title with no further hook-script change.

This issue is **blocked on #839**. No additional in-repo hook-script change is needed; once #839 lands, the dispatch-chain title problem is resolved as a downstream consequence.

## Acceptance criteria

- [x] In-repo `session-title.sh` hook exists and emits `hookSpecificOutput.sessionTitle` set to the worktree basename. (PR #828.)
- [x] Hook is wired into `SessionStart` (default) and `SessionStart` with `matcher: "clear"` in `.claude/settings.json`. (PR #828.)
- [x] Hook exits 0 silently if it cannot determine the worktree name. (PR #828.)
[...]

PR #863, opened 2026-05-28
dispatch: spawn worker from main with --name=worktree-basename

natb1:
Closes #860
Closes #828

## Summary

PR #851 (greenfield router/worker split) spawned each worker with `cwd=target-worktree` by wrapping `claude --bg` in `( cd "$WORKTREE_PATH" && ... )`. That worked for the worker, but it polluted the Claude daemon's launcher-default cwd — after every dispatch spawn, the TUI's "+ new session" prompt defaulted to the worker's worktree instead of `worktrees/main`.

Spawn the worker from `worktrees/main` (no pollution) and have the worker `cd` into its target worktree as Step 0. Pass the worktree path positionally. Set `--name` authoritatively at spawn time via `basename "$WORKTREE_PATH"` (e.g. `860-dispatch-spawn-worker-from-m`) — this also subsumes the SessionStart-title hook proposed in #825.

## Commits

- **aaf73f3** — `dispatch-spawn-worker`: drop the `cd` subshell, set `--name` to the worktree basename, pass the worktree path as a positional argument, switch dedup to name-keyed (worktree-basename uniqueness gives a one-line dedup). Plus `dispatch-worker` Step 0 changes from assertion to action (\`cd\` into the worktree first).
- **7229181** — Broaden the prune regex to \`^([0-9]+-|dispatch-)\` in both \`dispatch-spawn-worker\` and \`dispatch-spawn-router\` so stopped workers under the new name shape get reaped.
- **8ee378b** — \`restore-dispatch-skill.sh\`: gate on the worker \`--name\` shape (queried via \`claude agents --json\` filtered by sessionId) instead of the \`tmp/dispatch-worktree\` marker. Routers are skipped; branch-name fallback covers interactive sessions. The marker writer (\`worktree-create.sh:84\`) is unchanged.

## Test plan

- [x] \`test-dispatch-scripts.sh\` — 445/445 pass (4 new spawn-worker assertions, 2 new prune assertions).
- [x] \`bash -n\` on every modified shell script.
- [x] Manual sanity test on \`restore-dispatch-skill.sh\` — fake SessionStart payload triggers \`COMPACTION RECOVERY: Reload skill: /dispatch-worker 860\` via the branch-name fallback path.
- [ ] End-to-end: from an interactive Claude in \`worktrees/main\`, invoke \`/dispatch\`. After it spawns a worker, confirm the \`claude agents\` TUI's "+ new session" default cwd stays \`worktrees/main\`.
- [ ] End-to-end: \`claude agents\` lists the spawned worker with name \`<N>-<slug>\` (e.g. \`860-dispatch-spawn-worker-from-m\`), not \`dispatch-<N>-<short>\`.
- [ ] End-to-end: after a worker self-closes, the next router-spawn reaps it via the broadened prune regex.

### DIFF
--- a/.claude/hooks/restore-dispatch-skill.sh
+++ b/.claude/hooks/restore-dispatch-skill.sh
@@ -44,6 +44,14 @@ fi
 [ -n "$ISSUE_NUM" ] || exit 0
 [ -n "$WORKTREE_BASENAME" ] || exit 0
 
+# WORKTREE_BASENAME comes from session --name or git branch name; reject
+# path-traversal characters before composing the absolute path below. Git
+# refnames already disallow `..`, so a slash or `..` here indicates a
+# malformed or hostile source.
+case "$WORKTREE_BASENAME" in
+  *..*|*/*) exit 0 ;;
+esac
+
 # Resolve the absolute worktree path. The project root is the parent of the
 # shared git common dir (.bare). git is run from the hook's cwd, which is some
 # worktree of the project — --git-common-dir always returns the shared path.
--- a/.claude/hooks/restore-dispatch-skill.sh
+++ b/.claude/hooks/restore-dispatch-skill.sh
@@ -23,25 +23,39 @@ case "$NAME" in
 esac
 
 ISSUE_NUM=""
+WORKTREE_BASENAME=""
 
-# Primary path: session --name matches worker shape ^[0-9]+-.
+# Primary path: session --name matches worker shape ^[0-9]+-. The name IS the
+# worktree basename for workers spawned by dispatch-spawn-worker.
 if printf '%s\n' "$NAME" | grep -qE '^[0-9]+-'; then
   ISSUE_NUM=$(printf '%s\n' "$NAME" | grep -oE '^[0-9]+')
+  WORKTREE_BASENAME="$NAME"
 else
   # Fallback: derive from the branch name (worktree basename).
   BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null) || exit 0
   if printf '%s\n' "$BRANCH" | grep -qE '^[0-9]+-'; then
     ISSUE_NUM=$(printf '%s\n' "$BRANCH" | grep -oE '^[0-9]+')
+    WORKTREE_BASENAME="$BRANCH"
   else
     exit 0
   fi
 fi
 
 [ -n "$ISSUE_NUM" ] || exit 0
+[ -n "$WORKTREE_BASENAME" ] || exit 0
+
+# Resolve the absolute worktree path. The project root is the parent of the
+# shared git common dir (.bare). git is run from the hook's cwd, which is some
+# worktree of the project — --git-common-dir always returns the shared path.
+GIT_COMMON_DIR=$(git rev-parse --path-format=absolute --git-common-dir 2>/dev/null) || exit 0
+PROJECT_ROOT=$(dirname "$GIT_COMMON_DIR")
+WORKTREE_PATH="$PROJECT_ROOT/worktrees/$WORKTREE_BASENAME"
 
 # Emit reload instruction so Claude reloads /dispatch-worker for this worker
 # session. /dispatch-worker re-derives the phase from PR/CI ground truth, so
-# recovery is correct after a /clear in any phase.
-printf 'COMPACTION RECOVERY: Reload skill: /dispatch-worker %s\n' "$ISSUE_NUM"
+# recovery is correct after a /clear in any phase. The worker requires
+# <N> <worktree-path> — its Step 0 cd's into the path itself.
+printf 'COMPACTION RECOVERY: Reload skill: /dispatch-worker %s %s\n' \
+  "$ISSUE_NUM" "$WORKTREE_PATH"
 
 exit 0
--- a/.claude/hooks/restore-dispatch-skill.sh
+++ b/.claude/hooks/restore-dispatch-skill.sh
@@ -1,28 +1,47 @@
 #!/usr/bin/env bash
-# Context-clear recovery hook: if this is a dispatch session, emit a reload instruction.
-# Bound to SessionStart:clear.
+# Context-clear recovery hook: if this is a dispatch worker session, emit a
+# reload instruction. Bound to SessionStart:clear.
 # Always exits 0 — never blocks session recovery.
 set -uo pipefail
 trap 'echo "[restore-dispatch-skill] WARNING: unexpected error on line $LINENO" >&2; exit 0' ERR
 
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)" || {
-  echo "[restore-dispatch-skill] WARNING: cannot resolve script directory" >&2
-  exit 0
-}
+STDIN_JSON=$(cat 2>/dev/null) || STDIN_JSON=""
+SESSION_ID=$(printf '%s' "$STDIN_JSON" | jq -r '.session_id // empty' 2>/dev/null) || SESSION_ID=""
 
-# Extract issue number from branch name (e.g., "566-extend-dispatch-implement-wi" → "566")
-BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null) || exit 0
-ISSUE_NUM=$(printf '%s\n' "$BRANCH" | grep -oE '^[0-9]+') || exit 0
+# Query the running sessions to get the --name for this session.
+NAME=""
+if [ -n "$SESSION_ID" ]; then
+  NAME=$(claude agents --json 2>/dev/null | \
+    jq -r --arg sid "$SESSION_ID" '.[] | select(.sessionId == $sid) | .name' \
+    2>/dev/null) || NAME=""
+fi
 
-# Detect a dispatch session by checking for the worktree marker /dispatch creates.
-REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null) || exit 0
+# Router sessions (dispatch-<short-id>) restart via /dispatch, not skill
+# restoration — skip them explicitly.
+case "$NAME" in
+  dispatch-*) exit 0 ;;
+esac
 
-[ -f "$REPO_ROOT/tmp/dispatch-worktree" ] || exit 0
+ISSUE_NUM=""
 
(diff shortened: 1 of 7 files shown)


natb1:
## /dispatch-qa results

**Walkthrough type:** non-browser (infrastructure-only PR — dispatch scripts, worker SKILL, restore hook, test suite).

**Items walked:** 9
**PASS:** 9  **FAIL:** 0  **SKIP:** 0

### Verifications

1. **Test suite** — `test-dispatch-scripts.sh` reports 445/445 passing.
2. **Spawn cwd** — `dispatch-spawn-worker` invokes `claude --bg` from the script's inherited cwd; no `( cd "$WORKTREE_PATH" && ... )` wrapper at the spawn site.
3. **`--name` = worktree basename** — `AGENT_NAME="$(basename "$WORKTREE_PATH")"` is passed via `--name "$AGENT_NAME"`.
4. **Worker positional args** — invocation is `/dispatch-worker $ISSUE_NUM $WORKTREE_PATH` (both args).
5. **Worker Step 0** — `.claude/skills/dispatch-worker/SKILL.md` Step 0 parses `<N> <worktree-path>`, `cd`s into the worktree, runs the branch sanity check, and routes failures to Step 4 (early-stop).
6. **Name-keyed dedup** — `dispatch-spawn-worker` Step 3 dedup compares `[[ "$name" == "$AGENT_NAME" ]]`; no cwd-based comparison.
7. **Prune regex** — both `dispatch-spawn-router` and `dispatch-spawn-worker` use `^([0-9]+-|dispatch-)`.
8. **Restore hook** — `.claude/hooks/restore-dispatch-skill.sh` queries `claude agents --json` for the session's `--name`, matches `^[0-9]+-`, skips `dispatch-*` router sessions, falls back to branch name; no reference to `tmp/dispatch-worktree`.
9. **Test coverage of AC** — `spawn-worker:` (argv shape + `--name`), `spawn-worker-cwd:` (no `cd` to target), `dedup-worker:` (name-keyed dedup), and `prune-worker:` (worker-shape prune) assertions all present in the test file.

No bugs found.

natb1:
## /code-review max — code-review phase

Seven findings, all in-scope for this PR's spawn-from-main change. All fixed in-session; no Informational, Disregarded, or Deferred items.

### Findings that were fixed

Commit `17b7758` — script + hook + regression test:

- **`dispatch-spawn-worker:211` Step 5 verify queried under `WORKTREE_PATH`, but the new worker registers under the spawner's cwd (`worktrees/main`).** The daemon's `--cwd` filter is server-side (`lib-claude-agents.sh` line 16 / 67), so the verify query excluded the just-spawned worker, `registered` stayed empty, and the script exited 1 on every spawn in production. Fix: capture `SPAWN_CWD=$(pwd)` at script entry and use it for both Step 3 dedup (line 176) and Step 5 verify (line 211). The existing test fake ignored `--cwd` so the regression was invisible to Tests 1–8; Test 9 (new) installs a cwd-aware fake.
- **`restore-dispatch-skill.sh:45` recovery hook emitted `/dispatch-worker N` (one arg).** The new worker contract requires `<N> <worktree-path>`; every post-`/clear` recovery would die at Step 0 with `expected ARGUMENTS '<N> <worktree-path>'`. Fix: derive `WORKTREE_BASENAME` from the session `--name` (worker shape `<N>-<slug>` IS the worktree basename) or, on the fallback path, from the current branch; resolve `PROJECT_ROOT` from `git --git-common-dir`; emit `/dispatch-worker $ISSUE_NUM $WORKTREE_PATH`. Covers both the primary path and the fallback-from-branch path.

Commit `ce82034` — doc updates:

- **`dispatch/SKILL.md:122-129` per-worktree invariant section** claimed "The worker is born in the target worktree and dies there" — false after the spawn-from-main change. Rewrote to state that workers are born at the spawner cwd and `cd` into their target as their own Step 0.
- **`dispatch/SKILL.md:445` Step 6 spawn description** said `cwd=$WORKTREE_PATH` — contradicted the new semantics. Rewrote to state the spawn runs from `worktrees/main` and the worker `cd`s in Step 0.
- **`dispatch/SKILL.md:423-427` recovery-marker description** said `restore-dispatch-skill.sh` keys on the `tmp/dispatch-worktree` marker and emits `/dispatch-worker <N>` — both stale. The hook now keys on the session `--name` shape and emits two args; the marker is still written but only the lock script reads it.
- **`dispatch-qa/SKILL.md:11` attribution** said "cwd set by `dispatch-spawn-worker` at spawn time" — wrong after this PR. Rewrote to attribute the cwd to the worker's own Step 0 `cd`.

### Regression guard

Test 9 in `test-dispatch-scripts.sh` installs a cwd-aware fake `claude`: its `agents` handler filters its registry by `--cwd`, and its `--bg` handler records the new worker with `cwd = $(pwd)` at spawn time. Before the fix, verify would query under `WORKTREE_PATH`, the filter would drop the worker, and the script would exit 1. After the fix, both dedup and verify query under the spawner cwd, find the worker, and exit 0.

Full test suite: 447/447 pass.

natb1:
## Fixed
- Guard non-existent `<worktree-path>` in `dispatch-spawn-worker` Step 0 so a missing path surfaces a clear error at the spawner rather than failing in the worker's Step 0 `cd`; broadened the exit-2 docstring; added Test 8 sub-case E: 6fdb4ae

## Informational
- TOCTOU on name-keyed dedup — two concurrent routers could each pass the guard before either spawns. Inherited from the prior design; not introduced by this PR; the per-worktree invariant and #755 absorb it.
- Dedup scope is now global (all sessions under `SPAWN_CWD`) rather than worktree-scoped — concluded correct: per-worktree name uniqueness makes name-equality produce the same result.
- `restore-dispatch-skill.sh` falls through to the branch-name fallback when `claude agents --json` returns empty under sandbox. Behavior is safe; the hook always exits 0; no regression.

## Dismissed
- Add a comment noting the assumption that `WORKTREE_PATH` contains no spaces in `dispatch-worker` Step 0 — project convention defaults to no comments unless the WHY is non-obvious; the assumption is already enforced by the `<N>-<slug>` worktree-basename scheme.
- Add a comment in `restore-dispatch-skill.sh` acknowledging the sandbox edge case for `claude agents --json` — same rationale; the fallback path already handles the empty-result case correctly.

## Deferred
- Add a unit test for `restore-dispatch-skill.sh`'s `claude agents --json`-based name-lookup path → #875

natb1:
## Security review

`/security-review-fix` ran on this PR — 9 parallel subagents (6 domains, red team, built-in `/security-review`, CodeQL alerts) reviewed the diff vs `origin/main` (merge-base `7fd82be`).

### Required — fixed (commit `a63f8de`)

1. **Path traversal in `restore-dispatch-skill.sh`** — `WORKTREE_BASENAME` is derived from the session `--name` or git branch and concatenated into an absolute path that the recovery instruction emits to the session transcript. The existing `grep -qE '^[0-9]+-'` guard checks only the prefix and lets `..` or `/` through mid-string (e.g. `123-../../etc`). High confidence (flagged independently by subagents 1 and 7).
   - **Fix**: added a `case ... *..*|*/*) exit 0` guard between the basename derivation and the path composition.
   - **OWASP**: A01:2021 Broken Access Control. **STRIDE**: Tampering / Elevation of Privilege.

2. **Unsafe characters in `WORKTREE_PATH` reach `dispatch-spawn-worker`'s prompt-string interpolation** — the path is validated only by `[[ -d ... ]]`, then interpolated into `"/dispatch-worker $ISSUE_NUM $WORKTREE_PATH"` (the worker's `ARGUMENTS`, parsed via `read -r`) and into `basename` → `--name`. A path with whitespace or shell metacharacters would split the worker's `read -r` boundary or affect downstream parsing. Medium confidence (subagents 1, 7); the same fix also closes subagent 1's low-confidence `AGENT_NAME` finding.
   - **Fix**: added `[[ ! "$WORKTREE_PATH" =~ ^[A-Za-z0-9/_.-]+$ ]] && exit 2` after the existence check. New test sub-case F in `test-dispatch-scripts.sh` (458/458 pass).
   - **OWASP**: A03:2021 Injection. **STRIDE**: Tampering.

### Out-of-scope (genuine concerns, deferred)

3. **Prune regex `^([0-9]+-|dispatch-)` over-matches user sessions** — medium confidence (subagents 2, 7). The new pattern accepts any name starting with digits-and-hyphen, including user-named sessions like `42-research`; `claude rm` then reaps them when they hit a terminal state. This is the exact regex specified in #860 ("works for both shapes") and is intentional design, so narrowing it is a follow-up design decision rather than a security-review fix on this PR.

4. **Cross-project name-keyed dedup collision** — low confidence (subagent 2). Two unrelated projects with the same worktree basename both dispatching from a co-located `worktrees/main` would suppress each other. Unlikely setup; not in scope for this PR.

5. **Recovery hook emits full absolute path to session transcript** — medium confidence (subagent 3). Discloses the host username and directory layout. The path is functionally required for the worker to know its target; changing the recovery contract to pass only the basename and have the worker reconstruct is a separate design change.

6. **28 pre-existing `npm audit` findings** (subagent 4) — workspace-wide, not introduced by this PR. The one high-severity finding with a fix available (`fast-xml-builder` → `npm audit fix`) warrants a separate dependency-bump PR.

### False positives

7. **Wrong-issue recovery via cwd/name disagreement** (subagent 7, low confidence) — the analysis hypothesizes a cwd-drift state that the hook's data-flow doesn't actually expose; numeric-prefix collisions don't occur because the grep extracts leading digits and branches are uniquely numbered.

8. **Full path in `dispatch-spawn-worker` error message** (subagent 3, low confidence) — only fires when the spawner-validated path doesn't exist; useful for debugging the misconfiguration, no real disclosure risk.

9. **Tmpdir paths in test failure output** (subagent 3, low confidence) — synthetic `mktemp -d` paths in test-only output, no production data.

### Clean (no findings)

- **Subagent 5 (Firebase)** — diff touches no `firestore.rules`, Firebase config, or SDK code.
- **Subagent 6 (Secrets scan)** — no hardcoded credentials, no `.env` additions, no token-shaped strings in scripts or SKILL.md fixtures.
- **Subagent 8 (built-in `/security-review`)** — no findings on dispatch-orchestration shell scripts.
- **Subagent 9 (CodeQL alerts)** — `code-scanning/alerts?state=open&ref=refs/pull/863/head` returned `[]`.

### Coverage

All 9 subagents completed successfully on first attempt; no retries needed.

PR #865, opened 2026-05-28
dispatch: scale concurrent /dispatch-worker count to a target rate-limit-window usage

natb1:
## Summary

Adds a router-side concurrency budgeter that drives the live
`dispatch-worker-*` count toward a target weekly usage while never
exceeding a 5-hour budget schedule. The router now consults
`dispatch-target-workers` and the live worker count before spawning;
when `live_count >= target_N` the spawn is skipped and the chain
re-seeds on the next tick.

Closes #845.

## What's in this PR

- **`dispatch-target-workers`** — pure script implementing the two
  composed schedules. Reads
  `~/.local/share/productivity-tui/rate_limits.json` (already
  maintained by the productivity-tui statusline hook). All six
  tunables are optional; defaults baked in match the issue spec
  (`target_weekly=90`, `target_5h=50`, `max_workers=8`, `tau=86400`,
  `k=2`, `five_hour_window=18000`).
- **`dispatch-config-load target-workers`** — schema validator for
  the optional `dispatch.config/target-workers.json` tunable file.
- **`claude_agents_count_by_name_prefix`** — new helper in
  `lib-claude-agents.sh` that counts live sessions whose `name`
  starts with a given prefix. Same UNKNOWN-on-daemon-failure contract
  as `claude_sessions_under`.
- **Router Step 6 gate** — two-line bash composition in
  `dispatch/SKILL.md` that short-circuits the spawn when the live
  worker count meets the target. Step 7 recognizes the skip-over-cap
  outcome as a clean completion. Fails open on daemon UNKNOWN; the
  per-worktree dedup inside `dispatch-spawn-worker` is the last-line
  defense.
- **`dispatch/SKILL.md` documentation** — a new Concurrency Budgeting
  subsection with both math tables, every tunable's default, and
  recalibration notes.

## Test plan

- [x] `bash .claude/skills/dispatch/scripts/test-dispatch-scripts.sh`
      — 485/485 passing.
- [x] New `dispatch-target-workers` tests cover: weekly-target reached,
      weekly-refresh passed, 5h-cap tripped, both-clear spawn, missing
      telemetry, seven_day-absent fallback, five_hour-absent (no cap),
      absolute cap, weekly-ramp monotonic sweep, 5h-cap sweep across
      the threshold, config tunable override, and per-field env
      override.
- [x] New `dispatch-config-load target-workers` tests cover: valid
      schema, absent file, non-number rejection, empty-object accepted.
- [x] New `claude_agents_count_by_name_prefix` tests cover:
      prefix-matching, no-match, daemon-failure UNKNOWN, non-array
      UNKNOWN.
- [x] New router-gate composition tests cover: live>=target skip,
      live<target spawn, daemon UNKNOWN fail-open.
- [x] Spot-checked against the real `rate_limits.json` on the dev box:
      target_N=0 (5h cap is the binding constraint at 21% used vs.
      ~17% cap), gate routes to skip with live_count=0 since 0>=0.

### DIFF
--- a/.claude/skills/dispatch/SKILL.md
+++ b/.claude/skills/dispatch/SKILL.md
@@ -517,6 +517,13 @@ many concurrent `dispatch-worker-*` sessions should run:
   | 1h (3600s)     | 0.8             | 40%    |
   | 0              | 1.0             | 50%    |
 
+  At window-start, `cap_5h = 0` and the gate condition `used_5h >= cap_5h`
+  evaluates as `0 >= 0`, so spawning is briefly paused for the first slice of
+  every 5h window. This is intentional — the linear schedule enforces the
+  per-window ceiling from second zero. The blackout clears as soon as `cap_5h`
+  rises above the current `used_5h`; if `used_5h` was 0 at refresh, this
+  happens within seconds.
+
 Tunables (each optional in `dispatch.config/target-workers.json`; defaults
 baked into the script):
 
--- a/.claude/skills/dispatch/SKILL.md
+++ b/.claude/skills/dispatch/SKILL.md
@@ -442,25 +442,102 @@ the lock*). The worker in Step 6 onward runs lock-free.
 
 ## 6. Spawn the Worker
 
-Spawn a `/dispatch-worker <N>` background job with `cwd=$WORKTREE_PATH` (the
-path resolved in Step 5). Run with `dangerouslyDisableSandbox: true` — the
-script reaches the local Claude daemon over a socket; see
-`.claude/rules/sandbox.md`:
+Before spawning, consult the concurrency budgeter (see *Concurrency
+budgeting* below). If the live worker count already meets the target,
+**skip the spawn** for this tick — the chain re-seeds on the next router
+tick when budget reopens. Run with `dangerouslyDisableSandbox: true` — both
+the budgeter and `lib-claude-agents.sh` reach the local Claude daemon over a
+socket; see `.claude/rules/sandbox.md`:
 
 ```bash
-.claude/skills/dispatch/scripts/dispatch-spawn-worker <N> "$WORKTREE_PATH"
+source .claude/skills/dispatch/scripts/lib-claude-agents.sh
+TARGET_N=$(.claude/skills/dispatch/scripts/dispatch-target-workers)
+if LIVE_COUNT=$(claude_agents_count_by_name_prefix dispatch-worker-); then
+  if (( LIVE_COUNT >= TARGET_N )); then
+    echo "router: skipping spawn — $LIVE_COUNT live worker(s) >= target $TARGET_N"
+    # Skip-over-cap is a clean completion; proceed to Step 7.
+  else
+    .claude/skills/dispatch/scripts/dispatch-spawn-worker <N> "$WORKTREE_PATH"
+  fi
+else
+  # UNKNOWN — daemon query failed. Fail open and proceed to spawn; the
+  # per-worktree dedup inside dispatch-spawn-worker is the last-line
+  # defense if a worker is already there.
+  .claude/skills/dispatch/scripts/dispatch-spawn-worker <N> "$WORKTREE_PATH"
+fi
 ```
 
-It prints `spawned` (a worker was started) or `deduped` (another `dispatch-*`
-session is already live at `$WORKTREE_PATH` — the per-worktree invariant
-prevents two workers from racing on the same worktree) and exits 0; it exits
-non-zero when a worker was spawned but did not register.
+`dispatch-spawn-worker` prints `spawned` (a worker was started) or `deduped`
+(another `dispatch-*` session is already live at `$WORKTREE_PATH` — the
+per-worktree invariant prevents two workers from racing on the same worktree)
+and exits 0; it exits non-zero when a worker was spawned but did not register.
 
 The router never derives the phase, runs a phase skill, or invokes the
 pre-implementation relevance review — those are the worker's responsibilities
 (`/dispatch-worker` Steps 1–3).
 
-After the spawn returns, **proceed to Step 7** (terminal disposition).
+After the spawn (or skip) returns, **proceed to Step 7** (terminal disposition).
+
+### Concurrency budgeting
+
+`dispatch-target-workers` composes two open-loop schedules that decide how
+many concurrent `dispatch-worker-*` sessions should run:
+
+- **Weekly ramp** (sets the magnitude) — drives target_N
+  conservative-early/aggressive-late toward `target_weekly_usage_pct` by the
+  weekly window's `resets_at`. The shape is
+  `(tau / (remaining_weekly + tau)) ^ k`; smaller `tau` and larger `k` mean
+  more deferred consumption.
+
+  | remaining_weekly | ramp_weekly | candidate target_N |
+  |------------------|------------:|-------------------:|
+  | 7 days (604800s) | 0.0156      | 0                  |
+  | 3 days (259200s) | 0.0625      | 1                  |
+  | 1 day (86400s)   | 0.25        | 2                  |
+  | 6 hours (21600s) | 0.64        | 5                  |
+  | 1 hour (3600s)   | 0.922       | 7                  |
+  | 0                | 1.0         | 8                  |
+
+- **5-hour linear cap** (binary gate) — over each 5-hour window, the cap on
+  `used_5h` rises linearly from `0` at window-start to
+  `target_five_hour_usage_pct` at window-end. When `used_5h >= cap_5h`, the
+  budgeter prints `0` regardless of the weekly ramp's otherwise-positive
+  output, pausing spawning until either the window resets or the cap rises
+  enough to clear it.
+
+  | remaining_5h   | elapsed_5h_frac | cap_5h |
+  |----------------|----------------:|-------:|
+  | 5h (18000s)    | 0.0             | 0%     |
+  | 4h (14400s)    | 0.2             | 10%    |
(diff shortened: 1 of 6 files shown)


natb1:
## dispatch-qa: clean pass

**Items walked**: 8 PASS / 0 FAIL / 0 SKIP

Non-browser walkthrough — backend/shell-script change covering the new
`dispatch-target-workers` budgeter, the `target-workers` schema added to
`dispatch-config-load`, and the concurrency-gate integration in `/dispatch`
Step 6.

### Items

1. **Test harness** — `test-dispatch-scripts.sh` passes 485/485, covering the
   new `dispatch-target-workers` section (Tests 1–12), the three new
   `dispatch-config-load target-workers` tests (8–11), and the three new
   `lib-claude-agents` counter tests (10–16).
2. **Real-env invocation** — `dispatch-target-workers` against the host's
   `rate_limits.json` printed `0` because the 5-hour cap is currently
   tripped (`used_5h=43` vs `cap_5h≈30` at 0.6 elapsed window). Algorithm
   matches the SKILL.md table.
3. **Missing-telemetry fallback** — `DISPATCH_TARGET_WORKERS_RATE_LIMITS_PATH=/nonexistent`
   printed `1` to stdout and a one-line `dispatch-target-workers: …
   using fallback (1)` note to stderr, exit 0.
4. **Weekly-target gate** — `used_weekly=90 >= target=90` → `0`.
5. **5h binary cap** — late-week regime (`remaining_weekly=1h`,
   `remaining_5h=1h`, `cap_5h=40`): `used_5h=45` → `0`, `used_5h=30` → `7`.
6. **Absolute cap** — ramp-saturated (`remaining_weekly=1s`) clamps to
   default `max_concurrent_workers=8`.
7. **`dispatch-config-load target-workers`** — normalizes the example
   `target-workers.json` and reports all six tunables.
8. **Router gate snippet** — the inlined Step-6 snippet against a fake
   `claude agents --json` daemon: 3 sessions where 2 prefix-match
   `dispatch-worker-` and `target_N=2` → `skip`; 1 prefix-matching session
   and `target_N=2` → `spawn`. Non-worker sessions correctly excluded.

### Notes

- No browser component — pure backend change. Standard non-browser path.
- No `testOnly` data involved.

natb1:
## `/code-review-fix` report

`/code-review max` surfaced 5 findings. 2 fixed, 1 informational, 2 disregarded.

### Findings that were fixed (commit `ec64423`)

- **`.claude/skills/dispatch/scripts/dispatch-target-workers` (≈line 87)** — Comment claimed jq's `//` operator "only fires on null"; it also fires on `false`. Rewrote the comment to clarify the operator semantics and note that current tunables are all numeric so the boolean edge case does not affect this loop today, with a forward-looking note for adding boolean tunables.
- **`.claude/skills/dispatch/scripts/dispatch-target-workers` (former lines 239/243/244)** — Removed three dead post-rounding clamp guards. The math already guarantees `target_n ∈ [0, max_workers]` for `remaining_weekly > 0` (gated by Step 2 of the algorithm), so the runtime guards were dead code. Also collapsed the algorithm header comment from 8 steps to 7 to reflect the removed Step 7.

### Informational

- **`.claude/skills/dispatch/scripts/test-dispatch-scripts.sh` (router-gate Tests 14–16)** — The router-gate tests hardcode `TARGET_N=2` instead of invoking `dispatch-target-workers`; the bash-composition logic is fully covered, but there is no integration test that walks the full chain from `dispatch-target-workers` stdout to the gate comparison. The finding itself shows the `TARGET_N empty` failure mode is structurally safe (bash arithmetic syntax error → fail-open spawn path), so no defect surfaces from the current coverage. Surfaced for awareness if a future change adds parsing complexity at the boundary.

### Disregarded

- **`.claude/skills/dispatch/scripts/dispatch-target-workers` (line 227)** — 5h gate `used_5h >= cap_5h` blocks when both equal 0 at the exact moment a fresh 5h window opens. Implementation matches the spec exactly — the spec's algorithm step is literally `If used_5h >= cap_5h, print 0`, and the spec's math table shows `cap_5h=0%` at window-start. The corner case lasts a single router tick at the precise `remaining_5h == window_5h` instant; once one second elapses, `cap_5h > 0` and the gate clears with `used_5h=0`. Not worth deviating from the spec for a vanishingly narrow window.
- **`.claude/skills/dispatch/scripts/dispatch-target-workers` (lines 98–109)** — Six sequential `jq` subprocess forks to extract config values could be collapsed into one `@tsv`-emitting jq + bash split. The current per-field pattern is more readable, makes adding a new field trivially clear, and the per-router-tick fork cost is negligible against the ~minute heartbeat. Stylistic preference rather than an actual improvement.

### Verification

- `bash .claude/skills/dispatch/scripts/test-dispatch-scripts.sh` — 485/485 passing after the fixes.

natb1:
## Fixed

- awk `+ 0` coercion pattern: one-liner explains the explicit numeric coercion idiom: 2f17129
- `claude_agents_count_by_name_prefix` scope: corrected "repo-wide" → "machine-wide" in header doc + inline comment, with a fail-safe note about cross-checkout inflation: 2f17129
- 5h-cap window-start blackout: SKILL.md gains a short note that `cap_5h=0` ⇒ `0 >= 0` trips the gate for the first slice of every 5h window — intentional, clears as soon as the cap rises: 2f17129
- `ramp_tau_seconds=0` documented by test: new Test 13 asserts tau=0 disables spawning (test suite now 495/495): 2f17129

## Informational

- awk step numbering inside the script mirrors the outer step numbering (Step 1/2/...) — kept aligned; future renumbering of the outer steps could trip a reader, but no action needed today.
- `write_rl` test helper builds JSON via string interpolation; safe because all non-`absent` inputs are numeric — the existing `[[ != "absent" ]]` guard is sufficient.
- Router-gate composition lives in SKILL.md prose (not an executable script) — the inline `TARGET_N=...; if LIVE_COUNT ...` block is tested via the harness, but a future divergence between the prose and the production path would not be caught. Acceptable for the two-line composition.
- Tunable range validation: `dispatch-config-load`'s schema validates that fields are numeric but does not bound them (e.g. `max_concurrent_workers=0`, very large `ramp_shape_k`). Test 13 documents the most degenerate case (`tau=0`); broader bounds-checking is deferred to a future hardening pass if it surfaces in practice.
- `FILE_OK` flag interaction with env-override path: when the file is genuinely missing and only `five_hour` env overrides are set, the "rate_limits.json missing or unreadable" message is correctly named — the file *was* attempted unconditionally — so no message-accuracy issue exists.

## Dismissed

_None._

## Deferred

_None._

natb1:
## Security review

Structured fan-out: 6 domain subagents (input validation, auth/access control, data exposure, dependency audit, Firebase-specific, secrets scan) + red team + built-in `/security-review` scan + CodeQL alerts.

### Fixed

**Lower-bound validation on `target-workers` config tunables** — `dispatch-config-load`:253-272 — `A05:2021 Security Misconfiguration` / Denial of Service. The validator accepted `0` and negative values for all six numeric tunables. A typo or hostile config with `five_hour_window_seconds: 0` caused awk division-by-zero in `dispatch-target-workers`:226, emitting no stdout; the router gate `(( LIVE_COUNT >= "" ))` then always evaluated true → permanent skip-spawn DoS. `max_concurrent_workers: -1` and `ramp_tau_seconds: 0` had similar effects. Validator now rejects each present tunable that is `<= 0` with `"<field> must be > 0 if present, got <value>"`. Tests cover `five_hour_window_seconds: 0` and `max_concurrent_workers: -1` rejection; the existing `ramp_tau_seconds: 0` DoS-documentation test was updated to assert the post-fix safe behavior (validator rejects, baked-in defaults used). Commit `105fb5b`. Flagged by red-team subagent (high confidence) and auth/access-control subagent (medium).

### Out of scope (pre-existing, not introduced by this PR)

- **High-severity transitive npm CVEs** in `@xmldom/xmldom@0.7.13`, `fast-xml-parser@5.7.2`, `fast-xml-builder@1.1.5` (via `epubjs` and `firebase-admin`→`@google-cloud/storage`). Five XML-injection / DoS findings; `npm audit fix` available for the firebase-admin path. PR #865 does not change `package.json` or lockfiles.
- **`~/.local/share/productivity-tui/` directory mode 0755** (world-listable). `rate_limits.json` itself is `0600`. Directory mode is set by the `productivity-tui` writer hook, not by this PR.

### Out of scope (judgement-call deferrals)

- **No upper-bound on `max_concurrent_workers`** — the field IS the user's chosen absolute cap; imposing a project-level ceiling would conflict with its stated semantics. Per-tick spawn-one + per-worktree dedup are the real cost-burn backstops.
- **Router-side `TARGET_N` empty-stdout guard** (SKILL.md illustrative snippet) — with the Fixed item above, `dispatch-target-workers` no longer crashes on valid configs. The dispatching router's invocation pattern is established when the router caller lands.
- **Numeric-type validation of telemetry fields from `rate_limits.json`** — file is `0600`, written atomically by a trusted local hook; corruption risk is realistic but the file is not an attack surface in the documented single-user-dev-machine threat model.

### False positives

- **Stderr leak of `$RL_PATH`** in the fallback message (Information Disclosure). The path is `$HOME/.local/share/productivity-tui/rate_limits.json` — leakage is meaningful only on multi-user shared machines, which is not the supported deployment context. The diagnostic value to the operator outweighs the disclosure on single-user dev machines.
- **Unguarded awk stderr** — only Anthropic rate-limit percentages and epoch timestamps would leak in an awk fatal, and with the Fixed item above awk no longer fails on validated configs.

### Clean

Input validation, secrets scan, Firebase-specific, built-in `/security-review` scan, and CodeQL alerts (empty array on PR head ref) reported no findings.

PR #881, opened 2026-05-28
dispatch: cap-keyed re-seed replaces static 9 AM timer

natb1:
Closes #725

## Summary

Replaces the static daily 9 AM `systemd.user` timer with a **cap-keyed
dynamic re-seed**: when a dispatch tick's Step 6 skips the spawn because
the concurrency budget is at the cap, the chain schedules a one-shot
transient `systemd.user` timer to fire at the next rate-limit `resets_at`
read from `~/.local/share/productivity-tui/rate_limits.json`. The chain
resumes the instant the cap window reopens, rather than waiting on a
coarse static heartbeat.

The mechanism only covers the *cap-stall* class of chain stall. Empty
queue / all-parked stalls are handled by the office-hours queue
(#755 / #757 / #758) — engaging a parked item re-seeds via the
Step 4 hand-off.

## What changed

- **New** `.claude/skills/dispatch/scripts/dispatch-schedule-reseed` —
  reads `rate_limits.json`, picks the earliest blocking `resets_at`
  (`min(weekly_resets_at, 5h_resets_at)` when used%≥target on either
  axis), and runs `systemd-run --user --unit=dispatch-reseed-<epoch>` to
  schedule a one-shot timer that runs `dispatch-spawn-router` from the
  main worktree. The epoch suffix gives idempotency: a repeated call
  observing the same blocking cap collides on the unit name and is a
  no-op.

- **Wired** the trigger into `/dispatch` Step 6: the `drain
  concurrency-cap` skip path now calls `dispatch-schedule-reseed`
  immediately before proceeding to Step 7.

- **Removed** `nix/home/dispatch-timer.nix` (the static 9 AM
  `systemd.user` service+timer) and its import from
  `nix/home/default.nix`. Replaced the "#725 daily restart" SKILL.md
  subsections in both `/dispatch` and `/dispatch-worker` with a
  cap-keyed-re-seed description, and updated the three sites that
  referenced "9 AM" inline (queue-empty drain, worktree-conflict drain,
  worker `drain waiting`). Refreshed comment-only "heartbeat" references
  in `dispatch-spawn-router`, `dispatch-spawn-worker`, and
  `dispatch-select-target`.

- **Tests** added 9 new cases in
  `.claude/skills/dispatch/scripts/test-dispatch-scripts.sh`: cap-hit on
  weekly schedules at weekly `resets_at`, cap-hit on 5h schedules at 5h
  `resets_at`, both hit picks the earlier, neither hit is a silent no-op,
  missing telemetry is a no-op with stderr, `seven_day`-absent + 5h hit
  schedules at 5h `resets_at`, `five_hour`-absent + weekly hit schedules
  at weekly `resets_at`, repeated call is idempotent (single timer),
  already-passed `reseed_at` is a no-op.

## Verification split

**At implement-time** (this PR):

- [x] `test-dispatch-scripts.sh` — 548/548 passing (9 new schedule-reseed
      tests included).
- [x] No stale "9 AM" / "daily" references remain in the dispatch
      SKILL.md set or scripts (only intentional context in
      `dispatch-schedule-reseed`'s own comment that explains why
      `Persistent=true` mirrors the legacy timer's behavior).

**At QA** (post home-manager-switch, against live user systemd):

- [ ] After `home-manager switch`, the static `dispatch.service` and
      `dispatch.timer` units are gone from
      `systemctl --user list-unit-files`.
- [ ] When the budgeter returns 0 because used%≥target on either axis,
      Step 6's skip path schedules a `dispatch-reseed-<epoch>.timer`
      visible in `systemctl --user list-timers`.
- [ ] When the scheduled timer fires, the spawned session runs in
      auto-mode (not `--dangerously-skip-permissions`) and parks in
      `Needs input` when a phase needs it (resumable via
      `claude attach`).
- [ ] A tick firing while a previous dispatch is active exits via the
      `/dispatch` lock without starting an overlapping run.


natb1:
## `/dispatch-qa` results

Non-browser PR — dispatch infrastructure. Single user-acceptance pass.

**Items walked: 8 — PASS 8 / FAIL 0 / SKIP 0** (item 8 accepted by user as deferred to natural cap-hit; the test suite covers the systemd-run argv, and the live verification is operational).

### Walked items

1. **`dispatch-schedule-reseed` script exists and is executable** — PASS. `.claude/skills/dispatch/scripts/dispatch-schedule-reseed`, mode 0755, bash shebang.
2. **Static `nix/home/dispatch-timer.nix` is gone, no `dispatch.timer` references in `nix/`** — PASS. File absent; `nix/home/default.nix` imports do not include it; no `dispatch-timer` or `dispatch.timer` matches under `nix/`.
3. **Step 6's `drain concurrency-cap` path invokes `dispatch-schedule-reseed`** — PASS. Call wired at `.claude/skills/dispatch/SKILL.md:487` (cap-skip branch); design described at lines 501–504 and 666–670.
4. **`dispatch-schedule-reseed` test suite (9 new cases) passes** — PASS. Full `test-dispatch-scripts.sh` run: 565/565. Schedule-reseed cases:
   - weekly cap-hit schedules at the weekly `resets_at`
   - 5h cap-hit schedules at the 5h `resets_at`
   - both caps hit → schedules at the earlier `resets_at`
   - neither cap hit → silent no-op
   - missing `rate_limits.json` → no-op with stderr diagnostic
   - `seven_day` absent + 5h cap-hit
   - `five_hour` absent + weekly cap-hit
   - repeated call is idempotent (single timer)
   - already-passed `reseed_at` → no-op
5. **End-to-end with mocked `systemd-run` and synthetic `rate_limits.json` produces the right argv** — PASS. Covered by the schedule-reseed test suite's "both caps hit; picks earlier" + systemd-run argv assertion (unit name, calendar, cwd, exec).
6. **SKILL.md "9 AM" / "daily restart" prose replaced** — PASS. Zero matches for `9 AM`, `9AM`, `daily restart`, `daily (9 AM)` across `.claude/skills/dispatch/SKILL.md` and `.claude/skills/dispatch-worker/SKILL.md`.
7. **No live `dispatch.timer` / `dispatch.service` on user systemd** — PASS. `systemctl --user list-unit-files | grep dispatch` and `systemctl --user list-timers --all | grep dispatch` both empty. Note: the static timer was on branch only (added 704bbc8 → changed cf3c1d7 → removed 1d5a24e), never reached `main`, so nothing for `home-manager switch` to remove.
8. **Live cap-hit produces a `dispatch-reseed-<epoch>.timer`** — PASS (deferred). User accepted: the test suite covers the systemd-run argv, the live no-stale-units check passed, and operational verification will fall out naturally the next time the cap is hit.

### Bugs found

None.

### QA notes

- Step 0.5 merged `origin/main` into `725-run-dispatch-every-30-minute` cleanly. The `commit-merge-push` subagent emitted a verbose hallucination about implementing #824 — that report was discarded; the actual operation was the 4-commit ort merge into HEAD, then pushed as `6a27c62`.
- The full dispatch test suite was run (not just the schedule-reseed subset); 565/565 pass.

natb1:
## Code Review

### Findings that were fixed

- `.claude/skills/dispatch/scripts/dispatch-schedule-reseed` — the `if cmd; then exit 0; fi; RC=$?; exit "$RC"` pattern swallowed the real exit code: `$?` after `fi` is the exit status of the `if` construct (0 — the condition was evaluated successfully), not the failing command's. Every non-`already-exists` `systemd-run` failure (D-Bus down, malformed `--on-calendar`, missing binary) silently exited 0, so the documented "pass-through on unexpected systemd-run failure" contract was broken and the chain would stay stalled with no recovery scheduled. Moved the command out of the `if`, captured `RC=$?` immediately, and added regression test 10 in `test-dispatch-scripts.sh` asserting a 42 exit-code passthrough. Fix in commit 2649da9.
- `.claude/skills/dispatch/SKILL.md` — the `drain concurrency-cap` description still said the chain "re-seeds on the next router tick when budget reopens." There is no "next router tick" mechanism after this PR removes the static periodic timer; recovery is driven by the cap-keyed timer. Updated to point at *The #725 cap-keyed re-seed* section. Fix in commit 69bb35d.
- `.claude/skills/dispatch/scripts/dispatch-select-target` — comment edit in this PR claimed the cap-keyed re-seed fires `--health-only`, but the re-seed timer execs `dispatch-spawn-router` and no SKILL.md caller invokes `--health-only`. Reverted to a minimal accurate phrase. Fix in commit 69bb35d.
- `.claude/skills/dispatch/scripts/dispatch-handoff` — the `--early-stop` form's comment still said "the #725 heartbeat re-seeds the chain"; the heartbeat is gone after this PR. Replaced with an accurate two-class description (cap-keyed timer for cap-stalls, office-hours engagement for parked items). Fix in commit 69bb35d.

🤖 Code review run via `/code-review-fix`.

natb1:
## /review-fix

### Fixed

- SKILL.md sandbox prose clarified — the `dispatch-schedule-reseed` call in Step 6 is covered by the outer Bash block's `dangerouslyDisableSandbox: true`; the post-block sentence no longer reads like a separate "Run with…" instruction: f5285c7
- Already-passed-reset no-op now emits a stderr diagnostic (consistent with the missing-telemetry and idempotent-re-call no-ops); stdout-protocol header comment updated to accurately distinguish silent vs. stderr-diagnosed no-ops: f5285c7
- Test 9 extended to assert the new already-passed stderr diagnostic; new Test 11 added covering the partial-record path (`seven_day` present but `resets_at` null → block treated as absent): f5285c7

### Informational

- Reviewer noted a test-coverage gap for non-default `target_weekly_usage_pct` / `target_five_hour_usage_pct` config-file overrides. These tunables flow through `dispatch-config-load`, which has its own coverage; not added here.

### Dismissed

- `NOW="${DISPATCH_SCHEDULE_RESEED_NOW:-$(date +%s)}"` exit-code guard — defensive coding for a `date +%s` failure that essentially can't happen on any reachable system; reviewer themselves marked this "low severity, narrow failure mode."
- `awk`-wrap the `RESEED_AT`/`CAND_*` bash arithmetic on the speculative float-`resets_at` concern — the actual `rate_limits.json` schema is integer epoch seconds; the script already uses bash arithmetic for known-integer comparisons (`(( RESEED_AT <= NOW ))` at line 218) and `awk` only where percentages legitimately can be floats. Adding `awk` for known-integer fields would be defensive coding for a JSON schema variation that does not occur.

### Deferred

_None._

natb1:
## Security review

`/security-review-fix` fanned out 9 parallel subagents (6 domains + red team + built-in `/security-review` + CodeQL) against the diff vs current `origin/main`.

### Required — fixed

**Bash arithmetic-context command injection via tampered `rate_limits.json` or hostile env overrides**

- **Location**: `.claude/skills/dispatch/scripts/dispatch-schedule-reseed` — `WEEKLY_RESETS`/`FIVEH_RESETS` (then `CAND_WEEKLY`/`CAND_5H`/`RESEED_AT`) reached `(( CAND_WEEKLY <= CAND_5H ))` and `(( RESEED_AT <= NOW ))` without numeric validation. Bash arithmetic context evaluates array-index command substitution — a value like `a[$(touch /tmp/pwn)]` from a tampered `~/.local/share/productivity-tui/rate_limits.json` (or a hostile `DISPATCH_SCHEDULE_RESEED_RESETS_AT_*` env override) would execute the inner command when it reached `(( … ))`.
- **OWASP**: A03:2021 Injection
- **STRIDE**: Tampering → Elevation of Privilege
- **Confidence**: high
- **Found by**: red-team subagent (high confidence) + input-validation subagent (low confidence on the systemd-arg sink only — the arithmetic-context sink is the dangerous path).
- **Fix** (commit `657fc15`): validate `TARGET_WEEKLY`/`TARGET_5H` from `dispatch-config-load` as non-negative numbers (abort exit 2 on bad input); validate `NOW` (`$DISPATCH_SCHEDULE_RESEED_NOW` env override or `date +%s`) as an integer (abort exit 2); sanitize `WEEKLY_USED`/`FIVEH_USED`/`WEEKLY_RESETS`/`FIVEH_RESETS` after env overrides — any non-numeric/non-integer value is treated as missing telemetry (set to `""`) with a stderr diagnostic, falling through to the existing missing-telemetry no-op. Three new tests in `test-dispatch-scripts.sh` gate the RCE canary path (tampered weekly resets_at, tampered 5h resets_at, non-integer `NOW`, non-numeric used_percentage).

### Out-of-scope / informational

- **Home-directory path in stderr diagnostic** (`dispatch-schedule-reseed:158,232`) — data-exposure subagent flagged the `$RL_PATH` (and main-worktree path) appearing in missing-telemetry / not-in-a-repo diagnostics. Low confidence, intentional for debuggability, no exploit path on local developer tooling. Not fixed.
- **Theoretical runaway-timer DoS** (`dispatch-schedule-reseed:218,253–264`) — red-team subagent: a buggy telemetry producer that drifts `resets_at` by one second every tick would accumulate transient systemd-user timer units. Low confidence, requires a producer bug, and `dispatch-spawn-router`'s existing dedup-guard absorbs the spawn side. Not fixed; defensive caps deferred.
- **Pre-existing npm-audit high-severity advisories** (`@xmldom/xmldom`, `fast-xml-builder`/`fast-xml-parser`, `@google-cloud/storage` chain, `typescript-eslint` chain) — dependency-audit subagent: 5 high-severity advisories in transitive deps. None touched by this PR; pre-existing. Not actionable here.

### Clean

- Auth & access control, Firebase-specific, secrets scan — no findings.
- CodeQL code-scanning: no open alerts on the PR ref.

### Partial coverage

- The built-in `/security-review` subagent ran but reported "diff content is empty" — the inner `Skill` invocation appears to have resolved its working directory from the parent session's cwd (`worktrees/main`) rather than the subagent's intended cwd, so its diff scan saw nothing. The six domain subagents + red team + CodeQL cover the diff regardless; the built-in scan would have added overlap, not net new coverage. Noted for future skill-harness work.

### Dedup notes

- Input-validation subagent F1 (systemd-arg sink, low confidence) and red-team F1 (arithmetic-context sink, high confidence) name the same root issue at overlapping locations — collapsed into the single "Required — fixed" entry above, taking the higher confidence and broader fix scope.
- Input-validation F2 (env-var override path) is the same root issue reached via a different writer — covered by the same sanitizer block.

---
Generated by [/security-review-fix](https://github.com/natb1/commons.systems/blob/main/.claude/skills/security-review-fix/SKILL.md).

PR #897, opened 2026-05-29
budget: make updateJournalLegCleared atomic

natb1:
Closes #743

Wraps the read + guard + write in `updateJournalLegCleared` in a Firestore `runTransaction` so the guard against flipping `cleared` on a reconciled leg holds under concurrent reconciliation. Tests added for happy path, reconciled-leg rejection, and missing-leg cases.

PR #904, opened 2026-05-29
budget: make updateJournalLegCleared atomic

natb1:
Closes #897
Closes #743

## Summary

`updateJournalLegCleared` in `budget/src/firestore.ts` performed a non-atomic read-modify-write: a `getDoc` to read `reconciledAt`, an `assertLegStateTransition` guard, and an `updateDoc` to set `cleared`. Between the read and the write, a concurrent reconciliation can set `reconciledAt` on the same leg, so the subsequent `updateDoc` flips `cleared` on a now-reconciled leg — violating the "reconciled is terminal" invariant that `assertLegStateTransition` enforces but Firestore rules do not.

This PR wraps the read + guard + write in `runTransaction(db, async (tx) => { ... })` using `tx.get` and `tx.update`, so the guard re-runs against the value Firestore actually commits against. The existing error messages and call sites are preserved.

## Scope

- `budget/src/firestore.ts` — add `runTransaction` to the `firebase/firestore` import; rewrite `updateJournalLegCleared`'s body inside a transaction.
- `budget/test/firestore.test.ts` — mock `runTransaction` (passing a `tx` exposing `get` / `update`); add three tests: happy path, reconciled-leg rejection, missing-leg rejection.

Out of scope: the IDB sibling (`budget/src/data-source.ts`) — IDB is single-threaded in-process, so no TOCTOU exists there. Firestore security rules — adding rule-level enforcement of the reconciled-is-terminal invariant is a separate follow-up.

### DIFF
--- a/budget/src/firestore.ts
+++ b/budget/src/firestore.ts
@@ -1,4 +1,4 @@
-import { collection, doc, getDoc, getDocs, query, setDoc, updateDoc, where, increment, writeBatch, Timestamp, deleteDoc, type QueryDocumentSnapshot, type DocumentData } from "firebase/firestore";
+import { collection, doc, getDoc, getDocs, query, runTransaction, setDoc, updateDoc, where, increment, writeBatch, Timestamp, deleteDoc, type QueryDocumentSnapshot, type DocumentData } from "firebase/firestore";
 import { nsCollectionPath } from "@commons-systems/firestoreutil/namespace";
 import { requireOneOf } from "@commons-systems/firestoreutil/validate";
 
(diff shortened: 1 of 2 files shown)


natb1:
Closing as a duplicate of #897, which is the legitimate PR for #743.

This PR was spawned off branch \`897-budget-make-updatejournalleg\` — i.e. the dispatch chain treated PR #897 as if it were an issue and produced a second, identical fix (same \`runTransaction\` wrapping of \`updateJournalLegCleared\`). #897 (branch \`743-...\`) is further along: it carries \`dispatch:qa-done\` and has merged \`origin/main\`. Both PRs declared \`Closes #743\`, which left the issue marked as closed-by two PRs. Closing this one resolves that.

PR #912, opened 2026-05-29
fix(dispatch): derive dispatch-stop ISSUE_NUM from JOB_NAME not cwd branch

natb1:
Closes #910

## Problem

`.claude/hooks/dispatch-stop.sh` resolved the issue number from `git rev-parse --abbrev-ref HEAD` — the session cwd's branch. A background worker's cwd is the spawn cwd (`worktrees/main`), not the target worktree (the `cd` in `/dispatch-worker` Step 0 runs in a Bash subshell and does not persist). So `git rev-parse` returned `main`, the numeric grep yielded empty, and the hook hit `exit 0` before reading the marker, computing the phase, propagating the chain, or self-closing. Every worker leaked its session.

## Fix

Derive `ISSUE_NUM` from the already-validated `JOB_NAME` (`<N>-<slug>`, stable regardless of cwd). Discriminator 2 already guarantees `JOB_NAME` matches `^[0-9]+-`, so the numeric prefix is in hand. No `git rev-parse --abbrev-ref HEAD` issue-resolution path remains in the hook.

## Test

Adds a regression test in the `=== dispatch-stop ===` block reproducing the wrong-cwd scenario (`current-branch.txt=main`, `JOB_NAME=123-foo-bar`) and asserting the hook reaches Branch B (spawn + self-close each invoked once) rather than exiting early. Full suite: 596/596 passing.


natb1:
## Dispatch QA — #910 (dispatch-stop ISSUE_NUM from JOB_NAME)

Non-browser path (Stop-hook shell change). Single user-acceptance pass.

**Items walked:** 5 — all machine-verifiable.

**Counts:** 5 PASS / 0 FAIL / 0 SKIP. No bugs found.

| # | Item | Result |
|---|------|--------|
| 1 | Hook derives `ISSUE_NUM` from `JOB_NAME` (line 58 `${JOB_NAME%%-*}`) | PASS |
| 2 | Wrong-cwd (branch `main`) + valid `JOB_NAME` → hook exits 0, no early guard exit | PASS |
| 3 | Branch B advance: strip office-hours, spawn router, self-close each run once | PASS |
| 4 | Regression test (Test 12) present; full suite 596/596 passed | PASS |
| 5 | No remaining `git rev-parse --abbrev-ref HEAD` issue-resolution path in hook | PASS |

Verified against the live test harness (`test-dispatch-scripts.sh`), which stubs `git`/`gh`/dispatch scripts and drives the real hook. The wrong-cwd regression (Test 12) sets the stubbed branch to `main` with `JOB_NAME=123-foo-bar` and asserts the hook reaches Branch B (spawn + self-close) instead of the old line-58 early exit.

natb1:

No correctness bugs found in the diff. `${JOB_NAME%%-*}` is provably safe and
non-empty under `set -u` because discriminator 2's `^[0-9]+-` guard runs first,
so the dropped `[ -n "$ISSUE_NUM" ]` check is correctly redundant.

No code changes were applied.

### Informational findings

- **Dead `current-branch.txt` fixtures in the stop tests** — now that the hook
  no longer calls `git rev-parse --abbrev-ref HEAD`, the git stub in
  `stop_setup` and the per-test `current-branch.txt` writes are no longer read
  during a run. No action taken: the 11 pre-existing writes live on `main` and
  are outside this PR's diff (which only adds Test 12), and the git stub +
  Test 12's `current-branch.txt=main` fixture double as a deterministic
  regression guard — if a future change reintroduced a cwd-branch read, the
  stub would serve `main`, `ISSUE_NUM` would go empty, and Test 12 would fail.
  Surfaced for human reference only.

natb1:
## Review phase

`/review` found no issues. The change derives `ISSUE_NUM` from the
Discriminator-2-validated `JOB_NAME` (`^[0-9]+-`), removing the now-redundant
branch-derived fallback, and adds Test 12 as a wrong-cwd regression guard. Full
`test-dispatch-scripts.sh` suite passes (596/596), including the new test.

## Fixed
_None._

## Informational
_None._

## Dismissed
_None._

## Deferred
_None._

natb1:

Structured security review of this PR's pending changes (diff vs. `origin/main`
merge-base). Fan-out: 6 domain reviewers + red team + built-in `/security-review`
+ CodeQL alerts (9 subagents). The diff touches two files:
`.claude/hooks/dispatch-stop.sh` and a regression test in
`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`.

**Result: no required fixes. No code changes in this phase.**

### Findings & dispositions

| # | Finding | Location | Confidence | Disposition |
|---|---|---|---|---|
| 1 | `ISSUE_NUM` is not re-validated as a positive integer after `${JOB_NAME%%-*}` | `dispatch-stop.sh:58` | medium | **Not fixed — false-positive** |
| 2 | `dispatch-phase` does not validate its argument | `dispatch-phase` (not in diff) | low | **Not fixed — out-of-scope** |
| 3 | `jq` parse fallback silently swallows tool-absence | `dispatch-stop.sh:47` (not in diff) | low | **Not fixed — out-of-scope** |
| 4 | 40 pre-existing `npm audit` CVEs (5 high, 33 moderate, 2 low) | dependency tree | high–low | **Not fixed — out-of-scope** |
| 5 | Test 12 does not assert which issue number reached `gh` | test file | low | **Not fixed — informational** |

Clean domains (no findings): data exposure, Firebase-specific, secrets scan,
built-in `/security-review`, CodeQL (empty alert set for this ref).

### Why finding 1 is a false-positive

Three subagents (input-validation, auth, red-team) independently flagged the same
root concern at `dispatch-stop.sh:58`: the new `ISSUE_NUM="${JOB_NAME%%-*}"` is not
re-validated before flowing into `gh issue edit` / `dispatch-find-pr` /
`dispatch-phase`. It is not exploitable:

- **Already guaranteed.** Discriminator 2 (`dispatch-stop.sh:48`) gates execution
  on `[[ "$JOB_NAME" =~ ^[0-9]+- ]]`, so the numeric prefix `${JOB_NAME%%-*}` is
  always a non-empty digit run. This is the design decision recorded in the issue.
- **No injection surface.** `JOB_NAME` itself never reaches a command, and every
  consumer double-quotes `"$ISSUE_NUM"` — no word-splitting or command
  substitution is possible.
- **No privilege boundary.** `state.json`'s `.name` is written by the trusted
  Claude daemon at spawn. The attack scenarios require write access to
  `$CLAUDE_JOB_DIR/state.json`, which is equivalent to local shell access as the
  daemon's own user — no boundary is crossed.
- The locale "full-width Unicode digit" variant does not hold: glibc's `[0-9]`
  collation range does not match full-width digits in practice.

The defense-in-depth suggestion (add `[[ "$ISSUE_NUM" =~ ^[1-9][0-9]*$ ]]`) is
noted but declined here — it validates a scenario discriminator 2 already
precludes, and the issue explicitly reasoned it unnecessary.

### Out-of-scope notes

Findings 2–4 are in code this PR did not change (the `dispatch-phase` arg check,
the pre-existing `jq` fallback line, and the dependency tree — no `package.json`
or lockfile change in this diff). Finding 5 is a test-quality suggestion, not a
vulnerability.

PR #898, opened 2026-05-29
dispatch-finalize-selection writes marker in target worktree

natb1:
## Summary

- Make `dispatch-finalize-selection` take the target worktree path as
  its one required argument, `cd` into it, then write
  `tmp/dispatch-worktree` and `exec dispatch-acquire-lock --release`.
  Previously the wrapper wrote the marker into `PWD`, and every
  `/dispatch` router runs from `worktrees/main`, so the marker leaked
  into `worktrees/main/tmp/dispatch-worktree` and let any subsequent
  router reclaim the selection lock via the marker-bearing-cwd path —
  defeating per-tick selection serialization.
- Update `/dispatch` SKILL.md Step 5 to drop the duplicate manual
  marker-write block; the wrapper now owns both the marker write and
  the lock release on every proceed path. The `worktree-create.sh`
  hook still writes the marker as the outside-of-dispatch defense.
- Pin the cd-first contract in `test-dispatch-scripts.sh`. The
  existing router smoke now asserts the marker lands in the target
  and is absent from the router-equivalent cwd; four new
  `dispatch-finalize-selection` tests cover happy-path/regression,
  missing-arg, invalid-path, and extra-arg.

Closes #896.

**Manual cleanup after merging:** delete any pre-existing
`worktrees/main/tmp/dispatch-worktree` from environments running an
older `/dispatch` — the bug persists until that file is removed.

## Test plan

- [x] All 609 tests in `.claude/skills/dispatch/scripts/test-dispatch-scripts.sh` pass
- [ ] On merge, verify `worktrees/main/tmp/dispatch-worktree` is removed (manual)
- [ ] Next `/dispatch --bg` run does not re-create the file in `worktrees/main`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #913, opened 2026-05-29
dispatch-resolve-worktree: reconcile a reused worktree's branch against the PR head branch


natb1:
## Context

`dispatch-resolve-worktree` matches an existing worktree to a target purely by the issue-number prefix of its branch (`.claude/skills/dispatch/scripts/dispatch-resolve-worktree:39-49` — `WT_NUM == ISSUE`), then emits `enter` / `conflict` / (post-#905) a reuse outcome. It never checks that the matched worktree's *checked-out branch* is the branch the target's PR actually lives on.

`dispatch-find-pr` resolves a target's PR through GitHub's `closedByPullRequestsReferences` linkage, which is robust to branch naming. So the two resolvers use different ground truths: the PR is found by GitHub closing-linkage, but the worktree is matched by local branch-name prefix. When two `<issue>-*` branches exist for one issue, the two disagree and the worker lands on a branch that has none of the PR's commits.

## Expected vs actual (observed 2026-05-29, issue #896 / PR #898)

Two branches existed for #896:

- `896-dispatch-dispatch-finalize-s` — pre-created by the dispatch `implement` tick at 02:37 (title-derived slug), never took a commit, never switched branches. Registered as the `896-*` worktree.
- `896-finalize-selection-marker` — the human's actual implementation (3 commits), pushed and opened as draft PR #898 at 14:57 (`Closes #896`). Not checked out in any worktree.

On the `qa` tick, `dispatch-find-pr 896` correctly returned PR #898 (CI green → phase `qa`), but `dispatch-resolve-worktree 896` matched the stale `896-dispatch-dispatch-finalize-s` worktree by prefix and routed the qa worker there.

- **Expected:** the worker operates on the PR's head branch (`896-finalize-selection-marker`), so the phase runs against the PR's diff.
- **Actual:** the worker entered a worktree whose branch is one commit *behind* `origin/main` with zero of the PR's commits. `git diff origin/main...HEAD` was empty; QA had nothing to verify and parked #896 to `dispatch:office-hours`.

## Reproduction

1. For some issue `<N>`, have the dispatch `implement` phase pre-create a worktree on the title-slug branch `<N>-<title-slug>` (no commits land on it).
2. Implement `<N>` on a differently-named branch `<N>-<other-slug>` and open a PR that closes `<N>` (a hand-authored branch, or any flow where the PR head branch differs from the router's slug).
3. Run a `/dispatch <N>` tick at any green-CI phase.
[...]

Issue #837, opened 2026-05-27
dispatch-resolve-worktree: detect live-session conflict in explicit mode


natb1:
## Context

`/dispatch` accepts an optional issue-or-PR-number argument; with an
argument, the dispatcher resolves to that target and enters its worktree.
In `explicit` mode, `dispatch-resolve-worktree` emits `enter <path>` for
any existing `<issue>-*` worktree without checking whether another Claude
session is live in it.

Current code at `.claude/skills/dispatch/scripts/dispatch-resolve-worktree:42-46`:

```bash
if [[ "$MODE" == "explicit" ]]; then
  echo "enter $WT_PATH"
else
  echo "conflict $WT_PATH"
fi
```

After the greenfield router/worker split (#851) the worst-case outcome
of `enter <path>` against a live-owned worktree is no longer a race —
`dispatch-spawn-worker` keys per-worktree dedup on
`claude_sessions_under "$WORKTREE_PATH"` and catches a second spawn as
`deduped`. This is therefore a **diagnostic-accuracy fix**, not a
correctness fix: the upstream path reports the wrong outcome (`enter`
for a live worktree that should be `conflict`), the user-visible status
message is misleading, and the recycle-after-completion path is
mis-selected for a worktree whose previous session is still live. The
spawn-side belt remains; this issue tightens the pre-spawn path to
match the spawn-time invariant.

The intent of `explicit` is the recycle-after-completion case — re-use
a worktree whose previous session is done — which is correct when no
live session is present and emits the wrong outcome when one is.

The closing comment on #741 (2026-05-28) recorded the same framing
explicitly: "Sub-issues #740 (sweep liveness via `claude agents --json`)
and #837 (explicit-mode resolve-worktree liveness) remain open; they
would tighten the pre-spawn paths to match the spawn-time invariant —
defense-in-depth on diagnostics, no longer a correctness fix."

## Relation to #741 and PR #870

#741 (CLOSED 2026-05-28 as superseded by #851) framed the equivalent
queue-mode tightening: distinguish a live-session worktree from a
sessionless orphan. The greenfield router/worker split made the
queue-mode collision case moot at resolve time —
`dispatch-select-target` now filters via `other_live_sessions_under`
upstream (PR #870), so a queue-selected target whose worktree another
session owns never reaches `dispatch-resolve-worktree`. The queue-mode
`conflict <path>` outcome that remains in `dispatch-resolve-worktree`
[...]

PR #916, opened 2026-05-29
dispatch-resolve-worktree: reconcile a reused worktree's branch against the PR head branch

natb1:
## Summary

`dispatch-resolve-worktree` matched an existing worktree to a target purely by the issue-number prefix of its branch, while `dispatch-find-pr` resolves the target's PR through GitHub closing-linkage. When two `<issue>-*` branches exist for one issue, the worker could land on a stale branch with none of the PR's commits (observed 2026-05-29, issue #896 / PR #898: a pre-created title-slug worktree was reused for a `qa` tick, leaving `git diff origin/main...HEAD` empty).

This adds branch reconciliation to the resolver's `enter` path (explicit mode, and queue-mode orphan recycle):

- no PR for the issue (implement phase) → `enter` unchanged
- worktree already on the PR head branch → `enter` unchanged (no redundant checkout)
- worktree on a different branch with no unique commits → re-pointed losslessly to the PR head branch, then `enter`
- worktree on a different branch carrying unique commits → `conflict` (no silent discard)

The resolver's output vocabulary (`enter`/`create`/`conflict`) is unchanged, so the router and worker need no parsing changes.

## Changes

- `dispatch-resolve-worktree`: new `emit_enter_reconciled` helper on both `enter`-bound paths; restructured the match block so the queue live-session check still short-circuits to `conflict` first.
- `test-dispatch-scripts.sh`: gh stub for `pr view --json headRefName`; git stub for `-C fetch`/`rev-list --count`/`checkout`; five reconciliation tests (re-point, no-op, unique-commits conflict, no-PR unchanged, queue-mode orphan re-point).
- `SKILL.md`: documented reconciliation on the Step 5 `enter`/`conflict` bullets.

## Test plan

- [x] `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` → 627/627 passed

### DIFF
--- a/.claude/skills/dispatch-propagate/SKILL.md
+++ b/.claude/skills/dispatch-propagate/SKILL.md
@@ -381,8 +381,12 @@ worktree path that Step 6 will pass to `dispatch-spawn-worker`.
 
 - **`enter <path>`** → re-use an existing `<issue>-*` worktree
   (recycle-after-completion for an explicit argument, or recycle of an orphan
-  worktree — on disk, no live session — for a queue selection, #905). Set
-  `WORKTREE_PATH=<path>` for Step 6. Re-sync issue context from the worktree
+  worktree — on disk, no live session — for a queue selection, #905). A reused
+  worktree whose checked-out branch differs from the target PR's head branch
+  (resolved by `dispatch-find-pr`) is re-pointed to the PR head branch by the
+  resolver before it emits `enter` — lossless, since the case where the existing
+  branch carries commits not on the PR head yields `conflict` instead (#913).
+  Set `WORKTREE_PATH=<path>` for Step 6. Re-sync issue context from the worktree
   (`dangerouslyDisableSandbox: true` — `sync-issue-context` calls `gh`):
   ```bash
   (cd <path> && .claude/skills/dispatch-propagate/scripts/sync-issue-context <N>)
@@ -423,11 +427,14 @@ worktree path that Step 6 will pass to `dispatch-spawn-worker`.
      (cd "$WORKTREE_PATH" && .claude/skills/dispatch-propagate/scripts/sync-issue-context <N>)
      ```
 
-- **`conflict <path>`** → a queue-selected target already has a worktree, so
-  another session owns it. (`dispatch-select-target` resolves the `help wanted`
-  tier to a leaf with no worktree, so for a queue selection this arises only from
-  a race — another session created the worktree between selection and worktree
-  resolution.) Release the lock (see *Releasing the lock*), then proceed to
+- **`conflict <path>`** → the worktree at `<path>` cannot be safely entered.
+  Either a queue-selected target already has a worktree another live session
+  owns (`dispatch-select-target` resolves the `help wanted` tier to a leaf with
+  no worktree, so for a queue selection this arises only from a race — another
+  session created the worktree between selection and worktree resolution), or
+  the reused `<issue>-*` worktree's checked-out branch carries commits not on
+  the target PR's head branch, so re-pointing it would discard work (#913).
+  Release the lock (see *Releasing the lock*), then proceed to
   Step 7 with `notify worktree-conflict` — the user-visible report is mandatory
   there ("worktree at `<path>` owned by another live session for issue `<N>`;
   closing — the next baton-pass or office-hours hand-off will re-seed"), not
--- a/.claude/skills/dispatch-propagate/SKILL.md
+++ b/.claude/skills/dispatch-propagate/SKILL.md
@@ -439,9 +439,15 @@ worktree path that Step 6 will pass to `dispatch-spawn-worker`.
   the target PR's head branch, so re-pointing it would discard work (#913).
   Release the lock (see *Releasing the lock*), then proceed to
   Step 7 with `notify worktree-conflict` — the user-visible report is mandatory
-  there ("worktree at `<path>` owned by another live session for issue `<N>`;
-  closing — the next baton-pass or office-hours hand-off will re-seed"), not
-  optional. Do not spawn a worker.
+  there. The message depends on which conflict case fired:
+  - Live-session race: "worktree at `<path>` owned by another live session for
+    issue `<N>`; closing — the next baton-pass or office-hours hand-off will
+    re-seed"
+  - Unique-commits branch mismatch: "worktree at `<path>` for issue `<N>` is
+    on a branch with commits not on the PR head branch; manual inspection needed
+    before re-entry"
+
+  Not optional. Do not spawn a worker.
 
 As the **final action of this step on every non-`conflict` (proceed) path** —
 before Step 6 — run `dispatch-finalize-selection "$WORKTREE_PATH"`. The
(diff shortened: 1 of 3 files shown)


natb1:
## Dispatch QA — #913 branch reconciliation in `dispatch-resolve-worktree`

Non-browser path: the change is dispatch shell tooling
(`dispatch-resolve-worktree`, its SKILL.md docs, and `test-dispatch-scripts.sh`).
User-acceptance was verified against the behavior the acceptance criteria
describe, each backed by a dedicated regression test.

### Items walked

| # | Scenario | Result |
|---|----------|--------|
| 1 | Wrong branch + PR + no unique commits → re-point to PR head, emit `enter` | PASS |
| 2 | Worktree already on PR head branch → `enter`, no redundant checkout | PASS |
| 3 | Wrong branch carrying unique commits → `conflict`, no checkout | PASS |
| 4 | No PR (implement phase) → `enter` unchanged, no `gh pr view`, no checkout | PASS |
| 5 | Queue-orphan + wrong branch + PR → reconciliation applies in queue mode | PASS |
| 6 | Full `test-dispatch-scripts.sh` suite (660 tests) → no regressions | PASS |

**Counts:** 6 PASS / 0 FAIL / 0 SKIP

### Bugs found

None.

All items are machine-verifiable (script behavior under stubbed `gh`/`git`,
asserted via the test suite). The five reconciliation scenarios cover the
acceptance criteria one-to-one; the full suite passing confirms no regression
to the existing enter/conflict/orphan-recycle paths.

natb1:

### Findings that were fixed

All three findings were fixed in the working tree and committed in `9647ddf`:

1. **`dispatch-resolve-worktree:61` — defensive fallback removed.** `pr=$(dispatch-find-pr ...) || pr=""` silently treated a `gh` failure as "no PR", which would re-enter the wrong branch and reproduce the original #913 bug. Replaced with an explicit `if \! pr=$(...); then echo error >&2; exit 1; fi` so a hard `gh` failure aborts with a clear error. (The legitimate no-PR case still exits 0 with empty output, so the implement phase is unaffected.)

2. **`SKILL.md` (Step 5 conflict template) — message now covers both conflict causes.** The mandatory user-visible report previously only described the live-session race. Added a second template for the unique-commits branch-mismatch case so the worker reports the correct diagnosis instead of a phantom live session.

3. **`SKILL.md` (drain `worktree-conflict` description) — both causes documented.** Updated to cover both a live-session owner and a worktree branch carrying commits not on the PR head branch, with a pointer to the Step 5 `conflict` case.

No informational, disregarded, or deferred findings. `test-dispatch-scripts.sh` → 679/679 passed after the fixes.

natb1:
## Fixed
- Added a test documenting that a queue live-session conflict takes precedence over branch reconciliation (wrong-branch + live-session short-circuits to `conflict` before `emit_enter_reconciled` runs — no `gh pr view`, no checkout): ad581ee

## Informational
- Empty `headRefName` from `gh pr view` (e.g. a merged PR whose branch was deleted) is treated as "same branch" → `enter` unchanged. Reasonable fallback: there is nothing to reconcile against, and entering as-is is no worse than pre-fix behavior.
- `git checkout -B "$pr_head" "origin/${pr_head}"` is correct and idempotent: `-B` resets a pre-existing local branch (e.g. from an aborted prior run) rather than erroring, and the `origin/` ref is the fetched ground truth.
- The `enter` path now makes up to two extra `gh` round-trips (`dispatch-find-pr`, then `gh pr view`) per tick even when the worktree is already on the PR head branch. Acceptable cost; caching the PR head branch per-worktree is a possible future optimization, out of scope here.
- The `git fetch origin "$pr_head"` before `rev-list --count` is necessary so the unique-commit count is measured against the fetched tip; adds a third network call but is correct.
- The branch name from `gh pr view` is passed to `checkout -B` without explicit sanitization. Not a realistic risk: GitHub branch names are constrained and `jq -r` extracts the value safely.
- `emit_enter_reconciled` reads the script-global `$ISSUE` rather than taking it as a parameter. Consistent with the surrounding script's use of `$ISSUE`; passing it explicitly would only marginally improve self-containment.

## Dismissed
- The inline one-line comments on the reconciliation branches (`# no PR → unchanged`, etc.) are slightly compressed — they match the surrounding codebase's comment style and read clearly, so no change.

## Deferred
_None._

natb1:
## Security review — PR #916 (#913)

Structured security review of this PR's diff (3 files: `dispatch-propagate/SKILL.md`,
`dispatch-resolve-worktree`, `test-dispatch-scripts.sh`). Fan-out of 9 reviewers —
6 domains, red team, built-in `/security-review`, and CodeQL. The diff is dispatch
shell tooling; no app/Firestore/web surface is touched.

### Fixed

- **Unusable / injectable PR head branch name in `emit_enter_reconciled`** —
  `dispatch-resolve-worktree:66-69`. `pr_head` comes from `gh pr view … headRefName`
  (an external GitHub-API boundary) and was used in `git fetch` / `rev-list` /
  `checkout -B` after only an empty-string check that was *conflated with the
  same-branch no-op* — an empty `headRefName` when a PR exists is a failed lookup,
  but the code silently fell through to `enter`, entering the worktree unreconciled
  and defeating the reconciliation guard this PR exists to provide. The same path
  passed the name to git without a format guard, leaving option/refspec injection
  surface (a `--upload-pack=…` / leading-`-` / `..` name).
  - OWASP A03:2021 Injection · STRIDE Tampering · Confidence medium.
  - Fix (commit `c8e96e5`): once a PR is found, an empty/malformed `headRefName` is a
    clear error (`exit 1`) instead of a silent `enter`; names that are not
    `^[A-Za-z0-9._/-]+$`, begin with `-`, or contain `..` are rejected before any git
    call. Two tests added (`test-dispatch-scripts.sh`): empty-`headRefName` → error +
    no checkout; injection-shaped `headRefName` → error + no checkout. Full suite:
    699/699 pass.

### Informational / not changed

- **`$pr` (PR number) and `$wt_branch` not re-validated before use** —
  flagged by input-validation and red-team reviewers. `$pr` is internal
  (`dispatch-find-pr`, jq-filtered to an integer) and `$wt_branch` comes from local
  `git worktree list`; both are trusted internal sources, and a `..`-bearing branch
  name is not creatable under git's ref-format rules. The external-boundary value
  (`pr_head`) is the one now guarded. No change — guarding trusted internal output
  would be defensive bloat (`.claude/rules/code-style.md`).

### Out of scope

- **Explicit-mode skips the live-session guard; `emit_enter_reconciled`'s
  `git checkout -B` can now mutate a worktree under a live session** —
  `dispatch-resolve-worktree:83-89`. Real interaction, but explicit-mode
  live-session detection is **#837**'s domain (the issue scopes it out as
  orthogonal: session ownership vs. branch identity), and commit `ad581ee` already
  documents that the live-session conflict takes precedence over branch
  reconciliation. Tracked by #837 — not fixed here to keep this PR's scope to branch
  identity.
- **18 pre-existing `npm audit` advisories** (`@xmldom/xmldom`, `uuid`, `ws`, `qs`,
  `protobufjs`, etc.). This PR changes no `package.json`; all findings predate it and
  are unrelated to the diff. Not addressed here.

### Clean

Data-exposure, Firebase-specific, secrets, built-in `/security-review`, and CodeQL
(no open code-scanning alerts on the PR ref) all returned no findings.

PR #900, opened 2026-05-29
dispatch: headroom-driven concurrency budgeting

natb1:
Closes #878

## Summary

Replaces the time-based weekly ramp in `dispatch-target-workers`
(`(tau/(remaining+tau))^k`) with two **headroom-driven** linear ramps
(weekly and 5h) combined by `min`. Each ramp returns `max_workers` while
there is comfortable budget remaining, tapers linearly toward 1 as actual
usage approaches the cap, and returns 0 only when actual usage exceeds
the cap. A floor of 1 at `headroom == 0` prevents the `0 >= 0` spawn
deadlock the old time-based ramp produced for the first ~4 days of every
weekly window.

Smoke at real-world telemetry (`used_weekly=20, used_5h=2`, defaults):
`dispatch-target-workers` now prints `8` (was `0` under the old ramp,
stalling the chain).

## Schema changes

Removed (no longer accepted in `dispatch.config/target-workers.json`):
`ramp_tau_seconds`, `ramp_shape_k`, `five_hour_window_seconds`.

Added: `weekly_taper_window_pct` (default 30), `five_hour_taper_window_pct`
(default 15). Both validate as a number in `(0, 100]`.

**User action required:** users with a customised
`dispatch.config/target-workers.json` (gitignored) must update the field
names or delete the file to fall back to the new defaults; otherwise
`dispatch-config-load` will reject it and `dispatch-target-workers` will
silently use baked-in defaults.

## Test plan
- [x] `bash .claude/skills/dispatch/scripts/test-dispatch-scripts.sh` — 604/604 pass.
- [x] Smoke at `used_weekly=20, used_5h=2` → prints `8`.
- [x] `grep -rn 'ramp_tau_seconds|ramp_shape_k|five_hour_window_seconds' .claude/skills/` → no matches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #925, opened 2026-05-30
dispatch: pace-relative worker controller (weekly cumulative-pace curve)

natb1:
Closes #917

Replaces the flat-cap weekly stage of `dispatch-target-workers` (#900 / #878) with a cumulative weekly-**pace** curve, so the token budget spreads smoothly across the week instead of bursting early and idling. Keeps #900's 5-hour→worker headroom ramp, now fed from a pace-derived 5h target.

## What changed

The controller is now a sequential four-stage pipeline `W → F → N` (not a `min` of two ramps):

1. **Stage 1 — weekly pace curve `W`**: cumulative weekly-% target keyed to the elapsed fraction `x` of the weekly window; reaches `target_weekly_usage_pct` only at week end (`x=1`), never before. Per-window increment floored/hard-capped.
2. **Stage 2 — 5h target `F`**: `0` when usage runs ahead of the curve; otherwise scales `floor5..ceil5` with weekly headroom.
3. **Stage 3 — workers `N`**: scales `1..max_workers` with 5h headroom; `0` at/over target.

The stdout contract (single integer), router gate (`LIVE_COUNT >= TARGET_N`), and missing-telemetry fallback (`print 1`) are unchanged.

## Commits

- `c1227f4` controller + config schema + reseed consumer + tests
- `eefa106` docs: SKILL.md Concurrency budgeting rewrite + README token-budget pacing

## Test plan

- [ ] CI green (`test-dispatch-scripts.sh` — 710/710 locally)
- [ ] Curve reaches `target_weekly` only at `x=1`; `W < target` for all `x < 1`
- [ ] `F=0` ahead of pace; `N=0` at/over target; floor/ceiling clamps hold
- [ ] Early-week smoke (`used_weekly=20, used_5h=2`) yields `target_N >= 1` (no stall)

### DIFF
--- a/.claude/skills/dispatch-propagate/SKILL.md
+++ b/.claude/skills/dispatch-propagate/SKILL.md
@@ -526,80 +526,104 @@ After the spawn (or skip) returns, **proceed to Step 7** (terminal disposition).
 
 ### Concurrency budgeting
 
-`dispatch-target-workers` composes two **headroom-driven** linear ramps —
-one per axis (weekly and 5h) — and combines them with `min` to decide how
-many concurrent `dispatch-worker-*` sessions should run:
+`dispatch-target-workers` runs a sequential four-stage pace-relative pipeline
+(W → F → N) to decide how many concurrent `dispatch-worker-*` sessions should
+run. Instead of a flat weekly cap, the weekly budget follows a
+**cumulative-pace curve** keyed to how far through the weekly rate-limit window
+we are (in 5-hour-window terms), so token spend spreads smoothly across the
+week rather than bursting early and idling. The controller is intentionally
+more conservative early-week than a flat-cap design — it throttles whenever
+actual usage runs ahead of the curve, even when the weekly total is far below
+the cap.
 
 ```
-ramp(headroom_pp, taper_window_pp):
-  if headroom_pp <  0:                return 0
-  if headroom_pp >= taper_window_pp:  return max_workers
-  return max(1, round(max_workers * headroom_pp / taper_window_pp))
+WEEK_SECONDS = 604800
 
-target_N = min(
-  ramp(target_weekly - used_weekly, weekly_taper_window_pct),
-  ramp(target_5h     - used_5h,     five_hour_taper_window_pct),
-)
+# Stage 1 — weekly pace curve W (weekly %)
+remaining = resets_at_weekly - now;   if remaining <= 0: print 0; exit
+x   = clamp((WEEK_SECONDS - remaining) / WEEK_SECONDS, 0, 1)  # elapsed fraction
+T   = round(WEEK_SECONDS / 18000)                             # = 34 five-hour windows
+end = floor + (p+1)*(target_weekly/T - floor)                # solve so W(1)=target_weekly
+end = min(end, cap)                                          # per-window hard ceiling
+W   = T*( floor*x + (end-floor)*x^(p+1)/(p+1) )              # cumulative target now
+
+# Stage 2 — 5h target F (% of 5h limit)
+hw = W - used_weekly
+F  = (hw <= 0) ? 0 : floor5 + (ceil5-floor5)*clamp(hw/Hw, 0, 1)
+
+# Stage 3 — workers N
+h5 = F - used_5h
+N  = (h5 <= 0) ? 0 : clamp(round(max_workers * h5/H5), 1, max_workers)
+print N
 ```
 
-Each ramp returns `max_workers` while headroom comfortably exceeds the
-taper window, tapers linearly toward `1` as headroom approaches zero, floors
-at `1` when headroom is exactly zero, and returns `0` only when actual usage
-has crossed the configured target. The floor at `headroom == 0` prevents the
-`0 >= 0` deadlock the previous time-based ramp produced against Step 6's
-`LIVE_COUNT >= TARGET_N` gate — `target_N=0` now requires `used > target`.
+- The weekly increment `d(x) = floor + (end-floor)*x^p` rises from the floor
+  toward (but clamped at) the cap; `W` reaches `target_weekly` only at `x=1`
+  (week end), never before.
+- `F=0` when ahead of pace (`used_weekly >= W`); the `floor5..ceil5` band opens
+  with weekly headroom.
+- `N=0` only at/over the 5h target F (includes the `F=0` ahead-of-pace pause);
+  `1..max_workers` below. No `0 >= 0` router deadlock during healthy
+  under-budget operation.
 
-- **Weekly axis** (defaults: `target_weekly_usage_pct=90`,
-  `weekly_taper_window_pct=30`, `max_workers=8`):
+**Table A — weekly curve vs. elapsed** (defaults; `used_weekly=0`, `used_5h=0`):
 
-  | used_weekly (%) | headroom (pp) | target_N |
-  |---:|---:|---:|
-  | 60 | 30 | 8 |
-  | 70 | 20 | 5 |
-  | 80 | 10 | 3 |
-  | 85 |  5 | 1 |
-  | 89 |  1 | 1 |
-  | 90 |  0 | 1 |
-  | 91 | -1 | 0 |
+| elapsed `x` | W (weekly %) | target_N |
+|---:|---:|---:|
+| 0.25 | 12 | 8 |
+| 0.50 | 31 | 8 |
+| 0.75 | 57 | 8 |
+| 0.90 | 76 | 8 |
+| ~1.0 | 90 | 8 |
 
-- **5-hour axis** (defaults: `target_five_hour_usage_pct=50`,
-  `five_hour_taper_window_pct=15`, `max_workers=8`):
+At these defaults, `W >= Hw=20` for `x >= 0.5`, so `F=ceil5=80` and
+`h5=80 > H5=15`, giving `N=max_workers=8`. At `x=0.25`, `W=12 < Hw=20`,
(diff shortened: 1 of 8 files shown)


natb1:
## Dispatch QA — user-acceptance pass (clean)

**Issue #917** — pace-relative worker controller. Pure shell-script + docs PR, so this was the **non-browser path**: each item was machine-verified by invoking `dispatch-target-workers` / `dispatch-config-load` with telemetry and config env overrides and checking the operator-visible output (single-integer stdout, exit codes, stderr notes). No item required human judgment.

### Items walked

| # | Item | Result |
|---|---|---|
| 1 | Early-week under-pace does not stall the chain (`used_weekly=20, used_5h=2`, ~3d left) | **PASS** → prints `8` (≥1, the AC smoke) |
| 2 | Ahead-of-pace pauses spawning (early-week, `used_weekly` ≫ `W(now)`) | **PASS** → prints `0` |
| 3 | Missing-telemetry fallback | **PASS** → prints `1` + stderr note |
| 4 | Weekly window expired (`resets_at_weekly` in the past) | **PASS** → prints `0` |
| 5 | Config schema (new keys validate; `>100` rejected; dropped/renamed keys gone) | **PASS** |

**Counts:** PASS 5 · FAIL 0 · SKIP 0. **No bugs found.**

### Informational (not a PR defect)

The live on-disk `dispatch.config/target-workers.json` still carries the **pre-#917 schema** (`ramp_tau_seconds`, `ramp_shape_k`, `target_five_hour_usage_pct`). It validates fine — `dispatch-config-load` only type/range-checks known keys and leaves extras inert — but those keys are now ignored, so every new tunable falls back to its baked-in default. Worth refreshing that local config from `target-workers.example.json` to actually exercise the new curve tunables.

natb1:

Five findings surfaced; classification below. Fixes applied in `87f5575`.

### Findings that were fixed (`87f5575`)

- **`dispatch-config-load` missing `weekly_increment_floor_pct <= weekly_increment_cap_pct` cross-check.** Fields validated individually but not pairwise, so `floor > cap` validated field-by-field yet spiked `W` to hundreds of percent (budget never enforced). Added a cross-field check (fires only when both are explicitly present and numeric; absent fields fall back to script defaults the validator does not know).
- **`dispatch-config-load` missing `five_hour_target_floor_pct <= five_hour_target_ceiling_pct` cross-check.** `floor5 > ceil5` inverted the F band (max 5h budget at at-pace, min when budget is healthy). Added the same style of cross-field check. Tests `13c`/`13d`/`13e` cover both cross-checks (716/716 pass).
- **Docs overstated "W reaches `target_weekly` only at `x=1`".** False when `weekly_increment_cap_pct` clamps the per-window increment — the cap deliberately holds the terminal `W(1)` below `target_weekly`. Added the caveat to `reference.md` (Concurrency budgeting), its `target_weekly_usage_pct` tunable row, and the `dispatch-target-workers` header comment.

### Informational findings

- **Docs say the increment "rises from the floor toward the cap"; it instead *falls* when `end < floor`** (reachable only for `target_weekly < ~17`). This is the same low-target regime as the deferred finding below (#930): once the validator rejects floor-incompatible weekly targets, this claim becomes accurate again, so the doc was left unchanged pending that fix.

### Deferred findings

- **`dispatch-target-workers` weekly curve is non-monotone for `target_weekly < ~17`.** `end` is clamped above at `wcap` but never below, so for very low weekly targets `end` goes negative and `W` peaks above `target_weekly` mid-week — spawning workers while already over the weekly budget. The default (90) is safe. A sound fix needs a cross-field constraint coupling the validator to the algorithm's window constant `T`/`p` (or a floor↔target redesign) — a design decision out of scope here. Tracked in **#930**, blocked by #917.

natb1:
## Fixed
- Document the inverted-curve config footgun (`weekly_increment_floor_pct` too high vs. `target_weekly_usage_pct`, and floor > cap) in `reference.md`'s recalibration notes: f1fd924

## Informational
- `remaining <= 0` Stage-1 sentinel (`print 0`) is semantically distinct from the missing-anchor fallback (`print 1`); both paths are correct and separated. A clarifying one-line comment was considered but left out per the repo's default-no-comment style.
- `five_hour_target_floor_pct` is reused as the `dispatch-schedule-reseed` cap-hit threshold. The rename propagated correctly; the slight conservatism (reseed triggers at the F-band floor of 50, even when `F` may be 80) is pre-existing behavior carried over, not a regression.
- `WEEKLY_RESETS` float precision: the `^[0-9]+(\.[0-9]+)?$` guard admits a fractional `resets_at`, which awk would coerce with minor precision loss. Real-world timestamps are integers, so no practical risk.
- `NOW` is captured before the file read; a failed `date +%s` yields empty and fails closed to the `print 1` fallback — correct.
- `reference.md` Table A `~1.0` row: at `x≈1.0`, `W≈89.997` not exactly 90; the `~` notation is honest and `target_N=8` is correct.

## Dismissed
- Division-by-zero on `five_hour_headroom_taper_pct=0` — cannot occur: the config validator constrains it to `(0, 100]` and the baked-in default is 15.
- `floor5 == ceil5` degenerate F band — valid (if unusual) config; the linear band collapses to a constant and behaves correctly.

## Deferred
_None._

natb1:
## Security review — `dispatch-target-workers` pace-relative controller (#917)

Ran the full structured security pass: merged `origin/main`, then fanned out 9
parallel reviewers (6 domains + red team + built-in `/security-review` +
CodeQL). The diff is shell + awk + markdown for the dispatch concurrency
budgeter — it reads local telemetry/config JSON and prints an integer worker
count.

**Outcome: no required fixes.** All external inputs are numeric-regex sanitized
before reaching `awk`, every value is passed via `awk -v` (no injection into the
awk program text), and config is schema-validated by `dispatch-config-load`. No
`eval`/`system()`/shell-interpolated-command path exists. CodeQL: no open
alerts. Built-in `/security-review`, secrets, and Firebase domains: clean.

Every finding below requires either (a) same-user local write access to the
config/telemetry/env that the dispatch runner already owns — outside the threat
model for single-user internal tooling — or (b) an operator entering a config
value the docs already flag as unsafe. None are reachable across a privilege or
user boundary.

### Findings and dispositions

| # | Finding | Confidence | Disposition |
|---|---|---|---|
| 1 | Cross-field validator only fires when **both** `weekly_increment_floor_pct` and `weekly_increment_cap_pct` are present; a floor raised above the baked-in cap default (10) passes validation, drives the solved `end` negative, and inflates the W-curve mid-week → continuous max-worker spawn (`dispatch-config-load:300-313`, `dispatch-target-workers:291-295`) | high | **out-of-scope** |
| 2 | `max_concurrent_workers` / `weekly_curve_power` unbounded above; an absurd value (`1e300`) makes `jq -r` emit `1E+300`, which the router's `(( LIVE_COUNT >= TARGET_N ))` can't parse → gate bypass (`dispatch-config-load:264`) | medium | **out-of-scope** |
| 3 | Very large `weekly_curve_power` underflows `x^(p+1)` to 0 → curve collapses to floor (under-spawn) (`dispatch-target-workers:295`) | low | **out-of-scope** |
| 4 | Tampered `resets_at_weekly` / `NOW` / `five_hour_headroom_taper_pct` via the local telemetry file or env override can stall dispatch (`N=0`) or hit an awk divide-by-zero (clamped) (`dispatch-target-workers:154,283-284,310`) | medium–low | **out-of-scope** |
| 5 | awk `^` claimed to be XOR on non-gawk → wrong curve | medium (reported) | **false-positive** — `^` is exponentiation in all POSIX awk (gawk/mawk/nawk); verified `awk 'BEGIN{print 2^10}'` → 1024 |
| 6 | Rejection-path stderr echoes raw external telemetry/`NOW` values (log-injection / control chars) (`dispatch-target-workers:207,217,221`) | low | **out-of-scope** |
| 7 | Validation errors include the absolute config-file path and already-type-checked numeric values (`dispatch-config-load:305,312,321`) | low | **false-positive** — intentional operator diagnostics; numbers are type-validated, no injection |
| 8 | Test helper `write_rl()` builds fixture JSON by string interpolation (`test-dispatch-scripts.sh`) | low | **out-of-scope** — test scaffolding, no untrusted caller |
| 9 | `npm audit`: 40 pre-existing vulns (5 high, incl. 2 fixable via `npm audit fix`) | high | **out-of-scope** — pre-existing; this diff changes no dependency manifest |

### Why finding 1 is out-of-scope, not required

This is the only over-spend (cost-amplification) path, so it got the closest
look. It is a **deliberate, documented** design decision, not an oversight:

- `dispatch-config-load:294-299` comments explain the both-present-only check
  and why (absent fields fall back to baked-in defaults the validator can't see).
- `reference.md:258-269` (added by 87f5575 / f1fd924) documents the
  `floor <= target/T` and `cap >= floor` constraints and the validator's
  both-present limitation, with operator recalibration guidance.
- It requires the operator to misconfigure their **own** local config file
  with a floor above the documented safe range — within the trust boundary.

Tightening this (full single-field validation, or an awk lower-bound guard on
`end`) is a robustness change that would reverse a recently-litigated design
decision — it belongs in a follow-up issue if desired, not in the security
phase. One doc-accuracy nit worth a future tidy: reference.md's "still clamps"
wording understates the actual mid-week over-spawn magnitude.

Findings 2–4 are the same shape: real only under a local, same-user
misconfiguration or env/file tamper that is outside this tooling's threat model,
and `max_concurrent_workers` / `weekly_curve_power` being unbounded-above is
documented as intentional. Finding 9's two fixable highs (`fast-xml-parser` /
`fast-xml-builder`) are worth a separate `npm audit fix`, unrelated to this PR.

No code changes were made in this phase.

PR #922, opened 2026-05-29
dispatch-propagate: move explanatory prose to non-loaded reference.md

natb1:
Closes #918

Relocates dispatch-propagate's explanatory "why" prose out of the auto-loaded `SKILL.md` into a new sibling `reference.md` that the skill loader does **not** load. `SKILL.md` is read into the model's context on every tick of the self-perpetuating dispatch chain; moving reference-only prose out cuts that per-tick token cost.

## What moved to reference.md
- Lock semantics & reclaim deep dive
- Per-worktree invariant
- Concurrency budgeting (ramps, tunables, missing-telemetry fallback)
- Selection-ladder mechanics
- The #725 cap-keyed re-seed detail
- Scattered rationale (chain-mechanics intro, Step-5 marker deep dive, Step-6 spawn-cwd trade-off, per-command sandbox justifications)

## What stayed in SKILL.md
Every step header (Step 0–7), every script/command invocation with its `dangerouslyDisableSandbox`/`timeout` directive, every route, every terminal disposition (`propagate` / `notify` / `drain`) and disposition table, every prose-report template, and the office-hours idiom. The `### The #725 cap-keyed re-seed` heading survives (external ref from `dispatch-worker/SKILL.md`) with a stub forwarding to reference.md.

## Result
- `SKILL.md`: 685 → **411 lines** (40.0% drop, meets the ≥40% gate).
- No behavior change: all behavior lives in the scripts plus retained imperative directives; only reference prose moved, with pointers left behind.
- Merged `origin/main` (#900 headroom-driven budgeting) — the new budgeting tables now live in reference.md's *Concurrency budgeting* section.

Part A of a two-issue split; ships first because it is zero-risk and conflict-free against the in-flight branches editing this file.

PR #929, opened 2026-05-30
dispatch: fix pr-row worktree resolution keying on PR number instead of issue number

natb1:
Closes #926

For a queue `pr <num> <branch> <phase>` row, the only number is the PR number, but `/dispatch-propagate` Step 5 passed it to the issue-keyed `dispatch-resolve-worktree`. Since the REST issues endpoint silently serves PRs too, the create path slugged a stray `<pr-num>-*` branch from `origin/main`, stranding the PR's real `<issue>-*` worktree and mis-deriving the phase as `implement`.

## Changes (all under `.claude/skills/dispatch-propagate/`)

- **SKILL.md** — router keys Step 5/6 on the issue the PR closes (`N=${branch%%-*}`, mirroring `dispatch-select-target`'s `${pr_branch%%-*}`), never the PR number. Step 5 intro documents that `<N>` is always an issue number.
- **scripts/dispatch-resolve-worktree** — fail-loud guard rejecting an argument GitHub resolves to a PR, via the `has("pull_request")` discriminator already used by `dispatch-resolve-arg`. A gh failure (404/network) falls through; only a definitive `pull_request` key trips it.
- **scripts/test-dispatch-scripts.sh** — regression test 19(a)/(b) covering the #922 scenario where PR number (922) and closing-issue number (918) differ.

## Test plan

- [ ] `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — 704/704 pass, including new 19(a)/(b)

### DIFF
--- a/.claude/skills/dispatch-propagate/SKILL.md
+++ b/.claude/skills/dispatch-propagate/SKILL.md
@@ -152,9 +152,13 @@ but do not stop; continue to target selection.
 
   Route on `$SELECTED`:
 
-  - `pr <num> <branch> <phase>` — a PR to work on; `<phase>` is pre-derived by
-    the selection scan, so Step 6 reuses it instead of re-deriving. Proceed to
-    Step 5 with mode `queue`.
+  - `pr <num> <branch> <phase>` — a PR to work on; `<num>` is the **PR** number
+    and `<phase>` is pre-derived by the selection scan, so Step 6 reuses it
+    instead of re-deriving. The worktree-resolution key is the **issue** the PR
+    closes, not the PR number: set `N=${branch%%-*}` (the branch's `<issue>-`
+    prefix, mirroring `dispatch-select-target`'s own `${pr_branch%%-*}`) and use
+    that `N` — never the PR `<num>` — for Steps 5 and 6. Proceed to Step 5 with
+    mode `queue`.
   - `issue <num>` — a `help wanted` issue to implement, pre-resolved by the
     selection scan to a startable open leaf (not necessarily the top-level
     `help wanted` issue). Proceed to Step 5 with mode `queue`.
@@ -243,6 +247,11 @@ an explicit `/dispatch-propagate` argument, otherwise `queue`:
 ```bash
 .claude/skills/dispatch-propagate/scripts/dispatch-resolve-worktree <N> <explicit|queue>
 ```
+
+Here `<N>` is always the **issue** number, never a PR number. For a `pr` row it
+is the branch-prefix `N` set in Step 3, not the row's PR `<num>`;
+`dispatch-resolve-worktree` rejects a PR number with a clear error rather than
+fabricate a stray `<pr-num>-*` worktree from `origin/main` (#926).
 
 It prints exactly one decision line — act on it. Each branch resolves a
 worktree path that Step 6 passes to `dispatch-spawn-worker`.
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-resolve-worktree
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-resolve-worktree
@@ -37,6 +37,12 @@
 # accepted form ^[0-9]+-[a-z0-9]+(-[a-z0-9]+)*$, full name <= 32 characters.
 #
 # A bad argument or a failed `gh issue view` is a clear error, not a fallback.
+#
+# This script is issue-keyed. A PR number passed as `<issue-num>` is rejected up
+# front (#926): the REST issues endpoint serves PRs too, and a PR's JSON carries
+# a "pull_request" key (the same discriminator dispatch-resolve-arg uses). Caller
+# must pass the PR's closing-issue number, not the PR number — otherwise the
+# create path would fabricate a stray <pr-num>-* worktree from origin/main.
 set -euo pipefail
 
 ISSUE="${1:-}"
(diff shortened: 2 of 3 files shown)


natb1:
## Dispatch QA — #926 (pr-row worktree resolution keying fix)

Non-browser, dispatch-script change — QA verified operator-visible resolver
behavior. No app/browser surface, so the browser walkthrough and public-seed
data policy do not apply.

**Items walked:** 3 — **3 PASS / 0 FAIL / 0 SKIP**

| # | Item | Result |
|---|------|--------|
| 1 | Regression suite `test-dispatch-scripts.sh` (704/704, incl. the two new #926 tests) | PASS |
| 2 | Guard rejects a real live PR number (#922) — the exact bug scenario | PASS |
| 3 | Guard does not false-positive on a real issue (#918) | PASS |

**Detail**

- **#1** — `704/704 passed, 0 failed`. The new regression tests for the
  `pr`-row path where PR# ≠ issue# pass: PR number as issue key → exit 1 + no
  decision line; PR's closing-issue number → `enter` the real `<issue>-*`
  worktree.
- **#2** — `dispatch-resolve-worktree 922 queue` against the live repo exits
  `1`, prints no decision line, and emits
  `error: #922 is a PR, not an issue; pass the PR's closing-issue number…`.
  Before this fix the create path would have fabricated
  `create 922-dispatch-propagate-move-expl` from `origin/main`.
- **#3** — the guard's discriminator `jq -e 'has("pull_request")'` exits
  non-zero for issue #918 (passes through) and zero for PR #922 (rejects),
  exactly as the script applies it.

All four acceptance criteria from the issue are satisfied: router keys on the
closing-issue number (SKILL.md), resolver rejects a PR number with a clear
non-zero error (guard), a PR whose head branch differs from `<pr-num>-*`
resolves to the real `<issue>-*` worktree (test 19b), and a regression test
covers the PR#≠issue# path (tests 19a/19b).

natb1:

### Findings that were fixed

- **Missing regression coverage for the pre-scan guard's stale-worktree rationale** (`scripts/test-dispatch-scripts.sh`) — no test covered the case where a stale `<pr-num>-*` worktree already exists on disk when a PR number is passed as the issue key. The guard must fire *before* the worktree scan so the stale worktree is never entered; without a test, a regression that moved the guard to only the create path would silently route a worker into the wrong worktree. Added test 19c, which seeds a `922-stray` worktree plus an `arg-issue-922.json` carrying a `pull_request` key and asserts `dispatch-resolve-worktree 922 queue` exits 1 with no decision line. Fixed in `bdb6229`.

### Disregarded findings

- **Guard adds one `gh api` round-trip per invocation, including the enter path** — not an error. The pre-scan placement is correctness-critical (it prevents entering a stale `<pr-num>-*` worktree), so the extra call is justified and must not be moved after the scan.

natb1:
## Fixed
_None._

## Informational
- Branch-naming assumption `N=${branch%%-*}` relies on the `<issue>-<slug>` convention; a non-conforming branch would fail loud at the guard's arg-validation. Pre-existing convention constraint, not new to this PR.
- The new guard adds one `gh api .../issues/$ARG` round-trip per `dispatch-resolve-worktree` call (including issue rows). Negligible latency for an async dispatch system; noted for awareness.
- `jq -e 'has("pull_request")'` exit-code usage is correct — fires only when the `pull_request` key is present, so the guard trips only on a definitive PR answer.
- No test exercises a transient `gh api` failure (e.g. 503/network) at the guard step. The fall-through-on-gh-failure behavior is the stated intent and already correct; this is a minor coverage gap, not a blocker.

## Dismissed
_None._

## Deferred
_None._

natb1:
## Security review — #926 PR-number-as-issue-key guard

Reviewed the diff against `origin/main` merge-base `b7a9350` (3 files:
`dispatch-propagate/SKILL.md`, `dispatch-resolve-worktree`,
`test-dispatch-scripts.sh`). Fanned out 9 parallel reviewers — 6 domains, red
team, the built-in `/security-review` scan, and the PR's CodeQL alerts.

**Outcome: no required fixes.** CodeQL: 0 open alerts. Built-in scan: clean.
Input-validation, data-exposure, secrets, and Firebase domains: clean. The diff
is dispatch-orchestration bash + docs — no Firebase/Firestore surface, no
runtime npm dependencies.

### Findings and dispositions

**1. PR-number guard fails open on a `gh` error — not fixed (out of scope)**
- Location: `dispatch-resolve-worktree:59`
- OWASP A05:2021 / STRIDE Tampering. Flagged by the auth (medium) and red-team
  (high) reviewers.
- The new guard `gh api .../issues/$ISSUE 2>/dev/null | jq -e 'has("pull_request")'`
  suppresses `gh` failures and falls through to the create path, which the
  reviewers argued could re-fabricate a stray `<pr-num>-*` worktree during a
  GitHub API degradation window.
- **Why not fixed:** the scenario requires `gh api .../issues/<pr-num>` to fail
  while the create path's `gh issue view <pr-num>` (line 120) *succeeds* — but
  both target the same GitHub resource, so a transient failure is not
  endpoint-specific, and `gh issue view` already fails closed (`exit 1`) on a
  genuine outage. The upstream SKILL.md fix also removes the trigger at the
  source: the router now passes the branch-derived **issue** number, not the PR
  number. Converting the guard to fail-closed-on-404 would break every existing
  create-path test (they rely on 404 fall-through with no `arg-issue` fixture)
  and would only halt the chain during a real outage where the downstream
  already halts. The fail-open is documented as a deliberate tradeoff in the
  code comment (lines 56–58). Noted, not blocking.

**2. Branch-prefix issue-number extraction `N=${branch%%-*}` — false positive**
- Location: `SKILL.md` (router Step 5 instruction)
- OWASP A01:2021 / STRIDE Tampering, low confidence (auth reviewer).
- A non-numeric branch prefix would flow into `dispatch-resolve-worktree`, but
  that script validates `^[1-9][0-9]*$` (line 51) and exits with a clear error
  before any worktree action. Adequately defended; no change.

**3. `headRefName` regex admits git pseudorefs (`HEAD`, `refs/*`) — out of scope**
- Location: `dispatch-resolve-worktree:85`
- OWASP A03:2021 / STRIDE Tampering, medium confidence (red-team).
- This validation is pre-existing code from #913; it is not in this diff. Worth
  tracking separately, but outside this PR's scope.

**4. Missing test for the guard's failure-mode fall-through — out of scope**
- Location: `test-dispatch-scripts.sh` (tests 19a–c)
- Tied to finding 1: the proposed test would assert fail-closed behavior that
  this PR intentionally does not implement, so it is not added here.

### Dependency audit (informational)

`npm audit` reports 40 pre-existing advisories (5 high, 33 moderate, 2 low) —
none introduced by this diff, which adds no npm dependencies. Distinct vulnerable
roots: `@xmldom/xmldom`, `fast-xml-builder`, `protobufjs`, `qs`, `ws`. Out of
scope for this PR.

PR #811, opened 2026-05-26
Flaky CI: cache + retry for npm ci (#807)

natb1:
## Summary

Mitigates the transient `npm error code ECONNRESET` failure that broke PR #792's `acceptance` check by stacking two independent layers around every `npm ci` in the PR-check workflows:

- **Caching.** `cache: 'npm'` added to every `actions/setup-node@…` step in `pr-checks.yml` (`acceptance`, `preview-and-smoke`) and `unit-tests.yml` (`unit-tests`, `lint`, `typecheck`, `playwright-version-sync`). With the cache hot and `package-lock.json` unchanged, `npm ci` restores from `~/.npm/_cacache` and never hits `registry.npmjs.org`.
- **Retry.** `ensure_deps()` in `.claude/skills/dispatch/scripts/lib.sh` now sets per-invocation `npm_config_fetch_retries=5` (plus longer fetch timeouts) and wraps `npm ci` in a 3-attempt shell loop with 5s/15s backoff. Final-attempt failure still returns non-zero, preserving the `set -euo pipefail` contract for callers.

Together: cache reduces how often the install path needs the registry; retry backstops the residual transient failures that slip through (e.g. cold cache after a lockfile change).

The `hook-tests` job is intentionally untouched — it does not call `setup-node` or `npm ci`. Deploy workflows (firestore/storage/functions/prod) have the same exposure but a different blast radius and are deferred per the plan; scope here is PR checks only.

Closes #807.

## Test plan

- [ ] `acceptance`, `unit-tests`, `lint`, `typecheck`, `preview-and-smoke`, `playwright-version-sync` all pass on this PR (cold cache — first run).
- [ ] After a no-op push, the same checks pass with materially faster `npm ci` step durations (warm cache).
- [ ] `hook-tests` exercises `ensure_deps`'s new retry branch via 2 new tests in `test-dispatch-scripts.sh` (322/322 pass locally).

PR #947, opened 2026-05-30
Extend npm ci retry to bare workflow call sites

natb1:
Closes #815

PR #811 wrapped `ensure_deps()` with a 3-attempt retry and longer fetch
timeouts to recover from transient `ECONNRESET` during `npm ci`, but that
retry only covers code paths through `ensure_deps`. Two CI steps call
`npm ci` directly and inherit none of it:

- `pr-checks.yml` — the `preview-and-smoke` job's Cloud Functions deploy step.
- `unit-tests.yml` — the `playwright-version-sync` install step.

This adds a thin standalone wrapper
`.claude/skills/dispatch-propagate/scripts/npm-ci-with-retry.sh` that exports
the same fetch-retry envs and runs `npm ci` with the same 3-attempt 5s/15s
backoff, and wires both sites to it. The wrapper installs unconditionally in
the current directory (no `node_modules` guard, no `REPO_ROOT`/`cd`) since CI
always needs a clean install.

PR #949, opened 2026-05-30
ci: wrap preview-and-smoke deploy npm ci in retry

natb1:
## Summary
- The `Deploy Cloud Functions (if changed)` step of the `preview-and-smoke` job in `.github/workflows/pr-checks.yml` ran a **bare `npm ci`** that was never covered by the 3-attempt retry wrapper added in #811 — leaving the same cold-cache `ECONNRESET` exposure on this one step.
- Factored the retry loop out of `ensure_deps()` into a reusable `npm_ci_with_retry()` helper in `lib.sh` (uses `return`, not `exit`, so it works both standalone and inside `ensure_deps`'s subshell).
- The deploy step now sources `lib.sh` and calls `npm_ci_with_retry`; `ensure_deps()` delegates to the same helper, so its behavior is unchanged.

## Test plan
- [x] `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — 761/761 pass, including the existing `ensure_deps` retry tests and two new direct `npm_ci_with_retry` tests (succeeds-on-attempt-3, fails-after-3).
- [x] `bash -n` on `lib.sh`, the test file, and the extracted deploy `run:` body; `source lib.sh && type npm_ci_with_retry` resolves.
- [ ] CI: the `preview-and-smoke` job exercises the deploy step on this PR.

Closes #817


natb1:
## QA — #817 wrap preview-and-smoke deploy `npm ci` in retry

Non-browser QA pass (CI workflow + shell helper change; no app surface to drive).

**Items walked: 4 · PASS 4 · FAIL 0 · SKIP 0**

| # | Item | Result | Evidence |
|---|------|--------|----------|
| 1 | `npm_ci_with_retry` retries on transient failure | PASS | `test-dispatch-scripts.sh` → 824/824 pass, incl. 4 new assertions (retries to success on attempt 3; fails after exhausting 3 attempts; npm called exactly 3× each case) |
| 2 | `pr-checks.yml` Deploy step calls the helper, not bare `npm ci` | PASS | Step now runs `source .claude/.../lib.sh` + `npm_ci_with_retry` |
| 3 | `ensure_deps` still installs via the same retry path (regression) | PASS | `ensure_deps` delegates to `( cd "$REPO_ROOT" && npm_ci_with_retry )` |
| 4 | `lib.sh` is syntactically valid / sourceable | PASS | `bash -n lib.sh` clean |

No bugs found. Branch synced with `origin/main` before the pass.

The retry itself only fires on a real registry network drop (ECONNRESET) inside a CI runner, which can't be reproduced locally — the test suite exercises the helper's retry/backoff/exhaustion logic with an npm stub instead.

natb1:

### Findings that were fixed

- **`attempt` not declared `local` in `npm_ci_with_retry`** (`.claude/skills/dispatch-propagate/scripts/lib.sh:116`) — without `local`, the loop counter writes to the calling shell's scope. The CI step in `pr-checks.yml` calls `npm_ci_with_retry` directly (`source lib.sh; npm_ci_with_retry`, no subshell), so after the function returns `attempt` would linger in the step shell. Changed `attempt=1` → `local attempt=1`, confining it to function scope on both the subshell path (`ensure_deps`) and the direct-call path (CI). Fixed in `6420f5e`.

### Informational findings

- **`npm_config_*` exports propagate into the calling shell** (`.claude/skills/dispatch-propagate/scripts/lib.sh`) — `export npm_config_fetch_retries=5` and friends leak into the CI step shell and affect later `npm run` / `npx firebase-tools` calls. Elevated retry/timeout budgets for those commands are benign and possibly beneficial; no action recommended.

### Disregarded findings

- **Test exercises the subshell path, not the direct-call path** — false positive. The test subshell correctly isolates `attempt` to verify the retry loop's stub-counting and exit-code behavior; shell-variable leakage is a separate concern already addressed by the `local` fix above.

natb1:
## Code review — `/code-review max --fix`

Reviewed the diff for this PR (wrapping the `preview-and-smoke` "Deploy Cloud
Functions" `npm ci` in a 3-attempt retry helper). No correctness bugs found in
the workflow change: `npm_ci_with_retry` retries up to 3 times with 5s/15s
backoff, `ensure_deps` propagates failure via the subshell exit code, and the
deploy step calls `npm_ci_with_retry` at the workspace root before the build.

### Informational findings

- **`test-dispatch-scripts.sh` — inline pass/fail counter in `npm_ci_with_retry`
  test 2.** The new test uses an inline `TOTAL`/`PASS`/`FAIL` block rather than
  an `assert_eq` helper. This mirrors the pre-existing `ensure_deps` test 2
  verbatim, so the two parallel test blocks stay symmetric. Converting only the
  new test would create asymmetry; converting both is outside this diff's scope.
  No action recommended — surfaced for reference.

No findings were fixed (the one candidate — adding `local` to the `attempt`
variable in `npm_ci_with_retry` — was already present at `lib.sh:116`). No
findings disregarded or deferred.

natb1:
## /review-fix — parked for human review

This PR now has **zero net diff against `origin/main`**. The merge-base of
`origin/main` and this branch's HEAD *is* `origin/main` (`e2f3ba2`), so the
branch contributes no changes.

**Cause:** PR #947 ("Extend npm ci retry to bare workflow call sites") landed on
`main` and solved the same problem as #817 — wrapping the bare `npm ci` in the
`preview-and-smoke` deploy step with retry — using a cleaner standalone
`npm-ci-with-retry.sh` script instead of this branch's inline
`source lib.sh; npm_ci_with_retry` approach. Merging current `main` resolved all
three conflicting files (`pr-checks.yml`, `lib.sh`, `test-dispatch-scripts.sh`)
in main's favor, leaving this branch with nothing of its own.

**Why this is parked rather than advanced:** `/review` on an empty diff yields no
findings, after which the chain would apply `dispatch:reviewed` and march an
empty PR toward a ready/merge state. That is a no-op merge of a redundant PR — a
decision for a human, not the automated chain.

**Recommended action:** close #817 and PR #949 as superseded by #947, or
re-scope #817 if some residual gap remains that #947 did not cover.

natb1:
Closing as superseded by #947, which solved the same problem (npm ci retry on the preview-and-smoke deploy step) with a cleaner standalone npm-ci-with-retry.sh. This branch now has zero net diff against main.

PR #911, opened 2026-05-29
dispatch-select-target: orphan worktrees no longer lock priority items out of the ladder

natb1:
## Summary

A help-wanted issue or PR whose `<N>-*` worktree exists on disk but has no live
Claude session (an orphan, e.g. left by a crashed or swept session) was skipped
by the selector exactly as a session-owned worktree was. Orphan worktrees on
high-priority items could lock them out of the ladder, so a tick fell through to
lower-priority work that should have lost.

The skip is now keyed on **live-session ownership**, not mere worktree existence:

- `dispatch-select-target` sources `lib-claude-agents.sh`; the PR-branch row and
  the issue `<N>-*` row skip only when `worktree_has_live_session` owns the
  matching worktree. Worktree existence is a fast-path miss (no daemon
  round-trip); an orphan worktree no longer skips.
- `dispatch-resolve-worktree` queue mode emits `enter <path>` for an orphan
  worktree (recycle it) and `conflict <path>` only when a live session owns it.

`worktree_has_live_session` is fail-safe: an unqueryable daemon counts as owned,
so a row is skipped / a queue target conflicts rather than being double-claimed.

SKILL.md prose updated where it described the old worktree-existence skip.

## Test plan

- [x] `test-dispatch-scripts.sh` passes (598/598)
- [x] Selector PR-row & issue-row: live session skips; orphan worktree does not
- [x] Queue-mode resolve: orphan → `enter`, live → `conflict`, unqueryable daemon → `conflict`
- [x] 2026-05-29 reproduction: orphan-worktree priority PRs are selected over a lower-priority help-wanted issue with no worktree

Closes #905

Issue #945, opened 2026-05-30
dispatch: release the selection lock after worker registration to close the spawn-gap re-selection race


natb1:
## Context

`/dispatch-propagate` serializes target selection with a selection lock, but the lock is released too early to prevent two consecutive ticks from selecting the same worktree.

`dispatch-finalize-selection` releases the lock (its final `exec dispatch-acquire-lock --release`) at the end of worktree resolution — *before* the worker is spawned. The worker is a `claude --bg` job that takes ~1s to boot and register with the daemon. In that window the target worktree exists on disk with no live session, which `dispatch-select-target` treats as a recyclable **orphan** (per #905), not a skip. A concurrent tick therefore re-selects the same target and only discovers the collision one step later at `dispatch-resolve-worktree`, which now sees the live worker and emits `conflict` — a wasted `drain worktree-conflict` tick.

Observed: a manual `/dispatch 919` selected PR #942 (issue #919), finalized + released the lock, spawned its worker (named `919-dispatch-consolidate-router`), and self-closed; a second `/dispatch` then acquired the freed lock and re-selected the same PR during the worker's boot gap, hitting `drain worktree-conflict`.

This survives #919's router consolidation: PR #942 preserves "the lock-free post-finalize tail" (`finalize → CI gate → concurrency gate → spawn`), so the release still lands before the spawn. The fix therefore builds on the post-#919 `dispatch-materialize-spawn`.

## Root cause

The lock guarantees one selection at a time, but a selection's outcome is not *observable to the next selection* until the worker registers under its `--name=<worktree-basename>` — the exact name `worktree_has_live_session` keys on. Releasing at finalize ends the lock before that ownership handoff completes.

## Fix

Move the lock release from `dispatch-finalize-selection` to *after* worker registration, on the spawn path only:

1. `dispatch-finalize-selection` writes the `tmp/dispatch-worktree` marker and **no longer releases** the lock (drop the trailing `exec dispatch-acquire-lock --release`).
2. `dispatch-materialize-spawn` releases the lock at each terminal point in its tail:
   - **`propagate`**: release *after* `dispatch-spawn-worker` returns `spawned`/`deduped`. At that instant the worker is verified-registered (`verify_agent_registered_under`) under `<basename>`, so the next tick's `dispatch-select-target` skips it.
[...]

PR #967, opened 2026-05-30
dispatch: make trace-leaf queue-mode child-skip live-session-aware (#914)

natb1:
Closes #914

Queue-mode leaf tracing in `dispatch-trace-leaf` was worktree-existence-based, not liveness-aware. PR #911 (issue #905) fixed the symmetric problem for direct PR-rows / issue-rows in `dispatch-select-target`; this applies the same fix one level down, in issue-subtree leaf resolution.

An orphan child worktree (registered on disk, no live session) was miscounted as a conflict; when every reachable open leaf was thus conflicted the script exited 2 and the selector skipped the parent help-wanted issue. Live repro (2026-05-30): a tick reported issues 452/755/963 as all-leaves worktree-conflicted and skipped all three, while only two worktrees had live sessions and ~51 orphans were miscounted.

## Changes

- `dispatch-trace-leaf`: sources `lib-claude-agents.sh`; replaces the existence-keyed `CONFLICTED` map with a path-bearing `WORKTREE_PATHS_BY_NUM` (queue mode only); adds `child_has_live_worktree` (mirror of the selector's `issue_has_live_worktree`) — no worktree → fast-path miss, else live-session check per path; descent skips a child only when a live session owns its `<N>-*` worktree. Comments and the exit-2 message updated to live-session semantics (the `worktree-conflicted` token is retained so existing matchers stay valid).
- `dispatch-select-target`: comment-only — header bullet and leaf-trace block now describe the child-skip as live-session-owned, orphan children descendable.
- `test-dispatch-scripts.sh`: three new queue-mode trace-leaf cases (orphan child descendable, live child skipped, all-leaves-live → exit 2).

## Test plan

- [x] `test-dispatch-scripts.sh` — 952/952 passing

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -63,8 +63,9 @@
 #   5. Oldest open help-wanted issue whose <N>-* worktree, if any, is not owned
 #      by a live Claude session, and which is not already closed by a non-draft
 #      (ready) PR (#920), resolved to a startable open leaf (an issue whose
-#      entire open-leaf subtree is worktree-conflicted is skipped, exactly as a
-#      live-session-owned worktree is)
+#      entire open-leaf subtree is owned by live sessions is skipped, exactly as
+#      a live-session-owned direct worktree is — an orphan child worktree stays
+#      descendable, #914)
 #   6. Oldest "qa"          PR (draft, CI green, no dispatch:* label)
 #
 # A PR or help-wanted issue is skipped only when a live Claude session owns its
@@ -591,11 +592,12 @@ while IFS=$'\t' read -r issue_num issue_labels; do
   [[ -n "${FIRST_ISSUE[$issue_bucket]:-}" ]] && continue
   # Resolve a startable open leaf within the issue's subtree. dispatch-trace-leaf
   # in queue mode descends open blockers and sub-issues, skipping any child whose
-  # <N>-* branch is an existing local worktree (owned by another session):
+  # <N>-* worktree is owned by a live session (#914) — an orphan child worktree
+  # stays descendable:
   #   exit 0 — a startable open leaf was found; record it for this bucket.
-  #   exit 2 — every reachable open leaf is worktree-conflicted. Skip this issue
-  #            exactly as a direct worktree is skipped; a later issue in the
-  #            same bucket may still fill the slot.
+  #   exit 2 — every reachable open leaf is owned by a live session. Skip this
+  #            issue exactly as a direct live-owned worktree is skipped; a later
+  #            issue in the same bucket may still fill the slot.
   #   exit 1 — usage error; propagate as a hard failure, never swallow as a skip.
   if leaf=$("$SCRIPT_DIR/dispatch-trace-leaf" "$issue_num" queue); then
     FIRST_ISSUE[$issue_bucket]="$leaf"
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
@@ -209,8 +209,8 @@ fi
 # by a live session (or cycles block every path); orphan worktrees were
 # descended through. Surface this so /dispatch-propagate can stop with a clear
 # message instead of dispatching on a conflict. The "worktree-conflicted" token
-# is retained for the matchers in dispatch-select-target and
-# test-dispatch-scripts.sh that key on it.
+# is retained for the pattern-matcher in test-dispatch-scripts.sh that keys on it
+# (dispatch-select-target only checks the exit code, not the message text).
 if [[ "$MODE" == "queue" ]]; then
   echo "error: all open leaves reachable from $ISSUE_NUM are worktree-conflicted (owned by a live session)" >&2
   exit 2
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
@@ -12,13 +12,17 @@
 #              dispatch-resolve-worktree's explicit mode, which yields `enter`
 #              for the recycle-after-completion case, or `conflict` if a live
 #              session still owns the worktree — #837).
-#   queue    — descent filters out children whose <N>-* branch is an existing
-#              local worktree (owned by another session). If a node has open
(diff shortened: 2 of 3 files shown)


natb1:
## Dispatch QA — clean pass

**Issue #914** — make `dispatch-trace-leaf` queue-mode child-skip live-session-aware.

This is a non-browser PR (dispatch shell tooling), so QA verified behavioral acceptance criteria via the script test suite and syntax checks.

### Items walked

| # | Item | Result |
|---|------|--------|
| 1 | Orphan child worktree (on disk, no session) stays descendable → returns lowest leaf 701 | PASS |
| 2 | Live-session-owned child is skipped → returns sibling 702 | PASS |
| 3 | Fully live-owned subtree exits 2 with `worktree-conflicted` stderr | PASS |
| 4 | No regression across the dispatch script suite | PASS (952/952) |
| 5 | Changed scripts pass `bash -n` syntax check | PASS |

No bugs found. Acceptance criteria for orphan-vs-live child handling in queue-mode leaf tracing are covered by new tests 12a/12b/12c in `test-dispatch-scripts.sh`.

natb1:
## Code review

### Findings that were fixed

All findings were comment-accuracy corrections, fixed in commit `07f6a6b`:

1. **`dispatch-trace-leaf`** (exit-2 comment) — the comment claimed `dispatch-select-target` keys on the `worktree-conflicted` token, but `dispatch-select-target` only inspects exit codes, not the message text. Only `test-dispatch-scripts.sh` pattern-matches the token; the comment now says so.
2. **`test-dispatch-scripts.sh`** (test 8) — replaced the inaccurate "Pretend another session owns 701's worktree" comment. The child is skipped via the daemon-UNKNOWN fail-safe (non-existent binary path → occupied), not a confirmed live-session check.
3. **`test-dispatch-scripts.sh`** (test 10) — replaced the stale "every child is worktree-conflicted" framing. Both children are skipped because the daemon is UNKNOWN → occupied, not because worktree existence alone triggers a skip. The true all-live-owned case is covered by test 12c.

No correctness bugs were found in the core logic (`child_has_live_worktree`, `WORKTREE_PATHS_BY_NUM`, session-liveness delegation, fast-path miss, tests 12a/12b/12c). No findings were deferred, disregarded, or left informational.

natb1:
## Fixed
- Multi-path child coverage gap: added queue-mode test `12d` exercising `child_has_live_worktree`'s loop over a child with two `<N>-*` worktrees (one orphan, one live-owned) → skipped because any live match counts as owned: be0afe5

## Informational
- `WORKTREE_PATHS_BY_NUM` is declared unconditionally but populated only in queue mode — intentional, mirrors `dispatch-select-target`'s pattern; not a bug.

## Dismissed
- Test numbering `12a/12b/12c` — cosmetic; the `a/b/c` suffix is an accepted convention and the new cases sit correctly after test 12 / before test 13, so renumbering existing cases would add churn for no behavioral gain.

## Deferred
- Extract `child_has_live_worktree` / `issue_has_live_worktree` into a shared `lib.sh` helper to remove the structural duplication → #971

natb1:
## Security review — #914 (queue-mode leaf-skip liveness)

Fanned out 9 parallel reviewers (6 domains, red team, built-in `/security-review`, CodeQL) against the diff vs `origin/main` (3 bash scripts: `dispatch-trace-leaf`, `dispatch-select-target`, `test-dispatch-scripts.sh`). **No required fixes — no code changes.**

CodeQL: no open alerts. Built-in `/security-review`: no findings above threshold. Dependency / Firebase / secrets / data-exposure: clean (no dependency-file changes; no Firebase surface; no credentials; new diagnostics emit only the validated integer issue number).

### Findings and dispositions

All findings are **not fixed** — each is either pre-existing behavior in untouched code or not realistically exploitable. None are `required`.

| # | Finding | Location | Confidence | Disposition |
|---|---|---|---|---|
| 1 | `claude_sessions_with_name` does not filter `stopped` sessions, so a lingering stopped-session record could mark a worktree live and starve a child. | `lib-claude-agents.sh:183` (called from `dispatch-trace-leaf`) | medium | **out-of-scope** |
| 2 | Name-only liveness key (`basename`) is machine-wide; a same-named session in an unrelated checkout could false-positive as owning a child worktree (DoS/starvation). The diff extends this check to all subtree children. | `lib-claude-agents.sh:195-215`, `dispatch-trace-leaf:81-92` | medium | **out-of-scope** |
| 3 | Newline-joined `WORKTREE_PATHS_BY_NUM` accumulator splits on `\n`; a worktree path containing a newline would corrupt the per-child path loop. | `dispatch-trace-leaf:70,85-90` | low | **false-positive** |
| 4 | TOCTOU between the startup worktree snapshot and the live-session check could let a concurrently-created live worktree pass the fast-path miss. | `dispatch-trace-leaf:64-72,81-92` | low | **false-positive** |
| 5 | Child issue number used as bash array key without re-validation. | `dispatch-trace-leaf:83` | low | **false-positive** |
| 6 | `select_target_fake_claude` builds fixture JSON via string concat without escaping. | `test-dispatch-scripts.sh:906-910` | low | **false-positive** |

### Rationale

- **#1, #2 (out-of-scope):** Both live in `lib-claude-agents.sh`, which this diff does not touch. `dispatch-select-target` already calls `worktree_has_live_session` for direct PR/issue rows (shipped via #905/#911); this PR extends the *same* helper to leaf-tracing, so the behavior is symmetric and pre-existing rather than introduced here. The stopped-session gap (#1) is real (`verify_agent_registered_under:277` filters `stopped`; `claude_sessions_with_name:183` does not) and the cross-checkout name collision (#2) is a documented limitation of the daemon name-match scope — both warrant a separate hardening of the shared helper, not a change to this PR's diff.
- **#3 (false-positive):** Mirrors the existing `dispatch-select-target` accumulator pattern; `git worktree list --porcelain` does not emit newline-embedded paths and git resists creating them. `worktree_has_live_session` only feeds the value to `basename` + `jq --arg` (escaped) — no shell-execution boundary.
- **#4 (false-positive):** Acknowledged design trade-off (#945); the `dispatch-spawn-worker` dedup guard absorbs the window, and the new false-negative direction is strictly safer than the old existence-map false-positive (which had no in-tick recovery).
- **#5 (false-positive):** Child numbers come from `jq -r .number` (decimal integers) and feed an associative-array key lookup — a data-plane lookup, not a shell-execution boundary. The root argument is already validated `^[1-9][0-9]*$`.
- **#6 (false-positive):** Test harness only; all callers pass branch-name literals, which cannot contain `"`.

PR marked ready.

Issue #961, opened 2026-05-30
dispatch: reseed timer service exits 127 — systemd-run unit lacks bash on PATH


natb1:
## Context

`dispatch-schedule-reseed` (#725) schedules a one-shot `systemd.user` timer at the next rate-limit cap reset so the `/dispatch-propagate` chain resumes the moment a capped window reopens. The timer's `ExecStart` is `dispatch-spawn-router` (`.claude/skills/dispatch-propagate/scripts/dispatch-spawn-router`), whose shebang is `#!/usr/bin/env bash`.

On the NixOS/WSL dev host, `bash` and every other binary the script needs (`claude`, `git`, `jq`) live in the nix store (`/nix/store/…/bin`), not in `/usr/bin` or `/bin`. A transient unit created by `systemd-run --user` inherits the systemd user manager's minimal default `PATH`, which does not include the nix store. When the timer fires, `/usr/bin/env bash` cannot resolve `bash` and the service exits 127.

## Reproduction

1. Cross the 5h cap floor (>=50% on `five_hour.used_percentage`) so a tick schedules a reseed timer for the cap's `resets_at`.
2. Let the timer fire.

Observed on 2026-05-30: `dispatch-reseed-1780138800.service` fired at 11:00 UTC and failed, degrading the systemd user session.

### Expected

The reseed timer fires, `dispatch-spawn-router` runs, and the `/dispatch-propagate` chain is re-seeded.

### Actual

```
dispatch-spawn-router[…]: /usr/bin/env: 'bash': No such file or directory
dispatch-reseed-1780138800.service: Main process exited, code=exited, status=127/n/a
dispatch-reseed-1780138800.service: Failed with result 'exit-code'.
```

The chain is not re-seeded, and the failed transient unit lingers as a `failed` entry that degrades `systemctl --user`.

## Root cause

The `systemd-run --user` invocation (`dispatch-schedule-reseed` lines ~305-311) does not propagate a usable `PATH` into the transient unit. The unit runs with the user manager's default `PATH` (no nix store), so `env` cannot find `bash` — and even if it could, `dispatch-spawn-router`'s own `claude` / `git` / `jq` calls would fail for the same reason.

## Fix

Propagate the scheduling caller's `PATH` into the transient unit by adding `--setenv=PATH="$PATH"` to the `systemd-run --user` call. The caller (a `/dispatch-propagate` tick) always has the full nix-store `PATH`, so the unit becomes self-sufficient.

[...]

PR #991, opened 2026-06-01
dispatch: explicit single-target /dispatch waits for in-progress CI

natb1:
Closes #979

## Summary

`/dispatch <N>` targets one explicit issue/PR — an effective queue of size one. When that target's PR has CI in progress, `dispatch-phase <N>` returns `waiting` and the single-target driver emitted `drain ci-waiting`: the run self-closed without spawning a worker and nothing returned to the target (a silent no-op, reproduced on `/dispatch 961` and `/dispatch 829`). Multi-item queue mode already skips a `waiting` PR and picks the next ready priority — but the single-target path has no "next", so skipping equals doing nothing.

This makes the single-target path **wait** for its CI instead of draining: it schedules a target-keyed reseed timer that re-runs `/dispatch-propagate <N>` after a short delay, bounded by an attempt counter that escalates to `dispatch:office-hours` after a cap.

## Changes (three units)

- **Unit 1 — `dispatch-spawn-router`:** accepts an optional positive-integer target `<N>` and spawns `/dispatch-propagate <N>` (target-keyed) instead of the bare router. No-arg path unchanged.
- **Unit 2 — `dispatch-schedule-target-reseed <N>` (new):** resolves N's draft PR, reads/bumps a `dispatch:ci-wait-attempt-<n>` label, and schedules a transient `systemd.user` timer (`dispatch-reseed-target-<N>-<fire>`) whose ExecStart is `dispatch-spawn-router <N>` (`--setenv=PATH` + `--working-directory` mirror the #961 fix). At the cap (default 3) it parks the issue on `dispatch:office-hours` and prints `escalated`, scheduling no timer. Fail-closed input validation on every value reaching `(( ))` / `systemd-run` / `gh`.
- **Unit 3 — wire into `dispatch-materialize-spawn`:** the single-target driver (`MODE == explicit || GAP <= 1`) replaces `drain ci-waiting` with a call to the scheduler — `reseeded ...` maps to `drain ci-reseeded`, `escalated` to `notify ci-wait-exhausted`, and a hard scheduler failure falls back to `drain ci-waiting` (never wedges the tick). The fan-out path (`--gap N>1`) is unchanged: a `ci-waiting` target stays a skip. Updates `SKILL.md` Table 2 and `reference.md`.

## Test plan

- [x] `test-dispatch-scripts.sh` fully green (1049/1049): spawn-router target argv; schedule-target-reseed timer/counter-bump/escalation/bad-input; materialize-spawn single-target reseed → `drain ci-reseeded`, escalation → `notify ci-wait-exhausted`, and fan-out ci-waiting skip unchanged (scheduler not called).
- [x] `bash -n` clean on all changed/new scripts.
- [ ] Manual: `/dispatch <N>` against an in-progress-CI target schedules a reseed and reports it is waiting on N's CI (exercised structurally by the suite; full live run pending CI on this PR).

### DIFF
--- a/.claude/skills/dispatch-propagate/SKILL.md
+++ b/.claude/skills/dispatch-propagate/SKILL.md
@@ -109,7 +109,11 @@ for a fan-out run).
 | `resolve merge-conflict` | `origin/main` does not merge cleanly into the resolved worktree; `dispatch-merge-main` aborted the merge (tree left clean) and the script printed `issue:` / `worktree:` / `mode:` detail. Run the auto-resolve attempt (§2a). | `propagate` (resolved, after re-spawn) or `notify` (ambiguous) |
 | `notify spawn-failed` | `dispatch-spawn-worker` exited non-zero — a worker was spawned but did not register. | `notify` |
 | `drain worktree-conflict` | The target worktree cannot be safely entered (the script printed the `path:` detail): "worktree at `<path>` for issue `<N>` cannot be entered; closing — the next baton-pass or office-hours hand-off will re-seed". | `drain` |
-| `drain ci-waiting` | The target PR's CI is still in progress (the script printed the `#<N>:` line); echo it. | `drain` |
+| `drain ci-reseeded` | The target PR's CI is still in progress (the script printed the `#<N>:` line; echo it). A target-keyed reseed timer was scheduled, so the chain returns to `<N>` when its CI concludes: "waiting on #<N>'s CI; reseed scheduled". | `drain` |
+| `notify ci-wait-exhausted` | The CI-wait attempt cap was hit; the issue was parked on `dispatch:office-hours` and no further reseed will be scheduled: "#<N>'s CI did not conclude within the attempt cap; parked for review". | `notify` |
+
+> Fallback: if `dispatch-schedule-target-reseed` hard-fails, the single-target
+> path emits `drain ci-waiting` instead (route as `drain`); the reason is on stderr.
 
 #### Fan-out summary contract
 
--- a/.claude/skills/dispatch-propagate/reference.md
+++ b/.claude/skills/dispatch-propagate/reference.md
@@ -69,8 +69,10 @@ early-release window.
 
 `dispatch-materialize-spawn` releases at each terminal point: on the `propagate`
 path after `dispatch-spawn-worker` returns success (worker verified-registered),
-on `notify spawn-failed` after the failed spawn returns, and on `drain
-ci-waiting` at its stop.
+on `notify spawn-failed` after the failed spawn returns, and at the single-target
+CI-wait stop (now emitting `drain ci-reseeded` / `notify ci-wait-exhausted`, or
+`drain ci-waiting` on a scheduler hard-failure), which releases before scheduling
+the reseed.
 `dispatch-finalize-selection` writes the recovery marker but no longer releases.
 The pre-finalize stop paths — `notify target-blocked`, `resolve merge-conflict`,
 and `drain worktree-conflict`, plus the internal `exit 2` error paths — release
@@ -373,3 +375,22 @@ hand-off returns it to the dispatch chain and re-seeds from human action.
 A weekly fallback heartbeat for the edge case of a manual issue filed while
 the chain is stalled with no live session to receive it can be added in a
 follow-up if it matters in practice.
+
+## Target-keyed CI-wait reseed (#979)
+
+CI-wait handling splits on available queue size, not invocation mode. In a
+multi-item fan-out run (`--gap N>1`), a `ci-waiting` target is a `skipped`
+outcome: the loop moves on to the next ready priority, and other live workers'
+Stop-hooks bring the chain back to that target later. The single-target path —
+an explicit `/dispatch <N>`, or a `--gap 1` run whose only candidate is
+not-ready — has no "next". So instead of draining as a no-op, it calls
+`dispatch-schedule-target-reseed <N>`.
+
+That schedules a transient `systemd.user` timer
+(`dispatch-reseed-target-<N>-<fire>`) a short delay out that re-runs
+`dispatch-spawn-router <N>` → `/dispatch-propagate <N>`, returning the chain to
+N. A `dispatch:ci-wait-attempt-<n>` label on the PR counts the attempts; at the
+cap (default 3) the target is parked on `dispatch:office-hours` and no further
+reseed is scheduled. Because `dispatch-phase` returns `waiting` only for
+genuinely in-progress checks — a CI *failure* resolves to `verify`, an
+actionable phase — the wait terminates on its own when CI finishes.
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-materialize-spawn
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-materialize-spawn
@@ -46,8 +46,22 @@
 #   drain worktree-conflict  the target worktree cannot be safely entered; the
 #                          lock is released at the guard. The conflict path
 #                          precedes this token.
-#   drain ci-waiting       the target PR's CI is still in progress; no spawn this
-#                          tick. Lock released at this stop.
+#   drain ci-reseeded      single-target path (#979): the target PR's CI is still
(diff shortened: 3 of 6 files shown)


natb1:
## QA (autonomous) — clean pass

Scripts-only PR (no browser component) — ran the non-browser walkthrough.

**Items walked:** 4 — all machine-verifiable.

| # | Item | Result |
|---|------|--------|
| 1 | Dispatch script test suite (`test-dispatch-scripts.sh`) | PASS |
| 2 | `dispatch-schedule-target-reseed` syntax + bad-`<N>` validation | PASS |
| 3 | `dispatch-spawn-router` optional `<N>` arg validation | PASS |
| 4 | `dispatch-materialize-spawn` ci-waiting wiring syntax | PASS |

**PASS: 4 · FAIL: 0 · SKIP: 0**

**Coverage notes:**
- Suite passed **1049/1049**, including the four #979 acceptance scenarios:
  explicit-target `ci-waiting` → `drain ci-reseeded` (reseed scheduled),
  reseed-at-cap → `notify ci-wait-exhausted` (office-hours escalation),
  multi-item queue still skips a `waiting` PR, and the attempt-counter cap.
- `dispatch-schedule-target-reseed` rejects empty/non-numeric `<N>` with exit 2;
  `dispatch-spawn-router` rejects non-numeric and too-many-args with exit 2.

**Deferred to office-hours:** none.

**Bugs found:** none.

natb1:
## Code review

`/code-review max --fix` surfaced 3 findings, all fixed in `6c3eed7`.

### Findings that were fixed

- **`dispatch-schedule-target-reseed:263` — already-exists idempotency path emitted no stdout.** Two ticks for the same ci-waiting issue inside one 300s fire window produce the same unit name; the second `systemd-run` fails with `already exists`. The branch wrote only to stderr and exited 0, so `dispatch-materialize-spawn` matched neither `escalated` nor `reseeded *` and fell through to the `drain ci-waiting` no-op token (the pre-#979 behavior) even though a timer was correctly scheduled. Fixed by emitting `reseeded $UNIT_NAME at $FIRE` on this path so the caller routes to `drain ci-reseeded`. (`6c3eed7`)
- **`test-dispatch-scripts.sh` — no coverage for the scheduler-failure fallback or the already-exists path.** `MAT_RESEED_RC` was never set non-zero in any test, so a real scheduler failure emitting `drain ci-waiting` would go undetected. Added a test that sets `MAT_RESEED_RC=1` and asserts the `drain ci-waiting` fallback (scheduler called, no spawn, lock released), plus a test exercising the already-exists idempotency path (exit 0 + `reseeded …` stdout). (`6c3eed7`)
- **`dispatch-schedule-target-reseed:104` and `dispatch-spawn-router:82` — validation regex accepted `N=0`.** `^[0-9]+$` contradicted the "positive integer" requirement in the header comment and error message. Tightened both to `^[1-9][0-9]*$`, matching `dispatch-materialize-spawn`'s existing validation. (`6c3eed7`)

natb1:
## Fixed
- Test coverage gap: explicit-mode (`MODE=explicit`, the literal `/dispatch <N>` path) CI-waiting → `drain ci-reseeded` was untested; prior reseed tests only hit the `GAP<=1` queue-default disjunct. Added an explicit-mode test: 1b04a56

## Informational
- `dispatch-schedule-target-reseed` Step 3: a `gh pr view` failure leaves `CUR` empty and trips the integer guard with `got ''`. Behavior is correct (fails closed, exit 2) and the empty value is visible in the message; left as-is.
- `DELAY` validation accepts `^[0-9]+$`, so `DELAY=0` (immediate fire) passes — asymmetric with `CAP`'s positive-only check. The override is test-only (default 300) and the attempt cap bounds any fast re-fire, so no loop risk.
- `dispatch-spawn-router`'s dedup does not distinguish a bare vs target-keyed live dispatcher — pre-existing one-router-at-a-time behavior, intentional and conservative for the single-target reseed path.
- `Persistent=true` timers missed during a WSL-down window all fire on resume; each could schedule its own successor, but the attempt cap bounds total reseeds.
- Timer fire time is `NOW + DELAY` from the scheduling host's clock; clock skew is a non-issue on the single-machine dev setup.

## Dismissed
- `WAITING_N` as a script-scope global set inside `process_one_target` — consistent with the existing `CONFLICT_N`/`CONFLICT_WORKTREE` pattern, already carries an explanatory comment, and is read only when `OUTCOME == ci-waiting`; not a defect.
- "Late `WAITING_N` reassignment after lock release" — `WAITING_N` is process-local; a concurrent tick in another process cannot mutate it. Not reachable within a single script instance.
- Missing `*)` catch-all in the single-target `case "$OUTCOME"` — `OUTCOME` is a closed internal enum and every value it can take is handled; per `.claude/rules/code-style.md`, no handling is added for states that cannot occur.

## Deferred
- _None._

natb1:
## Security Review — `/security-review-fix`

Fanned out 9 parallel reviewers over the branch diff vs `origin/main`: 6 security
domains, a red team, the built-in `/security-review` scan, and the PR's CodeQL
alerts. The diff is shell scripts + docs under
`.claude/skills/dispatch-propagate/` (new `dispatch-schedule-target-reseed`,
modified `dispatch-materialize-spawn` / `dispatch-spawn-router`, tests, SKILL/reference docs).

**Clean domains** (no findings): dependency audit (no `package.json` /
lockfile change), Firebase, secrets scan, data exposure, built-in
`/security-review`, CodeQL (empty alert set for the PR ref).

### Findings & dispositions

| # | Finding | Location | OWASP / STRIDE | Conf. | Disposition |
|---|---------|----------|----------------|-------|-------------|
| 1 | `PR_NUM` from `dispatch-find-pr` was only emptiness-checked, then interpolated into the same `gh pr view/edit "$PR_NUM"` calls that `N` is numeric-guarded against | `dispatch-schedule-target-reseed:115` | A03 Injection / Tampering | medium | **Fixed** — `cd85879` |
| 2 | `WAITING_N` not re-validated in `dispatch-materialize-spawn` before passing to the reseed script | `dispatch-materialize-spawn` | A03 Injection / Tampering | low | **False positive** — callee validates `$1` as `^[1-9][0-9]*$` (line 104) and fails closed |
| 3 | `"already exists"` substring match could mask a genuine `systemd-run` failure | `dispatch-schedule-target-reseed:263` | A05 Misconfig / Repudiation | medium | **False positive** — systemd emits this only on `(N, FIRE)` unit-name collision, which is exactly the documented idempotency case |
| 4 | Cap bypass via TOCTOU on `dispatch:ci-wait-attempt-*` label removal | `dispatch-schedule-target-reseed:163` | A01 Broken Access Control / DoS | medium | **Out of scope** — requires repo label-write access that already permits arbitrary disruption; counter-in-labels mirrors the established #831 pattern; moving it off labels is a design change beyond this PR |
| 5 | Inherited `PATH` baked into the systemd timer unit (`--setenv=PATH="$PATH"`) | `dispatch-schedule-target-reseed:251` | A08 Integrity / EoP | low | **False positive** — this is the required #961 fix (timer otherwise exits 127 on NixOS/WSL); exploiting it needs prior control of the user's shell rc |
| 6 | `DISPATCH_TARGET_RESEED_MAIN_WORKTREE` env override unsanitized | `dispatch-schedule-target-reseed:211` | A03 Injection / Tampering | low | **False positive** — documented test-only override; production uses the git-derived path; env is trusted per project convention |
| 7 | No authorization check before `dispatch-apply-office-hours "$N"` | `dispatch-schedule-target-reseed:152` | A01 Broken Access Control / EoP | low | **Out of scope** — reviewer self-rated "no code change needed"; pre-existing `dispatch-apply-office-hours` pattern, not introduced by this diff |

### Fix applied

**Finding 1** — `cd85879`: added a `^[1-9][0-9]*$` guard on `PR_NUM` immediately
after the empty check, mirroring the script's existing `N` validation and its
documented rationale (a flag-like value would be parsed by `gh` as an option
rather than a PR number). A test (`test-dispatch-scripts.sh`) covers the new
guard: a non-numeric find-pr result exits 2 with no timer and no label edit. Full
suite green (1062/1062).

No high-confidence required findings remain unresolved.

PR #1012, opened 2026-06-02
dispatch: skip office-hours-parked leaves in dispatch-trace-leaf queue descent

natb1:
## Summary

`dispatch-trace-leaf` queue-mode descent filtered out children whose `<N>-*`
worktree is owned by a live Claude session (#914) but applied no equivalent
filter for `dispatch:office-hours`. A parked leaf reachable from an unparked
parent was therefore returned as a startable leaf, wedging queue selection: the
parked leaf occupied the top `priority`+`bug` bucket and was re-selected every
tick, starving the rest of the queue (#1011).

This extends the #914 live-session filter to office-hours parking — the same
skip #757 already applies to top-level queue issues in `dispatch-select-target`,
now applied during leaf resolution:

- Gather the open `dispatch:office-hours` set once up front (mirroring the
  worktree-path map), so the check is a map lookup with no per-child gh call.
  A gh failure is a hard error, not an empty set.
- Filter a parked child from the descent exactly like a live-owned one: neither
  descended into nor returned as a startable leaf.
- A subtree whose every reachable open leaf is parked or live-owned now exits 2,
  so `dispatch-select-target` falls through to the next candidate.
- Queue-mode only; explicit-mode descent is unchanged.

## Test plan

- [ ] `test-dispatch-scripts.sh` passes (1119 → 1123 tests)
- [ ] queue mode: a parked topological leaf is skipped in favor of an unparked
      sibling leaf
- [ ] queue mode: a subtree whose every open leaf is parked yields exit 2
- [ ] explicit mode: a parked leaf is still returned (skip is queue-only)
- [ ] end-to-end: `dispatch-select-target` resolves a help-wanted issue's parked
      leaf to the next unparked leaf

Closes #1011

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
@@ -222,7 +222,16 @@ find_leaf() {
   return 1
 }
 
-if LEAF=$(find_leaf "$ISSUE_NUM"); then
+# In queue mode the root issue is filtered exactly like a descended child
+# (#1011): if the starting issue is itself a parked topological leaf, the
+# descent loop never gets a chance to skip it, so guard it here before the
+# trace. This keeps dispatch-trace-leaf's standalone contract — "queue mode
+# never returns a leaf carrying dispatch:office-hours" — true for the root, not
+# just for children. Force the no-startable-leaf path (rc 2) so it flows into
+# the exit-2 surface below, matching the all-children-parked case.
+if [[ "$MODE" == "queue" ]] && child_is_office_hours_parked "$ISSUE_NUM"; then
+  trace_rc=2
+elif LEAF=$(find_leaf "$ISSUE_NUM"); then
   trace_rc=0
 else
   trace_rc=$?
@@ -249,7 +258,7 @@ fi
 # in test-dispatch-scripts.sh that keys on it (dispatch-select-target only checks
 # the exit code, not the message text).
 if [[ "$MODE" == "queue" ]]; then
-  echo "error: all open leaves reachable from $ISSUE_NUM are worktree-conflicted (owned by a live session)" >&2
+  echo "error: all open leaves reachable from $ISSUE_NUM are worktree-conflicted (owned by a live session or parked on dispatch:office-hours)" >&2
   exit 2
 fi
 
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
@@ -86,9 +86,12 @@ fi
 # mirroring WORKTREE_PATHS_BY_NUM above, so child_is_office_hours_parked is a map
 # lookup with no per-child gh round-trip. A gh failure here is a hard error, not
 # an empty set — silently returning a parked leaf is the very bug being fixed.
+# --limit 300 matches dispatch-select-target's open-issue cap: the parked set is
+# a subset of open issues, so the default limit of 30 could silently truncate it
+# and let a parked leaf past position 30 be returned as startable.
 declare -A OFFICE_HOURS_PARKED=()
 if [[ "$MODE" == "queue" ]]; then
-  parked_json=$(gh issue list --state open --label "dispatch:office-hours" --json number) \
+  parked_json=$(gh issue list --state open --label "dispatch:office-hours" --limit 300 --json number) \
     || { echo "error: dispatch-trace-leaf $ISSUE_NUM: gh issue list for the dispatch:office-hours parked set failed; aborting rather than risk returning a parked leaf" >&2; exit 1; }
   while IFS= read -r parked_num; do
     [[ -n "$parked_num" ]] && OFFICE_HOURS_PARKED["$parked_num"]=1
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
@@ -16,13 +16,18 @@
 #              LIVE Claude session (#914). An orphan <N>-* worktree (on disk, no
 #              session) stays descendable — it is a reuse signal, not a skip,
 #              matching the direct-row liveness fix in dispatch-select-target
-#              (#905). A child with no worktree at all is never filtered. If a
-#              node has open children but every one is live-owned, the descent
-#              bubbles up too — that subtree has no ready work. If every
-#              reachable open leaf in the subtree is owned by a live session,
-#              the script exits non-zero with a clear message on stderr so the
-#              /dispatch-propagate caller can stop instead of dispatching on a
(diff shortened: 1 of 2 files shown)


natb1:
## QA (autonomous) — #1011 dispatch-trace-leaf office-hours-parked leaf skip

Scripts-only change (`dispatch-trace-leaf` + `test-dispatch-scripts.sh`) — non-browser path. QA verified the acceptance criteria via the dispatch test suite.

**Items walked:** 5 — all machine-verifiable.

| Item | Result |
|---|---|
| 1. Queue mode: parked leaf skipped → unparked sibling (`701`→`702`) | PASS |
| 2. Queue mode: all reachable leaves parked → exit 2 (`worktree-conflicted`) | PASS |
| 3. Explicit mode: parked leaf returned unchanged (skip not applied) | PASS |
| 4. End-to-end: `dispatch-select-target` falls through parked leaf (`5500`→`5501`) | PASS |
| 5. No regression: full `test-dispatch-scripts.sh` suite | PASS |

**Counts:** PASS 5 / FAIL 0 / SKIP 0.

**Full suite:** `1119/1119 passed, 0 failed`.

**Deferred to office-hours:** none.

**Bugs found:** none.

All four acceptance criteria from the issue are covered by passing tests: the parked leaf is skipped in queue mode in favor of the next startable sibling; an all-parked subtree exits 2; the skip is queue-mode-only (explicit unchanged); and `dispatch-select-target` no longer returns a parked leaf.

natb1:

### Findings that were fixed

Both findings fixed in `ee427804`:

- **`dispatch-trace-leaf` queue-mode root guard.** The office-hours skip filtered only descended children, not the root issue passed to `find_leaf`. A parked root that is itself a topological leaf (no open children) was returned directly, violating the standalone contract that queue mode never returns a `dispatch:office-hours`-parked leaf. Added a queue-mode root guard before the trace that forces the no-startable-leaf path (rc 2), mirroring the descent child-skip. Explicit mode is unchanged.
- **Misleading exit-2 stderr message.** The message attributed the stop solely to "owned by a live session" even though #1011 also fires it when every reachable leaf is parked on `dispatch:office-hours`. Appended "or parked on dispatch:office-hours" to the parenthetical; the load-bearing `worktree-conflicted` token is preserved for the test pattern-matcher.

Added two tests (queue-mode parked-root → exit 2; explicit-mode parked-root → returned unchanged). Full suite: 1121/1121 passing.

natb1:
## Review (`/review-fix`)

## Fixed
- Add `--limit 300` to the `gh issue list --label dispatch:office-hours` parked-set fetch in `dispatch-trace-leaf`, preventing gh's default limit of 30 from silently truncating the parked set (a parked leaf past position 30 could otherwise be returned as startable). Matches `dispatch-select-target`'s open-issue cap; test stub invocation-string updated. Commit: 0f309114

## Informational
- `declare -A OFFICE_HOURS_PARKED=()` is declared unconditionally and populated only in queue mode — harmless empty map in explicit mode, consistent with how `WORKTREE_PATHS_BY_NUM` is handled.
- No test covers the gh-failure path of the parked-set fetch; the live-session map (`WORKTREE_PATHS_BY_NUM`) lacks one too, so this is consistent with existing test conventions, not a regression.
- The filter skips a parked *child leaf* but does not skip a parked *non-leaf intermediate node* (its children remain reachable). Within spec — the issue scopes the fix to a topological leaf carrying the label — noted for future reference.

## Dismissed
- Style nit suggesting process substitution over the `printf | jq` subshell pipeline for the parked-set parse — not a functional concern.

## Deferred
_None._

natb1:
## Security review

Surface: **code** (two changed files, both bash scripts in the dispatch chain). No application source, dependency, or Firestore-rules surface, so the app-domain reviewers and the dependency audit did not run; CodeQL ran on the PR ref.

Changed files:
- `.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf`
- `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`

Reviewers run: input-validation, secrets scan, red team, built-in `/security-review`, and CodeQL (inline). Dependency audit skipped (no `package.json`/lockfile change).

### Result: no required fixes

CodeQL reported **no open alerts**. The secrets scan and the built-in `/security-review` scan reported **no findings**. The input-validation and red-team reviewers raised four concerns, all dispositioned below as false-positive or out-of-scope — none is a real vulnerability in the changed code, so no code change was applied.

| # | Location | Finding | OWASP / STRIDE | Confidence | Disposition |
|---|---|---|---|---|---|
| 1 | `dispatch-trace-leaf:98` | `jq` failure in `< <(...)` process substitution is swallowed under `set -euo pipefail`, leaving an empty parked set | A09:2021 / Tampering | medium | **false-positive** |
| 2 | `dispatch-trace-leaf:97` | `parked_num` not validated as a positive integer before use as an array key | A03:2021 / Tampering | low | **false-positive** |
| 3 | `dispatch-trace-leaf:94` | `--limit 300` could silently truncate the parked set above 300 parked open issues | A05:2021 / Elevation of Privilege | medium | **out-of-scope** |
| 4 | `dispatch-trace-leaf:128,214,235` | `child_is_office_hours_parked` has no precondition guard asserting `OFFICE_HOURS_PARKED` was populated | A04:2021 / Elevation of Privilege | low | **false-positive** |

### Rationale

- **#1 (false-positive)** — The premise is that `gh issue list --json number` exits `0` while emitting non-JSON, so `jq` then fails silently. `gh --json` parses the GitHub API itself and exits non-zero on any API/HTML error response; the existing `|| { …; exit 1; }` guard already converts that into a hard abort. When `gh` exits `0`, `parked_json` is well-formed JSON, so `jq -r '.[].number'` cannot fail. The unreachable-input fallback is not worth guarding against (see `.claude/rules/code-style.md` — defend against real failure paths, not impossible states; the real path is already guarded).
- **#2 (false-positive)** — GitHub's `.number` is always a JSON integer; `jq -r` emits a bare integer. The value is used only as an associative-array key and in `${OFFICE_HOURS_PARKED[$1]:-}` lookups — never in a command, `eval`, or unquoted word-split context. Both reviewers acknowledged there is no injection path. Adding a numeric regex would be defensive validation for a value that cannot carry metacharacters.
- **#3 (out-of-scope)** — `--limit 300` is an intentional, documented choice that matches `dispatch-select-target`'s open-issue cap (the parked set is a subset of open issues). The exploit requires 300+ simultaneously-parked **open** issues — a state in which the whole queue is already degraded — and a repo collaborator with write access. This is a pre-existing system-wide cap convention, not a vulnerability introduced by this diff. If the cap is ever revisited it should be revisited uniformly across the dispatch scripts, not in this PR.
- **#4 (false-positive)** — `child_is_office_hours_parked` is only ever called under `[[ "$MODE" == "queue" ]]` guards, and `OFFICE_HOURS_PARKED` is populated under the same `MODE == queue` gate. The concern is explicitly "future refactor only" and not reachable in the current code.

No diff to the working tree resulted from this review.

PR #1042, opened 2026-06-02
dispatch: heartbeat scans shared statements folder, files parse-job issues

natb1:
Closes #1023

Extends the #725 heartbeat (the `/dispatch-propagate` router tick) with a third config-driven scan pass that files content-hash-idempotent bank-statement parse-job issues. A sibling of `dispatch-jit-engine`.

## What changed

- **`statements` config type** in `dispatch-config-load` — schema validation arm mirroring `jit` (required `key`/`dir`/`repo`/`label`/`project`; optional `debounce`, `extensions`), plus config-load tests.
- **`dispatch-statements-scan`** — new engine. Per statements entry: debounce, dir-absent skip, idempotent label-create, non-recursive file enumeration filtered by extension, per-file sha256, GitHub content-hash dedup (`gh search issues` covers open AND closed), in-run seen-set, issue create + project-add. Full test block against a stubbed `gh`.
- **Step 2c wiring** in `dispatch-select-tick` (best-effort, prefixed `statements: ` pass-through), an example config, and SKILL.md / reference.md docs.

## Idempotency

Keyed on GitHub state, not a side file (the #755 principle): for each file the scan asks GitHub whether an issue already carries that sha256; a hit (open or closed) skips, no hit files one issue. The filed issue body carries only the filename and sha256 — never the statement contents. The local `tmp/dispatch-statements-state.json` debounce timestamp is a per-machine rate-limiter, not an idempotency record.

Runs in the router tick with the dispatch machine's own `gh` auth — no browser GitHub credential, no PAT.

## Test plan

- [ ] `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` passes (config-load + dispatch-statements-scan blocks).


natb1:

Scripts-only change (statements scan added to the heartbeat tick). Filing,
sha256 dedup (open-or-closed), issue-body content, and debounce behavior are
already covered by the stubbed unit tests in `test-dispatch-scripts.sh`, which
CI runs green. This QA pass deliberately did not re-run those; it exercised the
**live** script in the real repo plus the shipped artifacts — the gap the stubs
don't cover.

### Items walked (3)

| # | Item | Result |
|---|------|--------|
| 1 | Live heartbeat no-op without `statements.json` (exit 0, empty stdout) | **PASS** |
| 2 | Shipped `statements.example.json` validates against the loader schema | **PASS** |
| 3 | Tick pipeline integration: `bash -n` clean + select-tick invokes the scan | **PASS** |

**PASS 3 / FAIL 0 / SKIP 0**

### Deferred to office-hours

None — every QA item was machine-verifiable and no human-judgment item arose.

### Bugs

None.

natb1:

Reviewed the `dispatch-statements-scan` engine and its wiring. Three findings were fixed; one is deferred.

### Findings that were fixed

Fix commit: `08777c7f`

- **NUL-unsafe sort pipeline** (`dispatch-statements-scan`, file enumeration) — the `printf '%s\n' … | sort` collapse split filenames containing a newline into phantom array entries, which then failed `sha256sum` and retried forever (debounce never stamped). Replaced with a NUL-delimited `printf '%s\0' … | sort -z` guarded by a non-empty check.
- **Unbounded dedup search** (`dispatch-statements-scan`, content-hash dedup) — added `--limit 1` to the `gh search issues "$HASH"` query. Existence is all the dedup guard needs, and capping the result avoids any dependence on result-page size.
- **Case-sensitive label-exists guard** (`dispatch-statements-scan`, label create) — the `*"already exists"*` glob would misread a future `gh` capitalization change as a hard error, aborting the scan for that entry. Now lowercases the output (`${LABEL_OUT,,}`) before matching, so a benign duplicate-label condition stays benign.

### Deferred findings

- **`parse_duration` duplicated across four scripts** — the helper is copied verbatim into `dispatch-statements-scan`, `dispatch-jit-engine`, `dispatch-jit-calendar-import`, and `dispatch-select-target`. Consolidating it into `lib.sh` touches three scripts outside this PR's diff, so it is out of scope here. Filed as #1065.

natb1:
## Fixed
- Drop the no-op `$(( num * 1 ))` in `parse_duration`'s `s` branch (use `$num`): c0f538ee
- Drop the redundant `declare -a` prefix on `EXTS_LC=()` / `MATCHED=()` inside the per-entry loop — the plain assignment is the reset idiom (`declare -A SEEN_HASHES` left untouched): c0f538ee
- Document that the non-atomic state-file write is safe: a crash mid-write only loses the debounce timestamp (recovered next run) and never affects idempotency, which is keyed on GitHub state: c0f538ee
- Harden malformed `gh search issues` output — validate the response parses as JSON and raise a hard error for the entry instead of falling through to a possible duplicate filing; add Test 9 pinning the failure mode: c0f538ee

## Informational
- `state_set` re-reads and `jq`-parses the state file on each call (2N reads per scan) — acceptable at typical statement-folder sizes
- `gh issue create` output parsed with `tail -n1` plus an `^[0-9]+$` guard — defensive and correct
- `sort -z` for deterministic, newline-safe file ordering — correct
- The `gh search issues` bare-hash query has a theoretical false-positive, but the `` `sha256: <hash>` `` body format makes a collision practically impossible — pragmatically fine

## Dismissed
- Label-casing comparison in the `gh label create` error match — correct and deliberate; the inline comment explains the intent, no change needed

## Deferred
- Add rate-limit backoff for `gh search` over large statement folders (one serial query per file, no 429 handling) → #1066

natb1:
## Security review — PR #1042 (dispatch-statements-scan)

**Surface:** code (bash/config under `.claude/skills/dispatch-propagate/scripts/` + `reference.md`). No dependency changes (`package.json`/lock untouched) → no dependency audit. CodeQL: **0 open alerts** on the PR ref. App/Firestore-rules reviewers N/A (no application code or rules touched).

Reviewers run: input-validation, secrets, red-team, built-in `/security-review`, inline CodeQL. Findings de-duplicated and classified below.

### Fixed (commit `7c32900d`)

**1. Symlink in the shared folder published an out-of-folder file's hash** — `required`, high
- **Location:** `dispatch-statements-scan` (file enumeration, the `[[ -f "$f" ]]` filter)
- **OWASP:** A01:2021 Broken Access Control · **STRIDE:** Information Disclosure
- `-f` follows symlinks, so a symlink in the synced statements folder pointing at e.g. `~/.ssh/id_rsa` would be hashed and its content-hash filed to GitHub. The local-first threat model treats the sync provider that populates this folder as untrusted.
- **Fix:** enumerate with `[[ -f "$f" && \! -L "$f" ]]` — symlinks are skipped; statements are in-place regular files. Regression test added (Test 10).

**2. Control-character filename → prompt injection + stdout-protocol corruption** — `required`, high
- **Location:** `dispatch-statements-scan` (per-file loop, filename → issue title/body)
- **OWASP:** A03:2021 Injection · **STRIDE:** Tampering
- A filename with an embedded newline breaks out of the issue body's Markdown code span. The filed issue is later read by an autonomous parse agent (the budget-etl consumer, #1024), so attacker-controlled text after the newline is a prompt-injection vector; it also corrupts the line-oriented stdout protocol.
- **Fix:** reject any filename containing control characters (`*[$'\001'-$'\037']*`) with a hard error before hashing/filing. Regression test added (Test 11).

### Not fixed — out of scope for this PR's threat model

**3. Full-text dedup search poisoning** — `out-of-scope`, medium
- **Location:** `dispatch-statements-scan` (the `gh search issues "$HASH"` dedup)
- **OWASP:** A04:2021 Insecure Design · **STRIDE:** Denial of Service
- GitHub search is full-text, so an issue merely *mentioning* the hash (forged with the parse-job label, or coincidental) could be treated as a dedup hit and silently block a statement from ever being parsed. Forging a labeled issue requires GitHub repo-write access — outside the local-first threat model, which trusts GitHub state as ground truth (the untrusted actor is the sync provider, addressed by #1 and #2). A body-structured-match tightening was considered but not adopted: marginal benefit against an out-of-model actor, at the cost of an extra per-hit API call and dedup-test rework. No stable identifier → no follow-up issue filed; noted here.

### Not fixed — false positives

**4. `dir` config not path-constrained** — `false-positive`, low. `statements.json` is machine-local config; whoever can write it owns the machine. Same trust tier as env vars/CLI flags.

**5. `gh label create "$LABEL"` flag-lookalike** — `false-positive`, low. `$LABEL` comes from trusted local config; the commonly-suggested `gh label create -- "$LABEL"` "fix" is itself wrong (it would consume the following `--repo`/`--color` flags as positionals).

**6. `SEARCH_JSON`/`CREATE_OUT` echoed to stderr on error** — `false-positive`, low. `gh` does not echo auth tokens in normal (non-`--verbose`) error output, and the logged content (filename + non-secret sha256 dedup key) is non-sensitive by design — statement contents never enter any variable or log.

---
Verification: full `test-dispatch-scripts.sh` suite green — **1321/1321 passed**, including the two new regression tests.

PR #1051, opened 2026-06-03
dispatch: arm reseed timer on pace-curve budget pause (#1050)

natb1:
Closes #1050

## Problem

`dispatch-target-workers` returns a worker target of 0 in two distinct cases:

1. **Absolute cap-hit** — `used_percentage >= target` on the weekly (90%) or
   5-hour (50% floor) window.
2. **Pace-curve pause** — cumulative weekly usage runs ahead of the smooth
   cumulative-spend curve (`used_weekly >= W(x)`) while still far below the
   absolute weekly cap. Stage 2 collapses the 5h band to `F=0`, so the target
   is 0.

`dispatch-schedule-reseed` (the resume-from-stall mechanism, #725) only armed a
`dispatch-reseed-<resets_at>` timer for case 1. Its "neither cap hit → no-op"
branch silently no-op'd on a pace-curve pause, so a chain that drained its last
worker during a pace pause stayed dead with no scheduled restart — even though a
pace pause is transient and self-clearing (the curve `W` rises to meet flat
`used_weekly` within the hour).

## Change

Make `dispatch-target-workers` the single authority for *when the target next
becomes ≥ 1*, reusing its own Stage-1 curve math so the two cannot diverge.

- **`dispatch-target-workers`** gains a `--reopen-at` mode: it solves
  `W(x) = used_weekly` by bisection over the monotonic Stage-1 curve and prints
  the pace-curve crossing epoch, or `none` when the pace curve is not the
  blocker (target already ≥ 1, transient 5h fill, or missing weekly anchor).
  Count mode (the only existing caller path) is unchanged — verified
  byte-identical across 448 input combinations.
- **`dispatch-schedule-reseed`**'s no-absolute-cap branch now consults
  `--reopen-at` and arms a `dispatch-reseed-<fire>` timer at the crossing, with
  a short-delay floor (`NOW + 300s`, the #979 pattern) when the crossing is
  at/before now. The absolute-cap path is unchanged (no regression). The unit
  name stays `dispatch-reseed-<fire>` so the #1010 continuation invariant's
  `dispatch-reseed*` pending-timer detection still recognizes pace-path timers.
- **Tests** in `test-dispatch-scripts.sh` cover `--reopen-at` (pace pause →
  crossing, under-pace/missing-anchor → `none`) and the reseed pace path
  (stubbed future/none/past cases plus an end-to-end real-budgeter case). Suite
  passes 1161/1161.
- **`reference.md`** documents both stall classes and the curve-crossing reopen.

## Acceptance criteria

- [x] Budgeter emits the earliest target-≥-1 epoch: `resets_at` for an absolute
      cap-hit, the `W(x)=used_weekly` crossing for a pace-curve pause, `none`
      when already ≥ 1.
- [x] Pace-curve pause arms a `dispatch-reseed-<fire>` timer at the crossing.
- [x] Absolute cap-hit still arms at the blocking window `resets_at` (no
      regression).
- [x] Crossing at/before now arms a short-delay timer, not a past-dated one.
- [x] `reference.md`'s #725 section documents the pace-curve `F=0` stall class
      and its curve-crossing reopen.
- [x] `test-dispatch-scripts.sh` covers all four reseed cases.


natb1:
## QA — autonomous pass ✅

Non-browser PR (dispatch scripts + reference doc). The QA pass covers the
acceptance criteria as machine-verifiable checks: the full `test-dispatch-scripts.sh`
suite plus direct exercise of the real scripts.

**Items walked: 6 — PASS 6 / FAIL 0 / SKIP 0**

| # | Item | Result |
|---|------|--------|
| 1 | Full `test-dispatch-scripts.sh` suite (incl. all #1050 tests) | **PASS** — 1206/1206 |
| 2 | `--reopen-at` on a pace-curve pause emits a numeric crossing epoch | **PASS** — `1026175`, strictly in `(1000000, 1302400)` |
| 3 | `--reopen-at` returns `none` when pace curve isn't the blocker (under-pace + missing anchor) | **PASS** — `none` in both cases |
| 4 | `--reopen-at` rejects an unrecognized mode arg | **PASS** — exit 2 + clear stderr |
| 5 | Count mode unchanged (pace pause→0, under-pace→≥1, missing telem→1 fallback) | **PASS** — `0` / `8` / `1` |
| 6 | `dispatch-schedule-reseed`: crossing timer on pace pause, short-delay floor on past crossing, no-op when pace permits | **PASS** — suite tests 15–18 (stub + real budgeter) |

Direct script exercise (beyond the in-suite harness) confirmed against the real
`dispatch-target-workers` in the worktree.

**Acceptance criteria mapping**
- Earliest reopen epoch (resets_at for cap-hit / curve-crossing for pace pause / `none` when ≥1) — items 2, 3, 5.
- Pace-curve pause arms a `dispatch-reseed-<fire>` timer at the crossing — item 6 (tests 15, 18).
- Absolute cap-hit still arms at `resets_at` (no regression) — suite reseed tests (no failures).
- Crossing at/before now → short-delay floor — item 6 (test 17).
- Chain resumes when target ≥ 1 — covered by count-mode behavior (item 5) + reseed timer.
- `reference.md` documents the pace-curve `F=0` stall class and curve-crossing reopen — present in the diff (new "Pace-curve pause" + "Short-delay floor" sections).
- `test-dispatch-scripts.sh` covers all four reseed scenarios — present and passing (item 1).

**Deferred to office-hours:** none — every item was machine-verifiable.

**Bugs found:** none.

natb1:

### Findings that were fixed
- **`dispatch-schedule-reseed`** — the no-cap-hit branch conflated the budgeter's `none` sentinel (pace curve genuinely permits) with an empty string (budgeter failed: missing binary or non-zero exit), printing "pace curve permits" in both cases. A failed budgeter therefore emitted a healthy-looking diagnostic, obscuring root-cause analysis of a stalled chain. Split into a `none` case (unchanged "permits" message) and an explicit "budgeter returned unexpected result" case. Fixed in `50627f9c`.
- **`test-dispatch-scripts.sh`** — added Test 17b covering the boundary where the pace-curve crossing epoch equals NOW exactly. The short-delay floor applies on `RESEED_AT <= NOW` (inclusive of equality); a future tightening to strict `<` would silently regress this boundary with no failing test. Test pins `dispatch-reseed-10300 at 10300`. Fixed in `50627f9c`.

### Informational findings
- **`dispatch-schedule-reseed`** — the reopen path with `FIVEH_PRESENT=0` is not directly exercised in this script's tests. The logic is correct (reopen mode ignores the 5-hour window entirely), so this is a minor coverage gap with no regression risk — noted for reference only.

### Disregarded findings
- **`dispatch-target-workers`** count-mode `remaining` "computed twice": the outer computation is the early-exit guard (`remaining <= 0 → print 0`), semantically distinct from `W_of` at that point. The duplication is load-bearing; collapsing it would change behavior. Not an improvement.

Full suite: 1208/1208 passing.

natb1:
## Fixed
- Correct the reopen-crossing clamp comment in `dispatch-target-workers` — it implied the clamp can trigger, but `W_of` is monotonic increasing and the bisect only runs when `W_of(now) <= used_weekly`, so `fire >= now` always and the clamp is unreachable: 4b38300a
- Pin the budgeter's own `rate_limits` path in the reseed test harness (`sr_setup` / `sr_write_rl` / `sr_teardown`) so reseed tests are isolated from the real `~/.local/share/.../rate_limits.json` even if the budgeter's read/override order ever changes (the reseed script passes telemetry via per-field env vars, which take precedence, so the path is never read in practice — defensive): 4b38300a
- Add a test that an unexpected (non-numeric, non-`none`) `--reopen-at` string makes `dispatch-schedule-reseed` no-op with exit 0 and a stderr diagnostic, making the numeric/`none`/else triad explicit: 4b38300a

## Informational
- `W_of`'s awk-local `remaining` shares a name with the count-mode `remaining` assigned in the `BEGIN` block. These are distinct awk scopes (the function parameter shadows), so there is no shared state or bug — purely a readability note for a future reader.
- A `hw == 0` short-circuit in the bisect (`print now; exit`) would save the 60 iterations that otherwise converge to `now`, but it is not a correctness issue: the bisect already lands on `now`, and the shell-layer short-delay floor handles the at/before-now case. Left as-is.

## Dismissed
_None._

## Deferred
_None._

natb1:
## Security review — pace-curve reseed (security phase)

**Surface classification:** `code` (deps=false, app_or_rules=false). Changed files are the dispatch chain's bash/awk scripts plus `reference.md`. No application source, Firestore rules, or dependency manifests changed.

**Reviewers run:** input-validation, secrets, red-team, built-in `/security-review` (the four unconditional-on-code reviewers). App-domain reviewers (auth, data-exposure, Firebase) skipped — `app_or_rules=false`. Dependency audit skipped — `deps=false`. CodeQL fetched inline.

**Inline scans:**
- **CodeQL** — no open alerts for the PR ref (clean).
- **Dependency audit** — not run (`package.json` / `package-lock.json` unchanged).

**Clean reviewers:** secrets, built-in `/security-review` — no findings.

### Findings and dispositions

All three findings are **not fixed** — each is a false-positive / out-of-scope within the project's threat model. No `required` findings; no code changes.

| # | Finding | Confidence | Disposition |
|---|---|---|---|
| A | `DISPATCH_SCHEDULE_RESEED_TARGET_WORKERS_CMD` lets an env-var override redirect the budgeter to an arbitrary executable (`dispatch-schedule-reseed`). OWASP A03 / STRIDE Elevation of Privilege. | medium | **false-positive** |
| B | `REOPEN_AT` echoed unsanitized into a stderr diagnostic in the "unexpected result" branch — terminal/log control-char injection (`dispatch-schedule-reseed`). OWASP A03/A09 / STRIDE Tampering. | low | **false-positive** |
| C | Reopen-mode bisection could arm a timer at `resets_weekly` if the curve never reaches `used_weekly`, pinning the chain to weekly restarts via a hostile `target-workers.json` (`dispatch-target-workers`). OWASP A05 / STRIDE Denial of Service. | low | **false-positive / out-of-scope** |

**Rationale:**

- **A** — `DISPATCH_SCHEDULE_RESEED_TARGET_WORKERS_CMD` is a documented env-var test override, structurally identical to the pre-existing `DISPATCH_SCHEDULE_RESEED_SYSTEMD_RUN_CMD` and `..._RATE_LIMITS_PATH` overrides this script already ships. Environment variables sit inside the dispatch chain's trust boundary; an attacker who can set them before the script runs already has session-level code execution. Consistent-with-convention, not a new attack surface. The `REOPEN_AT` output is then validated against `^[0-9]+$` / `none` before any arithmetic or unit-name use.

- **B** — only reachable when the budgeter emits output that is neither numeric nor `none`. In production the budgeter is the trusted sibling `dispatch-target-workers`; reaching this branch presupposes Finding A's precondition (a hostile budgeter binary). Terminal/log-control-char injection is outside the threat model — the built-in `/security-review` scan explicitly excludes log spoofing. The value never reaches stdout, a log file, or any arithmetic/exec context.

- **C** — at `t = resets_weekly` the terminal envelope gives `W_of(resets_weekly) = terminal` (default 100) and the smooth term equals `target_weekly` (default 90), so `W_of(resets_weekly) = 100`. `used_weekly` is a percentage ≤ 100, and a pace-pause requires `used_weekly < target_weekly` (below the absolute cap), so a crossing always exists strictly within `(now, resets_weekly]` — the bisection converges to a real crossing, never the boundary. The "curve never catches up" premise requires misconfiguring `terminal`/`target_weekly` below `used_weekly` (config-write access = already trusted), and the curve config is validated by #930.

### Conclusion

No required fixes. Input validation across the new surface is sound: `REOPEN_AT`, `SHORT_DELAY`, and `MODE` are all validated (`^[0-9]+$` / `case`) before reaching arithmetic, `env`, or `systemd-run` unit-name / `--on-calendar` construction. Marking the PR ready.

PR #1057, opened 2026-06-03
dispatch: synthesize a stable session id for the headless dispatch-tick (#1054)

natb1:
Closes #1054

PR #1081, opened 2026-06-03
dispatch: PID-backed sentinel so the selection lock serializes concurrent headless ticks (#1068)

natb1:
Closes #1068

## Problem

PR #1057 made the headless `dispatch-tick` synthesize a stable holder id
(`headless:${INVOCATION_ID:-$$}`) so it can acquire the dispatch selection lock
under systemd. But `dispatch-acquire-lock`'s `resolve_holder_state` resolves a
recorded holder's liveness **only** through the `claude agents --json` daemon
registry, which tracks model sessions, not bash processes. A synthetic
`headless:*` id never appears there, so any other concurrent tick that reads a
**live** headless holder queries the registry, finds it absent, classifies the
holder **dead**, and reclaims the lock mid-selection. Two ticks can then both
pass the lock and select the same target — and for a fresh `help wanted` issue
with no worktree yet, duplicate-spawn. A human-typed `/dispatch` (real UUID, in
the registry) likewise reclaims a running headless tick's lock.

Same-user concurrency-correctness defect — no adversary, no trust boundary.

## Fix

**Unit 1 — PID-backed sentinel.** The headless tick maintains a sentinel file
containing its own PID for its lifetime; `resolve_holder_state` treats a
`headless:*` holder as live iff the sentinel exists **and** its stored PID is
still alive (`kill -0`). Storing the PID (rather than relying on bare existence)
means a `SIGKILL`-orphaned sentinel still resolves **dead**, preserving the
self-healing reclaim the daemon path already gives real sessions.

- `lib.sh`: two shared helpers so the tick and the lock derive identical paths
  from one source of truth — `dispatch_lock_file` (canonical lock-file path) and
  `headless_sentinel_path` (slugged `dispatch-tick-<token>.live` beside the lock).
- `dispatch-acquire-lock`: a leading `headless:*` branch in `resolve_holder_state`
  that resolves liveness via the sentinel; Step 1 now calls `dispatch_lock_file`.
- `dispatch-tick`: writes `$$` to the sentinel on start and installs a
  `trap … EXIT INT TERM` that removes it on clean exit or systemd stop.

**Unit 2 — synthetic-id hardening.** Validate `INVOCATION_ID` against the systemd
hex shape (`^[0-9a-fA-F-]+$`) before admitting it into the holder id — a polluted
value with a newline would truncate the recorded holder or break the
`awk -F'\t'` comparison and the sentinel path. Replace the low-entropy bare-PID
fallback with `openssl rand -hex 16` (PID remains the last-resort fallback).

## Verification

`bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` passes
fully, including new acquire-lock headless-liveness cases (live PID → busy,
absent sentinel → acquired, dead PID → acquired) and tick-sentinel lifecycle
tests (written mid-run, removed by the trap after).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-acquire-lock
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-acquire-lock
@@ -35,6 +35,20 @@
 #     the lock is never stolen from a possibly-held foreign holder when the
 #     daemon cannot be queried.
 #
+# Headless holders bypass the registry (#1068). The systemd-launched dispatch-
+# tick has no real Claude session, so it synthesizes a `headless:<token>` holder
+# id (token = INVOCATION_ID or PID) — a bash string that never appears in
+# `claude agents --json`. Resolving its liveness through that registry would
+# always classify it dead, letting a concurrent tick steal a LIVE headless
+# holder's lock mid-selection (the duplicate-spawn defect of #1068). Instead the
+# tick writes a PID sentinel — `<lock-file dir>/dispatch-tick-<slug>.live`
+# containing its own PID — for the tick's lifetime, and resolve_holder_state
+# treats a `headless:*` holder as live iff that sentinel exists AND its recorded
+# PID is still alive (`kill -0`). PID-in-sentinel (not bare existence) keeps the
+# self-healing reclaim: a SIGKILL leaves a stale sentinel whose PID is gone, so
+# the next tick reads it dead and reclaims. The sentinel sits in the lock file's
+# own directory, so a concurrent tick in any worktree resolves the same path.
+#
 # Release is implicit and self-healing: when the session ends its sessionId
 # stops appearing in `claude agents --json`, so the next tick's read-check-
 # write reclaims the lock automatically. `--release` is an explicit optional
@@ -153,6 +167,23 @@ WAIT_INTERVAL="${DISPATCH_LOCK_WAIT_INTERVAL:-5}"
 # exit client, but the cost of defending against a future fork is one redir.
 resolve_holder_state() {
   local sid="$1" out rows match
+  # Headless holders (the systemd-launched dispatch-tick's synthetic
+  # `headless:<token>` id) never appear in the `claude agents --json` registry —
+  # it tracks model sessions, not bash processes. Resolve their liveness through
+  # the PID sentinel the tick writes alongside the lock file (#1068): live iff
+  # the sentinel exists AND its recorded PID is still alive. A SIGKILL leaves a
+  # stale sentinel whose PID is gone, so it resolves DEAD and the lock self-heals.
+  # No cwd is emitted (matches the live-with-empty-cwd convention — marker check
+  # stays strict), and the daemon is never queried for these ids.
+  if [[ "$sid" == headless:* ]]; then
+    local sentinel pid=""
+    sentinel=$(headless_sentinel_path "$sid" "$LOCK_FILE")
+    [[ -f "$sentinel" ]] || return 1
+    IFS= read -r pid <"$sentinel" 2>/dev/null || true
+    [[ "$pid" =~ ^[0-9]+$ ]] || return 1
+    kill -0 "$pid" 2>/dev/null || return 1
+    return 0
+  fi
   if ! out=$("${CLAUDE_AGENTS_CMD:-claude}" agents --json 2>/dev/null 9>&-); then
     return 0
   fi
@@ -200,18 +231,13 @@ marker_names_holder() {
 }
 
 # ---- Step 1: resolve the lock file ------------------------------------------
-# An explicit DISPATCH_LOCK_FILE is authoritative and bypasses the git lookup —
-# tests rely on this. Otherwise the lock lives at the shared project-root tmp/
-# (not a per-worktree tmp/) so concurrent ticks in different worktrees contend
-# on the same file. resolve_project_root (lib.sh) returns the project root in
-# both classic (.git) and bare (.bare) layouts.
-if [[ -n "${DISPATCH_LOCK_FILE:-}" ]]; then
-  LOCK_FILE="$DISPATCH_LOCK_FILE"
-else
-  PROJECT_ROOT=$(resolve_project_root) \
-    || { echo "ERROR: not in a git repo and DISPATCH_LOCK_FILE is unset" >&2; exit 2; }
-  LOCK_FILE="$PROJECT_ROOT/tmp/dispatch.lock"
-fi
+# dispatch_lock_file (lib.sh) honors an explicit DISPATCH_LOCK_FILE (tests rely
+# on this) and otherwise resolves the shared project-root tmp/dispatch.lock (not
+# a per-worktree tmp/) so concurrent ticks in different worktrees contend on the
+# same file. The tick's headless-liveness sentinel resolves through the same
+# helper, keeping the lock script and the tick on one lock-file directory (#1068).
+LOCK_FILE=$(dispatch_lock_file) \
+  || { echo "ERROR: not in a git repo and DISPATCH_LOCK_FILE is unset" >&2; exit 2; }
 
 # ---- Step 2: identify our own session ---------------------------------------
 # CLAUDE_CODE_SESSION_ID is set in every real .claude session's environment.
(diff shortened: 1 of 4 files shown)


natb1:
## QA pass — autonomous (#1068)

Scripts-only change (no browser/app surface), so the **non-browser QA path**
applied. The fix gives headless `dispatch-tick` selection-lock holders a
daemon-independent liveness signal (a PID sentinel beside the shared lock file),
so concurrent ticks serialize correctly instead of stealing a live holder's lock
mid-selection and duplicate-spawning.

CI's `unit-tests.yml` already runs the full `test-dispatch-scripts.sh` suite
(including the new #1068 sentinel tests) and is green. The QA walkthrough below
adds an **operator-level** confidence check: the real `dispatch-acquire-lock`
binary driven against a throwaway lock file, distinct from the unit harness's
fakes.

### Items walked (all machine-verifiable)

| # | Scenario | Expected | Result |
|---|---|---|---|
| 1 | Live-PID sentinel for a `headless:*` holder | caller sees `busy`; lock **not** stolen | **PASS** |
| 2 | Dead-PID sentinel (SIGKILL'd tick) | caller `acquired`; stale lock reclaimed | **PASS** |
| 3 | No sentinel (registry-invisible holder) | caller `acquired`; lock reclaimed | **PASS** |

**Counts:** PASS 9 assertions / 3 items · FAIL 0 · SKIP 0
**Deferred to office-hours:** none (no needs-human-judgment items)
**Bugs found:** none

Item 1 is the core fix — a live headless holder is now respected rather than
read dead and stolen. Items 2–3 confirm self-healing still works (a crashed or
sentinel-less holder is correctly reclaimed, so no permanent lock stall).

_An initial walkthrough run flagged Item 1; root-caused to a QA-harness artifact
(the Bash tool's zsh shell escaped `$!`, leaving a non-PID sentinel) and not a
product defect — re-run with a guaranteed-live PID passed cleanly._

natb1:

### Findings that were fixed

Applied in fix commit `82ea9a83`:

- **`dispatch-tick` (sentinel write):** A failed headless-sentinel write (`printf ... >"$SENTINEL"`) was silent — a disk-full or permission error would leave the tick with no liveness signal, letting a concurrent tick classify the holder dead and reclaim the lock mid-selection, the exact defect this PR prevents. Now emits a `WARNING` to stderr (journald-visible) on write failure; the tick still continues best-effort, consistent with the outer `dispatch_lock_file` guard.
- **`test-dispatch-scripts.sh` (headless tests A and C):** Two tests hardcoded the sentinel path (`"$(dirname "$DISPATCH_LOCK_FILE")/dispatch-tick-tok123.live"`) instead of calling the shared `headless_sentinel_path`. Replaced both with `headless_sentinel_path` invocations so the tests stay coupled to the function's slug format and won't silently diverge if it changes.

### Informational findings

- **PID reuse after `SIGKILL`:** If a tick is `SIGKILL`'d after writing the sentinel but before its EXIT trap fires, a stale sentinel with the dead PID remains; if the OS immediately reuses that PID for another same-user process, `kill -0` reads it as live and the lock stalls until that unrelated process exits. This is inherent to PID-based liveness and is an acknowledged design trade-off in the code comment — fixing it would require a fundamentally different mechanism (pidfd or systemd notification socket). PID reuse within the selection window is rare on Linux (sequential allocation, high `pid_max`). No change made.

natb1:
## Fixed
_None._

## Informational
- PID-in-sentinel design preserves the self-healing reclaim (SIGKILL-orphaned sentinel resolves dead on next `kill -0`) — sound choice.
- `dispatch_lock_file` / `headless_sentinel_path` lib.sh helpers give tick and lock one source of truth for the lock-dir path — correct abstraction.
- Sentinel content is written with a trailing newline (`printf '%s\n' "$$"`) and read with `IFS= read -r`, which strips it — consistent and correct; noted for future maintainers.
- EXIT/INT/TERM trap captures `$SENTINEL` from top-level scope (global), so it sees the right value at exit — safe.
- Empty/partial sentinel falls through to the `[[ "$pid" =~ ^[0-9]+$ ]]` guard and resolves dead — correct.
- `INVOCATION_ID` regex `^[0-9a-fA-F-]+$` admits hyphens as future-proofing for a UUID-style format; the comment could name that intent explicitly (optional).
- On a sentinel-write failure the trap is not installed; a one-line comment could make that explicit — low priority.
- `mkdir -p "$(dirname "$SENTINEL")" 2>/dev/null || true` double-silent-failure is acceptable under the best-effort framing.
- Race analyses (sentinel write vs. lock acquisition, sentinel removal vs. concurrent read, `openssl` availability, INT/TERM-then-EXIT double removal) all resolve correctly.
- Test coverage is thorough: parametric acquire-lock liveness cases, tick-lifecycle sentinel observation, polluted/absent `INVOCATION_ID` cases.

## Dismissed
- Test uses PID `2147483647` as the "dead" PID — reviewer: "No change needed"; reliable in practice for `kill -0`.
- No dedicated SIGTERM-mid-run sentinel-removal test — "acceptable omission"; the EXIT trap fires after TERM in bash, so behavior is covered.

## Deferred
_None._

natb1:

**Changed surface:** `code` (bash scripts only — the dispatch selection-lock mechanism). `app_or_rules=false`, `deps=false`.

**Reviewers run:** input-validation, secrets scan, red team, built-in `/security-review`, CodeQL (inline). App-domain reviewers (auth/access, data exposure, Firebase) and the dependency audit were skipped — no application code, Firestore rules, or `package.json`/`package-lock.json` in the diff.

Threat model is **same-user** (every tick is a same-user systemd unit; no external adversary) — findings are weighted accordingly.

### Findings

| # | Source | Finding | Confidence | Disposition |
|---|---|---|---|---|
| 1 | red team | `dispatch-tick` warn-and-continue on sentinel write failure silently reintroduces the #1068 duplicate-spawn defect | high | **Fixed** — `ce2febb2` |
| 2 | red team | PID reuse after SIGKILL → false-live read → bounded (≤300s) self-healing lock stall | medium | Out-of-scope |
| 3 | red team / input-validation | `INVOCATION_ID` regex has no length bound; over-long value → `ENAMETOOLONG` on the sentinel filename | low | Addressed indirectly by #1 |
| 4 | input-validation | `INVOCATION_ID` regex admits hyphen-only values → leading-hyphen sentinel filename | low | False-positive |
| 5 | input-validation | `openssl rand -hex 16` output not re-validated | low | False-positive |

Secrets scan, the built-in `/security-review` scan, and CodeQL (no open alerts on the PR ref) reported nothing. The dependency audit did not run (`deps=false`).

### Finding 1 — Fixed (required)

`dispatch-tick:170-171` printed a WARNING and **continued** when the headless-liveness sentinel write failed, with the cleanup trap installed unconditionally afterward. When the lock-file path resolves but the sentinel write fails — e.g. an over-long token yielding `ENAMETOOLONG` while the short `dispatch.lock` stays writable — the tick would acquire the selection lock with **no** sentinel. A concurrent tick then reads this live `headless:*` holder as dead via `resolve_holder_state` and reclaims the lock mid-selection: the exact duplicate-spawn defect #1068 exists to close, silently reintroduced.

This contradicts the PR's own intent and `.claude/rules/code-style.md` ("prefer clear errors over defensive fallbacks"). Fix: fail clear with `exit 2` on sentinel write failure (the trap, now installed before the write, clears any partial sentinel on abort). The dispatch chain self-heals on the next tick. The lock-file-path-unresolvable branch keeps its silent skip — `dispatch-acquire-lock` itself exits 2 there, so that tick never reaches selection. A focused test (`sentinel-write-fail`) covers the new behavior; full suite 1387/1387.

### Finding 2 — Out-of-scope (acknowledged design limitation)

A SIGKILL'd tick leaves a stale PID sentinel; if the OS recycles that PID to an unrelated process before the next liveness check, the dead holder reads live and a concurrent tick waits up to `DISPATCH_LOCK_WAIT_TIMEOUT` (300s default) before the recycled PID's process exits and the lock self-heals. This is a **bounded availability** concern, not a breach or duplicate spawn, and only under same-user assumptions. A PID-reuse-immune design (flock-on-sentinel, or PID + `/proc/<pid>` start-time) is a larger change beyond this PR's scope; #1068 frames PID-based liveness as the intended mechanism. Left for a possible future hardening pass.

### Findings 3-5 — not required

- **3:** With Finding 1 fixed, an over-long `INVOCATION_ID` now fails clear (clear error, tick aborts) rather than degrading the lock silently. The regex itself implements exactly the validation guard #1068's body recommended.
- **4:** A hyphen-only token is fully neutralized — every use is double-quoted and the filename slug runs through `tr -c '0-9A-Za-z._-' '_'`, so no option-injection or path escape is possible. Cosmetic only.
- **5:** `openssl rand -hex 16` emits hex by contract; the `tr` slug neutralizes any anomalous output, and the token is a liveness identifier, not a secret. Not a vulnerability.

No security follow-up issues were filed — no CodeQL alert or npm advisory met the meaningfulness threshold (CodeQL returned no open alerts; the dependency audit did not run).

PR #1084, opened 2026-06-03
dispatch: fan-out exclusion — re-surfaced targets no longer halt the loop (#1062)

natb1:
Closes #1062

## Problem

The autonomous fan-out loop in `dispatch-materialize-spawn` fills `gap` slots by
calling the stateless `dispatch-select-target` once per iteration. A `SEEN` guard
stopped the **entire** loop the moment selection re-returned any already-processed
target. Because `dispatch-select-target` is stateless and deterministic — it
always returns the single highest-priority target and cannot be told to exclude
one — a target that re-surfaces ended the whole fan-out even when distinct
startable leaves remained lower in the priority order.

A target re-surfaces via two common mechanisms, and the fix is agnostic to which:

1. **Registration-visibility lag** — a just-spawned worker is verified-registered
   but not yet reflected in the next selection's `worktree_has_live_session`
   daemon query, so its leaf re-surfaces.
2. **Fast self-close leaving an orphan worktree** — a worker whose phase completes
   in seconds self-closes before the next selection. Its worktree is now an orphan
   (on disk, no live session), which by design is a reuse signal, not a conflict —
   so its target re-surfaces.

## Fix

Turn the `SEEN` set into a true **exclusion** mechanism. The loop passes its
already-processed target numbers to `dispatch-select-target`, which threads them
into its queue scan and into `dispatch-trace-leaf`, so selection returns the
*next* distinct target instead of re-returning a seen one. The loop then stops
only on a genuine `empty` (or a deferred selection / hard error), never on a
re-selection.

The exclusion wire format is variadic integers (`--exclude 839 720 508`), so the
caller expands `--exclude "${!SEEN[@]}"` directly.

### Unit 1 — `dispatch-trace-leaf`
Accepts `--exclude <n>...`; the topological-leaf branch refuses to return an
excluded leaf, so the descent advances past it to the next startable
sibling/leaf. One change point covers both the root and every descended leaf.

### Unit 2 — `dispatch-select-target`
Accepts `--exclude <n>...`; applies it during the queue scan **before** bucket
recording (PR rows filtered by their branch issue-prefix before `FIRST_PR` is
populated, so the next-oldest PR in an already-tapped bucket surfaces) and threads
it into the leaf trace (so a multi-leaf subtree whose first leaf is excluded
advances to the next startable leaf).

### Unit 3 — `dispatch-materialize-spawn`
The fan-out loop passes its `SEEN` set as `--exclude`. The `no further distinct
targets` normal stop is removed; in its place is a defensive backstop that breaks
only if selection re-returns a `SEEN` target despite the exclusion (a selector
contract violation) — unreachable in normal operation, present to prevent an
infinite loop.

## Tests

`test-dispatch-scripts.sh` covers, across all three scripts:
- a fan-out where selection would otherwise re-return an already-spawned target —
  via a lagged-live worktree and via an orphan (self-closed) worktree — instead
  continues to the next distinct leaf and fills the gap;
- a multi-leaf subtree whose first leaf is excluded drains to the next startable
  leaf;
- the loop terminates cleanly on a genuinely empty frontier;
- the defensive backstop fires (no hang) if the selector violates its contract;
- the `no further distinct targets` string never appears on any path.

Full suite: 1394/1394 passing.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-materialize-spawn
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-materialize-spawn
@@ -388,13 +388,22 @@ if [[ "$MODE" == "explicit" ]] || (( GAP <= 1 )); then
 fi
 
 # Fan out: fill up to GAP slots in one held-lock run. Target 1 is the passed-in
-# N; targets 2..GAP come from successive dispatch-select-target calls. Each
-# spawned worker registers before the next selection (#945/#905), so the next
-# dispatch-select-target skips its now-live-owned worktree and yields a distinct
-# target. A SEEN guard additionally stops the run if selection ever returns an
-# already-processed target (a non-spawning outcome that did not remove the
-# target from the queue), bounding the loop. Per-target park/skip outcomes are
-# recorded and the loop continues; the lock is released ONCE at the end.
+# N; targets 2..GAP come from successive dispatch-select-target calls. The loop
+# passes its SEEN set as `--exclude` to dispatch-select-target so selection
+# yields the NEXT distinct target instead of re-returning an already-processed
+# one (#1062). A re-return would otherwise happen via two mechanisms: (1)
+# registration-visibility lag — a just-spawned worker is verified-registered but
+# not yet visible to the next selection's liveness query (#945/#905), so its
+# leaf re-surfaces; or (2) a fast self-close leaving an orphan worktree, which
+# selection treats as a reuse signal (not a conflict) and does not skip. The
+# exclusion defeats both: an excluded target is filtered during the queue scan
+# (and threaded into the leaf trace), so the loop advances to the next distinct
+# startable leaf. The loop stops only on a genuine `empty`, a deferred selection
+# (main-broken / jit-reminder), or a hard error — plus a defensive backstop if
+# selection ever re-returns a SEEN target despite the exclusion (a selector
+# contract violation), to prevent an infinite loop. Per-target park/skip
+# outcomes are recorded and the loop continues; the lock is released ONCE at the
+# end.
 declare -A SEEN=()
 spawned=0
 stop_reason="gap filled"
@@ -437,7 +446,9 @@ while :; do
       ;;
   esac
   (( spawned >= GAP )) && break
-  SELECTED=$("$SCRIPT_DIR/dispatch-select-target") || true
+  exclude_args=()
+  (( ${#SEEN[@]} > 0 )) && exclude_args=(--exclude "${!SEEN[@]}")
+  SELECTED=$("$SCRIPT_DIR/dispatch-select-target" "${exclude_args[@]}") || true
   case "$SELECTED" in
     pr\ *)
       # Extract the branch from "pr <pr_num> <branch> <phase>" without clobbering $@.
@@ -451,8 +462,14 @@ while :; do
        echo "dispatch-materialize-spawn: dispatch-select-target emitted unexpected '$SELECTED' in fan-out" >&2
        exit 2 ;;
   esac
+  # Defensive backstop (#1062): the loop passes SEEN as --exclude, so
+  # dispatch-select-target must never return an already-processed target. If it
+  # somehow does, that is a selector contract violation — break rather than spin
+  # forever. This stop is unreachable in normal operation; the loop's real
+  # terminators are `empty`, a deferred selection, and hard errors above.
   if [[ -n "${SEEN[$cur_n]:-}" ]]; then
-    stop_reason="no further distinct targets"
+    echo "dispatch-materialize-spawn: dispatch-select-target re-returned already-processed target #$cur_n despite --exclude (contract violation); stopping fan-out to avoid an infinite loop" >&2
+    stop_reason="selection re-returned an excluded target (unexpected)"
     break
   fi
 done
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -136,8 +136,8 @@ fi
 
 # Build the leaf-trace passthrough once: the excluded numbers thread into
 # dispatch-trace-leaf (#1062) so a multi-leaf subtree descends past an excluded
-# leaf. Only constructed when non-empty so "${EXCLUDE_ARGS[@]}" expands to
-# nothing under set -u when no exclusion was requested.
+# leaf. Only constructed when non-empty so the call site passes no --exclude
+# flag at all when no exclusion was requested.
 EXCLUDE_ARGS=()
 (( ${#EXCLUDED[@]} > 0 )) && EXCLUDE_ARGS=(--exclude "${!EXCLUDED[@]}")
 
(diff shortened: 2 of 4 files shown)


natb1:
## QA pass — autonomous (#1062)

**Path:** non-browser — the change is confined to dispatch shell scripts
(`dispatch-materialize-spawn`, `dispatch-select-target`, `dispatch-trace-leaf`)
plus their test suite. No frontend surface, so the walkthrough is machine-verifiable
checks rather than a browser demo.

### Items walked

| # | Item | Result |
|---|---|---|
| 1 | Full `test-dispatch-scripts.sh` regression | **PASS** (1394/1394) |
| 2 | `dispatch-trace-leaf --exclude` argument parsing | **PASS** |
| 3 | `dispatch-select-target --exclude` argument parsing | **PASS** |
| 4 | `bash -n` syntax check (3 scripts + test) | **PASS** |

**PASS: 4 · FAIL: 0 · SKIP: 0**

### Acceptance-criteria coverage (all green in the suite)

- **(a)** registration-lag re-surface → fan-out still fills the gap (`fanout-lag`)
- **(a)** orphan self-closed re-surface → fan-out still fills the gap (`fanout-orphan`)
- **(b)** multi-leaf subtree, first leaf excluded → drains to the next startable leaf
  (`fanout-subtree`; trace `701→702`; select `5500→5501`)
- **(c)** genuinely empty frontier → loop stops cleanly (`fanout-empty`)
- Backstop: a selector contract violation breaks the loop and names the violation (`backstop`)
- `no further distinct targets` stop is no longer reachable in normal operation

### Deferred to office-hours

None — no needs-human-judgment items.

### Bugs

None.

natb1:

Ran `/code-review max --fix` over the branch diff. Two findings, both fixed in-tree.

### Findings that were fixed

- **`dispatch-select-target:139` — inaccurate guard comment.** The comment claimed the `EXCLUDE_ARGS` empty-array guard exists so the array "expands to nothing under `set -u`", but empty indexed arrays are nounset-safe on bash 4.4+. Corrected the rationale to describe the actual behavior: the guard ensures the call site passes no `--exclude` flag at all when no exclusion was requested. (commit ddeb5c4c)
- **`dispatch-trace-leaf:92` — stray args after `--exclude` silently dropped.** The integer-consuming loop stopped at the first non-integer token but nothing checked for leftover positional args, so a typo'd flag after the `--exclude` run was silently ignored. Added a usage-error guard that exits non-zero on any remaining args. No current caller triggers this (the only caller passes pure integers), but a future caller with a typo now gets a clear diagnostic. (commit ddeb5c4c)

No informational, disregarded, or deferred findings — the core exclusion mechanism (`--exclude` threaded through the queue scan before `FIRST_PR`/`FIRST_ISSUE` population and into the leaf trace) reviewed as correct.

natb1:
## Fixed
_None._

## Informational
- Defensive backstop: `cur_n` is parsed from `SELECTED` before the `SEEN[$cur_n]` backstop check fires — ordering is correct.
- Parser asymmetry: `dispatch-trace-leaf` rejects any trailing arg after the `--exclude` integer run, while `dispatch-select-target` stops the integer run at the next flag. Safe given the controlled caller, which only ever passes `--exclude <n>...` as the final args.
- `exclude_args` is rebuilt from `SEEN` each fan-out iteration so the latest spawned target is included — a cached array would be stale.
- The backstop stop-reason string (`selection re-returned an excluded target (unexpected)`) is asserted verbatim by its test; the test label is just a human name.
- `MAT_QUEUE` is read space-separated into an array (`read -r -a`); the tests pass space-separated input, so this is correct.
- Test 23's `2>&1 1>/dev/null` ordering captures only stderr — the intended behavior, used consistently elsewhere in the suite.
- `--exclude` is not forwarded in the `--qa` / `--health-only` paths, but the QA path selects only PRs (whose exclusion is handled directly in the PR scan), so this is not a gap.
- `SEEN` stores the branch issue-prefix as its key (`cur_n="${_pr_branch%%-*}"`), matching `dispatch-select-target`'s `pr_issue_num="${pr_branch%%-*}"` — the two sides key consistently.
- `"${!SEEN[@]}"` expands associative-array keys in unspecified order; the exclusion is an order-independent set, so this is fine.

## Dismissed
_None._

## Deferred
_None._

natb1:
## Security review — `dispatch` `--exclude` fan-out exclusion (#1062)

**Changed surface:** `code` (four bash scripts under
`.claude/skills/dispatch-propagate/scripts/`: `dispatch-materialize-spawn`,
`dispatch-select-target`, `dispatch-trace-leaf`, `test-dispatch-scripts.sh`).
`deps=false`, `app_or_rules=false`.

**Reviewers run:** input-validation, secrets, red-team, built-in
`/security-review` (all four unconditional-on-code reviewers), plus CodeQL
inline. App-domain reviewers (auth, data-exposure, Firebase) and the dependency
audit were skipped as structurally N/A — no application code, Firestore/Storage
rules, or `package.json`/`package-lock.json` changes.

**Scans:** CodeQL — `0` open alerts for this PR ref. Dependency audit — not run
(`deps=false`). Secrets — none. Built-in `/security-review` — no findings.

### Findings & dispositions

No `required` findings — no code change. Every reviewer finding was traced to a
pre-existing convention assumption or an unreachable precondition in the changed
code:

| # | Finding | Disposition |
|---|---|---|
| A | Non-numeric PR branch prefix (a branch without an `<N>-` prefix) selected as a fan-out continuation target produces a non-integer `cur_n`, which the diff then expands into `--exclude`, theoretically aborting the fan-out (DoS). *(red-team: high; input-validation: medium)* | **Not fixed — false positive.** The exploit path is unreachable: in the fan-out loop, `process_one_target` runs `dispatch-resolve-worktree "$cur_n" queue` (`dispatch-materialize-spawn:413` → `:242`) **before** the `--exclude` is built (`:450`). `dispatch-resolve-worktree` rejects any non-numeric arg with `^[1-9][0-9]*$` (`:54`) and exits 1, so a non-numeric `cur_n` already aborts the loop one step earlier — independent of this diff. The branch-prefix=issue-number assumption is a pre-existing convention, not introduced here. The receiving-side parsers in the changed code both correctly validate `^[0-9]+$` before storing an exclude token. |
| B | Empty `cur_n` from a branch name starting with `-` (DoS). *(red-team: medium)* | **Not fixed — false positive.** Same `dispatch-resolve-worktree` guard rejects an empty arg first; Git refs cannot begin with `-`. |
| C | The `--exclude` integer guard `^[0-9]+$` accepts `"0"` rather than `^[1-9][0-9]*$`. *(input-validation: low)* | **Not fixed — false positive / non-issue.** GitHub issue numbers start at 1; `EXCLUDED["0"]` can never match a real target, so it cannot cause a false exclusion. A pure hardening nit with no reachable security impact. |
| D | A flat help-wanted issue already in the exclude set still incurs a `dispatch-trace-leaf` subprocess (two `gh` calls) before the leaf is found excluded. *(red-team: high-confidence, but performance)* | **Not fixed — out of scope.** A latency/efficiency observation, not a security weakness; no follow-up filed (the follow-up threshold is for CodeQL/npm advisories). |
| E | `SELECTED=$(... ) \|\| true` masks a non-zero `dispatch-select-target` exit as an empty selection. *(red-team: low)* | **Not fixed — out of scope.** Observability nit; the selector's stderr still surfaces the real error, and the catchall already exits rather than spinning. Not a security weakness. |

### Verdict

The exclusion mechanism is sound: both receiving-side parsers
(`dispatch-select-target`, `dispatch-trace-leaf`) validate each `--exclude`
token as a digit-run before use, the associative-array keys are pre-validated
integers expanded under double-quotes (no word-splitting/glob/injection), and
no secrets, no new attack surface, no auth/data-exposure surface, and no
CodeQL alerts are introduced. No required fixes; PR marked ready.

Issue #1087, opened 2026-06-03
dispatch-security-followup: group npm advisories by package so one remediation files one issue


natb1:
## Context

`/security-review-fix` files out-of-scope dependency findings as tracking issues via `.claude/skills/dispatch-propagate/scripts/dispatch-security-followup`. The script maps npm findings 1:1 to issues, keyed by advisory ID:

```
identifier: ("npm advisory " + .advisory_id)
title:      ("security: " + .identifier + " in " + .package)
```

`npm audit` reports each GHSA in a coordinated disclosure as a separate advisory on the same package node. So a multi-advisory disclosure on one package — one package, one resolved version, one `overrides`/upgrade fix — fans out to N separate issues, even though they share a single unit of remediation.

## Reproduction

The security review of PR #1002 surfaced the `@xmldom/xmldom` serialization-injection family (5 advisories: GHSA-wh4c-j3r5-mjhp, GHSA-2v35-w6hq-6mfw, GHSA-f6ww-3ggp-fr8h, GHSA-x6wf-f3px-wcqx, GHSA-j759-j44w-7fr8), all fixed by a single `overrides` pin to `@xmldom/xmldom@^0.8.13`. The follow-up step filed 5 separate issues (#1071, #1073, #1074, #1075 plus #1072), since manually consolidated onto #1072.

- **Expected:** one tracking issue per vulnerable package needing a version bump, listing every advisory it resolves.
- **Actual:** one issue per advisory ID — 5 issues for 1 remediation.

`/file-issue`'s keyword dedup cannot catch this: it keys on title keywords and is deliberately conservative ("prefer creating a near-duplicate"), and distinct advisory IDs read as distinct problems. The fix belongs upstream in the follow-up emitter, not the dedup layer.

## Proposed change

In `dispatch-security-followup`, change the npm follow-up unit from advisory to package remediation:

- `group_by(.package)` over qualifying npm findings; emit one follow-up per package.
- Qualify a group if any member is `out-of-scope` + severity `high`/`critical` + `introduced_by_diff == false`; label with the group's max severity.
- Body lists every advisory in the group (advisory ID, GHSA URL, severity).
- Dedup identifier becomes package-scoped (e.g. `npm advisories in <package>`), so re-runs don't refile.
- CodeQL findings unchanged — one issue per alert remains the correct unit.

[...]

Issue #50, opened 2026-02-21
Fix acceptance test CI: hosting emulator fails without Firebase auth


natb1:
## Problem

The acceptance test CI workflow (`hello-acceptance-tests.yml`) is failing because the Firebase hosting emulator cannot fetch web app configuration in the unauthenticated CI environment.

Error from CI:
```
⚠  hosting: Could not fetch web app configuration and there is no cached configuration on this machine.
Check your internet connection and make sure you are authenticated.
ERROR: Firebase hosting emulator did not start within 30s
```

This appears to be a regression from a newer version of `firebase-tools` (15.7.0) being installed via `npx` without version pinning.

## Parent issue

Sub-issue of #47

## Fix

Pin `firebase-tools` version in the acceptance test script or workflow, or configure the hosting emulator to not require web app config fetch.

PR #1110, opened 2026-06-04
security: idempotent security-followup filing across open and closed issues

natb1:
Closes #1100

Add a deterministic, state-spanning exact-identifier existence check in front of
`/file-issue` on the `/review-fix` Step 6b security-followup filing path, so
pre-existing out-of-scope advisories are never re-filed across review runs.

## Problem

The security fan-out re-discovers pre-existing out-of-scope advisories on every
PR review and re-files them. The only dedup guard was `/file-issue`'s fuzzy
keyword+judgment check, scoped to open issues only. For the machine-stable
identifiers this path emits, that guard leaked even on exact-title matches (the
vitest advisory was filed 3×: #1077, #1094, #1096) and re-filed after a tracking
issue was closed.

#1087 collapsed N-advisories-per-package to one issue per package but left the
dedup mechanism and its open-only state scope unchanged. This closes that gap.

## Change

- New script `.claude/skills/dispatch-propagate/scripts/dispatch-followup-exists`:
  runs `gh issue list --search '"<identifier>" in:title' --state all` and
  post-filters with jq for exact substring containment, printing the first
  matching issue number (or nothing). Spans open and closed issues. The pure
  `dispatch-security-followup` emitter stays gh-free; this is its gh-calling
  companion.
- `/review-fix` Step 6b: the per-follow-up subagent runs `dispatch-followup-exists`
  before `/file-issue`; on a match (open or closed) it skips filing and maps the
  identifier to the existing issue. `/file-issue`'s fuzzy dedup stays as
  defense-in-depth for free-text callers.
- Tests in `test-dispatch-scripts.sh` cover open-match, closed-match,
  CodeQL-match suppression, no-match, and fuzzy-but-not-exact (exactness) cases.

## Tradeoff

Because the guard spans `--state all`, a closed tracking issue — already-resolved
or already-rejected — suppresses re-filing of its identifier. Re-filing is noise
either way. Documented in Step 6b.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-followup-exists
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-followup-exists
@@ -37,9 +37,20 @@
 # MATCHING
 #   gh's --search is tokenized/fuzzy, so it only narrows the candidate set
 #   server-side (wrapping the identifier in literal double-quotes and adding the
-#   `in:title` qualifier). The authoritative match is the jq post-filter, which
-#   uses `contains($id)` — a LITERAL substring test — so a mere shared-token hit
-#   does NOT match. The first matching issue number is printed; nothing if none.
+#   `in:title` qualifier). The authoritative match is the jq post-filter.
+#
+#   The post-filter is BOUNDARY-AWARE, not a bare substring test. A title matches
+#   only when the identifier sits at a token boundary — it either ends the title
+#   (`endswith`) or is immediately followed by a space (`contains($id + " ")`).
+#   The emitter's titles are `security: <identifier>` (npm) and
+#   `security: <identifier> in <location>` (CodeQL), so the real identifier is
+#   always followed by end-of-title or a space. A bare `contains($id)` would
+#   false-match on a prefix collision — identifier `npm advisories in lodash`
+#   is a substring of title `... lodash-es`, and `CodeQL <rule> alert #5` is a
+#   substring of `... alert #50 ...` — silently suppressing a genuine follow-up
+#   for the shorter-named package or lower-numbered alert. The boundary check
+#   rejects those: the char after the candidate is `-` / `0`, not a space or
+#   end-of-string. The first matching issue number is printed; nothing if none.
 set -euo pipefail
 
 if [[ "$#" -lt 1 || -z "${1:-}" ]]; then
@@ -50,4 +61,5 @@ fi
 identifier="$1"
 
 gh issue list --search "\"$identifier\" in:title" --state all --json number,title --limit 100 \
-  | jq -r --arg id "$identifier" '[.[] | select(.title | contains($id))][0].number // empty'
+  | jq -r --arg id "$identifier" \
+      '[.[] | select((.title | endswith($id)) or (.title | contains($id + " ")))][0].number // empty'
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-followup-exists
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-followup-exists
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+# Print the number of the first issue whose title contains <identifier>, or
+# nothing if none exists.
+# Usage: dispatch-followup-exists <identifier>
+#
+# WHY THIS EXISTS
+#   This is a deterministic, state-spanning, exact-identifier existence check
+#   used by /review-fix Step 6b's security-followup filing path to make
+#   re-filing idempotent. Before filing each emitted follow-up, the caller
+#   probes for an issue already tracking the same stable identifier and skips
+#   filing when one exists.
+#
+#   It complements two other pieces and replaces neither:
+#     - The PURE `dispatch-security-followup` emitter, which derives the stable
+#       identifiers but must stay free of gh/network calls. The existence check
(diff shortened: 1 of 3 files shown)


natb1:
## QA pass — #1100 (autonomous, non-browser)

Scripts/docs-only change to the dispatch tooling (`dispatch-followup-exists`, its
tests, and `review-fix` Step 6b docs). No browser component — non-browser
walkthrough.

**Items walked: 5 — all machine-verifiable.**

| # | Item | Result |
|---|------|--------|
| 1 | `test-dispatch-scripts.sh` covers re-file suppression (open + closed match) and passes | **PASS** — full suite 1485/1485; all 5 `dispatch-followup-exists` cases (open npm, closed npm, CodeQL, no-match, fuzzy-overlap) pass |
| 2 | Exact-identifier match returns the issue number (npm + CodeQL) | **PASS** — npm token → `1095`; `CodeQL <rule> alert #<n>` token → `1077` |
| 3 | Idempotency across two simulated review runs → exactly one issue | **PASS** — run1 (no existing issue) → empty (would file); run2 (issue now exists) → number (suppress re-file) |
| 4 | Fuzzy/near-token overlap does not match (literal-substring authority) | **PASS** — shared leading tokens with a broken substring → empty |
| 5 | State-spanning tradeoff documented in `review-fix` SKILL.md Step 6b | **PASS** — notes `--state all` (open + closed) and that a closed tracking issue suppresses re-filing |

**Counts:** PASS 5 / FAIL 0 / SKIP 0.

**Deferred to office-hours:** none — every plan item was machine-verifiable.

**Bugs found:** none.

Acceptance criteria for #1100 are all satisfied by the diff and verified here.

natb1:
## Review phase — consolidated findings (code-review + review + security)

One terminal review pass over the diff (`dispatch-followup-exists`,
`test-dispatch-scripts.sh`, `review-fix/SKILL.md`). Surface classified `code`
(`deps=false`, `app_or_rules=false`), so the always-on security reviewers ran
plus an inline CodeQL fetch; no dependency audit and no app-domain reviewers.

### Fixed
- **`dispatch-followup-exists` false-positive suppression on prefix collisions** (security, red-team, high) — the jq post-filter used a bare `contains($id)` substring test. Identifier `npm advisories in lodash` is a literal substring of an existing `…lodash-es` title, and `CodeQL <rule> alert #5` is a substring of `…alert #50 …` — so the dedup guard would match the wrong issue and **silently drop a genuine follow-up** for the shorter-named package / lower-numbered alert. Replaced with a boundary-aware match: `endswith($id) or contains($id + " ")`. The emitter's titles are `security: <identifier>` (npm) and `security: <identifier> in <location>` (CodeQL), so the real identifier is always followed by end-of-title or a space; a colliding longer token has `-`/`0` there instead. Fixed in `558a1bc6`.
- **Regression tests for the prefix collision** (security/code-review) — added CASE 7 (`lodash` vs `lodash-es` → empty) and CASE 8 (CodeQL `#5` vs `#50` → empty) to `test-dispatch-scripts.sh`. All 1488 tests pass. `558a1bc6`.
- **CASE 6 multiple-match test** (code-review) — added a test asserting the script returns the first issue number when several titles match. `558a1bc6`.
- **CASE 2 comment accuracy** (code-review) — the comment misdescribed what the test exercises (the stub is state-agnostic); corrected. `558a1bc6`.
- **SKILL.md Step 6b line reflow** (code-review/review) — a mid-phrase line break split the identifier description; reflowed. `558a1bc6`.

### Informational
- **`--limit 100` candidate ceiling** (review/red-team, low) — the server-side `gh issue list` search is capped at 100 candidates before the jq post-filter. Theoretical at this repo's scale; no observable failure today. Left as-is.
- **Caller treats a nonzero `gh` exit as "no match"** (code-review/red-team) — on a transient `gh` failure the script exits nonzero with empty stdout (`set -euo pipefail`), and the Step 6b caller's "otherwise proceed" path would then file rather than abort. A rare failure-path edge that errs toward not-dropping an advisory; out of scope for this PR's idempotency feature.

### Dismissed
- **"`--state all` is not asserted by the tests"** (review, high) — incorrect. The `gh` stub's case pattern is `issue list *--state all --json number,title --limit 100`; if the script dropped `--state all` the invocation falls through to the stub's error branch and the test fails. The flag is enforced by the suite.
- **Test-helper variable naming (`TMPDIR_TEST`, `SAVED_PATH_FE`)** (review, low) — setup/teardown are always paired, so there is no real clobber; a style preference, not an improvement.

### False-positive (security)
- **Shell injection via `$identifier` in `gh --search`** (input-validation, review; medium) — not exploitable. `$identifier` is expanded inside a double-quoted string, so the whole `"…" in:title` is a single argument: no word-splitting and no argument injection (`--repo …` etc.). Bash does not recursively re-evaluate a variable's value, so `$( )`/backticks inside the identifier are not executed. Worst case is a malformed `gh` search query, parsed by `gh`, not the shell.
- **jq `--arg id` injection** (input-validation) — `--arg` binds a typed jq string variable; the value is never interpolated into the filter program.

### Security scans
- **CodeQL** — fetched inline (`surface=code`); 0 open alerts on the PR ref.
- **Dependency audit** — not run (`deps=false`; the diff touches no `package.json`/lockfile).

PR #1004, opened 2026-06-01
Retire the orphaned productivity app code

natb1:
Closes #776

Removes the dead footprint left by the retired productivity app (#456 and its
sub-issues), now superseded by the /dispatch JIT reminder engine (#726) and the
agenda app (#775).

## What changed

- **Relocated the dispatch rate-limit telemetry writer** out of the retiring
  package: `productivity-tui/hooks/update-rate-limits.sh` →
  `.claude/skills/dispatch-propagate/scripts/update-rate-limits.sh`, and renamed
  its state dir `~/.local/share/productivity-tui` → `~/.local/share/commons-dispatch`.
  All consumers (`statusline.sh`, `dispatch-target-workers`,
  `dispatch-schedule-reseed`, `reference.md`) point at the new path.
- **Removed the productivity-tui package**: the Go TUI, `internal/`,
  `go.mod`/`go.sum`, and its lifecycle hooks; unwired its 4 hook entries from
  `.claude/settings.json` (the now-empty `SessionEnd` key dropped); dropped its
  `.gitignore` line; renamed the go-module test fixture to a neutral name.
- **Removed the Nix build**: `nix/packages/productivity-tui.nix`, the `flake.nix`
  packages output + devShell PATH wiring, and the `wezterm.nix` update-status
  handler. The `go` toolchain stays for budget-etl and scaffolding/firebase.
- **Removed the productivity Firestore rules + tests**: the
  `isProductivityGroupMember` helper, the 5 `match /productivity/{env}/` blocks
  and their validators, and `rules-test/test/firestore/productivity.test.ts`. The
  deny-all catch-all remains last.

## Notes

- **Telemetry-writer relocation bundled in.** The dispatch concurrency-budgeting
  system had repurposed the productivity-tui rate-limit writer and its state file
  as live telemetry without relocating them. A naive directory removal would not
  crash but would silently regress the dispatch chain to 1 worker/tick, so the
  relocation lands as the first commit here.
- **Firestore rules follow the serialized rules workflow.** The rule change ships
  in this feature branch; the standalone rules-only PR + manual deploy is a later
  step at submission time, per .claude/rules/firestore.md.
- **Retained `marketplace.json` budget category.** The `"category": "productivity"`
  at `.claude-plugin/marketplace.json:19` is the budget plugin's legitimate
  marketplace category, not app residue — a deliberate deviation from the issue's
  literal "no productivity references" criterion.

PR #1111, opened 2026-06-04
dispatch: flush stranded local merges before a review-fix PR goes ready (#1105)

natb1:
Closes #1105

## Problem

The dispatch chain follows a "never push a bare merge commit" contract:
`dispatch-merge-main` and `/dispatch-resolve-conflict` merge `origin/main` into a
worktree locally and never push, relying on each phase skill's own push point to
carry the merge to origin. `/review-fix` is the terminal phase; its only push
point is the single `/commit-merge-push` in Step 5. On a zero-findings review the
worker can skip that fork, then still apply `dispatch:reviewed` + `gh pr ready`,
stranding any earlier local merge commit permanently — every later tick routes
`STOP done`, no push point fires again, and GitHub reports the PR `CONFLICTING`
forever. (Observed on PR #1004 / issue #776.)

## Fix

**Primary — `/review-fix` Step 8 terminal flush.** A guard runs unconditionally
as the first action of Step 8, before `gh pr ready`: it fetches the PR branch and
pushes `origin HEAD` whenever `origin/<branch>..HEAD` is non-empty. This is the
terminal flush of the contract — `/review-fix` is the chain's terminal phase, so
any local merge left behind must be carried to origin here. The guard cannot be
reasoned away on a zero-findings run. The `dispatch:reviewed` re-entry path now
routes through Step 8 rather than readying the PR inline, so the flush also covers
an interrupted prior run.

**Defensive — `dispatch-route` refuses to park stranded commits.** The `done`
case no longer emits a bare `STOP done`: it checks `origin/<branch>..HEAD` and
emits a new `PUSH-STRANDED` directive when the worktree is ahead. `/dispatch-worker`
handles `PUSH-STRANDED` by pushing `origin HEAD`, writing an office-hours reason,
and parking with no marker — healing the PR and surfacing the anomaly. Fails safe
to `STOP done` when the remote branch cannot be resolved. With the Step 8 guard in
place this is a true backstop that should essentially never fire.

## Tests

`test-dispatch-scripts.sh`: 1482/1482 pass, including the unchanged Test 8
(non-draft → `STOP done`) and the new Test 8b (non-draft + worktree ahead →
`PUSH-STRANDED`).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-route
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-route
@@ -12,6 +12,7 @@
 #   INVOKE /review-fix           phase review       (dispatch:qa-done / reviewed re-entry)
 #   RELEVANCE-REVIEW             phase implement    (no PR) — routes to SKILL.md Step 2
 #   STOP done                    phase done         (non-draft PR)
+#   PUSH-STRANDED                phase done, but worktree ahead of origin/<branch> — push then park
 #   STOP waiting                 dispatch-ci-ready reports not-ready (draft PR CI pending)
 #   STOP wrong-worktree          cross-check failed (diagnostic on stderr); exits non-zero
 #
@@ -88,7 +89,19 @@ case "$PHASE" in
   verify)       echo "INVOKE /verify-pr" ;;
   qa)           echo "INVOKE /qa-fix" ;;
   review)       echo "INVOKE /review-fix" ;;
-  done)         echo "STOP done" ;;
+  done)
+    # #1105: never park a `done` PR whose worktree is ahead of its remote branch —
+    # those unpushed dispatch-merge-main / resolve-conflict merges leave the PR
+    # permanently CONFLICTING with no later tick able to push them.
+    CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+    git fetch origin "$CURRENT_BRANCH" >/dev/null 2>&1 || true
+    AHEAD=$(git rev-list --count "origin/${CURRENT_BRANCH}..HEAD" 2>/dev/null || echo 0)
+    if [[ "$AHEAD" -ne 0 ]]; then
+      echo "PUSH-STRANDED"
+    else
+      echo "STOP done"
+    fi
+    ;;
   *)
     echo "dispatch-route: unexpected phase '$PHASE' for $N" >&2
     exit 1
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -461,6 +461,18 @@ case "$args" in
     # dispatch-resolve-worktree reconciliation: re-point to the PR head branch.
     echo "$args" >> "$STUB_DIR/git-checkout.log"
     ;;
+  fetch\ *)
+    # dispatch-route done-case: fetch the current branch from origin. No-op stub.
+    : ;;
+  rev-list\ --count\ *)
+    # dispatch-route done-case: commits the worktree is ahead of origin/<branch>.
+    # Default 0 (remote up to date → STOP done); route-ahead-count.txt overrides to N.
+    if [[ -f "$STUB_DIR/route-ahead-count.txt" ]]; then
+      cat "$STUB_DIR/route-ahead-count.txt"
+    else
+      echo "0"
+    fi
+    ;;
   *)
     echo "git stub: unknown invocation: $args" >&2
     exit 1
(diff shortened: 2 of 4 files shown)


natb1:
## QA pass — clean (autonomous, non-browser)

This PR changes the dispatch chain (`.claude/skills/` scripts + SKILL.md docs)
to close the issue #1105 hole where `/review-fix` could mark a PR ready while
unpushed local merge commits stranded it as permanently `CONFLICTING`. No
frontend package is touched, so QA ran the non-browser path: the dispatch test
suite plus targeted inspection of the changed behaviors.

**Items walked: 4 — PASS 4 / FAIL 0 / SKIP 0**

| # | Item | Result |
|---|------|--------|
| 1 | Dispatch script test suite passes (incl. new `PUSH-STRANDED` route test) | PASS — 1482/1482 |
| 2 | `dispatch-route` re-routes a stranded `done` PR to `PUSH-STRANDED` instead of `STOP done` | PASS |
| 3 | `/review-fix` Step 8 flushes unpushed commits (`git fetch` + conditional `git push origin HEAD`) before `gh pr ready`, unconditionally | PASS |
| 4 | SKILL.md docs state *why* the push is unconditional — terminal flush of the "never push a bare merge commit" contract; `dispatch-worker` documents the `PUSH-STRANDED` directive | PASS |

**Deferred to office-hours:** none — all items were machine-verifiable.

**Bugs found:** none.

natb1:
## Review-fix summary (PR #1111)

One consolidated pass: `/code-review max --fix`, `/review`, and the surface-gated
security fan-out (input-validation, secrets, red-team, built-in `/security-review`,
plus inline CodeQL) over the same diff. Surface classified `code` (bash scripts);
`app_or_rules=false` and `deps=false`, so app-domain reviewers and the dependency
audit did not apply. CodeQL: no open alerts. All findings de-duplicated below.

### Fixed
- **`test-dispatch-scripts.sh` — Test 8b branch stub** (code-review, high): the new
  PUSH-STRANDED test omitted `current-branch.txt`, so the route resolved the default
  stub branch `main` instead of the test's `42-my-feature`; a regression hardcoding
  `main` would still have passed. Added the `current-branch.txt` stub so the route
  resolves the intended branch. Fixed in `4dfbbf48`.
- **`review-fix/SKILL.md` — Step 8 snippet quoting** (code-review, medium): the Step 8
  flush example used a nested-double-quote `if [[ "$(... "...")" ]]` form; split into
  the capture-then-test `AHEAD=$(...)` / `[[ "$AHEAD" -ne 0 ]]` pattern, matching the
  actual `dispatch-route` code. Fixed in `4dfbbf48`.

### Required (security)
_None._

### Informational
- **Fetch-failure fallback is untested** (`test-dispatch-scripts.sh`, review, low): no
  test covers the `done)` case's `|| echo 0` fail-safe (unresolvable ref → `STOP done`).
  The fallback is intentional and correct for network hiccups; added coverage is a
  nice-to-have, not a correctness gap in this PR.
- **PUSH-STRANDED handler is prose-only** (`dispatch-worker/SKILL.md`, review, low): a
  future model could reason away the worker's push step. This is acknowledged by design
  — the primary `/review-fix` Step 8 guard is mechanical and unconditional, and
  `dispatch-route` enforces the routing; PUSH-STRANDED is the belt-and-suspenders
  backstop. No change needed here.
- **Fetch failure routes `STOP done` instead of `PUSH-STRANDED`** (`dispatch-route`,
  red-team, low): a disrupted `git fetch` falls back to `STOP done`, deferring the push.
  Extreme, high-noise attack surface; the fallback is correct for the common case.
  Pre-existing fail-safe behavior, out of scope for this PR.

### Dismissed
- **Detached-HEAD `git push origin HEAD`** (`dispatch-worker/SKILL.md`, red-team): a
  detached HEAD could push a stray `refs/heads/HEAD`. Not reachable — dispatch worktrees
  are never manually manipulated into detached HEAD. Not a security issue.

### False-positive (security)
- **Command substitution via `"origin/${CURRENT_BRANCH}..HEAD"`** (`dispatch-route`,
  mirrored in `review-fix/SKILL.md`; input-validation + red-team): bash does **not**
  re-evaluate a variable's expanded value for `$(...)` command substitution — the `$()`
  syntax must appear literally in the source. The branch name also comes from
  `git rev-parse --abbrev-ref HEAD`, and git rejects branch names with shell
  metacharacters. The outer double-quotes already neutralize word-splitting/globbing.
  Non-exploitable.

### Deferred / Out-of-scope (security)
_None._

---
Applied `dispatch:reviewed`; marking the PR ready. The Step 5 commit merged
`origin/main` (`97991644`), resolving directive-comment conflicts in `dispatch-route`
and prose conflicts in `dispatch-worker/SKILL.md` by keeping both sides' directives
(PUSH-STRANDED alongside the new `INVOKE /dispatch-resolve-conflict` / `STOP
provision-failed`). Full suite green: 1545/1545.

PR #1128, opened 2026-06-04
budget-etl: enforce JournalLeg and Account invariants at serialization (#800)

natb1:
Closes #800

The Go double-entry export structs (`JournalLeg`, `Account`) in
`budget-etl/internal/export/export.go` had no construction-time validation, so
the ETL could write a `budget.json` that the budget app's TypeScript upload path
would later refuse to upload.

This adds, mirroring the TS rules:

- A typed `AccountType` enum (`asset`/`liability`/`equity`/`income`/`expense`)
  with a `Valid()` predicate, matching `ACCOUNT_TYPES` in
  `budget/src/schema/enums.ts`. `Account.AccountType` is now this type (JSON tag
  unchanged — still serializes as a string).
- `JournalLeg.Validate()` — rejects NaN/Inf/negative debit or credit and a
  both-positive leg; mirrors `parseRawJournalLeg`.
- `Account.Validate()` — rejects unrecognized account types; mirrors
  `parseRawAccount`.
- `Output.Validate()` — iterates `journalLegs` and `accounts`, wrapping each
  failure with its index.
- A `data.Validate()` call at the top of `export.WriteFile` — the Go analog of
  the TS upload boundary — so `budget-etl` never writes a rejectable file (no
  temp file is created on a validation failure).

The producer `journal.buildAccounts` is updated to use the typed constants.

Verification: `go test ./internal/export/... ./internal/journal/...`, `go vet
./...`, and `go build ./...` all pass.

### DIFF
--- a/budget-etl/internal/export/export.go
+++ b/budget-etl/internal/export/export.go
@@ -335,12 +335,18 @@ func ReadFile(path, password string) (Output, error) {
 	if out.Transactions == nil {
 		return Output{}, fmt.Errorf("parsing %s: missing required field 'transactions'", path)
 	}
+	// Note: ReadFile intentionally does not call out.Validate(). Validation is
+	// enforced at the serialization boundary (WriteFile) so that a file written
+	// by an older ETL version still loads; refusing to read it would strand the
+	// user with no migration path. Callers that re-emit a read Output go back
+	// through WriteFile, which re-validates before writing.
 	return out, nil
 }
 
 // Validate checks the double-entry invariants on the journal legs and accounts
 // before serialization, mirroring the TypeScript upload-path validation so the
-// two halves stay in lockstep.
+// two halves stay in lockstep. WriteFile calls Validate before serializing;
+// callers that only need the check can call Validate directly.
 func (o Output) Validate() error {
 	for i, l := range o.JournalLegs {
 		if err := l.Validate(); err != nil {
--- a/budget-etl/internal/export/export.go
+++ b/budget-etl/internal/export/export.go
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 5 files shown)


natb1:
## QA — autonomous pass (#800)

Pure-backend `budget-etl` change adding serialization-time validation
(`JournalLeg.Validate`, `Account.Validate`, `AccountType` enum,
`Output.Validate` wired into `WriteFile`). No browser component — non-browser
QA path. Validator-in-isolation behavior is covered by unit tests (CI's job);
these items target end-to-end, user-visible binary behavior.

**Items walked: 2 — PASS 2 / FAIL 0 / SKIP 0**

### 1. Real ETL run emits a valid budget.json through the new gate — PASS
Built the binary (`go build`, `go vet` clean) and ran it on a constructed
statement directory:
`budget-etl --dir <statements> --output budget.json --plaintext --group test-group`.
The real `journal.go` producer wrote 3 transactions / 3 journal entries / 6
legs / 3 accounts and exited 0 — the new validation gate accepts genuine ETL
output (no happy-path regression). Verified in the produced `budget.json`:
- account types all valid: `expense`, `income`, `asset`
- no leg with both debit and credit positive; no negative leg
- every journal entry balanced (debits == credits): 5=5, 12.5=12.5, 200=200
- every `transaction` carries a `journalEntryId`

### 2. Gate rejects malformed output at the real WriteFile path — PASS
Confirmed the validation is wired into the real write path (not just the
validator in isolation): a both-positive debit/credit leg and an unrecognized
account type each make `WriteFile` return a descriptive error and write **no**
file, while valid output round-trips. (`TestWriteFileValidatesOutput`,
`TestJournalLegValidate`, `TestAccountValidate` — all subcases pass.)

**Bugs found:** none

natb1:

One pass over the branch diff (vs merge-base with `origin/main`): `/code-review max --fix`, `/review`, and the surface-gated security fan-out (input-validation, secrets, red-team, built-in `/security-review`, auth, data-exposure, firebase), plus inline CodeQL. No dependency audit (the diff touches no `package.json`/lockfile). CodeQL: no open alerts on the PR ref.

Fixes are in commit `3f62897`.

### Fixed (code-review / review)
- **`budget-etl/main.go:365` & `:1073`** — the two `priorLiabilities` loops compared `a.AccountType == "liability"` with a raw string literal; switched to the typed `export.AccountTypeLiability` constant this PR introduces, so a future enum-value rename can't silently misclassify credit-card accounts. (`/code-review`)
- **`budget-etl/internal/export/export_test.go`** — `TestWriteFileValidatesOutput` now asserts the error names the offending `journalLegs[0]` / `accounts[0]` index, pinning the `fmt.Errorf` index-wrapping so its accidental removal would fail the test. (`/review`)
- **`budget-etl/internal/export/export.go`** — documented that `Output.Validate()` is the check `WriteFile` runs before serializing (callers may call it directly), and added a comment on `ReadFile` explaining the deliberate read-path/write-path asymmetry (below). (`/review`)

### Informational
- **`ReadFile` does not validate** (review + red-team, medium) — `WriteFile` validates but `ReadFile` does not. This is intentional: validating on read would strand a user holding a `budget.json` written by an older ETL with no migration path, whereas refusing to *write* invalid data is the right boundary. Re-emitting a read `Output` goes back through `WriteFile`, which re-validates. Documented in `ReadFile` rather than changed.
- **Float-point `> 0` strictness** (red-team, low) — a notionally-zero amount with a tiny floating-point residual could trip the "both debit and credit" check. Low likelihood: amounts derive from parsed integer cents. Noted; not changed (matching the TS side's exact comparison keeps the two in lockstep).
- **`Account.Validate()` does not check non-empty `ID`/`Institution`/`Account`** (red-team, medium) — out of #800's scope (the issue is the account-type enum); the emitter always populates these. Worth folding into the referential-integrity follow-up below.

### Dismissed / false-positive (security)
- **`-0.0` accepted as a valid debit/credit** (red-team, medium) — `-0.0 < 0` is `false`, so negative zero passes. This is symmetric with the TypeScript `parseRawJournalLeg` (same `< 0` check), and `-0.0` is value-equal to `0` for the both-zero-allowed rule. Matching TS is the explicit goal; not a real defect.
- **Validation error messages embed debit/credit values and the account-type string** (data-exposure, low) — `budget-etl` runs locally and processes the user's own statements on their own machine; surfacing the offending value in an error to that same user is helpful, not a disclosure. Not a vulnerability in this local-binary context.

### Deferred → follow-ups
- **`Output.Validate()` should check `JournalLeg`→`Account` referential integrity** (red-team, high) — a leg can name a phantom `accountId` and still pass per-object validation. Beyond #800's per-object scope (the TS counterpart doesn't check it either). Filed as **#1138** (blocked by #800).
- **`accountID()` underscore join can collide** (red-team, high) — `institution + "_" + account` collides when either field contains an underscore, merging distinct accounts' legs. Pre-existing in `journal.go`; this PR only swapped literals to typed constants there. Filed as **#1139** (independent).

### Out-of-scope (security)
- None beyond the two deferred items above. CodeQL returned no open alerts; no dependency surface to audit.

PR #1132, opened 2026-06-04
dispatch: headless rate-limit telemetry refresh for dispatch-tick

natb1:
Closes #1127

## Problem

The autonomous dispatch chain runs headlessly: `dispatch-tick` (pure bash,
launched by systemd timers and worker Stop hooks) computes the budget via
`dispatch-target-workers`, which reads cached telemetry in
`~/.local/share/commons-dispatch/rate_limits.json`. That file has **exactly one
writer**: the `statusLine` hook (`statusline.sh` → `update-rate-limits.sh`),
which lifts the `.rate_limits` payload Claude Code passes on stdin **only when a
session renders its status line**.

When the chain fully pauses — the last worker exits, the tick arms a one-shot
reseed timer at the cap's `resets_at`, and no live sessions remain — and no
interactive session is open, nothing renders a status line, so
`rate_limits.json` freezes at the cap-hit snapshot. The reseed timer then fires
at `resets_at`, but the tick re-reads the *frozen* file: `dispatch-target-workers`
returns 0, nothing spawns. The budget has reopened but the chain can't observe
it — it stalls until a human opens a session. Observed live 2026-06-04: a 5h cap
reopened at 13:10, but the chain sat idle ~3h until an unrelated interactive
session refreshed telemetry.

A retry can't fix this — re-reading a frozen file yields the same stale answer.
The chain needs a telemetry source independent of status-line renders.

## Fix

Give the headless path its own rate-limit telemetry refresh, run at the start of
the tick before any budget read.

- **`dispatch-refresh-rate-limits`** (new) — a probe that reads the stored OAuth
  access token from `~/.claude/.credentials.json`, fetches the unified
  rate-limit response headers from the Anthropic messages endpoint
  (`anthropic-ratelimit-unified-{5h,7d}-{utilization,reset}` — the same data
  Claude Code forwards into `.rate_limits`), and writes `rate_limits.json` in the
  existing `five_hour`/`seven_day` schema by piping a `{"rate_limits":{…}}`
  payload through `update-rate-limits.sh`, reusing its atomic write. The schema is
  byte-identical to today's file: `used_percentage = round(utilization*100)`,
  `resets_at = <window>-reset` verbatim.

- **`update-rate-limits.sh`** — the output path is now overridable via
  `DISPATCH_RATE_LIMITS_STATE_FILE` (default unchanged), so the probe and tests
  can redirect the write. Its behavior as the status-line writer is unchanged, so
  there is no format divergence — the hook remains a valid writer.

- **`dispatch-tick`** — runs the probe as Step 0.5, before
  `dispatch-select-tick` / `dispatch-target-workers`, so every budget read this
  tick (and any reseed it arms) is computed against fresh telemetry.

**Fail-safe contract.** Every probe failure path — missing/unreadable
credentials, missing/expired token, probe request failure, missing/malformed
headers — logs exactly one stderr diagnostic, exits non-zero, and leaves the
cached file untouched; the access token is never echoed. The tick treats a
non-zero probe exit as log-and-proceed, so a probe outage degrades to the prior
cached-file behavior rather than erroring the tick. Header parsing is
status-independent: a 429 cap response still carries the unified headers and is
consumed, so the capped state is recorded too.

**Out of scope.** OAuth token refresh — the probe uses the stored access token
as-is. Hand-rolling refresh would race Claude Code's own credential writes
(rotating refresh tokens → auth corruption) and exceeds the small-helper scope.
An expired token degrades safely (fall back to cache + log), which the
fail-safe contract permits.

## Tests

`test-dispatch-scripts.sh` gains a `=== dispatch-refresh-rate-limits ===` group
and two `dispatch-tick` tests:

- **probe:** success writes the canonical telemetry; expired token, missing
  creds, missing headers, and malformed values each fail safe with no write.
- **refresh→budget regression (the reported stall, end-to-end with real
  scripts):** seed a *frozen pre-reset* `rate_limits.json` (5h used ≥ target,
  `resets_at` in the past), run the probe with headers showing the **reopened**
  window, then run the **real** `dispatch-target-workers` against the refreshed
  file — target is `>= 1`. Frozen → refresh → reopened budget observed, no
  interactive session needed.
- **tick wiring:** the refresh runs *before* select (ordering), and a probe
  failure still lets the tick route its decision and exit 0 (fail-safe).

Full suite: 1655/1655 passing.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-refresh-rate-limits
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-refresh-rate-limits
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+# dispatch-refresh-rate-limits — headless rate-limit telemetry probe for #1127.
+#
+# The autonomous dispatch chain runs headlessly. Its budget telemetry
+# (rate_limits.json) has historically had exactly ONE writer: the statusLine
+# hook (statusline.sh -> update-rate-limits.sh), which lifts the .rate_limits
+# field Claude Code passes on stdin when a session renders its status line.
+# When the chain fully pauses with zero live sessions and no interactive
+# session open, nothing renders a status line, so rate_limits.json freezes at
+# the cap-hit snapshot — and a reseed-fired tick re-reads that frozen file,
+# computes target 0, and stalls even though the budget has reopened.
+#
+# This probe gives the headless path its own telemetry source. It fetches the
+# unified rate-limit response headers from the Anthropic messages endpoint
+# (the same 5h/7d windows Claude Code forwards into the status-line hook) and
+# writes rate_limits.json in the existing schema (five_hour/seven_day ->
+# used_percentage, resets_at), reusing update-rate-limits.sh's atomic write.
+# The status-line hook remains a valid writer — no format divergence.
+#
+# FAIL-SAFE CONTRACT (per .claude/rules/code-style.md; AC #4 of #1127 makes the
+# fall-back-to-cache-and-log degradation an explicit requirement, NOT an
+# accidental fallback): every failure path logs exactly one stderr diagnostic,
+# exits non-zero, and never touches the cached file or echoes the access token
+# in output. Exit 0 ONLY on a successful write. A probe outage (network,
+# missing creds, expired token, missing/malformed headers) therefore degrades
+# to today's behavior — the tick keeps the cached telemetry — rather than
+# erroring the tick.
+#
+# Token refresh is out of scope: this uses the stored access token as-is and
+# degrades safely on expiry (logs and exits non-zero).
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+CREDS="${DISPATCH_REFRESH_RATE_LIMITS_CREDS:-$HOME/.claude/.credentials.json}"
+if [[ ! -r "$CREDS" ]]; then
+  echo "dispatch-refresh-rate-limits: credentials not found at $CREDS; keeping cached telemetry" >&2
+  exit 1
+fi
+
+TOKEN=$(jq -r '.claudeAiOauth.accessToken // empty' "$CREDS")
+EXPIRES_AT=$(jq -r '.claudeAiOauth.expiresAt // empty' "$CREDS")
+if [[ -z "$TOKEN" || -z "$EXPIRES_AT" ]]; then
+  echo "dispatch-refresh-rate-limits: access token or expiry missing from credentials; keeping cached telemetry" >&2
+  exit 1
+fi
+
+now_ms=$(( $(date +%s) * 1000 ))
+if [[ ! "$EXPIRES_AT" =~ ^[0-9]+$ ]] || (( EXPIRES_AT <= now_ms )); then
+  echo "dispatch-refresh-rate-limits: token expired; keeping cached telemetry" >&2
+  exit 1
+fi
+
+ENDPOINT="${DISPATCH_REFRESH_RATE_LIMITS_ENDPOINT:-https://api.anthropic.com/v1/messages}"
+MODEL="${DISPATCH_REFRESH_RATE_LIMITS_MODEL:-claude-haiku-4-5-20251001}"
+
+# TEST SEAM: read headers from a file instead of the network (no eval).
+if [[ -n "${DISPATCH_REFRESH_RATE_LIMITS_HEADERS_FILE:-}" ]]; then
+  HEADERS="$(cat "$DISPATCH_REFRESH_RATE_LIMITS_HEADERS_FILE")"
+else
+  HEADERS=$(curl -sS -D - -o /dev/null --connect-timeout 5 --max-time 10 \
+    -H "authorization: Bearer $TOKEN" \
+    -H "anthropic-version: 2023-06-01" \
+    -H "anthropic-beta: oauth-2025-04-20" \
+    -H "content-type: application/json" \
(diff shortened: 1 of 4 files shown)


natb1:
## QA pass — autonomous (non-browser path)

Scripts-only change (dispatch chain internals); no app UI, so the QA data policy
and browser walkthrough do not apply. Verified the user-visible outcome — the
headless dispatch chain self-resuming once a rate-limit window reopens — through
the script test suite and shell checks.

### Items walked (all machine-verifiable)

| # | Item | Result | Evidence |
|---|------|--------|----------|
| 1 | Probe writes fresh telemetry in the existing schema | **PASS** | suite CASE 1 — exit 0, canonical `five_hour`/`seven_day` → `used_percentage`+`resets_at` (no format divergence) |
| 2 | Reseed-fired tick observes reopened window → positive target (reported stall regression) | **PASS** | suite CASE 6 — frozen pre-reset file (used 95%, past `resets_at`) overwritten; real `dispatch-target-workers` → target **8** |
| 3 | Probe failure degrades safely; tick still routes | **PASS** | refresh-failsafe + CASEs 2–5b — tick exits 0 after a non-zero probe; expired token / missing creds / missing+malformed headers each exit non-zero **without** writing the state file |
| 4 | Refresh runs before the budget read (ordering) | **PASS** | refresh-ordering — `refresh` logged before `select` |
| 5 | Changed scripts syntactically valid | **PASS** | `bash -n` on all four changed scripts |

### Counts

- PASS: 5 · FAIL: 0 · SKIP: 0
- Full `test-dispatch-scripts.sh` suite: **1655/1655 passed, 0 failed**

### Deferred to office-hours

None — every QA item was machine-verifiable.

### Bugs found

None.

natb1:
## Review phase — unified findings

`/code-review max --fix`, `/review`, and the surface-gated security fan-out (input-validation, secrets, red-team, `/security-review`) ran over the same diff (4 bash scripts). Security surface = `code`; no app/rules surface (no auth / data-exposure / Firebase reviewers warranted), `deps=false` (no dependency audit). **CodeQL: 0 open alerts.**

### Fixed
- **`dispatch-refresh-rate-limits` `header_val`** — removed the dead `s/\r$//` sed expression; the preceding `s/[[:space:]]//g` already strips CR. (`/code-review`) — `93fa78d8`

- **Token character-set validation** — the OAuth token is now validated against the base64url/JWT character set before it is placed in the `Authorization` header, so a corrupted/tampered credentials file (e.g. an embedded newline) can't inject a header. (input-validation) — `93fa78d8`
- **Model-name validation** — the model name is validated before being interpolated into the JSON request body, preventing malformed/injected JSON from a bad override. (input-validation) — `93fa78d8`
- **https-only endpoint** — the probe now refuses to send the bearer token over a non-TLS endpoint, closing the high-confidence red-team finding (a non-https `DISPATCH_REFRESH_RATE_LIMITS_ENDPOINT` would otherwise leak the token in cleartext). (red-team) — `93fa78d8`

Added test cases 5c/5d/5e covering all three guards; full suite 1661/1661 passing.

### Informational
- **Probe runs unconditionally, incl. manual `dispatch` ticks** (code-review + review, medium) — adds one `max_tokens:1` API call and up to ~10s latency to manual runs, where the status-line hook already refreshes telemetry. The PR deliberately documents the unconditional design ("every budget read this tick computed against fresh telemetry"); a headless-only guard is a reasonable alternative but reverses that documented choice, so left as-is.
- **OAuth token in `curl` argv → `/proc/<pid>/cmdline`** (secrets, medium) — visible to same-machine users for the ~10s request window. On the single-user dispatch host this is low real risk; the `curl --config` rewrite was deferred because it changes the probe's only un-tested (real-network) code path, where a silent breakage would disable the feature with no test to catch it. Worth revisiting if the probe gains integration coverage.
- **No dedicated partial-header test case** (review, low) — largely covered already: an absent header yields an empty value that fails the existing `^[0-9]+...$` validation, the same path exercised by cases 5a/5b.

### Dismissed
- **Default model ID may be invalid** (code-review, low) — verified valid: `claude-haiku-4-5-20251001` is the current Haiku 4.5 model ID. False positive.
- **`EXPIRES_AT` unbounded / 32-bit overflow** (review, low) — target is 64-bit Linux; an out-of-range `expiresAt` requires write access to the user's own `~/.claude/.credentials.json` (same privilege). Not a real concern.

### False-positive (security)
- **`DISPATCH_RATE_LIMITS_STATE_FILE` path traversal** (input-validation, low) — env-var seam; abuse requires env-var control over the dispatch process = same privilege as the credentials owner. A prefix guard also risks breaking the `mktemp`-based test seam.
- **`DISPATCH_REFRESH_RATE_LIMITS_HEADERS_FILE` arbitrary-file read / not disabled in prod** (input-validation + red-team, low) — env-var test seam, same-privilege; content is regex-validated before use. The input-validation pass rated this a false positive itself.
- **`ENDPOINT` override SSRF** (input-validation, low) — superseded by the https-only guard now applied; the residual requires same-privilege env-var control.

### Out-of-scope (security) — pre-existing, tracked as #1136
These are properties of the `rate_limits.json` → budget/reseed pipeline that **pre-date this PR**: the `statusLine` hook has always written the same unbounded schema from Claude Code's `.rate_limits`, and the new probe writes the identical format — neither writer introduces the gap.
- **`resets_at` unbounded → far-future reseed timer** (red-team, medium) — a far-future epoch is passed straight to `systemd-run --on-calendar`; `dispatch-schedule-reseed` bounds only the past case.
- **`used_percentage` unbounded → budget poisoning** (red-team, medium) — a value far above 100 pins the worker target at 0.
- **`ENDPOINT`-override forged telemetry → excess worker spawn** (red-team, high) — requires same-privilege env-var control over the dispatch process; an attacker who can set its environment already owns the chain and the credentials, so this grants no incremental capability. The https-only guard closes the cleartext-exfil variant.

Filed #1136 to bound the telemetry values at the budget/reseed boundary.

PR #1130, opened 2026-06-04
dispatch: binary weekly-pace gate + linear 5-hour worker ramp

natb1:
Closes #1125

Replaces the magnitude-coupled weekly→F→N pace pipeline in `dispatch-target-workers` with two decoupled controls:

- **Weekly pace is a binary gate.** `hw = W - used_weekly`; `hw <= 0` (usage at/over the pace curve) → 0 workers (pause), regardless of 5-hour usage. Otherwise the gate is open.
- **When open, worker count is a pure function of 5-hour usage.** `span = ceil5 - floor5`; `h5 = ceil5 - used_5h`; `N = clamp(round(max_workers * h5/span), 1, max_workers)` — `max_concurrent_workers` at/below `floor5` (50%), ramping linearly to 0 at/above `ceil5` (80%). A `span <= 0` guard returns the ramp's limit instead of dividing by zero (the validator allows `floor5 == ceil5`).

Weekly pace decides *whether* to spend; the 5-hour ramp decides *how many*.

Ramp at defaults (8 / 50 / 80): used_5h ≤50→8, 55→7, 60→5, 65→4, 70→3, 75→1, ≥80→0; used_5h strictly under `ceil5` that rounds below 1 still clamps to ≥1 (79→1).

### Removed knobs

`weekly_headroom_taper_pct` and `five_hour_headroom_taper_pct` are gone — the gate no longer scales by weekly-headroom magnitude and the ramp width is now `ceil5 - floor5`. Removed from `dispatch-target-workers`, `dispatch-config-load` (schema + validation), `reference.md`, and `target-workers.example.json`. The `floor5 <= ceil5` ordering check is retained; `--reopen-at` is unchanged (it already computed the `hw=0` crossing).

### Behavior change

Telemetry that yields 0 workers today (thin weekly headroom) now yields workers per the 5-hour ramp whenever under pace. Intended.

### Commits

1. Algorithm rewrite + knob removal (scripts/config/example)
2. Test suite rewritten for the new gate/ramp model
3. Reference doc updated

Full dispatch test suite: 1623/1623 passing.

PR #1137, opened 2026-06-04
dispatch: priority/main-broken bypass the worker-count gate, floored by token exhaustion

natb1:
Closes #1134

## What

`priority`-labeled work — including `dispatch:main-broken` issues, which are
already filed with the `priority` label — now bypasses the worker-count gate on
the autonomous dispatch path, exactly like a manual `/dispatch`. The one hard
stop is genuine token exhaustion.

Before this change, the autonomous tick computed `TARGET_N` from
`dispatch-target-workers` (the binary weekly-pace gate + linear 5-hour ramp from
#1130) and, when `LIVE_COUNT >= TARGET_N`, emitted `concurrency-cap` and exited
*before any selection work*. A waiting priority item was throttled by the pacing
budget, and a main that broke while at the cap was never surfaced — the early
exit fired before the main-broken health gate ran.

## How

Four units, one commit each:

1. **`dispatch-target-workers --exhausted`** — a telemetry-only query that prints
   `exhausted` when the 5-hour or weekly window is at/near 100% used
   (`used_percentage >= exhaustion_threshold_pct`, default 98) with `resets_at`
   still in the future, else `ok`. Missing/unreadable telemetry → `ok` (fail
   open). This distinguishes "paced to 0" (a self-imposed throttle priority work
   overrides) from "genuinely out of tokens" (a hard stop), computed
   independently of the pace/ramp math. Adds the `exhaustion_threshold_pct`
   config knob (validated as a `(0,100]` percentage).

2. **`dispatch-select-target --priority-only`** — a selection mode that returns
   the main-broken health-gate result if any, else the highest-priority
   `priority=1` PR or help-wanted issue, else `empty`. JIT scan suppressed; every
   existing skip rule (worktree ownership/reservation, `blocked_by`,
   ready-PR-closes-issue, not-ready PR) preserved. Non-priority items are skipped
   wholesale before the expensive per-item `gh` calls.

3. **At-cap three-way bypass in `dispatch-select-tick`** — on the autonomous
   no-arg path, when `LIVE_COUNT >= TARGET_N`:
   - exhausted → hard stop: release the lock, arm the reseed at the window reset,
     emit `concurrency-cap`, spawn nothing.
   - else a priority/main-broken item → `main-broken <sha>` (release lock, the
     lock-free diagnose job spawns) or `pr …`/`issue …` (hold the lock, append
     `GAP=1` — one gate-exempt worker, exactly like manual dispatch).
   - else `empty` → `concurrency-cap` + reseed, unchanged.

   This also closes the latent gap where a newly-red main at the cap was never
   diagnosed.

4. **`dispatch-tick` header doc** — documents the third gate-bypass case and the
   exhaustion hard floor. Doc-only.

Priority/main-broken workers are named `<N>-slug` like any worker, so
`claude_agents_count_busy_workers` still counts them — they suppress
non-priority fan-out on later ticks. The count is no longer a *gate* on selecting
them, but they still *count* toward the budget that paces non-priority dispatch.
The `--manual` and explicit `dispatch <N>` paths are untouched (they already skip
the gate).

## Testing

`test-dispatch-scripts.sh`: 1706/1706 pass. New coverage: `--exhausted`
ok/exhausted (incl. past-reset and missing-telemetry fail-open, config
threshold); `--priority-only` (priority PR, priority issue, main-broken gate,
empty, JIT-suppressed, mutual-exclusion guard); at-cap priority PR/issue spawn
(`GAP=1`, lock held, no reseed); at-cap main-broken (lock released); at-cap
exhausted hard stop (probe not consulted); at-cap empty → `concurrency-cap` +
reseed; and the worker-count regression (a priority-named busy worker is still
counted).

### DIFF
--- a/.claude/skills/dispatch-propagate/reference.md
+++ b/.claude/skills/dispatch-propagate/reference.md
@@ -385,6 +385,7 @@ baked into the script):
 | `five_hour_target_floor_pct` (`floor5`) | 50 | `used_5h` % at which workers reach max (gate open) |
 | `five_hour_target_ceiling_pct` (`ceil5`) | 80 | `used_5h` % at which workers reach zero (gate open) |
 | `max_concurrent_workers` | 8 | max worker count |
+| `exhaustion_threshold_pct` | 98 | used_% (either window) at/near 100, with resets_at in the future, treated as genuine token exhaustion — a hard stop even for priority/main-broken work |
 
 Recalibration: raise `weekly_curve_power` to back-load spend later in the
 week (a higher `p` makes the curve concave up, so `W` grows slowly early and
@@ -406,6 +407,13 @@ clamps the terminal `W(1)` below `target_weekly_usage_pct`. The config validator
 only cross-checks floor vs. cap when both appear in the same config file, so a
 floor raised against the baked-in cap default passes validation but still
 clamps.
+
+Keep `exhaustion_threshold_pct` above `five_hour_target_ceiling_pct` (default 98
+> 80) so the 5h ramp's zero-point and the exhaustion floor stay distinct: the
+band between them (80→98) is "paced to 0", which priority/main-broken work
+overrides, while only used_% at/above the threshold is genuine token exhaustion,
+a hard stop for all work. Setting the threshold at or below the ceiling would
+collapse that override band into a hard stop.
 
 **Missing-telemetry fallback.** When
 `~/.local/share/commons-dispatch/rate_limits.json` is missing or unreadable,
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-config-load
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-config-load
@@ -354,7 +354,7 @@ case "$CONFIG_TYPE" in
     # top-level value must be an object; all tunables are optional numbers.
     # `dispatch-target-workers` substitutes a baked-in default for any field
     # absent here, so absence is silent — only the type of present fields is
-    # checked. The six percentage fields carry an additional `<= 100` bound;
+    # checked. The seven percentage fields carry an additional `<= 100` bound;
     # weekly_curve_power and max_concurrent_workers have no upper bound.
     ERRORS=$(jq -r '
       if type != "object" then
@@ -408,6 +408,13 @@ case "$CONFIG_TYPE" in
             and ($cfg.five_hour_target_ceiling_pct | type) == "number"
             and ($cfg.five_hour_target_floor_pct > $cfg.five_hour_target_ceiling_pct)
          then "five_hour_target_floor_pct must be <= five_hour_target_ceiling_pct, got floor \($cfg.five_hour_target_floor_pct) > ceiling \($cfg.five_hour_target_ceiling_pct)"
+         else empty end),
+        (if ($cfg | has("exhaustion_threshold_pct"))
+            and ($cfg | has("five_hour_target_ceiling_pct"))
+            and ($cfg.exhaustion_threshold_pct | type) == "number"
+            and ($cfg.five_hour_target_ceiling_pct | type) == "number"
+            and ($cfg.exhaustion_threshold_pct <= $cfg.five_hour_target_ceiling_pct)
+         then "exhaustion_threshold_pct must be > five_hour_target_ceiling_pct (the override band collapses otherwise), got threshold \($cfg.exhaustion_threshold_pct) <= ceiling \($cfg.five_hour_target_ceiling_pct)"
          else empty end)
       end
     ' "$CONFIG_FILE")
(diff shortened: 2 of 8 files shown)


natb1:

This PR changes only dispatch orchestration scripts and docs (no app UI), so QA
ran the non-browser path against the new CLI query modes and the project's own
`test-dispatch-scripts.sh` harness.

**Items walked: 4 — PASS 4 / FAIL 0 / SKIP 0**

| # | Item | Result |
|---|------|--------|
| 1 | Full `test-dispatch-scripts.sh` suite | PASS — 1706/1706 passed, 0 failed |
| 2 | `dispatch-target-workers --exhausted` distinguishes genuine exhaustion from paced-zero | PASS |
| 3 | `dispatch-select-target --priority-only` mode recognized/implemented | PASS |
| 4 | `dispatch-target-workers --bogus` → exit 2 naming valid modes | PASS |

**Item 2 detail** (telemetry-injected via env overrides, `NOW=1000`):
- 5h used 99% + reset in future → `exhausted`
- weekly used 100% + reset in future → `exhausted`
- weekly used exactly 98% (threshold) → `exhausted`
- 5h used 85% (over the 80% ramp ceiling, under 98% exhaustion) → `ok` (paced-zero, a priority override — not a hard stop)
- missing/unreadable telemetry → `ok` (fails open)
- used 99% but reset already in the past → `ok` (window not actually exhausted)

**Deferred to office-hours:** none — every item was machine-verifiable.

**Bugs found:** none.

Clean autonomous pass — applying `dispatch:qa-done`.

natb1:
## Review phase (`/review-fix`)

One consolidated pass over the diff vs `origin/main`: `/code-review max --fix`, `/review`, and the surface-gated security fan-out (input-validation, secrets, red-team, built-in `/security-review`, plus inline CodeQL). Surface classified `code`; `deps=false` (no dependency audit) and `app_or_rules=false` (no app-domain reviewers). CodeQL returned no open alerts. Findings de-duplicated below; fixes landed in `a3b0d617`.

### Fixed
- **`dispatch-config-load`: missing cross-field validation** (`exhaustion_threshold_pct` vs `five_hour_target_ceiling_pct`) — flagged independently by code-review, `/review`, and the red-team pass. Without it, a config setting `exhaustion_threshold_pct <= five_hour_target_ceiling_pct` silently collapses the 80→98 priority-override band into a hard stop, neutering the bypass. `reference.md` documented the requirement but the validator never enforced it. Added a cross-field check (reject when `<=`, both present and numeric), mirroring the existing `floor5 <= ceil5` guard. — `a3b0d617`
- **`dispatch-config-load`: stale comment** — "The six percentage fields" corrected to "seven" after `exhaustion_threshold_pct` was added. — `a3b0d617`
- **`test-dispatch-scripts.sh`: PO6 coverage gap** — the `--priority-only` rejection guard covers three sibling modes but only `--qa` was tested. Added cases for `--priority-only --health-only` and `--priority-only --main-broken-sha`, each asserting exit 1. — `a3b0d617`
- **`dispatch-select-target`: pri=0 dead-loop clarity** — the output loop iterates `pri in 1 0`, but in `--priority-only` mode the scan loops filter every priority=0 item, so the pri=0 pass is a guaranteed no-op. Added a comment documenting the invariant (left in the shared loop intentionally — the empty-map iteration is harmless and keeps the two output paths identical). — `a3b0d617`

Full suite: **1723/1723 passing**.

### Informational (no change — deliberate design)
- **`dispatch-select-tick:206` — `--exhausted` probe fails open** (`|| EXHAUSTED="ok"`). Red-team flagged this as fail-open. Left as-is: it is consistent with the module's explicit fail-open convention (daemon UNKNOWN → `GAP=1`), and failing *closed* would be strictly worse — on a transient probe failure it would route to the exhausted branch, whose `dispatch-schedule-reseed` no-ops when telemetry is unreadable, risking a multi-hour dispatch stall. A single possible noise spawn (fail-open) is cheaper than a stalled chain (fail-closed). The exploit scenario requires already-privileged local replacement of the script binary.
- **`dispatch-select-tick:218` — `--priority-only` fails open** (`|| PRIO="empty"`). On a non-selection failure (e.g. transient `gh` error) the tick treats it as "no priority work" → `concurrency-cap` + reseed, which is the correct, benign degraded behavior; the next reseed re-probes. Garbage (non-empty unexpected) output is still caught by the `*)` arm (`exit 2`).
- **`dispatch-select-tick:189` — `dispatch-target-workers` invoked twice** at cap (once for the count, once for `--exhausted`). Efficiency only; the cost is bounded (conditional, off the hot path) and the two queries are semantically independent.

### Out-of-scope (security)
- **Priority-label access control** (`dispatch-select-target`, red-team, medium) — any collaborator with triage permission can queue `priority`-labeled issues to sustain one at-cap bypass spawn per tick. This is the intended bypass mechanism (#1134); the access-control surface is pre-existing GitHub repo-permission posture, not introduced by this diff, and the primary defense is repo label-permission configuration rather than dispatch code. Not filed as a follow-up (no stable scanner ID; single-maintainer repo).
- **Telemetry-file tampering** (`dispatch-target-workers`, red-team, low) — local write access to `rate_limits.json` can set a valid low `used_percentage` to silence the exhaustion floor. Requires already-privileged local user-level write access; numeric sanitization is correct. No code change warranted.

Secrets, input-validation, and the built-in `/security-review` pass returned no findings.

Issue #1029, opened 2026-06-02
media: shared local-first infrastructure (@commons-systems/local-first + MediaSource)


natb1:
## Charter framing

The media apps should embody local-first data ownership: a user's PDFs, EPUBs, and
audio files live as in-place files on their own disk, synced by a provider they choose,
read as the source of truth, with no SaaS custody and no lock-in. This epic factors the
File System Access (FSA) primitives — shared by the budget local-first program
(epic #1026) and now extended to `print` and `audio` — into a reusable
`@commons-systems/local-first` package plus a generic `MediaSource<T>` abstraction, so
all three apps (and future media apps) reuse one implementation rather than
re-deriving it per app.

The enabling primitive is the **File System Access API** (Chromium-approved, with OPFS
as fallback), which lets the browser persist a handle to an on-disk file or directory
and read it across sessions.

## Migration-safety constraint (binds every issue in this program)

The local-folder / unauthenticated path is **purely additive**. The Firebase Storage +
authenticated path stays the default and **fully functional in prod**. Both paths must
run in prod simultaneously, and FSA / unauthenticated usage must be **fully confirmed in
prod**, before any deprecation or removal of the Firebase Storage / authenticated path is
even considered. **Removing or deprecating Firebase Storage is OUT OF SCOPE** for this
program — a separate, gated future decision.

## Architecture overview

- An FSA handle on disk (a folder of the user's media files) is the **source of truth**.
- The `@commons-systems/idbutil` LRU blob cache is a **derived cache** — re-derivable
  from the on-disk files, never authoritative.
- **Firebase Storage** + `createMediaQueries` cloud metadata is **the shared /
  public-domain library** — untouched by this program and the default path in prod.
- A generic `MediaSource<T>` interface unifies the cloud library and the local folder
  behind one list / resolve-to-blob / metadata contract; apps union the two sources.

## What does NOT transfer from the budget program

The dispatch-heartbeat parse-job loop (budget #1023 / #1024) is budget-only — bank
statements need Claude to parse before they are usable. PDFs and audio files are directly
usable, so `print` / `audio` get the local-first benefit with **no ingestion-job
machinery**. No analog issues are filed.

## Existing substrate this builds on

`print` and `audio` already share substrate: both build their cache on
`@commons-systems/idbutil/lru-blob-cache` and read cloud metadata through
`@commons-systems/firestoreutil/media-queries` (`createMediaQueries`). The shared package
formalizes the rest of the local-first stack around that substrate.

[...]

PR #1142, opened 2026-06-05
media: @commons-systems/local-first — FsaHandleStore + OPFS fallback

natb1:
Closes #1032

## Summary

Adds a new shared workspace package `@commons-systems/local-first` (F1) — the
foundation primitive for the shared media local-first stack (parent epic #1029).
It persists a File System Access (FSA) handle across browser sessions so a
Chromium browser can re-open the same on-disk file/directory the user picked,
with no re-prompt. Pure infrastructure — no app wiring; `print`, `audio`, and
budget (#1018, #1021) consume it later.

## What's in the package

- **`capabilities`** — `detectFsaCapabilities()` (`{ filePicker, directoryPicker }`)
  and `isFsaSupported()`. Every probe is a guarded `typeof` check, so detection
  never throws on a browser lacking FSA.
- **`fsa-handle-store`** — `createFsaHandleStore({ app, dbName?, version? })`:
  - Persist / retrieve / remove a handle in an IndexedDB meta store, keyed
    `app:purpose` (reuses `@commons-systems/idbutil`'s memoized connection).
  - Permission flow: `queryPermission` / `requestPermission` wrappers, and
    `ensurePermission` — zero prompts when already `granted`, one request when in
    the `prompt` state.
  - `load(purpose)` — re-open a persisted handle and resolve its permission.
- **`src/fsa-types.d.ts`** — ambient declarations for the WICG FSA surface that
  TS's `lib.dom.d.ts` does not yet ship (the `Window` pickers and
  `FileSystemHandle.queryPermission` / `requestPermission`).

## Scope change during review: OPFS dropped

The issue originally proposed an OPFS fallback on non-FSA browsers. That was
dropped after review (issue #1032 updated to match). OPFS is an origin-private,
provider-unsynced browser store — not the user's on-disk files — so it cannot
stand in as the local-first source of truth that an FSA handle provides. On a
non-Chromium browser the bytes still come from the cloud, which is the existing
Firebase Storage path (already the default and fully functional in prod per the
epic's migration-safety constraint). So the fallback is simply: when
`isSupported()` is false, the consuming app routes to its existing cloud path —
no new storage layer here. OPFS's only legitimate role is as a derived
blob-cache backend, a separate optimization decided on caching merits and out of
scope for F1.

## Acceptance criteria

- [x] A picked handle persists across sessions in the IndexedDB meta store.
- [x] Permission already granted → zero prompts; `prompt` state → one request.
- [x] Non-FSA browser → `isSupported()` reports false so the app routes to its
      existing cloud path (no OPFS).
- [x] Capability detection never throws on a browser lacking FSA.
- [x] Covered by unit tests (16 tests across capabilities + handle store).

## Testing notes

`fake-indexeddb` uses `structuredClone`, which throws on method-bearing objects,
and there is no real FSA in the node test env. So persistence tests round-trip a
structured-clone-safe plain stand-in (`{ kind, name }`) — the IDB put/get code
path is identical for a real handle — while permission tests pass method-bearing
mocks directly to the permission methods. The full real-browser handle round-trip
is validated later when an app consumes the package as a PWA (#1034).

The root `package.json` workspaces edit means CI's changed-app resolution treats
this as a global change, so unit-tests / lint / typecheck run repo-wide on this
PR — expected, not a regression.

### DIFF
--- a/local-first/eslint.config.js
+++ b/local-first/eslint.config.js
@@ -1,0 +1,1 @@
+export { default } from "@commons-systems/config/eslint";
--- a/local-first/package.json
+++ b/local-first/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "@commons-systems/local-first",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "exports": {
+    "./capabilities": "./src/capabilities.ts",
+    "./fsa-handle-store": "./src/fsa-handle-store.ts"
+  },
+  "scripts": {
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "lint": "eslint src/"
+  },
+  "dependencies": {
+    "@commons-systems/idbutil": "*"
+  },
+  "devDependencies": {
+    "fake-indexeddb": "^6.0.0"
+  }
+}
--- a/local-first/src/capabilities.ts
+++ b/local-first/src/capabilities.ts
@@ -1,0 +1,35 @@
+/**
+ * Capability detection for the File System Access (FSA) API. Every check is a
+ * guarded `typeof` probe so it never throws on a browser (or test environment)
+ * that lacks the API.
+ */
+
+export interface FsaCapabilities {
+  /** `window.showOpenFilePicker` is available (pick an on-disk file). */
+  filePicker: boolean;
+  /** `window.showDirectoryPicker` is available (pick an on-disk directory). */
+  directoryPicker: boolean;
+}
+
+/**
+ * Probe the current environment for FSA support. Never throws — on any non-FSA
+ * browser (or non-browser context) the corresponding flag is `false`. When both
+ * flags are false the caller should route to its existing cloud path.
+ */
+export function detectFsaCapabilities(): FsaCapabilities {
+  const win = typeof window !== "undefined" ? window : undefined;
+  return {
+    filePicker: typeof win?.showOpenFilePicker === "function",
+    directoryPicker: typeof win?.showDirectoryPicker === "function",
+  };
+}
+
+/**
+ * True when the browser can let the user pick a persistable on-disk handle
+ * (file or directory). When false, callers should route to their existing
+ * cloud (Firebase Storage) path rather than the local-first on-disk path.
+ */
+export function isFsaSupported(): boolean {
+  const caps = detectFsaCapabilities();
+  return caps.filePicker || caps.directoryPicker;
+}
--- a/local-first/src/fsa-handle-store.ts
+++ b/local-first/src/fsa-handle-store.ts
@@ -1,5 +1,5 @@
 import { createDbConnection } from "@commons-systems/idbutil/connection";
-import { detectFsaCapabilities, type FsaCapabilities } from "./capabilities.js";
+import { detectFsaCapabilities, isFsaSupported, type FsaCapabilities } from "./capabilities.js";
 
 const HANDLE_STORE = "handles";
 const DEFAULT_DB_NAME = "fsa-handle-store";
@@ -54,6 +54,12 @@ export interface FsaHandleStore {
    * Load the persisted handle for (app, purpose) and resolve its permission.
    * Returns null when no handle is persisted. `permission` reflects the
    * post-`ensurePermission` state.
+   *
+   * Because this calls `ensurePermission`, which issues a permission request
+   * when the handle is in the `prompt` state, call `load` from within a user
+   * gesture. Outside a gesture the browser silently denies the request and
+   * `permission` resolves to `denied`; a non-null return therefore does not
+   * imply the handle is usable — check `permission === "granted"` before I/O.
    */
   load(
     purpose: string,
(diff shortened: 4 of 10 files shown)


natb1:
## QA pass — #1032 `@commons-systems/local-first` (FsaHandleStore + capability detection)

Autonomous QA (`/qa-fix`). Non-browser path — pure infrastructure library, no app wiring, no UI surface. The user-acceptance surface is the package's library behavior, verified by its unit suite plus type-check and lint. Each item maps to an acceptance criterion.

**Items walked (machine-verifiable):**

| # | Acceptance criterion | Result |
|---|---|---|
| 1 | A picked handle persists across sessions in the IndexedDB meta store | PASS |
| 2 | Permission already granted → zero prompts; `prompt` state → one re-grant | PASS |
| 3 | On a non-FSA (non-Chromium) browser, `isSupported()` reports false | PASS |
| 4 | Capability detection never throws on a browser lacking FSA | PASS |
| 5 | Package type-checks and lints clean | PASS |

**Counts:** PASS 5 · FAIL 0 · SKIP 0

**Evidence:**
- Unit suite: `vitest run --project local-first` → 2 files, 16 tests, all passing.
- `tsc --noEmit --project local-first` → exit 0.
- `eslint src/` (local-first) → exit 0.

**Deferred to office-hours:** none — every acceptance criterion is machine-verifiable for a headless library package; there is no visual/subjective surface.

**Bugs found:** none.

_Note: running vitest with cwd inside the package (`--root local-first` / `npm run test --prefix local-first`) fails because the root `vitest.config.ts` resolves per-project `vite.config.ts` paths relative to the discovered config dir; the canonical root invocation `vitest run --project local-first` (what CI uses) passes cleanly. No code issue._

natb1:

Single terminal review pass over the diff vs `origin/main`: `/code-review max --fix`, `/review`, and the surface-gated security fan-out (input-validation, secrets, red-team, built-in `/security-review`, auth, data-exposure, firebase) plus inline CodeQL and a differential `npm audit`. Findings unified and de-duplicated below. Fixes landed in commit `43ac29b0`.

### Fixed (code quality)
- **`fsa-handle-store.ts` — `isSupported()` was a construction-time snapshot.** Constructing the store before `window` exists (SSR / module init) would pin it to `false` forever even in capable Chromium. Now delegates to `isFsaSupported()` so it probes `window` on each call. _(code-review, review)_
- **`fsa-handle-store.ts` — `get()` could hang on transaction abort.** It registered `req.onerror` but not `tx.onerror`/`tx.onabort` (unlike `put`/`remove`); an externally aborted read left the promise unsettled. Both handlers added. _(code-review, review)_
- **`load()` user-gesture requirement was undocumented.** `load` calls `ensurePermission`, which requests permission in the `prompt` state; called outside a user gesture the browser silently denies. JSDoc now states this and that a non-null return does not imply `permission === "granted"`. _(review, red-team, auth)_
- **Test coverage for `isSupported()` added** (live-window probing across SSR→hydration). _(code-review)_

### Required (security) — fixed in `43ac29b0`
- **IndexedDB key-namespace collision via `:` (A01/A03, Tampering — medium).** The key was `` `${app}:${purpose}` `` with no separator validation, so `app='a:b'`/`purpose='c'` collided with `app='a'`/`purpose='b:c'`, breaking app-namespace isolation. `app` and `purpose` now reject a `:` with a clear error; covered by tests. _(input-validation, red-team, auth, code-review)_
- **Unchecked cast of the value read back from IndexedDB (A03, Elevation of Privilege — medium).** `get()` cast `req.result` straight to `FileSystemHandle`. A same-origin actor able to write IDB (malicious extension / XSS) could substitute a non-handle object. `get()` now validates the stored value is handle-shaped (`kind` ∈ {file, directory}) and rejects with a clear "malformed / may be tampered with" error rather than silently returning a bogus handle; covered by a test. _(red-team)_

### Informational
- **`load()` granted/denied unit coverage gap.** The composition (`get` → `ensurePermission`) for a real, permission-bearing handle cannot round-trip through the unit IDB layer — fake-indexeddb's `structuredClone` rejects method-bearing objects (the same constraint the PR's testing notes document). `get` and `ensurePermission` are independently covered; the full real-handle round-trip is validated when an app consumes the package as a PWA (#1034). No change. _(review)_
- **`store.capabilities` remains a construction-time snapshot.** The actionable part — `isSupported()` — is now dynamic; the residual `capabilities` field is non-actionable here. No change. _(red-team)_

### False-positive
- **Unbounded `purpose` length → quota DoS.** Requires same-origin code execution; a DoS-of-self, not a vulnerability. _(red-team)_

### Out-of-scope (security)
These are inherent FSA-API / IndexedDB design properties, not weaknesses the diff introduced, and none meets the follow-up-filing bar on its own — recorded here, not filed:
- `capabilities` field enables finer browser fingerprinting than the binary `isSupported()`. _(data-exposure)_
- Default shared `dbName` → isolation is key-prefix-only within an origin (origin-scoped IDB). _(auth)_
- `requestPermission` auto-denies outside a user gesture — a browser-API constraint, JSDoc-only mitigation. _(auth)_
- `load()` returns the raw handle whose `.name` is the user's on-disk name — inherent to the FSA contract (callers need the handle to read files). _(data-exposure)_

### Scans
- **CodeQL:** no open code-scanning alerts on the PR ref.
- **Dependency audit (differential vs merge-base):** the diff introduced no new advisories; the lockfile's pre-existing advisories are all `moderate`/`low` (no `high`/`critical`). No security follow-ups filed.
- **Secrets / firebase / built-in `/security-review`:** no findings.

All reviewers and scans ran — no partial coverage.

Issue #1022, opened 2026-06-02
budget: read-only directory access to display source statement files


natb1:
## Context

Part of the local-first data-ownership program (see parent epic). Transactions in the
budget originate from source statement files (`.qfx`, CSV, PDF) that live in the user's
synced statements folder. Showing the raw source file behind a transaction gives the
user provenance — a way to verify a parsed transaction against its original statement —
without any upload or re-parsing.

## Approach

Persist a directory handle (FSA) for the statements folder. Resolve a transaction's
recorded source-statement filename against that directory handle and render the raw
file in the UI (`.qfx`/CSV as text, PDF in an embedded viewer). Access is read-only —
no writes, no parsing, no upload.

## Acceptance criteria

- [ ] A statement-linked transaction can open its source file from the persisted
      directory handle.
- [ ] Access is read-only (no write, no parse, no upload).
- [ ] A missing or moved source file is handled gracefully (clear message, no crash).

## References

- Parent epic: local-first data ownership via the File System Access API.

PR #1143, opened 2026-06-05
budget: persist FSA file handle for picker-free .benc load

natb1:
Closes #1018

## What

First slice of the local-first data-ownership epic (#1026): persist a Chromium
**File System Access API** `FileSystemFileHandle` so the budget app re-opens the
same on-disk synced `.benc` across sessions without a fresh upload every time.

The real filesystem stays the single source of truth on every browser. Chromium
re-opens it via the persisted handle; non-Chromium browsers keep re-reading it via
the existing `<input type=file>` picker each session, unchanged. No browser-private
copy (OPFS) is kept — IndexedDB remains a cache of the parsed working data, never a
source of truth. BENC stays encrypted at rest; the password is prompted via the
existing `promptPassword` and held in memory only.

## How

Three units:

1. **`budget/src/idb.ts`** — persist the handle in the `meta` store
   (`FileHandleMeta` + `putFileHandle`/`getFileHandle`/`clearFileHandle`), and stop
   `storeParsedData` from wiping it: the clear loop now skips the `meta` store, so a
   data load overwrites the `upload` record while the separate `fileHandle` record
   survives. `clearAll` (the "Clear data" button) still drops the handle — the
   correct unlink-on-clear behavior.

2. **`budget/src/local-file.ts`** (new) — the entire FSA platform surface behind a
   small interface: `isFsaSupported`, `pickBencFile` (returns `null` on user cancel),
   `queryReadPermission`/`requestReadPermission` (mode `"read"`; #1019 upgrades to
   `"readwrite"`), and `readFileFromHandle`. Owns the minimal ambient type
   declarations the API needs (`showOpenFilePicker`, `queryPermission`/
   `requestPermission`), which are not in baseline `lib.dom.d.ts`.

3. **`budget/src/main.ts`** — the startup ladder and picker UI. `loadFromFile` is
   extracted from `handleFileUpload`'s core pipeline; `loadFromHandle` reads the
   handle (stale-handle read failures clear the handle and surface a re-link picker)
   then reuses `handleFileUpload`'s parse/validation error handling. On launch: a
   `granted` handle auto-loads with no picker; a `prompt` handle shows cached/seed
   data plus a one-click "Reload <name>" button; otherwise the existing cached/seed
   path runs unchanged. When FSA is unsupported, the `<input>` upload path is the
   active path, untouched.

## Acceptance criteria

- [x] The `FileSystemFileHandle` is persisted in the IDB `meta` store.
- [x] A returning session with permission `granted` auto-loads with no picker; with
      permission in `prompt` state, one click re-grants and loads.
- [x] A stale/invalid handle falls back to the re-link picker rather than erroring.
- [x] Non-Chromium browsers use the existing upload `<input>` path unchanged.
- [x] BENC decrypt path is unchanged.

## Verification

- `npx vitest run --root budget` — full suite green (892 passed, 7 skipped),
  including the new idb, local-file, and main ladder tests.
- `npx tsc --noEmit --project budget` — clean (ambient FSA declarations type-check,
  no `any` leaks).
- `npm run build --prefix budget` — succeeds.

Out of scope (later epic slices): write-back (#1019), `lastModified` reload (#1020),
PWA install (#1021), directory access (#1022).

### DIFF
--- a/budget/src/idb.ts
+++ b/budget/src/idb.ts
@@ -110,7 +110,12 @@ export async function storeParsedData(data: ParsedData): Promise<void> {
     stores[name] = tx.objectStore(name);
   }
 
-  // Clear all stores first (skip "meta" to preserve the persisted fileHandle record)
+  // Clear all stores first. The "meta" store is special: it mixes the
+  // parsed-data cache (the "upload" record) with capability records (the
+  // persisted "fileHandle"). A reload replaces the former and must preserve the
+  // latter, so delete only the "upload" key here rather than clearing the whole
+  // store — clearing it blindly would also drop the fileHandle, and skipping it
+  // blindly would silently retain any future meta key.
   const clearPromises: Promise<void>[] = [];
   for (const name of STORE_NAMES) {
     if (name === "meta") continue;
@@ -122,6 +127,13 @@ export async function storeParsedData(data: ParsedData): Promise<void> {
       }),
     );
   }
+  clearPromises.push(
+    new Promise((resolve, reject) => {
+      const req = stores.meta.delete("upload");
+      req.onsuccess = () => resolve();
+      req.onerror = () => reject(req.error);
+    }),
+  );
   await Promise.all(clearPromises);
 
   // Write all records
--- a/budget/src/idb.ts
+++ b/budget/src/idb.ts
@@ -67,6 +67,24 @@ export interface UploadMeta {
   exportedAt: string;
 }
 
+export interface FileHandleMeta {
+  key: "fileHandle";
+  handle: FileSystemFileHandle;
+}
+
+export async function putFileHandle(handle: FileSystemFileHandle): Promise<void> {
+  await put("meta", { key: "fileHandle", handle });
+}
+
+export async function getFileHandle(): Promise<FileSystemFileHandle | undefined> {
+  const record = await get<FileHandleMeta>("meta", "fileHandle");
+  return record?.handle;
+}
+
+export async function clearFileHandle(): Promise<void> {
+  await deleteRecord("meta", "fileHandle");
+}
+
 export interface ParsedData {
   transactions: IdbTransaction[];
   budgets: IdbBudget[];
@@ -92,9 +110,10 @@ export async function storeParsedData(data: ParsedData): Promise<void> {
     stores[name] = tx.objectStore(name);
   }
 
-  // Clear all stores first
+  // Clear all stores first (skip "meta" to preserve the persisted fileHandle record)
   const clearPromises: Promise<void>[] = [];
   for (const name of STORE_NAMES) {
+    if (name === "meta") continue;
     clearPromises.push(
       new Promise((resolve, reject) => {
         const req = stores[name].clear();
--- a/budget/src/local-file.ts
+++ b/budget/src/local-file.ts
@@ -1,0 +1,70 @@
+// File System Access API (FSA) surface for the budget app.
+//
+// `window.showOpenFilePicker` and `FileSystemHandle.queryPermission` /
+// `requestPermission` are not in baseline `lib.dom.d.ts`, so this module owns
+// the minimal ambient declarations. The FSA primitives let a Chromium browser
+// persist a `FileSystemFileHandle` (stored in IDB) and re-open the same on-disk
+// `.benc` across sessions; non-Chromium browsers fall back to the existing
+// `<input type=file>` upload path (detected via `isFsaSupported`).
+
+interface FileSystemHandlePermissionDescriptor {
+  mode?: "read" | "readwrite";
+}
+interface OpenFilePickerOptions {
+  multiple?: boolean;
+  types?: { description?: string; accept: Record<string, string[]> }[];
+}
+declare global {
+  interface Window {
+    showOpenFilePicker(options?: OpenFilePickerOptions): Promise<FileSystemFileHandle[]>;
+  }
+  interface FileSystemHandle {
+    queryPermission(descriptor?: FileSystemHandlePermissionDescriptor): Promise<PermissionState>;
+    requestPermission(descriptor?: FileSystemHandlePermissionDescriptor): Promise<PermissionState>;
+  }
+}
+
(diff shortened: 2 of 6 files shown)


natb1:
## Autonomous QA — #1018 (FSA persisted file handle, upload fallback)

**Result: clean autonomous pass.** All machine-verifiable items PASS; 0
needs-human-judgment items; no bugs.

The change is purely client-side (IndexedDB, FSA file picker/handle, startup
ladder). Its private-data / native-dialog flows (the FSA picker, the persisted
on-disk `.benc`, cross-session permission state) are out of scope for the
public-data walkthrough, so the QA plan focused on the public-data-reachable
**startup regression** and deferred the FSA-specific flows to the unit suite.

### Pre-QA acceptance check
`run-acceptance-tests.sh budget` — **443 passed, 1 skipped.**

### Browser walkthrough
Skipped — the Chrome extension permission could not be granted in this
autonomous background session (navigate denied repeatedly; no user present for
the permit popup). Fell through to non-browser verification using the passing
acceptance suite as machine-verifiable evidence (qa-fix Step 3.d.1).

### Items walked (machine-verifiable) — PASS 3 / FAIL 0 / SKIP 0
1. **App startup renders the seed view** — the new `initialize()` checks
   `getFileHandle()` first and, with no persisted handle, falls through to the
   seed branch. Asserted by every spec that loads `/transactions` with empty IDB
   (`#seed-data-notice` visible). PASS.
2. **No startup console errors from the FSA wiring** — `isFsaSupported()`
   detection and the conditional `uploadLabel` handler register without throwing.
   PASS.
3. **Legacy `<input>` upload path intact** — the new FSA `preventDefault()` on
   the label does not break the legacy `.upload-input`; upload (valid, invalid,
   encrypted correct/wrong password, clear-to-seed) tests pass, and
   `storeParsedData`'s new meta-skip does not break parsed-data storage/reload.
   PASS.

### Deferred to automated tests (private-data / native-dialog flows)
Not needs-human-judgment items — covered by the unit suite CI runs:
- `main.test.ts` — auto-load when permission `granted`; one-click reload button
  when permission is `prompt`; stale-handle clear + re-link error when the file
  is gone.
- `local-file.test.ts` — `isFsaSupported()` capability detection (non-Chromium
  `<input>` fallback path).
- `upload.spec.ts` — BENC decrypt path (encrypted-fixture tests), unchanged.

natb1:
## Review phase — consolidated findings

Single pass over the diff combining `/code-review max --fix`, `/review`, and the surface-gated security fan-out (input-validation, secrets, red-team, `/security-review`, auth, data-exposure, firebase), plus inline CodeQL. Findings de-duplicated into one disposition. All fixes landed in `a5528137`.

### Fixed
- **`main.ts` reload button stranding** (`showReloadPrompt`, ~main.ts:414) — when the re-grant picker was canceled, `button.remove()` ran unconditionally, leaving the user with no reload affordance and a hidden upload control. Now guarded on an actual transition to `local` state. _(code-review)_
- **`main.ts` permission-`denied` dead end** (`initialize`) — a `denied` handle was kept in IDB forever while showing stale cached/seed data with no explanation or recovery. The browser never re-prompts for a denied handle, so it is now dropped on startup and the user gets an "unlinked" notice. _(review)_
- **`idb.ts` meta-store clear intent** (`storeParsedData`) — replaced the blanket "skip the whole `meta` store" with an explicit `delete("upload")`, so a reload still preserves the persisted `fileHandle` without silently retaining any future `meta` key. Behavior-equivalent for the current key set; clearer intent. _(review)_
- **Tests** — added coverage for the content-validation handle-clear, the `denied` startup path, and the reload-button→picker fallback; reset `getMeta`/`storeParsedData` mocks per test to avoid cross-test pollution. _(review)_

### Required (security)
- **Persisted handle survives a failed content-validation load** (`loadFromHandle`, ~main.ts:320) — high confidence. A handle pointing at a non-budget file (wrong file picked) was auto-loaded, failed `parseUploadedJson`, but stayed in IDB, so every subsequent startup re-attempted and re-failed the load — a permanent failing-load loop with no self-recovery. Fixed by running the parse pipeline directly in `loadFromHandle` and calling `clearFileHandle()` on `UploadValidationError`. A wrong decryption password throws a different error and keeps the handle (the file is valid; the user can retry). OWASP A04:2021 Insecure Design / STRIDE DoS. Fixed in `a5528137`. _(red-team)_

### Informational
- **`pickAndLoad` outer try/catch is currently unreachable** (~main.ts:315) — `loadFromHandle` now surfaces its own errors and does not rethrow, so the wrapping catch never fires. Harmless; left in place as defense-in-depth. No change. _(review)_

### Dismissed
- **Reload button appended to `authContainer`** (~main.ts:405) — cosmetic nav layout placement, no functional impact. _(review, low)_
- **`isFsaSupported()` checked at load-time for the click handler vs per-event for keydown** (~main.ts:329) — it reads a static runtime property (`window.showOpenFilePicker`) that cannot change mid-session; the inconsistency is cosmetic. _(review, low)_
- **`pickBencFile` returns `handles[0]` without an empty-array guard** (`local-file.ts:43`) — the FSA spec guarantees at least one handle when `multiple: false` and the user confirms; the empty case is unreachable. _(review, low)_

### False-positive (security)
- **`handle.name` rendered in the reload button text** (~main.ts:404) — `textContent` prevents HTML injection; injecting a misleading filename requires attacker-controlled IDB write on the same origin, which already implies full origin compromise, making this moot. _(red-team, low)_

### Deferred
- **FSA picker accepts unencrypted `.json`, enabling passwordless auto-load** (`local-file.ts:41`) — genuine concern, but the existing `<input>` upload path already accepts `.json` (not a regression from this PR); restricting the picker or adding an unencrypted-file warning is a behavior change beyond this slice's scope. Filed as #1147 (blocked by #1018). _(red-team, medium)_

### Security scan coverage
- input-validation, secrets, auth, data-exposure, firebase, and the built-in `/security-review` all returned **clean**.
- CodeQL: **0 open alerts** for this PR ref.
- Dependency audit: **skipped** — the diff touches no `package.json` / `package-lock.json`.

PR #1157, opened 2026-06-05
budget: encrypted write-back of .benc via FileSyncingDataSource

natb1:
Closes #1019

## Summary

S2 of the local-first data-ownership epic (#1026). S1 (#1018) gave the budget app a
persisted FSA `FileSystemFileHandle` so it re-opens the user's on-disk `.benc` across
sessions — but load-only. This PR closes the loop: UI edits now write back, encrypted
and in place, to that same on-disk file, so the user's synced file becomes the durable
source of truth instead of a re-uploaded copy.

A new `FileSyncingDataSource` decorator over `IdbDataSource` mirrors every mutation to
disk via a debounced (`800ms`), single-flight encrypted write:
`exportToJson → encrypt → handle.createWritable`. Reads delegate straight through. The
scheduler state (handle, password, debounce timer) lives at module scope in a new
`file-sync.ts` — `createDataSource()` builds a fresh `IdbDataSource` on every render, so
per-instance state would drop a pending write across a navigation.

## What changed

- **`local-file.ts`** — upgraded the FSA permission helpers to `readwrite`
  (`queryReadWritePermission` / `requestReadWritePermission`) — a single grant covers
  both load and write-back rather than prompting twice. Added `writeFileToHandle`, which
  overwrites the on-disk file via `createWritable` (closing the stream in a `finally`).
- **`file-sync.ts`** (new) — module-scoped write-back coordinator: `configureFileSync`
  arms it, `scheduleWriteBack` debounces, `flushWriteBack` force-runs pending writes
  (single-flight), `resetFileSync` disarms. `doWrite` re-checks/requests permission
  defensively, then `exportToJson → encrypt → writeFileToHandle`. Write failures are
  logged, never thrown, so a failed write can't crash the editing UI.
- **`data-source.ts`** — thin `FileSyncingDataSource` decorator: 13 reads delegate
  straight through; 16 mutations delegate and then call `scheduleWriteBack()`.
- **`main.ts`** — wraps the local `DataSource`; arms sync on a successful encrypted
  handle load (`configureFileSync(handle, importPassword)`); secures readwrite inside the
  picker user-gesture in `pickAndLoad`; force-flushes on `visibilitychange → hidden`;
  disarms via `resetFileSync()` on clear-data.

## Acceptance-criteria mapping

- **UI edits persist back to the on-disk `.benc` in place (a handle write, not a
  download)** → `doWrite` → `writeFileToHandle` / `createWritable`, triggered by the
  decorator on every mutation.
- **Derived key in memory for the session only, never persisted** → reuses the in-memory
  `importPassword`; passed to `configureFileSync`, never written to IDB/storage.
- **Pending write flushed on `visibilitychange → hidden`** → the top-level listener calls
  `flushWriteBack`.
- **Seed mode (no file handle) unaffected** → `SeedDataSource` stays unwrapped;
  `scheduleWriteBack` no-ops when unconfigured; `resetFileSync` runs on clear.
- **Written bytes remain BENC-encrypted** → `doWrite` always `encrypt`s before writing,
  and write-back is armed only when `importPassword` (an encrypted source) is present.

## Out of scope

The Export button stays as the manual download path. The non-FSA `<input>` upload path
keeps no handle, so it does not write back (users still Export). Reload-on-external-change
is the sibling issue #1020.

## Verification

- `npm run test --prefix budget` — green (new `file-sync` suite; updated
  `local-file` / `data-source` / `main` suites).
- `npm run build --prefix budget` — production build succeeds.


natb1:
## QA pass — encrypted write-back of `.benc` (#1019)

**Autonomous QA: clean pass.** No bugs, no needs-human-judgment items.

### Pre-QA acceptance check
PASS — **441 passed, 3 skipped, 0 failed** (Playwright, run in-WSL). Covers seed-mode boot, navigation across all views, and the upload / encrypted-decrypt path.

### Items walked
| # | Item | Verification | Result |
|---|------|--------------|--------|
| 1 | Seed mode is unaffected by the write-back wiring | machine-verifiable | **PASS** |

- **PASS: 1 · FAIL: 0 · SKIP: 0**

Item 1 evidence: the acceptance suite boots the app in seed mode and navigates every view (transactions, budgets, accounts, rules). The new wiring — the `FileSyncingDataSource` decorator, the `file-sync.js` imports, and the top-level `visibilitychange` listener — introduced no boot regression. `scheduleWriteBack()` is a no-op until `configureFileSync` arms a handle, so seed mode never touches the write-back path. Satisfies the "Seed mode (no file handle) is unaffected" criterion.

### Interactive browser walkthrough — not run
The Chrome extension was unavailable this session: only a Windows browser is connected, and the `navigate` permission was denied on both the call and its retry. Per the QA fallthrough, the one reachable item was verified via the non-browser path instead (the in-WSL acceptance suite above, which exercises the same seed-mode boot + navigation more reliably than the Windows-Chrome→WSL-emulator bridge).

### Deferred — not reachable via public seed data
The remaining acceptance criteria are only reachable by picking a real on-disk `.benc` through the FSA file picker and entering a session password (private data behind a user file-picker interaction):
- UI edits write back to the on-disk `.benc` in place (handle write, not a download)
- The derived key lives in memory for the session only, never persisted
- A pending write is flushed on `visibilitychange → hidden`
- Written bytes remain BENC-encrypted

These are out of scope for the public-data walkthrough per the QA data policy and are covered by the unit suite — `file-sync.test.ts`, `local-file.test.ts`, `main.test.ts`, `data-source.test.ts` — green in CI.

natb1:

`/code-review max --fix`, `/review`, and the surface-gated security fan-out
(input-validation, secrets, red-team, `/security-review`, auth, data-exposure,
firebase) ran over the same diff. CodeQL: **0 open alerts**. Dependency audit:
**not run** (no `package.json`/lockfile change). All findings below are
de-duplicated across sources.

### Fixed — `63a2b8eb`

- **`local-file.ts` `writeFileToHandle` could destroy the on-disk file on a failed
  write.** `createWritable()` defaults to `keepExistingData:false`, so a `close()`
  after a failed `write()` commits a truncated/empty file over the user's `.benc`.
  Now `abort()`s the writable on failure and rethrows.
  _(code-review; independently confirmed by red-team [high] and auth.)_
- **`file-sync.ts` flush did not drain mutations made during an in-flight write, and
  double-wrote on concurrent flush.** The "needs another write" signal now comes from
  mutations (`scheduleWriteBack` sets a `dirty` flag), not from flush calls, so a
  concurrent flush (e.g. `visibilitychange→hidden`) waits until every pending write —
  including one queued mid-flight — is on disk, without redundantly re-writing
  unchanged state. _(code-review, review.)_
- **Clear-data could clobber the on-disk file with a stale/empty snapshot.** A write
  in flight when the user clears data captured the handle/password locally and would
  still run, re-reading the now-cleared IDB. A generation counter now makes `doWrite`
  bail before touching disk if the session was reset or re-armed while it ran.
  _(review, red-team, auth — all the `resetFileSync`-vs-in-flight findings.)_

Tests added: concurrent-flush drain, mutation-during-write, reset-mid-write.

### Deferred → follow-up #1166

- **Write-back failures are logged but never surfaced to the user.** A persistent
  failure (revoked permission, disk full) silently diverges IDB from the on-disk
  `.benc`. Filed as #1166 (blocked_by #1019) — the error-surfacing layer is out of
  scope for the write-back mechanism this PR delivers. _(review, red-team [high],
  auth — folded into one follow-up.)_

### Informational — no change

- **`visibilitychange→hidden` flush is best-effort.** If the OS kills the tab before
  the async write completes, the last edit can still be lost. The drain fix above
  tightens this, but the page-termination window is an inherent browser limitation;
  the acceptance criterion (flush on hidden) is met.
- **Plaintext `.benc` loaded via an FSA handle does not write back.** Write-back is
  armed only when an `importPassword` exists (`main.ts`), i.e. encrypted files —
  consistent with the epic's "written bytes remain BENC-encrypted" constraint. No
  user indication that a plaintext file is read-only; minor, deliberate.
- **`requestReadWritePermission` re-request inside `doWrite` runs outside a user
  gesture** (debounced timer / visibilitychange) and harmlessly no-ops; the granted
  case is covered by the `queryReadWritePermission` check and the pick-time grant.
- **Raw error object passed to `logError` on write failure** (`doWrite` catch).
  WebCrypto/`DOMException` errors do not echo plaintext or the password, and no remote
  log sink is registered (console only) — exposure is speculative. _(data-exposure.)_
- **Session password held in module scope** is readable via DevTools / a malicious
  extension. This is the intended in-memory-only session-key design (#1019/#1026); the
  trade-off was decided in the epic. _(data-exposure.)_
- **Password is structure-cloned to the crypto worker on each encrypt** (`crypto.ts`,
  pre-existing, not touched by this diff). Deriving the key once in the worker would
  narrow the window; out of scope (pre-existing architecture). _(red-team.)_

### Dismissed

- **Empty-password write-back failure via `configureFileSync`** (input-validation,
  medium). Unreachable: `decrypt` (`crypto.ts`) rejects an empty password, so
  `importPassword` is never an empty string when `configureFileSync` is called.

### Security (no real vulnerabilities)

- secrets, `/security-review`, firebase: **no findings**.
- No CodeQL alerts; no dependency changes to audit.
- No `required` security finding left unresolved.

PR #1165, opened 2026-06-06
budget-etl: config-driven ingest from Downloads with snapshot rotation

natb1:
Closes #1158

## Summary

Makes the interactive `/budget-etl` skill config-driven. The skill previously
hardcoded three machine-specific paths — `~/Downloads`, `~/statements`, and
`~/Desktop/budget*.enc.json`. On this machine the statement archive and
snapshots live on a Google Drive mount and downloads land in the Windows
profile, so these locations now come from an uncommitted, per-machine config
file instead of being baked into the repo.

## What changed

- **`budget-etl` config type** in `dispatch-config-load` (+
  `budget-etl.example.json`). A flat top-level object with four required
  non-empty string path fields: `downloads`, `statements`, `snapshotDir`,
  `current`. Loaded from the shared `dispatch.config/budget-etl.json`; absent →
  `no-config`, invalid → exit 1 with a field-naming error. Paths may contain
  spaces (`/mnt/g/My Drive/...`).

- **`ingest-downloads.sh`** — a pure, testable mover invoked as
  `ingest-downloads.sh <downloads-dir> <statements-dir>`. It enumerates
  statement files (`qfx,ofx,csv,pdf`, non-recursive), classifies each via the
  existing `identify-qfx.sh` into `institution/account`, and moves them into
  `statements/<institution>/<account>/`, printing one `<src> to <dest>` line per
  move. Classify-all-first: if any file's `<ORG>` is unknown it surfaces the
  file(s) and moves nothing, rather than guessing. A target-name collision
  appends a timestamp before the extension instead of overwriting. Moving a file
  out of downloads is the "new statement" dedup.

- **`budget-etl/SKILL.md` rewritten config-driven.** New Step 0 resolves and
  validates config (a missing/unmounted Drive dir fails loudly — no silent
  fallback to a stale local path). Step 1 calls `ingest-downloads.sh`. The merge
  reads rules and prior transactions from `current`, writes a new immutable
  `snapshotDir/budget-<ts>.enc.json`, then replaces `current` with a copy of it
  (`cp -f`, not `mv -f`, so the snapshot stays in `snapshotDir` as history). A
  symlink is deliberately not used — `ln -s` fails on the `/mnt/g` Drive mount.
  No `~/Downloads`, `~/statements`, or `~/Desktop` literals remain.

- **Tests** in `test-dispatch-scripts.sh`: 25 new assertions covering the
  `budget-etl` schema validation (valid round-trip with spaces, missing /
  non-string / empty-string field rejection, `no-config` on absent file) and the
  `ingest-downloads.sh` move / collision / unknown-ORG-stops-and-moves-nothing
  behavior. Full suite: 1822/1822, 0 failures.

## Out of scope

Converging the autonomous parse-job snapshot model (#1024, in-place `.benc`
overwrite) onto this immutable-snapshot-plus-`current`-pointer scheme — tracked
separately.

## Test plan

- `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` →
  1822/1822 passed, 0 failed.
- Manual: `dispatch-config-load budget-etl` against a temp config dir — valid
  config prints normalized JSON (exit 0); missing/non-string/empty field exits 1
  with a descriptive error; absent file prints `no-config`.
- Manual ingest smoke: synthetic QFX in a temp downloads dir → correct
  `statements/<institution>/<account>/` placement and `<src> to <dest>` line;
  collision re-run produces a timestamp-suffixed name; unknown-ORG run exits
  non-zero and moves nothing.


natb1:

Backend/scripts PR (config-driven `budget-etl` ingest). No browser component in
the diff, so QA ran the non-browser path against the issue's acceptance
criteria. The new behavior lives in `dispatch-config-load` (the `budget-etl`
schema), `ingest-downloads.sh`, `budget-etl.example.json`, and the rewritten
`budget-etl` SKILL.md — all verified directly with synthetic data.

### Items walked (machine-verifiable)

| # | Item | Result |
|---|------|--------|
| 1 | `dispatch-config-load budget-etl` validates a good config | **PASS** — exit 0; all four paths round-trip, spaces in `My Drive` preserved |
| 2 | Missing config → `no-config`, exit 0 | **PASS** |
| 3 | Invalid config → exit 1, stderr names field | **PASS** — missing `current`, non-string `downloads`, empty `snapshotDir` each named |
| 4 | `budget-etl.example.json` exists beside other examples | **PASS** — four keys present |
| 5 | `ingest-downloads.sh` moves to `<institution>/<account>/`, prints `<src> to <dest>` | **PASS** — `amex.qfx` → `american_express/12345/amex.qfx` |
| 6 | Unknown `<ORG>` stops, moves nothing | **PASS** — classify-all-first; both files left in downloads, exit 1, error surfaces `NOPE` |
| 7 | Name collision → timestamp suffix, original untouched | **PASS** — `amex.<ts>.qfx` created; original `OLD-SENTINEL` preserved |
| 8 | No `~/Downloads` / `~/statements` / `~/Desktop` literals in SKILL.md | **PASS** — grep: no matches |
| 9 | Paths containing spaces handled end-to-end | **PASS** — ingest moved file under spaced statements root |

Bonus edge cases: empty downloads dir → graceful `no statement files`;
nonexistent downloads dir → clear error, exit 1.

### Counts

- **PASS: 9 / FAIL: 0 / SKIP: 0**
- Deferred to office-hours (needs human judgment): **0**
- Bugs found: **0**

### Out of scope — private-data flow (per QA data policy)

The end-to-end encrypted merge (final merge reading rules + prior transactions
from `current`, writing `snapshotDir/budget-<ts>.enc.json`, then rotating
`current` by file copy) drives the `budget-etl` binary against the user's
**encrypted** snapshot on the Google Drive mount and requires
`BUDGET_ETL_PASSWORD` (GPG/`pass`). This is a private-data flow not reachable by
the autonomous walkthrough; coverage is deferred to the user's own `/budget-etl`
runs and the SKILL.md-documented flow. It is noted as out-of-scope, not recorded
as a needs-human-judgment item.

Clean autonomous pass.

natb1:
## Review phase — unified findings

One pass over the branch diff combining `/code-review max --fix`, `/review`, and the surface-gated security fan-out (input-validation, secrets, red-team, built-in `/security-review`, plus inline CodeQL). Surface classified `code`; `deps=false` so no dependency audit ran. Fixes applied in `fa55374`.

### Fixed
- **`ingest-downloads.sh` / `SKILL.md` Step 0 — `$statements` existence guard** (`code-review`). The skill verified `downloads`, `current`'s parent, and `snapshotDir` but not `statements`; an unmounted Drive would let `mkdir -p` silently create a local tree. Both the script and the Step 0 doc block now fail loudly when `statements` is absent. — `fa55374`
- **`SKILL.md` Step 0 — config-load exit code was swallowed** (`review`, high). `CFG=$(dispatch-config-load …)` discarded the non-zero exit on an invalid config, so the prose's "stop and surface that error" never fired. Now `… || { echo …; exit 1; }`. — `fa55374`

### Required (security)
- **`ingest-downloads.sh` — path traversal via crafted `<ACCTID>`** (`input-validation`, `red-team`; OWASP A03:2021 Injection; STRIDE Tampering; **high**). `identify-qfx.sh` canonicalizes `<ACCTID>tok|../../escape</ACCTID>` to the account component `../../escape` without stripping path separators; that value flowed into `mkdir -p "$statements/$institution/$account"` and `mv`, so a crafted statement file dropped in `Downloads` (a realistic drive-by-download vector) escaped the archive root for arbitrary file write. Fix: PASS 1 now rejects any `institution`/`account` that is empty or contains `/` or `..`, routing it through the existing classify-all-first failure path so nothing moves. — `fa55374`
- **`dispatch-config-load` — `budget-etl` schema accepted relative / `..` paths** (`red-team`; OWASP A05:2021; STRIDE EoP; **medium**). The four path fields were type/emptiness-checked only. They become filesystem destinations the skill writes to, so the schema now also requires each to be an absolute path with no `..` component. — `fa55374`

### Out-of-scope (security)
- **`identify-qfx.sh:45` — ACCTID canonicalization does not strip path chars** (`input-validation`, `red-team`; high). Same root issue as the path-traversal finding, but in a pre-existing file the diff does not touch. The in-scope `ingest-downloads.sh` guard fully closes the traversal at the move boundary, so this defense-in-depth change is redundant for the vulnerability and is not filed as a separate follow-up.

### Informational
- **`ingest-downloads.sh` collision branch — no-extension `else` is unreachable** for the enumerated `*.qfx/*.ofx/*.csv/*.pdf` set. Left in place as harmless defensive handling.
- **`ingest-downloads.sh` collision name — no second existence re-check.** Two runs within the same one-second timestamp could overwrite a prior timestamped file; a narrow edge case, not changed.
- **`ingest-downloads.sh` enumerates `csv`/`pdf` that `identify-qfx.sh` cannot classify**, so any such file triggers the classify-all-first stop. This is the issue's intended "surface to the user to map" design.
- **`SKILL.md` Step 2 — first-run `$current` absence** has no bootstrap note.

### Dismissed
- **`test-dispatch-scripts.sh` move-output tests use manual `TOTAL/PASS/FAIL`** instead of `assert_eq` — a style inconsistency with surrounding tests, not a defect.

### Other passes
- **Secrets:** clean — no hardcoded credentials; the `BUDGET_ETL_PASSWORD` reference is to `pass`, not a committed value.
- **CodeQL:** no open alerts on the PR ref.

New tests added in `fa55374` cover the config abs-path/`..` rejection and the ingest path-traversal rejection; full suite passes (1833/1833).

PR #1167, opened 2026-06-06
budget: reload .benc on external change via lastModified watermark

natb1:
Closes #1020

Reloads the on-disk `.benc` when an external writer (another device's filesystem
sync, or a dispatched `budget-etl` parse-job) changes it underneath the app, so a
newer external write is never overwritten by stale in-memory state.

`file-sync.ts` now stashes the on-disk file's `lastModified` as a watermark: at each
load (via `configureFileSync`) and immediately after each write-back (re-reading the
just-written file in `doWrite`). On window `focus`, `main.ts` re-reads the handle's
`lastModified` and reloads through the existing `decrypt → parse → store` pipeline
when it exceeds the watermark. Stamping the watermark on the app's own write-back is
what prevents a write-back from self-triggering a reload.

Conflict resolution is last-write-wins at whole-file granularity — no CRDT, no
per-field merge. Launch reload is already handled by the existing `initialize()`
auto-load; this adds the `focus` path and the watermark baseline. The watermark is
in-memory only (no drift-prone persisted side-value). Seed and non-FSA upload
sessions are unaffected (no armed handle → the watcher is a no-op).

Closes #1020.


natb1:
## QA (autonomous) — #1020: reload `.benc` on external change via `lastModified` watermark

App: **budget** · QA server: Vite dev server, **public seed data only**.

### Result: clean autonomous pass ✅

| Item | Verification | Result |
|---|---|---|
| 1. App loads cleanly with public seed data | DOM render + console + network | **PASS** |
| 2. Window `focus` in a seed/non-FSA session is a safe no-op | event dispatch + DOM/console/network assertions | **PASS** |

- **PASS: 2 · FAIL: 0 · SKIP: 0 · needs-human-judgment: 0**

**Item 1** — Full budget UI rendered (header, nav, "Viewing example data" seed banner, budget charts). Console: only DEBUG (vite connect, analytics-disabled); no errors. The seed banner confirms **seed / non-FSA mode** (no armed FSA handle).

**Item 2** — Dispatched a window `focus` event. The new handler (`reloadIfExternallyChanged`) early-returns because `getSyncHandle()` is `null`: no synchronous throw, no async console error/exception, no reload-triggered network request, and the app stayed in its seed state (banner + header intact, `readyState: complete`). Confirms the watermark-reload wiring **does not disturb non-FSA sessions** — the user-visible guarantee reachable on the public-data path.

### Deferred to automated tests (out of scope for the public-data walkthrough)

The **core** behavior — detect a newer on-disk `.benc` via the `lastModified` watermark on focus and reload (decrypt → parse → store, last-write-wins), and stamp the watermark on the app's own write-back so it doesn't self-trigger — is reachable only through a **real FSA handle to a user-owned `.benc`** with an external write. That is private-data + real-filesystem flow, not present in public seed data and not scriptable through the file picker. It is covered by the **vitest unit suite (CI)**:
- `main.test.ts`: reload-when-newer, no-reload-when-not-newer, no-op-in-seed-mode.
- `file-sync.test.ts`: watermark stamped from the post-write file; `getSyncHandle`/`getLastSyncedModified` reset on `resetFileSync`.

### Pre-QA acceptance check — note (not a blocker)

The pre-QA acceptance run reported **436 passed, 3 failed**. The 3 failures are all `e2e/prerender.spec.ts` "static HTML output", failing in `beforeAll` with `ENOENT … budget/dist/index.html`. This is a **pre-existing test-harness gap, not a #1020 regression**, and was treated as a documented environment limitation rather than a QA blocker:

- `prerender.spec.ts` is **unchanged by this PR** (last touched in #634 / #362).
- `run-acceptance-tests.sh` **skips `npm run build`** when given an external base URL (the path `/qa-fix` uses), so `budget/dist/` is never produced; the spec reads it from disk and ENOENTs on **every branch** — branch-independent.
- The block isn't tagged `@hosting`/`@testonly`, so the URL-path `--grep-invert` doesn't exclude it.
- The #1020 diff touches only `file-sync.ts` and `main.ts`'s focus handler — nothing related to prerendering / `dist`.
- CI's full `acceptance` job builds `dist/` and covers prerender properly — **CI is authoritative** for this coverage.

**Recommended permanent fix:** tag the `prerender.spec.ts` "static HTML output" describe block `@hosting` so the dev-server URL path of `/qa-fix` skips it (matching how the build-artifact dependency is handled elsewhere).

<sub>Walkthrough verified via DOM/console/network assertions; screenshot capture hit a CDP timeout, so the attached GIF (`qa-fix-walkthrough-1020.gif`) has a single frame.</sub>

natb1:

One pass over the branch diff combining `/code-review max --fix`, `/review`, and the surface-gated security fan-out (input-validation, secrets, red-team, `/security-review`, auth, data-exposure, firebase) plus inline CodeQL. Findings de-duplicated across sources into one disposition.

### Fixed — `95c3c558`
- **Password re-prompt loop on every external reload** (`main.ts`): reload now reuses the already-read `File` and the cached session password via `loadFromFile(file, cachedPw)`, so the decrypt dialog no longer reappears on every external write.
- **Permanent FSA-handle drop on bad external content** (`main.ts`): the reload path no longer calls `clearFileHandle()` when an external write fails content validation — the handle is valid, only that file version is bad. Previously a single bad `budget-etl` write would unlink the synced file.
- **Unguarded concurrent focus reloads** (`main.ts`): added a `reloadInFlight` coalescing guard so rapid focus events don't stack concurrent `storeParsedData`/`transition` pairs.
- **Double disk read per external change** (`main.ts`): reload reuses the already-read `File` instead of reading the handle a second time.
- **Fragile null-watermark guard** (`main.ts`, flagged by `/review` + two security reviewers): inverted `watermark !== null && file.lastModified <= watermark` to `watermark === null || file.lastModified <= watermark`, making "no watermark → no reload" explicit rather than an unconditional-reload trap. Added a covering test.

### Required (security)
_None._ No high- or medium-confidence security finding warranted an in-PR fix; the medium-confidence integrity edges are out-of-model under the encrypted-by-design epic and are deferred (see below). The in-model encrypted reload path is correctly protected — a successful encrypted reload bumps `generation` (`configureFileSync`), abandoning any stale in-flight write (`file-sync.ts:114`).

### Informational
- **TOCTOU on the post-write watermark re-read** (`file-sync.ts:119-121`): a concurrent external write landing between `writeFileToHandle` and the re-read could stamp the watermark with the external timestamp, suppressing the next detection. Re-reading the file's own `lastModified` is a deliberate choice for cross-device sync (comparing the file's mtime, not a local clock, avoids clock-skew false positives). Narrow race; left as designed.
- **Focus-handler serial flood** (`main.ts`): the `reloadInFlight` guard coalesces concurrent reloads. A serial flood of focus events touching `lastModified` could still drive repeated PBKDF2 decrypts, but that requires local filesystem write access — at which point the attacker already controls the `.benc`. The in-flight guard is the proportionate mitigation.
- **Raw error logged on the reload error path** (`main.ts:449,470`): the FSA error may embed the on-disk file path in its message. This matches the existing `load-from-handle` convention and the path is the user's own on their own machine; left consistent with the established pattern rather than diverging one call site.
- **Post-write readback failure** (`file-sync.ts:119-121`): if the re-read fails (file deleted between write and read), it surfaces as a write error and the watermark stays stale; the app recovers on next startup. Acceptable graceful degradation.

### Deferred — `#1179` (blocked_by #1020)
- **Plaintext / encryption-mode reload hardening**: the reload path does not enforce the session's encryption mode, so (a) a session that loaded an encrypted `.benc` silently accepts a *plaintext* external replacement with no password challenge, and (b) a plaintext reload sets `importPassword = null`, skipping `configureFileSync` (`main.ts:464`) so `generation` is not bumped — a queued write-back can then overwrite a newer external file with stale data. Both are out-of-model edges (the epic mandates encrypted-at-rest `.benc`); the encrypted in-model path is unaffected. Filed for deliberate handling of the rejection policy.

### Out-of-scope (security)
_None._ CodeQL: 0 open alerts on the PR ref. Dependency audit not run — the diff touches no `package.json` / `package-lock.json`.

---
_Coverage: all reviewers ran; no partial coverage. Unit suite: `budget/test/main.test.ts` + `file-sync.test.ts` — 35 passed._

PR #1175, opened 2026-06-06
dispatch: stop jq control-char errors from echo-into-jq in qa-fix/review-fix

### LINKED ISSUE #1170: dispatch: recurring jq control-character parse error in worker gh-output handling
## Context

Dispatch worker sessions repeatedly hit:

```
jq: parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line N
```

Transcript analysis (Jun 4–6, 2026) found **~43 occurrences in 2 days (~22/day)**, spread across nearly every active issue — so it sits in a hot path. It arises from piping raw text (likely `gh` issue/PR bodies or labels, or git output containing literal control characters) into `jq` as if it were JSON, rather than using `gh --json … --jq` (which emits already-escaped JSON) or `jq -R`/`--raw-input`. Each failure costs a tool round-trip plus a model correction; some cascade into larger retries.

The known safe callsites (`gh pr view --json … | jq` in `qa-fix`/`review-fix`) are not the source — `gh --json` emits valid JSON. The error signature (a parse failure immediately followed by an unrelated section marker such as `---LABELS---`) points to an ad-hoc text→`jq` pipeline; the exact source must be localized first.

## Expected vs actual

- Expected: structured data is parsed without `jq` control-character errors.
- Actual: ~22 `jq` control-character parse failures per day in worker sessions.

## Acceptance criteria

- [ ] The source(s) of the control-character `jq` failure are localized (script callsite vs. model-authored pipeline), with at least one captured reproduction from a transcript.
- [ ] The source is fixed to parse structured data safely — `gh … --json … --jq`, `jq --raw-input`, or reading the field via `gh`'s native template — so raw control characters never reach a bare `jq` string parse.
- [ ] If the source is a model-authored pattern rather than a script, the governing SKILL.md / rule is updated to prescribe the safe form.
- [ ] Zero `jq` control-character parse errors over a sample of subsequent dispatch sessions.

Provenance: surfaced by the dispatch token-usage analysis (`dispatch-token-usage-analysis.md`).

natb1:
Closes #1170

## Problem

Dispatch worker sessions hit `jq: parse error: Invalid string: control
characters from U+0000 through U+001F must be escaped` ~22×/day (#1170) — 298
matched error tool-results across saved transcripts.

## Localized source

Model-authored documentation, not a script. Two SKILL.md idempotency preambles
literally prescribed the corrupting pattern, and workers copy it verbatim:

- `.claude/skills/review-fix/SKILL.md`
- `.claude/skills/qa-fix/SKILL.md`

```bash
PR_JSON=$(gh pr view "$BRANCH" --json number,labels,body)
PR_NUM=$(echo "$PR_JSON" | jq -r .number)   # ← fails
```

## Mechanism (reproduced)

The Bash tool runs in **zsh**, whose builtin `echo` interprets backslash escapes
by default. `gh --json` emits valid JSON — a tab inside a PR body is the
two-character escape `\t`. `echo "$PR_JSON"` rewrites it to a *literal* control
character, which `jq` then rejects. PR bodies routinely contain tabs/newlines,
so the `review-fix` variant (which fetches `body`) fires constantly.

The issue's stated hypothesis (`gh --json … | jq` is unsafe) was wrong about the
cause: `gh` does emit valid JSON; the zsh `echo` round-trip is the corruption.

Reproduced locally:

```
$ echo '{"n":1,"body":"a\tb"}' | jq -r .body
jq: parse error: Invalid string: control characters ...   # exit 5
$ jq -r .body <<<'{"n":1,"body":"a\tb"}'
a	b                                                        # exit 0
```

## Fix

Two docs-only changes:

1. **New rule `.claude/rules/shell-json.md`** — auto-loaded into every session's
   context, steering every worker away from the pattern. Documents the
   here-string / `gh --jq` / `printf '%s'` alternatives.
2. **Fix the two SKILL.md callsites** — replace `echo "$VAR" | jq` with
   `jq … <<<"$VAR"` (passes bytes through unchanged, preserves the single-fetch
   design) and add a one-sentence reference to the new rule.

A lexical sweep (`grep -rn 'echo "\$VAR" | jq' .claude/ --include='*.md'`)
returns zero executable bad pipelines after the fix — remaining matches are the
new rule's BAD example and prose references.

## Out of scope

`run-acceptance-tests.sh` and `run-preview-deploy.sh` also use `echo "$VAR" |
jq`, but on firebase/header output (not gh JSON with escaped control chars), and
never appear in the 298 errors. The new rule covers them conceptually.

Acceptance criterion 4 (zero errors over subsequent sessions) is satisfied going
forward by the corrected guidance; it can only be measured after merge by
re-running the transcript grep on post-merge worker sessions.


natb1:
## QA summary — autonomous pass (qa-fix)

Instructions-only change (the #1170 jq control-character fix): a new
`.claude/rules/shell-json.md` rule plus safe-pattern updates to the `qa-fix` and
`review-fix` skill callsites. **Non-browser path** — no app/frontend code changed.

The QA pass treats a future Claude session as the "user": success means the
prescribed patterns actually work in zsh and the bad pattern is gone from the
callsites.

### Items walked (4 machine-verifiable)

| # | Item | Result |
|---|------|--------|
| 1 | New rule `shell-json.md` present & documents the error signature + three GOOD forms | **PASS** |
| 2 | BAD pattern (`echo "$VAR" \| jq`) reproduces the documented error in zsh | **PASS** |
| 3 | GOOD patterns (`jq <<<`, `printf '%s' \| jq`, direct pipe) parse the same tab-bearing JSON cleanly | **PASS** |
| 4 | `qa-fix`/`review-fix` callsites no longer use the bad `echo "$PR_JSON" \| jq` form | **PASS** |

**Counts:** PASS 4 / FAIL 0 / SKIP 0

### Captured reproduction (Item 2)

Against valid JSON containing a string with the two-char escape `\t` (exactly how
`gh --json` emits an embedded tab):

```
jq: parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 1, column 33
```

zsh `echo` un-escaped the `\t` into a raw TAB before `jq` saw it. All three GOOD
forms returned the field value (`1175`) with no error — confirming the fix.

### Deferred to office-hours

None.

### Bugs found

None.

— autonomous QA pass clean; applying `dispatch:qa-done`.

natb1:

One terminal review pass over the docs-only diff (`.claude/rules/shell-json.md`,
`.claude/skills/qa-fix/SKILL.md`, `.claude/skills/review-fix/SKILL.md`). Ran
`/code-review max --fix` and `/review` as direct subagents; the security pass was
surface-gated and skipped (see note below).

### Fixed
_None._

### Required (security)
_None._

### Informational
_None._

### Dismissed
- `.claude/rules/shell-json.md:3` — *(review, low)* Suggestion to hedge the
  "interprets `\uXXXX` by default" claim. Dismissed: the Bash tool runs in zsh,
  whose builtin `echo` does interpret `\t`/`\n`/`\uXXXX` escapes by default —
  the very environment the rule documents — so the claim is accurate as written.
- `.claude/skills/qa-fix/SKILL.md` / `.claude/skills/review-fix/SKILL.md`
  *(comments)* — *(review, low)* Suggestion to enumerate `gh --jq` / `printf
  '%s'` alongside the here-string in the SKILL.md comments. Dismissed: both
  comments already link to `.claude/rules/shell-json.md`, which lists every safe
  form; a redundant in-line enumeration is not an improvement.

### False-positive (security)
_None._

### Deferred
_None._

### Out-of-scope (security)
_None._

---

Security review: no attack surface — docs-only diff (no executable, config,
dependency, or Firestore-rules changes).

PR #1195, opened 2026-06-07
Test restore-dispatch-skill.sh name-lookup vs branch-name derivation

natb1:
Closes #875

## What

Adds two unit tests to `test-dispatch-scripts.sh` covering the `claude agents
--json` name-lookup path in `.claude/hooks/restore-dispatch-skill.sh`.

PR #863 added a primary name-lookup path to the dispatch context-recovery hook:
it calls `claude agents --json`, selects the session by `session_id`, and reads
`.name` to derive the worktree basename, only falling back to `git rev-parse
--abbrev-ref HEAD` (the branch name) when that returns empty. PR #903 built the
test harness for this hook, but every existing `restore-dispatch-skill` test
sets the session `.name` and the git branch to **consistent** values, so none
prove the basename comes from `.name` rather than from `git rev-parse`. This is
the coverage gap #875 addresses.

## Tests added

Both route through `current-phase.txt = ERROR` (the `dispatch-worker` fallback),
so the derived worktree path appears on the `ARGUMENTS:` line.

- **Test 7b — name-lookup wins over a divergent branch.** Sets `.name =
  875-from-name` and branch = `999-from-branch` so the two sources diverge.
  Asserts the emitted `ARGUMENTS:` path uses the `.name`-derived basename
  (`.../worktrees/875-from-name`), and a negative check that `999-from-branch`
  never appears — proving `git rev-parse` was not the source.
- **Test 7c — empty-registry fallback.** Omits `set_agents_name` so the `claude`
  stub returns `[]` and `.name` is empty; asserts the basename falls back to the
  branch (`999-from-branch`), exercising the `git rev-parse` path.

## Scope

Test-only, additive. No change to `restore-dispatch-skill.sh`, existing tests,
or test helpers — reuses the existing `restore_setup` / `set_agents_name` /
`run_restore` / `restore_teardown` harness from #903.

## Verification

```
test-dispatch-scripts.sh 2>&1 | grep -E 'name-lookup|empty-registry fallback'
```

emits three `PASS` lines (Test 7b has two assertions, Test 7c one) and zero
`FAIL`. The full suite has ~27 pre-existing, unrelated failures in environments
lacking `python3`; CI (`unit-tests.yml`) has `python3` and is authoritative.


natb1:
## QA pass — #875 (PR #1195)

**Non-browser path** — only `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` changed.

### Items walked

| # | Item | Result |
|---|------|--------|
| 1 | Test 7b: name-lookup path wins over divergent branch | ✅ PASS |
| 2 | Test 7c: empty registry → branch-name fallback | ✅ PASS |
| 3 | Full suite regression check (1949 assertions) | ✅ PASS |

**3/3 PASS · 0 FAIL · 0 SKIP · 0 deferred to office-hours**

### Detail

- **Test 7b**: set `.name=875-from-name`, `current-branch.txt=999-from-branch`, `phase=ERROR`. Confirmed `ARGUMENTS` line uses `.../worktrees/875-from-name` (name-derived) and `999-from-branch` is absent from output — proves `claude agents --json` path consulted ahead of `git rev-parse`.
- **Test 7c**: no `claude-agents.json` fixture (stub returns `[]`), `current-branch.txt=999-from-branch`. Confirmed `ARGUMENTS` uses `.../worktrees/999-from-branch` — exercises the `git rev-parse` fallback.
- **Full suite**: `1949/1949 passed, 0 failed` across `test-dispatch-scripts.sh` plus all other suites.

No bugs found. No needs-human-judgment items.

natb1:
## Review findings

### Fixed

| Finding | Location | Fix commit |
|---|---|---|
| Test 7c echo label used ASCII `->` instead of Unicode `→`, inconsistent with all other restore-dispatch-skill test labels | `test-dispatch-scripts.sh:15589` | `dd62c76` |

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

### Security

**Surface classification:** `code`, no app/rules surface, no dependency changes.

**Reviewers run:** input-validation, secrets, red-team, security-review (built-in), CodeQL (inline fetch).

**Required:** _None._

**False-positive:** _None._

**Out-of-scope (pre-existing code, diff did not touch):**

| Finding | Location | Confidence | Note |
|---|---|---|---|
| `set_agents_name` writes JSON without encoding the argument — a future test author passing a name with `"` or `\` could produce malformed JSON fixture, silently defeating what the test claims to prove | `test-dispatch-scripts.sh:15310` | medium | Pre-existing helper; new tests 7b/7c use only safe hardcoded literals |
| Negative assertion in Test 7b checks full output for `999-from-branch` absence — fragile if `restore_setup`'s SKILL.md fixture body is later changed to include that literal | `test-dispatch-scripts.sh:15576` | low | Assertion is correct for current fixture; hypothetical fragility only if harness is refactored |
| `TMPDIR_TEST` (from `mktemp -d`) lacks mode-700 restriction — low-confidence race on shared CI hosts | `test-dispatch-scripts.sh:15244` | low | Pre-existing harness setup; no change in this diff |

No out-of-scope CodeQL or npm advisories to file as follow-up issues.

PR #1207, opened 2026-06-07
dispatch: emit recover unit WorkingDirectory= unquoted so OnFailure recovery loads

### LINKED ISSUE #1203: dispatch: dispatch-tick-recover.service fails to load — WorkingDirectory written double-quoted
## Context

`dispatch-spawn-tick` and `dispatch-schedule-reseed` register `--property=OnFailure=dispatch-tick-recover.service` so an abnormally-exiting tick re-seeds the chain. The recovery unit is installed by `ensure_recover_unit` in `lib.sh`, which writes its path directives double-quoted, including `WorkingDirectory="$main_worktree"`.

systemd's `WorkingDirectory=` does not accept a quoted value the way `ExecStart=`/`Environment=` do: the literal leading `"` makes the path non-absolute, so the unit is rejected at load:

```
dispatch-tick-recover.service:8: WorkingDirectory= path is not absolute: "/home/n8/.../worktrees/main"
Unit configuration has fatal error, unit will not be started.
```

`systemctl --user status dispatch-tick-recover.service` reports `Loaded: bad-setting`. Because the unit never loads, the `OnFailure=` recovery both launchers depend on **never fires** — a failed tick does not re-seed the chain, so the chain can wedge with nothing to recover it. The comment at `lib.sh:826` ("path-bearing directives are double-quoted: systemd unescapes C-style quotes...") is correct for `ExecStart=` but wrong for `WorkingDirectory=`, which drove the defect.

## Reproduction

1. Trigger `ensure_recover_unit` (run a tick/reseed launcher).
2. `systemctl --user status dispatch-tick-recover.service` reports `Loaded: bad-setting`, `Active: inactive (dead)`.
3. `systemctl --user cat dispatch-tick-recover.service` shows `WorkingDirectory="/abs/path"` (quoted).

**Expected:** the recovery unit loads and runs `dispatch-tick-recover` on a tick's `OnFailure`.

**Actual:** unit is `bad-setting`; `OnFailure` recovery is silently non-functional.

## Fix

Emit `WorkingDirectory=` unquoted in `ensure_recover_unit` (`WorkingDirectory=$main_worktree`); the worktree path has no spaces, and `WorkingDirectory=` takes the bare path as a single value. Correct the misleading comment at `lib.sh:826` to scope the quote-unescaping claim to `ExecStart=`/`Environment=`.

## Acceptance criteria

- [ ] `ensure_recover_unit` writes `WorkingDirectory=` without surrounding double quotes.
- [ ] After install, `systemctl --user status dispatch-tick-recover.service` reports `Loaded: loaded` (not `bad-setting`); a forced `OnFailure` actually starts `dispatch-tick-recover`.
- [ ] The `lib.sh:826` comment no longer claims `WorkingDirectory=` supports quote-unescaping.
- [ ] `test-dispatch-scripts.sh` asserts the emitted unit has an unquoted, absolute `WorkingDirectory=` line.

natb1:
Closes #1203

## Problem

`dispatch-spawn-tick` and `dispatch-schedule-reseed` register
`--property=OnFailure=dispatch-tick-recover.service` so an abnormally-exiting
tick re-seeds the chain. That recovery unit is installed by `ensure_recover_unit`
in `lib.sh`, which emitted its path directives double-quoted, including:

```
WorkingDirectory="$main_worktree"
```

systemd's `WorkingDirectory=` does **not** unescape a quoted value the way
`ExecStart=`/`Environment=` do. The literal leading `"` makes the path
non-absolute, so the unit is rejected at load:

```
dispatch-tick-recover.service:8: WorkingDirectory= path is not absolute: "/.../worktrees/main"
Unit configuration has fatal error, unit will not be started.
```

`systemctl --user status` reported `Loaded: bad-setting`. Because the unit never
loaded, the `OnFailure=` recovery both launchers depend on **never fired** — a
failed tick did not re-seed, so the chain could wedge with nothing to recover it.

## Fix

- Emit `WorkingDirectory=$main_worktree` bare (unquoted) in `ensure_recover_unit`.
  The worktree path has no spaces and `WorkingDirectory=` takes the bare path as a
  single value. `Environment=`/`ExecStart=` stay double-quoted — those directives
  *do* unescape C-style quotes and are correct.
- Rescope the preceding comment so the quote-unescaping claim no longer covers
  `WorkingDirectory=`, and note the exception explicitly.
- Add a regression test to `test-dispatch-scripts.sh` asserting the emitted unit
  has an unquoted, absolute `WorkingDirectory=` line.

## Verification

`test-dispatch-scripts.sh` (shell only, no build): `Results: 1909/1909 passed, 0 failed`,
including the two new `ensure_recover_unit WorkingDirectory=` assertions.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -823,9 +823,12 @@ ensure_recover_unit() {
   # git/jq/claude. The caller carries the full nix-store PATH, so baking it into
   # the unit at write time makes the unit self-sufficient.
   #
-  # The path-bearing directives are double-quoted: systemd unescapes C-style
-  # quotes, so a path containing spaces is parsed as a single token rather than
-  # split into an executable + spurious arguments.
+  # ExecStart= and Environment= are double-quoted: systemd unescapes C-style
+  # quotes for these two directives, so a path containing spaces is parsed as a
+  # single token rather than split into an executable + spurious arguments.
+  # WorkingDirectory= is the exception — it does NOT unescape quotes; a leading
+  # `"` makes the path non-absolute and systemd rejects the unit (bad-setting),
+  # so it takes the bare path (safe here: the worktree path has no spaces).
   #
   # Deliberately NO OnFailure= on this unit — the recover handler must not chain
   # to itself, or a failing recovery would recurse.
@@ -838,7 +841,7 @@ Description=Dispatch chain continuation recovery (OnFailure handler)
 Type=oneshot
 Environment="PATH=$safe_path"
 ExecStart="$RECOVER_SCRIPT"
-WorkingDirectory="$main_worktree"
+WorkingDirectory=$main_worktree
 EOF
 )
 
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -20494,6 +20494,40 @@ unset DISPATCH_TARGET_WORKERS_RATE_LIMITS_PATH DISPATCH_TARGET_WORKERS_NOW
 rr_teardown
 
 # ============================================================================
+# ensure_recover_unit: WorkingDirectory= is unquoted and absolute (#1203)
+# ============================================================================
+echo ""
+echo "=== ensure_recover_unit WorkingDirectory= quoting ==="
+# Regression: systemd's WorkingDirectory= does not unescape quotes; a quoted
+# value makes the path non-absolute and the unit loads as bad-setting, so the
+# OnFailure= recovery never fires. Assert the emitted line is bare + absolute.
+eru_tmp=$(mktemp -d)
+mkdir -p "$eru_tmp/bin"
+cat > "$eru_tmp/bin/systemctl" <<'STUB'
+#!/usr/bin/env bash
+exit 0
+STUB
+chmod +x "$eru_tmp/bin/systemctl"
+(
+  export DISPATCH_RECOVER_UNIT_DIR="$eru_tmp/systemd-user"
+  export DISPATCH_RECOVER_SYSTEMCTL_CMD="$eru_tmp/bin/systemctl"
+  source "$SCRIPT_DIR/lib.sh"
+  ensure_recover_unit "$eru_tmp/main-worktree"
+)
+eru_unit="$eru_tmp/systemd-user/dispatch-tick-recover.service"
+eru_wd_line=$(grep '^WorkingDirectory=' "$eru_unit")
+assert_eq "WorkingDirectory= is the bare absolute path (no quotes)" \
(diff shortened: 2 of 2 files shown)


natb1:
## QA Pass — #1203: dispatch-tick-recover.service WorkingDirectory= fix

Non-browser (scripts-only) QA. All 3 plan items verified autonomously.

| # | Item | Result |
|---|------|--------|
| 1 | Test suite regression — `WorkingDirectory=` unquoted and absolute | ✅ PASS |
| 2 | Comment accuracy — `lib.sh` comment scope for `WorkingDirectory=` | ✅ PASS |
| 3 | Unit file emission — heredoc emits bare `WorkingDirectory=$main_worktree` | ✅ PASS |

**Full test suite:** 1909/1909 passed, 0 failed — including both new `ensure_recover_unit` assertions:
- `WorkingDirectory=` is the bare absolute path (no quotes)
- `WorkingDirectory=` value is not double-quoted

**Bugs found:** none

natb1:
## Review findings — PR #1207

### Fixed

- **test-dispatch-scripts.sh:20518** — `grep '^WorkingDirectory='` was called without an `if/else` guard under `set -euo pipefail`; a missing unit file or absent line would abort the test script with no `FAIL` recorded, making a broken test look clean. Guarded with `if/else` that records `FAIL` on either path. (commit b353c304)
- **test-dispatch-scripts.sh:20511** — the subshell running `ensure_recover_unit` was unguarded; a non-zero return would abort under `set -e` with no `FAIL` increment. Wrapped with `if (...); then ... else FAIL++ fi`, consistent with the rest of the test file. (commit b353c304)
- **test-dispatch-scripts.sh:20513** — `ensure_recover_unit` was called with `$eru_tmp/main-worktree` as the path argument but that directory was never created. Added `mkdir -p "$eru_tmp/main-worktree"` so the test is robust to any future existence check in the function. (commit b353c304)

### Informational

_None._

### Dismissed

_None._

### False-positive (security)

_None._

### Deferred

- **lib.sh:844** — `WorkingDirectory=$main_worktree` is unquoted; a space in the path would cause systemd to truncate the value silently. The newline guard at line 804 exists but does not block spaces. The 'no spaces' constraint is only documented, not machine-checked. Valid hardening but out of scope for the quoting fix this PR targets; filed as #1211 (blocked by #1203).
- **lib.sh:817** — `safe_path` strips newlines from `PATH` before embedding in `Environment="PATH=$safe_path"`, but not backslashes. systemd applies C-string unescaping to `Environment=` values; a backslash in a PATH entry would be silently corrupted. Not a realistic concern on Linux/WSL but worth enforcing; filed as #1212 (blocked by #1203).

### Out-of-scope (security)

_None. No open CodeQL alerts. No dependency changes (deps=false)._

---

Security review: surface=code (shell scripts). Input-validation, secrets, red-team, and built-in `/security-review` reviewers ran. No app/rules surface (app_or_rules=false) — auth, data-exposure, and Firebase reviewers not applicable. No high-confidence required security findings.

PR #1204, opened 2026-06-07
dispatch: KillMode=process on transient tick/reseed units so cgroup-kill spares the bg fleet

natb1:
Closes #1196

## Problem

`/dispatch`'s autonomous chain launches every tick and reseed as a
`systemd-run --user` transient unit (`dispatch-spawn-tick`,
`dispatch-schedule-reseed`). These units run with systemd's default
`KillMode=control-group`.

Claude Code's per-user bg supervisor daemon is spawned on demand by the first
`claude` call that needs one. In the autonomous chain (no human `claude agents`
view open) that first call comes from inside a tick/reseed unit — so the daemon,
and the `--bg-spare` / `--bg-pty-host` children hosting each agent, is born
**inside the transient unit's cgroup**. When the short tick finishes and the unit
deactivates, `KillMode=control-group` makes systemd SIGTERM the entire cgroup:
daemon plus every worker. The next reseed re-spawns into a fresh cgroup and is
reaped again — a self-perpetuating crash loop.

The bg hosts are spawned `detached: true` (setsid), but setsid changes only the
session/process-group, not the cgroup, so cgroup-kill ignores it.

## Fix

Pass `--property=KillMode=process` to the `systemd-run --user` invocations so a
finishing tick kills only its own `dispatch-tick` process and leaves the detached
bg hosts running as init orphans (what `detached: true` is designed for). A later
tick attaches to the surviving daemon via the existing lock.

This is the minimal correct stop-gap; the greenfield durable daemon-ownership
design is tracked separately in #1197.

## Changes

- **`dispatch-spawn-tick`** — `--property=KillMode=process` added to both
  `systemd-run --user` invocations (target-keyed branch and no-target `else`
  branch), adjacent to the existing `--property=OnFailure=...` line; header
  rationale block added.
- **`dispatch-schedule-reseed`** — `--property=KillMode=process` added to the
  single `systemd-run --user` invocation, adjacent to `--property=OnFailure=...`;
  Step-6 comment block extended with the rationale.
- **`test-dispatch-scripts.sh`** — the three launcher argv-assertion blocks
  (no-arg spawn-tick, target-keyed spawn-tick, weekly cap-hit reseed) now also
  require `--property=KillMode=process`, with `KillMode` in their PASS/FAIL
  labels.

## Verification

`test-dispatch-scripts.sh` passes (1907/1907), including the three extended
KillMode assertions. The runtime end-to-end check — an autonomous reseed chain
[...]

Issue #1196, opened 2026-06-07
dispatch: transient-unit cgroup kill reaps the bg supervisor daemon and the whole fleet


natb1:
## Context

`/dispatch`'s autonomous chain launches every tick and reseed as a `systemd-run --user` transient unit (`dispatch-spawn-tick`, `dispatch-schedule-reseed`), which run with systemd's default `KillMode=control-group`.

Claude Code hosts background sessions under a per-user **supervisor daemon** spawned on demand by the first `claude` call that needs one. In the autonomous chain (no human `claude agents` view open) that first call comes from inside a tick/reseed unit — the `claude --bg` worker kick in `dispatch-spawn-job`, or the `claude agents --json` liveness checks in `lib-claude-agents.sh` / `dispatch-acquire-lock`. So the daemon — and the `--bg-spare` / `--bg-pty-host` children that host each agent — is born **inside the transient unit's cgroup**.

When the short tick finishes, the unit deactivates and `KillMode=control-group` makes systemd SIGTERM the entire cgroup: daemon plus every worker. The next reseed timer re-spawns into a fresh transient cgroup and is reaped again — a self-perpetuating crash loop. The bg hosts are spawned `detached: true` (setsid) to outlive their launcher, but setsid changes only the session/process-group, not the cgroup, so cgroup-kill ignores it.

## Reproduction

1. Run the autonomous dispatch chain (reseed timer active) with no persistent `claude agents` view.
2. `~/.claude/daemon.log` shows `[supervisor] shutting down` / `daemon start` / `bg adopt: adopted=0 dead=N` cycling on the reseed cadence (~2-5 min); jobs stamped `process gone while supervisor was down`; `[bg-pty] host crash` errors.
3. `systemctl --user show <dispatch-reseed-*.service> -p KillMode` returns `KillMode=control-group`.

**Expected:** workers spawned by a tick survive the tick unit's completion; the daemon persists across reseeds.

**Actual:** daemon + fleet reaped on every tick/reseed completion; the chain crash-loops.

## Fix (this PR)

[...]

PR #1217, opened 2026-06-07
dispatch: own the bg supervisor daemon in a durable systemd --user service

### LINKED ISSUE #1197: dispatch: host the bg supervisor daemon in a durable cgroup so the fleet never depends on tick lifecycle
## Context

The autonomous chain relies on Claude Code's per-user supervisor daemon to host every `claude --bg` worker. That daemon is spawned on demand by whichever `claude` call first needs it; in the autonomous chain that's a short-lived `systemd-run --user` tick/reseed unit, binding the daemon's life to an ephemeral unit (see #1196 for the cgroup-reap failure that results).

#1196 stops the immediate crash with `KillMode=process` (a finishing tick leaves the daemon orphaned-but-alive). That is correct but leaves the daemon **unowned** — it survives only until Claude Code's idle timeout or a binary-takeover restart (`tengu_bg_binary_takeover`, e.g. a Nix profile bump), and orphan daemons can accumulate across reseeds with no clean owner.

## Greenfield design

Own the supervisor daemon explicitly, in a stable non-transient cgroup, so the fleet never depends on any tick's lifecycle:

- A long-lived `systemd --user` **service** (`Type=simple`, restart-on-failure) that starts and holds the supervisor daemon, OR Claude Code's own persistent-daemon install path (the binary exposes a `via_service` route — `tengu_bg_daemon_install`).
- Every tick/reseed/worker `claude` call then attaches to the running durable daemon via the existing lock instead of spawning one in its own transient cgroup.
- The chain ensures the service is installed/active (idempotently) before scheduling ticks, mirroring how `dispatch-schedule-reseed` already installs the recover unit.

This makes tick-unit `KillMode` irrelevant for fleet survival (the fleet no longer lives in tick cgroups), supersedes the #1196 stop-gap, and removes idle-timeout / binary-takeover fragility for the autonomous chain.

## Acceptance criteria

- [ ] The chain idempotently installs/activates a durable `systemd --user` service (or invokes Claude Code's `via_service` persistent install) hosting the daemon in a non-transient cgroup, before ticks run.
- [ ] With the service active, `/proc/<daemon-pid>/cgroup` is the durable service unit, not a `dispatch-reseed-*`/`dispatch-tick` transient unit.
- [ ] An autonomous reseed chain runs across multiple cycles with a single stable daemon PID (no tick-completion churn).
- [ ] Degrades safely when `systemd --user` is unavailable (clear diagnostic; falls back to #1196 behavior, no crash loop).
- [ ] `test-dispatch-scripts.sh` covers the idempotent install and the attach-to-existing-daemon path.

natb1:
Closes #1197

## Summary

Owns Claude Code's per-user bg supervisor daemon explicitly, in a stable
non-transient cgroup, so the autonomous fleet never depends on any tick's
lifecycle.

The autonomous chain relies on the bg supervisor daemon to host every
`claude --bg` worker. That daemon is spawned *on demand* by whichever `claude`
call first needs it — in the chain, the first call inside a short-lived
`systemd-run --user` tick/reseed unit, so the daemon (and the `--bg-spare` /
`--bg-pty-host` children hosting each worker) is born in that unit's **ephemeral
transient cgroup**. When the tick finishes, systemd reaps the cgroup and kills
the fleet (#1196). PR #1204 added `KillMode=process` as a stop-gap — a finishing
tick then leaves the daemon orphaned-but-alive — but the daemon stays **unowned**:
it survives only until idle-timeout or a binary-takeover restart, and orphans
accumulate across reseeds.

## Design

A dispatch-owned, long-lived `systemd --user` service —
`dispatch-claude-daemon.service` (`Type=simple`, `Restart=always`,
`ExecStart="<abs-claude>" daemon run`) — installed and activated **idempotently
before ticks run**, exactly mirroring how `ensure_recover_unit` (`lib.sh`)
installs the static `dispatch-tick-recover.service`. With it holding the daemon
lock (`~/.claude/daemon.lock`), every tick/reseed/worker `claude` call attaches
to the running daemon instead of spawning its own, and the fleet lives
permanently in the service's durable cgroup.

Why this shape:

- **Claude Code's own `via_service` install is disabled in 2.1.168.** `claude
  daemon --help` states service install is disabled; the `daemon` subcommand
  exposes `run`/`status`/`logs`/`uninstall`/`stop` but **no `install`**. So we
  host the daemon ourselves rather than delegating.
- **`claude daemon run` runs the supervisor in the foreground**, defaulting its
  origin to `foreground` when `--origin` is omitted — a long-lived foreground
  supervisor, not the on-demand idle-exiting daemon.
- **No `--origin service`.** That origin carries the binary's
  self-uninstall-on-binary-takeover path — the exact fragility #1197 removes.
  Foreground origin has no self-uninstall logic; we own the lifecycle via
  systemd `Restart=always`.

`KillMode=process` on the tick/reseed units is **kept as the documented
degraded-path fallback**: when `systemd --user` is unavailable and no durable
service can be installed, the on-demand daemon still survives a finishing tick.

## Changes

- `lib.sh`: new `ensure_daemon_service` helper (no args — `claude daemon run`
  chdirs to `$HOME`, so no `WorkingDirectory=`, sidestepping the #1203/#1207
  quoting hazard). Best-effort contract: warns + returns non-zero, never `exit`.
  Steady-state hot path is the **attach-to-existing-daemon path** — when the
  unit matches byte-for-byte *and* `is-active` reports running, it no-ops.
- `dispatch-spawn-tick` / `dispatch-schedule-reseed`: call
  `ensure_daemon_service || true` before scheduling, with comments demoting
  `KillMode=process` to the degraded-path fallback (this supersedes the #1196
  stop-gap as the load-bearing mechanism).
- `test-dispatch-scripts.sh`: new block covering idempotent cold install,
  attach-to-existing-daemon no-op, inactive self-heal, and degrade-safe paths,
  using a recording `systemctl` stub (no live systemd).

## Testing

  1923/1923 pass, including the new `ensure_daemon_service` block.
- `bash -n` clean on all four edited scripts.

Live cgroup/PID-stability validation (daemon in `dispatch-claude-daemon.service`
cgroup, stable PID across reseed cycles) is user-validated per the acceptance
criteria, not in CI.


natb1:
## QA Summary — #1197 (PR #1217)

**Path**: non-browser (scripts-only: `lib.sh`, `dispatch-spawn-tick`, `dispatch-schedule-reseed`, `test-dispatch-scripts.sh`)

### Items walked

| # | Item | Result |
|---|------|--------|
| 1 | `ensure_daemon_service` cold install — unit written with `Type=simple`, `Restart=always`, `WantedBy=default.target`, `ExecStart="<claude>" daemon run` (no `--origin service`), `daemon-reload` + `enable --now` ran | PASS |
| 2 | `ensure_daemon_service` attach path — steady-state no-op: no `daemon-reload`, no `enable` when content matches and service is active | PASS |
| 3 | `ensure_daemon_service` inactive self-heal — `enable --now` runs; no unit rewrite when content unchanged | PASS |
| 4 | `ensure_daemon_service` degrade safely — `enable` failure returns non-zero without aborting caller; empty claude path returns non-zero | PASS |
| 5 | `dispatch-spawn-tick` calls `ensure_daemon_service \|\| true` before `systemd-run` (line 163) | PASS |
| 6 | `dispatch-schedule-reseed` calls `ensure_daemon_service \|\| true` before `systemd-run` (line 385) | PASS |
| 7 | Full test suite: `test-dispatch-scripts.sh` — 1998/1998 passed, 0 failed | PASS |

### Counts
- PASS: 7
- FAIL: 0
- SKIP: 0
- Deferred to office-hours (needs human judgment): 0
- Bugs found: 0

natb1:
## Review pass — PR #1217

### Fixed (commit 4a4b2ea0)

- **`lib.sh:1037`** — daemon-reload stuck-state: after `mv` places the new unit file, a failed `daemon-reload` left systemd unaware of it. Subsequent calls would skip the write (content matches) but never retry the reload, leaving the service permanently unloadable. Fixed: retry `daemon-reload` unconditionally on the slow path regardless of whether the file write was skipped. _(sources: code-review)_
- **`lib.sh:994`** — `CLAUDE_CMD` double-quote injection: the binary path was checked for newlines but not for `"`. A `"` in the path would break the `ExecStart="…"` unit-file field. Fixed: added a guard that rejects `CLAUDE_CMD` containing a double-quote, mirroring the existing newline-rejection pattern. _(sources: code-review, input-validation, review, secrets, red-team)_
- **`lib.sh:993`** — `safe_path` double-quote stripping: `$PATH` was sanitized of newlines only; an embedded `"` would corrupt the `Environment="PATH=…"` unit-file field. Fixed: strip `"` (alongside `\n`) from `safe_path` in both `ensure_daemon_service` and `ensure_recover_unit`. _(sources: input-validation, red-team)_
- **`test-dispatch-scripts.sh:21904`** — Missing test for daemon-reload failure: no case exercised `STUB_RELOAD_RC=1` on the first `ensure_daemon_service` call. The stuck-state bug above was therefore untested. Fixed: added test case 3b that sets `STUB_RELOAD_RC=1`, confirms the first call returns non-zero with the unit on disk, then sets `STUB_RELOAD_RC=0` and verifies a second call retries `daemon-reload` and reaches `enable --now`. _(sources: code-review)_

### Deferred (follow-up filed)

- **`dispatch-tick-recover:353`** — `dispatch-tick-recover` schedules `dispatch-tick` directly via `systemd-run`, bypassing `dispatch-schedule-reseed` and its `ensure_daemon_service` call. If the daemon service enters systemd's failed state (StartLimit exhausted) during a recover path, `ensure_daemon_service` is never called, so the next tick spawns the daemon inside the transient reseed cgroup — reverting to the pre-#1197 behavior. Recommended fix: add `ensure_daemon_service || true` to `dispatch-tick-recover` before the recovery-reseed `systemd-run` call. Filed as #1370. _(sources: code-review)_

### Informational

- **`test-dispatch-scripts.sh:21905–21925`** — Test 4a (degrade path: `enable` fails) sets `STUB_IS_ACTIVE_RC=0` in a fresh `eds_degrade_dir`. After `mv` succeeds but `enable --now` fails, the test relies on checking degrade-path output rather than directly verifying the service unit's activation state. Not incorrect but worth noting for future test clarity. _(sources: review)_
- **`lib.sh:1037–1040`** — When `daemon-reload` fails after `mv` has already placed the new unit file at `UNIT_PATH`, the function returns 1 without removing or reverting the file. The unit file on disk is now correct but systemd has not loaded it. This is safe (the next call retries reload), but leaves a transient partial state. No action needed; documented for awareness. _(sources: review)_
- **`lib.sh:939`** — `DISPATCH_DAEMON_CLAUDE_CMD` controls what binary the durable daemon runs. A compromised environment variable lets an attacker establish persistent user-level code execution via the installed systemd service. Mitigated by the new double-quote and newline guards; residual risk is inherent to environment-variable-controlled service install. _(sources: red-team)_
- **`lib.sh:999`** — `WantedBy=default.target` causes the service to be enabled at user-session login. An attacker who plants a malicious unit file gains login persistence. This is expected behavior for a daemon meant to survive across sessions; the content-hash check limits the attack surface. _(sources: red-team)_

### Dismissed

- **`lib.sh:993`** — `$PATH` embedded in unit file flagged as a potential secret leak. `$PATH` contains directory entries, not credentials. The sanitization (newline + double-quote stripping) is adequate. _(sources: secrets)_
- **`lib.sh:950`** — Red-team scenario: stripped `CLAUDE_CMD` could be used to disable the durable daemon. The existing newline-rejection guard already prevents this; the new double-quote guard closes the remaining injection surface. _(sources: red-team)_
- **`lib.sh:1009`** — Red-team scenario: steady-state hot path skips rewrite if content matches, which could allow a pre-installed malicious unit to persist. The content-hash check is a write-skip optimization on already-correct content; by design the unit is trusted at install time. _(sources: red-team)_

PR #965, opened 2026-05-30
dispatch-target-workers: terminal-sprint envelope drives weekly usage to 100% at week end

natb1:
Closes #962

## Summary

`dispatch-target-workers` Stage 1's smooth power curve tops out at `target_weekly_usage_pct` (default 90) and, with the per-window increment clamped at `weekly_increment_cap_pct`, never drove weekly usage to 100% near week end — it idled the final stretch below the cap, leaving weekly budget unspent.

This adds a terminal "you-can-still-make-it" envelope that floors the cumulative curve from the end:

    W = max(smooth_curve, weekly_terminal_pct - weekly_increment_cap_pct * remaining_windows)

The envelope term is deeply negative mid-week (smooth curve dominates, output unchanged) and only lifts the final ~2 windows — 80 at r=2, 90 at r=1, 100 at r=0 — so the closing windows sprint weekly usage to `weekly_terminal_pct`. Live `used_weekly` remains the backstop.

- **Unit 1** (ea62778): add `weekly_terminal_pct` knob (default 100, validated in (0,100]) to the `target-workers` config schema, example, and config-load tests.
- **Unit 2** (86f9243): consume the knob in the controller's Stage 1 awk; update the header-comment algorithm/`W(1)` invariant; add envelope-floor + mid-week-unchanged tests.

## Test plan

- [ ] `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — all tests pass (963/963 locally)
- [ ] Envelope-floor test pins W to 80/90/100 at remaining windows r=2/1/0 (discriminating probes vs. the pre-envelope curve)
- [ ] Mid-week tick (x=0.5) gating boundary is unchanged from the smooth curve

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1226, opened 2026-06-07
dispatch-config-load: reject target-workers configs whose floor makes the pace curve non-monotone

natb1:
Closes #930

## Problem

`dispatch-target-workers` builds a weekly cumulative-pace curve whose terminal
per-window increment is `end = wfloor + (p+1)*(target_weekly/T - wfloor)`,
clamped from above at `wcap` but never from below. When
`target_weekly_usage_pct < T*wfloor*p/(p+1)` (≈17 at defaults: `T=34`, `p=1`,
`wfloor=1`), `end` goes negative, the per-window increment
`d(x)=wfloor+(end-wfloor)·x^p` crosses zero, and the cumulative curve `W(x)`
becomes non-monotone — it overshoots `target_weekly` mid-week before descending
back to it at `x=1`. The binary weekly gate then opens and spawns workers while
`used_weekly` sits between `target_weekly` and the mid-week peak — i.e. while
already over the weekly budget. (The terminal-sprint envelope from #965 only
floors `W` from the end via `max()`; it cannot lower the mid-week overshoot, so
the defect persisted.)

`dispatch-config-load`'s `target-workers` validator bounded each percentage
field to `(0,100]` and had cross-field *ordering* checks, but nothing rejected
the floor↔target contradiction, so `target_weekly_usage_pct: 10` (or a large
`weekly_increment_floor_pct`) was accepted and silently broke the curve.

## Fix

The floor and a low target are genuinely contradictory: "always spend ≥ `wfloor`
per window" forces a cumulative minimum of `T*wfloor*p/(p+1)`, which a lower
`target_weekly` cannot honor. Per `.claude/rules/code-style.md` (prefer clear
errors over defensive fallbacks), the right resolution is to reject the config
at load rather than silently reconcile it.

- `dispatch-config-load`: add a fourth cross-field check to the `target-workers`
  validator — a curve-monotonicity constraint that rejects configs where
  `target_weekly_usage_pct < T*weekly_increment_floor_pct*p/(p+1)` (`T=34`,
  `p=weekly_curve_power`). Unlike the three ordering checks (which fire only when
  both members of a pair are explicitly present), this check applies the
  algorithm defaults for absent fields, because the canonical misconfiguration
  is `target_weekly_usage_pct: 10` set alone with floor/power defaulted. The
  error names the field and its effective value, states the computed floor and
  `T`/`p`/`floor`, and gives the remedy. The schema header comment documents the
  divergence.
- `dispatch-target-workers`: document at the `end` upper-clamp that the lower
  bound (`end >= 0`, curve monotonicity) is guaranteed upstream by the validator,
  which is why no lower clamp is applied here (a clamp would be dead code masking
  a rejected config). Documentation only — no behavior change.

This necessarily couples the validator to the algorithm's window constant
`T = round(WEEK_SECONDS/18000) = 34` and the `p/(p+1)` factor — a coupling the
issue anticipates. It is a single backwards-compatible change: it only rejects
configs that were already broken. The shipped `target-workers.example.json`
(target 90) and the all-defaults config (`90 >= 17`) remain accepted.

## Tests

`test-dispatch-scripts.sh`: update Test 13d (a large floor alone is now
rejected) and add Tests 13k–13o — low-target rejection (`10 < 17`), the
inclusive boundary (`17` accepted, `16` rejected), the power-shifted threshold
(`p=3` rejects target `20`; default `p=1` accepts it), and a valid combined
config. Full suite: 1994/1994 passing.

Surfaced during code review of #925.


natb1:
## QA Summary — #930 dispatch-target-workers: monotonicity constraint

**Result: All checks passed. No browser component — shell-command lane only.**

### Items executed (7/7)

| # | Title | Result |
|---|-------|--------|
| 1 | Full dispatch-scripts test suite passes | PASS |
| 2 | target_weekly=10 rejected with clear error | PASS |
| 3 | Boundary: 17 accepted, 16 rejected | PASS |
| 4 | Power=3 shifts rejection threshold | PASS |
| 5 | Large floor alone rejected | PASS |
| 6 | Golden path: valid configs accepted | PASS |
| 7 | Non-numeric field: type error (no jq crash) | PASS |

### Detail

1. **Full test suite** — `test-dispatch-scripts.sh` completed 2411/2411 passing, 0 failed.

2. **target_weekly=10 rejected** — `dispatch-config-load` exits 1 with stderr naming `target_weekly_usage_pct`, effective value (10), computed floor (17), `monotonicity` keyword, and the remedy ("Raise target_weekly_usage_pct, or lower weekly_increment_floor_pct / weekly_curve_power").

3. **Boundary** — target=17 accepted (exit 0); target=16 rejected (exit 1) with monotonicity error. Strict `<` gives inclusive-accept at the math boundary.

4. **Power coupling** — `{"target_weekly_usage_pct": 20, "weekly_curve_power": 3}` rejected (min=25.5 > 20); default power=1 accepts target=20 (min=17 ≤ 20).

5. **Large floor alone** — `{"weekly_increment_floor_pct": 50}` rejected (min=850 > default target=90). Confirms new check applies algorithm defaults for absent fields.

6. **Golden path** — full example config (target=90) and `{"target_weekly_usage_pct":90,"weekly_increment_floor_pct":5}` (min=85≤90) both accepted (exit 0).

7. **Type guard** — `{"target_weekly_usage_pct":"ten"}` rejected with `must be a number if present, got string`; no jq arithmetic crash from the monotonicity constraint.

### Deferred to office-hours

None.

### Bugs found

None.

natb1:
## Review

**Coverage note:** The code-review finder had a transient socket failure and could not run this pass. The `review` finder, security surface classifier, and CodeQL scan completed normally. CodeQL: no open alerts.

### Dismissed

- **`dispatch-config-load:478` — T=34 hardcoded in jq** (`review`): The awk script derives `T` dynamically via `int(WEEK_SECONDS / 18000)`, while the jq validator hardcodes `34` with an inline derivation comment. The comment explains the coupling explicitly (`# round(WEEK_SECONDS/18000) = round(604800/18000) = 34`) and the PR description acknowledges it. The validator runs at config-load time (not in the hot path), so the two constants do not diverge at runtime. Dismissed — the comment is the intentional documentation of the coupling rather than a hidden inconsistency.

### Informational

- **`test-dispatch-scripts.sh:9042` — Test 13l boundary comment** (`review`): Test 13l asserts `target_weekly_usage_pct=17` is accepted at the exact boundary `min = 34*1*1/2 = 17`, using the condition `$tw < $min` (strict less-than). The test comment says "17 < 17 is false" which is correct. No action needed — surfaced for human reference.

### Fixed

_None._

### Required (security)

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1318, opened 2026-06-10
firestore.rules: clamp reading-position writes and close the read existence oracle

natb1:
Closes #1297

Fixes two security defects in the `match /print/{env}/reading-position/{docId}`
Firestore rule.

## Read — close the cross-user existence oracle

The old read rule

```
allow read: if request.auth != null
  && (resource == null || request.auth.uid == resource.data.uid);
```

let any authed user probe `getDoc("print/prod/reading-position/{victimUid}_{mediaId}")`:
success (`resource == null`) meant the doc was absent, permission-denied meant it
was present — leaking whether another user had opened a specific media item.

The read rule now authorizes by **docId prefix** (the requester's own uid),
without touching `resource`:

```
allow read: if request.auth != null
  && docId.split("_")[0] == request.auth.uid;
```

Firebase uids are alphanumeric (no `_`), so the prefix before the first `_` is
exactly the owner uid. This returns the same answer whether or not the doc
exists — closing the oracle — and keeps first-visit `getDoc()` on the user's own
`{uid}_{mediaId}` doc working.

## Write — clamp keys and type-check fields

The old write rule used `hasAll(['uid','mediaId','position'])` with no `hasOnly`
clamp and no type validation, so an authed user could store arbitrary extra
fields under their own key. This was the only client-writable collection lacking
a `hasOnly` clamp. The write rule now adds `hasOnly` plus `is string` checks on
`uid`, `mediaId`, and `position` (the `is string` checks also enforce presence,
so `hasAll` is dropped). `position` is a string per
`print/src/reading-position.ts`. The existing uid-match and docId-format
constraints are unchanged.

## Tests

`rules-test/test/firestore/print.test.ts` covers both: a new test proves another
user is now denied reading a non-existent victim-prefixed docId (the oracle), and
three new write tests cover the `hasOnly` clamp and non-string `position` /
`mediaId` rejection. Existing client-write tests were updated to send string
positions to match the new type check. All 23 print reading-position tests pass
locally (`rules-test` is not in CI — run manually).

Rules deploy automatically on merge to `main` via `firestore-deploy.yml`.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -256,14 +256,20 @@ service cloud.firestore {
     }
 
     // Reading positions are private per-user documents keyed by {uid}_{mediaId}.
-    // Only the owning user may read or write their own position.
+    // Reads are authorized by docId prefix (the requester's own uid), not by
+    // resource.data.uid — this returns the same answer whether or not the doc
+    // exists, closing the cross-user existence oracle while keeping first-visit
+    // getDoc() on the user's own doc working. Writes are clamped to exactly the
+    // three expected fields, each type-checked.
     match /print/{env}/reading-position/{docId} {
       allow read: if request.auth != null
-        // resource == null allows getDoc() on a non-existent doc (first visit)
-        && (resource == null || request.auth.uid == resource.data.uid);
+        && docId.split("_")[0] == request.auth.uid;
       allow write: if request.auth != null
         && request.auth.uid == request.resource.data.uid
-        && request.resource.data.keys().hasAll(['uid', 'mediaId', 'position'])
+        && request.resource.data.keys().hasOnly(['uid', 'mediaId', 'position'])
+        && request.resource.data.uid is string
+        && request.resource.data.mediaId is string
+        && request.resource.data.position is string
         && docId == request.auth.uid + "_" + request.resource.data.mediaId;
     }
 
--- a/rules-test/test/firestore/print.test.ts
+++ b/rules-test/test/firestore/print.test.ts
@@ -63,6 +63,14 @@ describe("print reading-position", () => {
       );
     });
 
+    it("denies reading another user's non-existent position (no existence oracle)", async () => {
+      const ctx = env.authenticatedContext("other-user");
+      const db = ctx.firestore();
+      await assertFails(
+        getDoc(doc(db, `print/${ENV}/reading-position/${UID}_nonexistent`)),
+      );
+    });
+
     it("denies unauthenticated read", async () => {
       const ctx = unauthenticatedContext(env);
       const db = ctx.firestore();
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — #1297 (PR #1318)

**Path:** Non-browser (pure Firestore rules + rules-test change)

### Results

| Item | Result |
|---|---|
| Rules unit test suite (`print.test.ts`) | ✅ PASS — 23/23 |
| Deny-all catch-all is last rule | ✅ PASS |
| Only reading-position block changed | ✅ PASS |

**3 PASS · 0 FAIL · 0 SKIP · 0 deferred**

### Notes

- All 23 `print.test.ts` tests pass, including the new existence-oracle denial test and the `hasOnly`/type-check write tests.
- 1 pre-existing failure in `office-hours.test.ts` (`denies non-member list query on test-env collection`) — not in this PR's diff; no office-hours files changed.
- The deny-all catch-all (`match /{document=**} { allow read, write: if false; }`) is confirmed as the last rule.
- `git diff` confirms only the `match /print/{env}/reading-position/{docId}` block changed — comment, read rule (docId-prefix authorization), and write rule (`hasOnly` + `is string` type checks).

natb1:
## Review findings

Reviewed with `/code-review max`, `/review`, and full security fan-out (input-validation, secrets, red-team, `/security-review`, auth & access control, data-exposure, firebase-specific). CodeQL: no open alerts. No dependency changes.

### Fixed

_None._

### Required (security)

_None._

### Informational

- **CR-1** `firestore.rules:269` — The `hasOnly` clamp creates a migration hazard: any existing reading-position doc with extra fields would block the owner from using `updateDoc` until they do a full `setDoc` overwrite. In practice, the production client (`saveReadingPosition`) uses `setDoc` exclusively (full overwrite), so this scenario is not reachable via normal client flows. Noted for awareness in case admin-SDK-written docs exist.

- **CR-2** `rules-test/test/firestore/print.test.ts:98` — The "allows owner to update their position" test uses `updateDoc`, but the production client always calls `setDoc`. The test covers a write path no production code exercises. Not a regression introduced by this PR; noted as a test design FYI.

### Dismissed

- **RV-2** `rules-test/test/firestore/print.test.ts:108` — No test for `updateDoc` adding a new key to verify `hasOnly` on the merged post-update state. Low-confidence gap for a path the production client never uses; out of scope for this PR's test additions.

- **SEC-1** `firestore.rules:265-266` — The `docId.split("_")[0] == request.auth.uid` read rule implicitly assumes Firebase UIDs are alphanumeric (no underscores). This assumption is correct per Firebase Auth documentation; adding a comment would be cosmetic only and does not change security behavior.

### False-positive (security)

_None._

### Out-of-scope (security)

- **SEC-2** `firestore.rules:264` — The `{env}` wildcard allows writes to any environment path (prod, preview-*, etc.). Pre-existing behavior consistent with all other client-writable collections; the diff did not introduce this and the plan explicitly excludes env restriction.

- **SEC-3** `firestore.rules:272` — `position is string` enforces type but not length, allowing up to ~1 MB. Pre-existing concern not introduced by this diff; `hasOnly` prevents extra fields but doesn't cap per-field size.

### Deferred

_None._

PR #1320, opened 2026-06-10
firestore.rules: pin error-log env + fully validate/bound the error-doc schema

natb1:
Closes #1296

## Problem

`firestore.rules` exposed the error-log collections to fully **unauthenticated**
`create` — `match /<app>/{env}/errors/{errorId} { allow create: if isValidErrorLog(); }`
for five apps (budget, landing, fellspiral, print, audio) — with two abuse levers:

1. **Fabricated envs.** The `{env}` path wildcard accepted *any* string, so an
   anonymous script could sprawl writes across unbounded distinct env subtrees
   (`budget/zzz/errors/...`, `budget/aaa/errors/...`, …).
2. **Unbounded doc size + shape.** `isValidErrorLog()` validated only
   `message`/`operation`/`timestamp`/`stack` and capped total keys at 15.
   `kind`/`code`/`userAgent`/`url`/`uid`/`email` and *arbitrary caller extras*
   passed through unchecked up to the ~1 MB doc limit — the client sink literally
   spread every non-reserved context key as a new top-level field.

Together: anonymous, unbounded ~1 MB writes × five apps × unlimited fabricated
envs — billable cost / DoS on this Blaze-plan project. Rules cannot rate-limit,
and the client-side throttle is useless against an attacker who bypasses the SDK.

## Fix

Make the error document a **fixed, fully-enumerated, size-bounded schema** and
**pin `{env}` to a known-env pattern** — the rules-level path the issue's "and/or"
acceptance permits. This closes the fabricated-env vector and cuts the worst-case
doc from ~1 MB to ~73 KB.

- **Unit 1 — `firestore.rules`.** New `isKnownErrorEnv(env)` pins `{env}` to the
  real env taxonomy (`prod`/`demo`/`test`, `preview-pr-N`, `qa(-id)?`,
  `emulator(-id)?`). `isValidErrorLog()` rewritten to `keys().hasOnly([…11 keys])`
  (replacing the loose `keys().size() <= 15`) plus a reusable
  `isOptionalString(data, field, maxLen)` helper enforcing per-field UTF-16 caps.
  Every match block now gates `allow create: if isKnownErrorEnv(env) &&
  isValidErrorLog();`.
- **Unit 2 — `firebaseutil/src/error-sink.ts`.** The sink no longer spreads
  arbitrary context as top-level keys (which the stricter rule would reject).
  Non-reserved context is folded into a single JSON-stringified `extras` field
  (try/catch → `null` on circular-ref/BigInt so logging never throws), and every
  variable-length field is truncated to the exact caps the rule enforces. The doc
  now has exactly the 11 enumerated keys.
- **Unit 3 — `rules-test/test/firestore/errors.test.ts`.** Coverage for the new
  env-pin (fabricated/malformed envs denied; real envs allowed) and the doc
  size/shape limits (`hasOnly`, each per-field cap, non-string optional fields,
  all-optionals-null, valid populated `extras`).

The `qa`/`emulator` suffix is **optional** (`qa(-id)?`): `get_env_suffix` returns
the bare base suffix outside a git worktree, so bare `qa`/`emulator` are real env
shapes and must be allowed. The patterns never widen to accept arbitrary tokens.

Tradeoff: `kind` is length-capped (32) rather than enum-restricted, to avoid
coupling the rule to the `ErrorKind` union.

## Verification

- `firebaseutil` unit tests: 69 pass (extras-as-JSON, truncation, reserved
  `extras`).
- `rules-test` (run manually — not in CI): all 34 `errors.test.ts` cases pass.
  (One unrelated, pre-existing `office-hours.test.ts` failure is outside this
  change's scope — the `firestore.rules` diff touches only the error-log blocks.)
- Rules syntax check (`run-rules-check.sh`) passes.

## Follow-up (out of scope)

A stronger gate — App Check *enforcement* on Firestore (initialized client-side
but not enforced) or a Cloud Function with per-IP quota — would actually
rate-limit anonymous writes. Both are infrastructure changes risky to bundle into
a rules PR; they complement, not replace, the size/shape/env hardening here. The
review phase may file one as a `blocked_by` follow-up.

## Deployment note

Per `.claude/rules/firestore.md`, `firestore.rules` deploys on merge to `main`
(preview branches do not deploy rules), so smoke tests on this feature branch
permission-deny until the rules land. Units 1 and 2 are functionally coupled —
sequence the app redeploys carrying the new sink to land before/with the rules
deploy. The rules change is ultimately split into a standalone rules-only PR at
QA/merge time.

### DIFF
--- a/firebaseutil/src/error-sink.ts
+++ b/firebaseutil/src/error-sink.ts
@@ -99,7 +99,8 @@ export function createFirestoreErrorSink(options: ErrorSinkOptions): ErrorSink {
     if (Object.keys(extrasObj).length > 0) {
       let extrasStr: string | null = null;
       try {
-        extrasStr = truncate(JSON.stringify(extrasObj), CAPS.extras);
+        const serialized = JSON.stringify(extrasObj);
+        extrasStr = serialized.length <= CAPS.extras ? serialized : null;
       } catch {
         extrasStr = null;
       }
--- a/firebaseutil/src/error-sink.ts
+++ b/firebaseutil/src/error-sink.ts
@@ -20,7 +20,26 @@ export interface ErrorSinkOptions {
 // with these names are dropped so caller-provided extras cannot overwrite canonical
 // error values like the original Error message. operation and kind are included
 // for safety even though they are always written from structured fields.
-const RESERVED_KEYS = new Set(["operation", "kind", "message", "stack", "code", "timestamp", "userAgent", "url", "uid", "email"]);
+// "extras" is reserved for the synthesized JSON blob of non-reserved context.
+const RESERVED_KEYS = new Set(["operation", "kind", "message", "stack", "code", "timestamp", "userAgent", "url", "uid", "email", "extras"]);
+
+// Size caps (UTF-16 code units) — must mirror the firestore.rules isValidErrorLog() caps exactly.
+const CAPS = {
+  message: 10000,
+  operation: 200,
+  kind: 32,
+  stack: 50000,
+  code: 200,
+  userAgent: 500,
+  url: 2000,
+  uid: 128,
+  email: 320,
+  extras: 10000,
+} as const;
+
+function truncate(value: string | null | undefined, max: number): string | null {
+  return value == null ? null : String(value).slice(0, max);
+}
 
 export function createFirestoreErrorSink(options: ErrorSinkOptions): ErrorSink {
   const { db, namespace, getCurrentUser } = options;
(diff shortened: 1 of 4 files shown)


natb1:
## QA Pass — #1296 Firestore error-log hardening

**Result: PASS** — 6/6 machine-verifiable items, 0 fails, 0 deferred.

### Items walked

| # | Item | Result |
|---|---|---|
| 1 | Rules syntax check | ✅ PASS |
| 2 | Client unit tests — `error-sink.test.ts` (18 tests) | ✅ PASS |
| 3 | Firestore rules unit tests — `errors.test.ts` (74 tests) | ✅ PASS |
| 4 | `extras` folded into single JSON field (structural) | ✅ PASS |
| 5 | `isKnownErrorEnv` env allowlist + all five app blocks gated | ✅ PASS |
| 6 | `isValidErrorLog` uses `hasOnly` + `isOptionalString` caps | ✅ PASS |

### Notes

**Minor plan deviation (non-security):** `isKnownErrorEnv` allows bare `qa` and `emulator` (the suffix group is optional: `qa(-[a-z0-9._-]+)?`) rather than the plan's required-suffix patterns. This is a broader allowlist but not exploitably so — attackers targeting cost/DoS use arbitrary strings like `zzz`, not `qa` or `emulator`.

**Pre-existing failure (not caused by this PR):** `rules-test/test/firestore/office-hours.test.ts` has one failing test (`denies non-member list query on test-env collection`). `office-hours.test.ts` is unchanged in this branch — confirmed via `git diff origin/main...HEAD`.

### Counts

- PASS: 6 | FAIL: 0 | SKIP: 0
- Deferred to office-hours: 0
- Bugs found: 0

natb1:
## Review findings

### Required (security)

- **extras JSON truncation** (`firebaseutil/src/error-sink.ts:102`) — `truncate(JSON.stringify(extrasObj), CAPS.extras)` could slice the serialized JSON string at an arbitrary offset, storing unparseable (mid-cut) JSON in Firestore. The rule accepts any string ≤10000 chars so malformed JSON passed. Fixed by checking serialized length before assignment: when it exceeds `CAPS.extras` the field is set to `null` (same as the circular-ref/BigInt catch path), preserving the invariant that `extras` is either `null` or valid JSON. _(commits 265efd00)_

### Fixed

- **Missing client truncation tests** (`firebaseutil/test/error-sink.test.ts`) — No unit tests covered truncation for `operation`, `code`, `url`, `userAgent`, `uid`, or `email` fields. Added tests for each at one character over the cap, plus a test asserting an oversized `extrasObj` (serializing to >10000 chars) produces `null` rather than truncated JSON. _(commits 265efd00)_
- **Missing bare `qa`/`emulator` env tests** (`rules-test/test/firestore/errors.test.ts`) — The `isKnownErrorEnv` patterns `qa(-[a-z0-9._-]+)?` and `emulator(-[a-z0-9._-]+)?` intentionally allow bare `"qa"` and `"emulator"` (matching `get_env_suffix` behavior on the main worktree), but no test asserted this. Added `assertSucceeds` cases for both. Added an inline comment in the rule documenting the intent. _(commits 715452c5)_
- **Missing `prod` env test** (`rules-test/test/firestore/errors.test.ts`) — The most critical environment (`env == "prod"`) had no `assertSucceeds` test case. Added it. _(commits 715452c5)_
- **Missing oversized `operation` rules-test** (`rules-test/test/firestore/errors.test.ts`) — The rule enforces `operation.size() <= 200` but the test suite covered every other size-bounded field except this one. Added `assertFails` for 201 chars. _(commits 715452c5)_
- **Missing `emulator-` denial test** (`rules-test/test/firestore/errors.test.ts`) — No test asserted that `emulator-` (trailing dash, no suffix chars) is denied, mirroring the existing `qa-` denial test. Added it. _(commits 715452c5)_

### Informational

- **`extras` serialization failure is silent** (`firebaseutil/src/error-sink.ts:103`) — The circular-ref/BigInt catch branch drops context silently with no `console.warn`. Low-impact (logging is fire-and-forget) but could obscure data loss in dev. No change; noted for awareness.
- **`uid`/`email` are self-reported, not auth-validated** (`firestore.rules:331-332`) — The rule validates `uid`/`email` as optional strings but does not cross-check against `request.auth`. This is the intentional design — creates are unauthenticated — but means attribution in error telemetry is not trusted identity.

### Deferred (follow-up issues filed)

- **PII risk in `extras` field** (`firebaseutil/src/error-sink.ts:102`) — Callers can inadvertently pass sensitive keys through non-reserved context. Tracked in #1327, blocked by #1296.
- **`preview-pr-[0-9]+` env pattern has no upper bound on PR number length** (`firestore.rules:310`) — An attacker can iterate arbitrarily many `preview-pr-N` envs; the per-doc cap limits per-write cost but does not close the namespace. Tracked in #1326, blocked by #1296.

### Out-of-scope (pre-existing, no follow-up)

- **`location.href` captures full URL including sensitive query params** (`firebaseutil/src/error-sink.ts:85`) — Pre-existing behavior, not introduced by this diff.

### Security review

CodeQL: no open alerts on this PR's ref.
Dependency audit: no dependency changes in this diff.
Secrets scan: no hardcoded credentials.

PR #1321, opened 2026-06-10
functions: use Firestore BulkWriter in office-hours-sync to clear the 500-op batch limit

natb1:
Closes #1299

## Problem

`syncOfficeHoursCore` in `functions/src/office-hours-sync.ts` mirrors open
`jit:`-labelled GitHub issues into the `office-hours/{env}/items` Firestore
collection on a 30-minute schedule. It accumulated one `set` per open jit issue
plus one `delete` per stale doc into a **single** `firestore.batch()`.

Firestore caps a `WriteBatch` at **500 operations**. Once
`written + deleted > 500`, `batch.commit()` threw on every run and the sync
collection went permanently stale — the function could never recover on its own
because each run re-attempted the same over-cap batch.

## Fix

Replace the single batch with a Firestore **`BulkWriter`**
(`firestore.bulkWriter()`), which is purpose-built for bulk writes that exceed
the batch limit: it internally chunks at 500 ops, parallelizes, and retries.
This removes all manual op-counting. Atomicity is not a correctness requirement
here — the sync is idempotent and reconciles on the next run — and a single
batch already forfeited all-or-nothing semantics by failing entirely past 500.

Each enqueued write's promise is collected into a `writes` array; after
`writer.close()` flushes the chunks, `await Promise.all(writes)` surfaces any
permanent write failure as a thrown error rather than swallowing it (per
`.claude/rules/code-style.md`). No `onWriteError` handler is added.

The `written` / `deleted` / `skippedNoDate` counters, the `isValidJitKey` /
`dueAt` skip logic, and the `SyncResult` contract are unchanged — only the write
mechanism changes.

## Tests

- The in-memory firestore mock gains a `bulkWriter()` factory; existing
  `syncOfficeHoursCore` tests now drive that path.
- The mock's `batch()` is kept and hardened to throw on the 501st op
  (`INVALID_ARGUMENT: maximum 500 writes allowed per request`) as a deliberate
  regression guard — reverting production to a single batch fails the new test.
- New test `handles more than 500 operations in a single sync` exercises both
  the 600-issue set-only path and the combined set+delete >500 path
  (600 stale deletes + 5 fresh writes = 605 ops in one run).

`npx vitest run --project functions … office-hours-sync` — 20 tests green;
`npx tsc --noEmit --project functions` — clean.

### DIFF
--- a/functions/src/office-hours-sync.ts
+++ b/functions/src/office-hours-sync.ts
@@ -150,7 +150,7 @@ export async function syncOfficeHoursCore(deps: {
   }
 
   await writer.close(); // flushes all enqueued writes in ≤500-op chunks
-  await Promise.all(writes); // surface any write failure as a thrown error
+  await Promise.all(writes); // BulkWriter routes per-op failures to the individual op promises, not to close(); this is what re-raises them
 
   return { written, deleted, skippedNoDate };
 }
--- a/functions/src/office-hours-sync.ts
+++ b/functions/src/office-hours-sync.ts
@@ -100,7 +100,8 @@ export async function syncOfficeHoursCore(deps: {
 
   const itemsPath = `${deps.namespace}/items`;
   const itemsCollection = deps.firestore.collection(itemsPath);
-  const batch = deps.firestore.batch();
+  const writer = deps.firestore.bulkWriter();
+  const writes: Promise<unknown>[] = [];
 
   const writtenKeys = new Set<string>();
   let written = 0;
@@ -124,15 +125,17 @@ export async function syncOfficeHoursCore(deps: {
 
     const dueAt = Timestamp.fromDate(issue.dueAt);
     const docRef = itemsCollection.doc(issue.jitKey);
-    batch.set(docRef, {
-      title: issue.title,
-      dueAt,
-      repo: issue.repo,
-      issueNumber: issue.number,
-      jitKey: issue.jitKey,
-      memberEmails: deps.memberEmails,
-      updatedAt: FieldValue.serverTimestamp(),
-    });
+    writes.push(
+      writer.set(docRef, {
+        title: issue.title,
+        dueAt,
+        repo: issue.repo,
+        issueNumber: issue.number,
+        jitKey: issue.jitKey,
+        memberEmails: deps.memberEmails,
+        updatedAt: FieldValue.serverTimestamp(),
+      }),
+    );
     writtenKeys.add(issue.jitKey);
     written += 1;
   }
@@ -141,12 +144,13 @@ export async function syncOfficeHoursCore(deps: {
   let deleted = 0;
   for (const doc of existing.docs) {
     if (!writtenKeys.has(doc.id)) {
-      batch.delete(doc.ref);
+      writes.push(writer.delete(doc.ref));
       deleted += 1;
     }
   }
 
-  await batch.commit();
+  await writer.close(); // flushes all enqueued writes in ≤500-op chunks
+  await Promise.all(writes); // surface any write failure as a thrown error
 
   return { written, deleted, skippedNoDate };
 }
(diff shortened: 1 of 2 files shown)


natb1:
## QA Summary — #1299: BulkWriter in office-hours-sync

**Path:** non-browser (functions-only change)

### Items walked

| # | Item | Result |
|---|------|--------|
| 1 | Type-check passes | PASS |
| 2 | Existing `syncOfficeHoursCore` tests pass with BulkWriter | PASS |
| 3 | New >500-op test passes | PASS |
| 4 | Regression guard on `batch()` mock enforces 500-op cap | PASS |

### Counts
- PASS: 4 / 4
- FAIL: 0
- SKIP: 0
- Needs human judgment: 0

### Details
- `npx tsc --noEmit --project functions`: clean, no type errors.
- Unit tests: **31/31 passed** (2 test files). The new `"handles more than 500 operations in a single sync"` test is included and passes. This covers both the 600-write set-only path and the combined set+delete >605-op path.
- Regression guard: `batch()` mock throws `INVALID_ARGUMENT: maximum 500 writes allowed per request` on the 501st op; production now uses `bulkWriter()` which chunks internally and never hits the cap.
- Note: the `npm run build` step fails in the sandbox due to an EPERM on tsx's IPC pipe (`/tmp/claude-1000/tsx-1000/22.pipe`) — this is a sandbox restriction, not a code defect. CI runs build outside the sandbox.

### Deferred to office-hours

natb1:
## Review findings

### Fixed
- **`functions/src/office-hours-sync.ts:153`** — Expanded the `Promise.all(writes)` comment to explain why `writer.close()` alone is insufficient: BulkWriter routes per-op failures to the individual op promises, not to `close()`. Without this context a reader might remove the line as redundant. (commit d5d16217)

### Informational
- **`functions/test/office-hours-sync.test.ts:114`** — The mock `bulkWriter`'s `set()`/`delete()` return pre-resolved `Promise.resolve()` rather than deferred promises that settle inside `close()`. The production ordering (`close()` then `Promise.all`) is clearly correct, so this is an observation for future maintainers rather than a current bug.
- **`functions/test/office-hours-sync.test.ts:108`** — The mock `close()` applies ops serially; the real BulkWriter parallelizes sub-batches. Intentional simplification; no behavioral difference for the scenarios under test.
- **`functions/src/office-hours-sync.ts:152-153`** — BulkWriter flushes in independent chunks, so readers may observe a mix of old and new documents between commits. The design plan explicitly accepts this non-atomicity: the sync is idempotent and reconciles every 30 minutes.

### Dismissed
- **`functions/test/office-hours-sync.test.ts:244`** — `number: i` produces issue number 0 in the 600-issue fixture. Harmless: production has no truthiness check on `issueNumber` and the test's purpose is the 600-op regression guard.

### False-positive (security)
- **`functions/src/office-hours-sync.ts:152-153`** (red-team) — Suggested registering `writer.onWriteError(err => { throw err; })` to ensure errors propagate. Not needed: `BulkWriter.set()`/`delete()` return per-op promises that reject after retry exhaustion, and `await Promise.all(writes)` after `await writer.close()` is the correct and intended error-surface mechanism per the design plan.
- **`functions/src/office-hours-sync.ts:143-149`** (red-team) — Suggested the deletion pass might erroneously delete documents written in the same run because BulkWriter may flush chunks before the collection read. Not an issue: `writtenKeys` is populated synchronously before the `get()`, so documents written this run are always in `writtenKeys` and will never be erroneously scheduled for deletion.

### Out-of-scope (security)
- **`functions/src/office-hours-sync.ts:128-138`** — `memberEmails` is written verbatim from config without email-format validation; downstream Firestore rules use it for access control. Pre-existing concern not introduced by this diff (diff changes only the write mechanism, not what is written).

### Deferred
- **`functions/test/office-hours-sync.test.ts`** — No test covers the error-propagation path where a BulkWriter op promise rejects and `Promise.all(writes)` surfaces it. Valid coverage gap but out of scope for this PR's acceptance criteria. Tracked as #1325 (blocked by #1299).

_Security review: CodeQL — no open alerts. No dependency changes (deps audit not applicable)._

PR #1332, opened 2026-06-10
budget: disarm prior file-sync session on every load path

natb1:
Closes #1261

## Problem

`budget` links a budget snapshot to an on-disk `.benc` file via the File System
Access API and debounces encrypted write-backs. `file-sync.ts` holds the active
`(handle, password)` as module globals; a write-back exports the current
IndexedDB dataset, encrypts it with the held password, and overwrites the held
handle in place.

The invariant that must hold: the armed `(handle, password)` must always
correspond to the dataset currently in IndexedDB. But `resetFileSync()` (the
disarm) had exactly one call site — the Clear-data button — and arming was gated
on the stale module global `importPassword`, which cannot distinguish "this load
set the password" from "a prior session's password lingers." So switching the
linked file left sync armed to the *previous* file, and the next UI edit silently
destroyed it:

- **Stale-handle overwrite.** Loading file B (plaintext FSA, or any non-FSA
  upload that never re-armed) swapped IndexedDB to dataset B but left the globals
  at handle A + password A. The next edit wrote dataset B, encrypted with password
  A, to file A — destroying file A.
- **Cancel-arms-wrong-session.** Cancelling the password dialog for encrypted file
  B returned early without touching `importPassword`, so the `if (importPassword)`
  guard armed `configureFileSync(handleB, passwordA, …)`. The next edit overwrote
  file B with the still-displayed dataset A.

This is the load/arm path — distinct from #1179, which hardens the external
*reload* path.

## Fix

Two root-cause changes, both in `budget/src/main.ts`:

1. **Disarm at the single point of dataset divergence.** `loadFromFile` now calls
   `resetFileSync()` on the line immediately before `storeParsedData(data)` — after
   every throwing decrypt/parse step and after the password-cancel early return, so
   a wrong-password / validation failure / cancel leaves the prior session armed and
   still correct (IndexedDB is unchanged in those cases). This one placement covers
   every load path: upload, plaintext-FSA, encrypted-FSA, and external reload.
2. **Arm from an explicit load outcome, never the stale global.** `loadFromFile`
   now returns a discriminated `LoadOutcome` (`{ committed: true; password: string
   | null } | { committed: false }`) instead of signalling cancel via an
   indistinguishable `void` early-return. `loadFromHandle` and
   `reloadIfExternallyChanged` arm `configureFileSync` only when the outcome reports
   a committed encrypted load, using the password the outcome carries — not
   `importPassword`. `importPassword` keeps its existing meaning (the current
   session's encryption password, read by export and reload); it is simply no longer
   abused as the arm-decision proxy.

`file-sync.ts` is unchanged — its generation guard already bails an in-flight
write whose generation moved.

## Tests

Three regression cases added to `budget/test/main.test.ts`, asserting the
arm/disarm contract through the existing mocked `file-sync.js` spies:

- **AC1** — a committed plaintext upload calls `resetFileSync` and leaves sync
  disarmed (no `configureFileSync` re-arm).
- **AC2** — cancelling the password prompt for a new encrypted file does not arm
  write-back with the prior session's password.
- **AC3** — switching encrypted A → encrypted B disarms A, arms B only, and never
  re-arms handle A after the switch.

Full suite: 970 passed, 7 skipped, 0 failed. `npx tsc --noEmit --project budget`
clean.

### DIFF
--- a/budget/src/main.ts
+++ b/budget/src/main.ts
@@ -277,15 +277,17 @@ async function loadFromFile(file: File, cachedPassword?: string): Promise<LoadOu
   }
   const parsed = parseUploadedJson(text);
   const data = toParsedData(parsed);
-  // This is the single point of dataset divergence: the line below replaces the
-  // in-memory dataset, so the prior armed (handle, password) no longer matches
-  // what IndexedDB holds. Disarm sync *before* the swap, after every throwing
-  // decrypt/parse step and the password-cancel early return, so a failure leaves
-  // the prior session armed and still correct (IndexedDB is unchanged). This one
-  // placement covers every load path: upload, plaintext-FSA, encrypted-FSA, and
-  // external reload — callers re-arm afterward only when appropriate.
-  resetFileSync();
+  // storeParsedData is the single point of dataset divergence: it replaces the
+  // in-memory dataset, so afterward the prior armed (handle, password) no longer
+  // matches what IndexedDB holds. Disarm sync *after* the swap commits, so any
+  // throwing step that leaves IndexedDB unchanged — a decrypt/parse failure, the
+  // password-cancel early return, or a storeParsedData failure — leaves the prior
+  // session armed and still correct. The generation guard in file-sync protects
+  // any write started before resetFileSync runs. This one placement covers every
+  // load path: upload, plaintext-FSA, encrypted-FSA, and external reload —
+  // callers re-arm afterward only when appropriate.
   await storeParsedData(data);
+  resetFileSync();
   importPassword = pw;
   transition({ source: "local", groupName: parsed.groupName });
   return { committed: true, password: pw };
--- a/budget/src/main.ts
+++ b/budget/src/main.ts
@@ -247,12 +247,20 @@ function promptPassword(message: string): Promise<string | null> {
   });
 }
 
+// Outcome of a load attempt. `committed: true` means this call replaced the
+// in-memory dataset (and disarmed any prior file-sync session); `password` is
+// the password used for the load — `null` for plaintext, the actual string for
+// an encrypted file. `committed: false` means the load did not mutate state
+// (e.g. the user cancelled the password prompt), so any prior sync session is
+// intentionally left armed and still correct.
+type LoadOutcome = { committed: true; password: string | null } | { committed: false };
+
 // Core load pipeline shared by the upload path and the FSA handle path:
 // read bytes, decrypt-or-decode, parse, store, and transition to the local
 // state. Throws on any failure; callers wrap it in error handling.
 // Pass `cachedPassword` to skip the password dialog when the session password
 // is already known (e.g. on an external-change reload).
-async function loadFromFile(file: File, cachedPassword?: string): Promise<void> {
+async function loadFromFile(file: File, cachedPassword?: string): Promise<LoadOutcome> {
   const buffer = await file.arrayBuffer();
   let text: string;
   let pw: string | null = null;
@@ -261,7 +269,7 @@ async function loadFromFile(file: File, cachedPassword?: string): Promise<void> 
       pw = cachedPassword;
     } else {
       pw = await promptPassword("Enter password to decrypt");
-      if (pw === null) return;
+      if (pw === null) return { committed: false };
     }
     text = await decrypt(buffer, pw);
   } else {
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — #1261 budget: disarm prior file-sync session on every load path

**PR**: #1332
**Result**: User-input blocker — deferred to office-hours (judgment item)

## Items Walked

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests (AC1/AC2/AC3 + existing suite) | **PASS** |
| 2 | TypeScript type-check | **PASS** |
| 3 | App loads and renders in browser (regression) | **SKIP** — WSL2 networking |
| 4 | File upload UI reachable (smoke) | **SKIP** — WSL2 networking |
| 5 | File-switch arm/disarm behavior (FSA flow) | **Deferred to office-hours** |

**PASS: 2 · FAIL: 0 · SKIP: 2 · Deferred: 1**

## Machine-Verifiable Results

**Item 1 — Unit tests: PASS**
All 970 tests passed (7 skipped) across 47 test files. The three new acceptance-criteria cases all passed:
- **AC1**: Committed plaintext upload calls `resetFileSync` and leaves sync disarmed. ✓
- **AC2**: Cancelling the password prompt for encrypted file B does not arm B with the prior session's password. ✓
- **AC3**: Switching from encrypted A to encrypted B disarms A and arms B only. ✓

**Item 2 — Type-check: PASS**
`npx tsc --noEmit --project budget` exited clean. The `LoadOutcome` discriminant type threads correctly through all three `loadFromFile` callers (`loadFromHandle`, `reloadIfExternallyChanged`, `handleFileUpload`).

**Items 3 & 4 — Browser checks: SKIP**
Chrome on Windows cannot reach the Vite dev server at `localhost:43327` in WSL2. This is a known environment limitation (WSL2 NAT forwarding is unreliable without mirrored networking). The server responds correctly from within WSL (`curl` 200). No bugs — environment gap only.

Pre-QA acceptance tests could not run (Playwright browser installation requires `apt-get`, unavailable in Nix). CI has already passed these tests.

## Deferred to Office-Hours

**Item 5 — File-switch correct arm/disarm behavior (FSA flow)**

Verifying the end-to-end file-switch path requires:
- A real encrypted `.benc` file (handle A + password A)
- Loading a second file (handle B) via the FSA picker
- Confirming that after the switch, edits do NOT overwrite file A

This requires real files and FSA interaction in a Chromium browser — it is not part of the QA seed data and cannot be automated by the Chrome extension without those files. The arm/disarm contract is fully covered by the AC1/AC2/AC3 unit tests; this item is end-to-end confirmation with real on-disk files.

**Steps for the office-hours walkthrough:**
1. Open the budget app in Chrome (or installed PWA).
2. Load an encrypted `.benc` file (file A) via the file picker; enter its password.
3. Make a small edit (e.g. add a transaction). Confirm file A is updated on disk.
4. Load a second file (file B) via the file picker.
5. **Sub-case A (cancel):** Cancel the password prompt for file B. Make another edit. Confirm file A is NOT overwritten (the prior session remains armed to file A).
6. **Sub-case B (switch):** Load file B successfully (enter its password). Make an edit. Confirm file A is NOT overwritten and file B IS updated.

natb1:
## Review findings

### Fixed

- **`budget/src/main.ts:287` — `resetFileSync()` ordering before `storeParsedData()`** (commit `716b99fd`)
  `resetFileSync()` was called before `await storeParsedData(data)`. If `storeParsedData` throws (storage quota exceeded, IDB corruption, transaction abort), the prior file-sync session was disarmed even though IndexedDB is unchanged — breaking the invariant that a failure should leave the prior session armed and correct. Fixed by moving `resetFileSync()` to after `storeParsedData` completes, and updating the block comment to accurately describe the new ordering. The generation guard in `file-sync.ts` already protects any write started before the disarm.

### Informational

- **`budget/src/main.ts:334` — Plaintext FSA file leaves sync disarmed with no UI indication**
  A plaintext file loaded via the FSA picker produces `{ committed: true, password: null }`. The guard `outcome.committed && outcome.password !== null` is false, so `configureFileSync` is never called and write-back is permanently disarmed. This is intentional (write-back only encrypts, so plaintext files cannot be written back), but there is no UI indication that the linked file is read-only. This was the same behavior before this PR; surfacing it here as a follow-up UX consideration.

- **`budget/test/main.test.ts:843` — AC2 coverage gap**
  AC2 verifies the cancel-does-not-arm contract by asserting `mockResetFileSync` and `mockConfigureFileSync` were not called, but does not verify that the session remaining armed to `handleA` is still usable after the cancel. Low risk given the mock structure, but worth noting for future test hardening.

### Security review

No attack surface: `surface=code` with no new network endpoints, injection surfaces, or privilege boundaries. All security reviewers (input validation, secrets scan, red team, built-in security review, auth, data exposure, Firebase-specific) and CodeQL returned no findings.

PR #1333, opened 2026-06-10
budget: export weeklyAggregates so write-back round-trips them

natb1:
Closes #1262

## Problem

`exportToJson` (`budget/src/export.ts`) is the FSA write-back path — it's called
from the auto-save (`file-sync.ts`) and the manual download (`main.ts`), and the
first UI edit in an FSA session rewrites the whole `.benc` from IndexedDB through
it.

It never fetched or emitted `weeklyAggregates`. `budget-etl` emits that store
(`budget-etl/internal/export/export.go:117`) and the app parses it on load
(`parseUploadedJson`, `upload.ts:115`), but the export side was missing. So the
first write-back rewrote the file **without** `weeklyAggregates`; on the next
load `parseUploadedJson` saw none (`raw.weeklyAggregates ?? []`) and silently
emptied the store. The budgets pie chart's `averageWeeklyCredits` and the area
chart's "Other" series dropped to zero until the ETL regenerated the snapshot.

Root cause: `weekly-aggregate.ts` had no `…ToRawJson` export adaptor (its header
comment explicitly said "no export path"), and `export.ts` never read the store.

## Fix

- **`entities/weekly-aggregate.ts`** — add the `weeklyAggregateToRawJson` adaptor
  (`id`, `weekStart: msToISO(weekStartMs)`, `creditTotal`, `unbudgetedTotal`),
  mirroring the sibling `…ToRawJson` adaptors; correct the stale header comment.
- **`export.ts`** — fetch `getAll("weeklyAggregates")` and emit the mapped
  collection in the output object.
- **`test/export.test.ts`** — add the `weeklyAggregates` mock fixture (keeps the
  existing suite green) plus two regression tests:
  - `weeklyAggregates` round-trips unchanged through
    export → `parseUploadedJson` → `toParsedData`.
  - an ETL-snapshot round-trip empties no store — a generic guard looping over
    all 10 ETL-emitted store names, so any future store `export.ts` forgets to
    emit also trips the test.

## Verification

- `npx vitest run --root budget test/export.test.ts` — 16 passed (14 existing +
  2 new).

## Out of scope

The `statementItems` / `reconciliationNotes` mirror is the same trap but latent
today — the ETL snapshot never carries those two collections, so no ETL-sourced
file exercises the drop. Full export/import symmetry for them is a separate,
larger change with no current trigger; recommend a follow-up issue.

### DIFF
--- a/budget/src/entities/weekly-aggregate.ts
+++ b/budget/src/entities/weekly-aggregate.ts
@@ -1,13 +1,14 @@
 /**
  * Single source of truth for the WeeklyAggregate entity.
- * WeeklyAggregate is a read-only computed entity: it has Firestore read, IDB storage,
- * upload parse, and seed declaration layers, but no export path (not in exportToJson).
+ * WeeklyAggregate is a computed entity: it has Firestore read, IDB storage,
+ * upload parse, export (exportToJson), and seed declaration layers.
  * All per-entity representations are defined or imported here; adaptor functions live alongside them.
  */
 import { Timestamp } from "firebase/firestore";
 import type { QueryDocumentSnapshot, DocumentData } from "firebase/firestore";
 import type { GroupId } from "@commons-systems/authutil/groups";
 import {
+  msToISO,
   optionalString,
   parseISOTimestamp,
   requireMs,
@@ -104,6 +105,17 @@ export function idbToWeeklyAggregate(row: IdbWeeklyAggregate): WeeklyAggregate {
     creditTotal: row.creditTotal,
     unbudgetedTotal: row.unbudgetedTotal,
     groupId: null as GroupId | null,
+  };
+}
+
+// ── IdbWeeklyAggregate → RawWeeklyAggregate (export) ─────────────────────────
+
+export function weeklyAggregateToRawJson(a: IdbWeeklyAggregate): RawWeeklyAggregate {
+  return {
+    id: a.id,
+    weekStart: msToISO(a.weekStartMs),
+    creditTotal: a.creditTotal,
+    unbudgetedTotal: a.unbudgetedTotal,
   };
 }
 
--- a/budget/src/export.ts
+++ b/budget/src/export.ts
@@ -24,9 +24,11 @@ import type { IdbRule } from "./entities/rule.js";
 import { ruleToRawJson } from "./entities/rule.js";
 import type { IdbNormalizationRule } from "./entities/normalization-rule.js";
 import { normalizationRuleToRawJson } from "./entities/normalization-rule.js";
+import type { IdbWeeklyAggregate } from "./entities/weekly-aggregate.js";
+import { weeklyAggregateToRawJson } from "./entities/weekly-aggregate.js";
 
 export async function exportToJson(): Promise<string> {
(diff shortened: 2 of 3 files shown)


natb1:
## QA — Issue #1262 / PR #1333

**Changed files:** `budget/src/entities/weekly-aggregate.ts`, `budget/src/export.ts`, `budget/test/export.test.ts`

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests pass with new `weeklyAggregates` export (`npx vitest run --root budget test/export.test.ts`) | ✅ PASS (16/16) |
| 2 | Type-check clean (`npx tsc --noEmit --project budget`) | ✅ PASS |
| 3 | Browser FSA write-back round-trip (load `.benc`, edit, reload, verify charts) | ⬜ Out of scope |

**Counts:** 2 PASS · 0 FAIL · 0 SKIP · 0 deferred-to-office-hours

Item 3 is out of scope for automated QA — it requires the user's private `.benc` ETL snapshot, which is not Firestore-seeded data. Both acceptance criteria (#1262) are fully covered by the unit tests. The implementation plan marks the browser end-to-end check as optional.

### Acceptance criteria status
- [x] A write-back round-trips `weeklyAggregates` unchanged — covered by `export.test.ts` round-trip test
- [x] A round-trip test loads an ETL snapshot, exports, reloads, and asserts no store is silently emptied — covered by ETL-snapshot round-trip test in `export.test.ts`

**Autonomous QA: clean pass.** Proceeding to apply `dispatch:qa-done`.

natb1:
## Review findings

### Fixed

**F1 — Mock implementation not restored after ETL-snapshot test override** (`budget/test/export.test.ts:485`, confidence: high)

`vi.clearAllMocks()` in `afterEach` clears call counts but not mock implementations. The `mockGetAll.mockImplementation(...)` override inside `"ETL-snapshot round-trip empties no store"` was not restored, so any test added after it in the same `describe` block would inherit the overridden implementation for `journalEntries`, `journalLegs`, and `accounts`. Wrapped the test body in `try { ... } finally { mockGetAll.mockImplementation(original); }`. _(commit 91764565)_

**F2 — Misleading "generic guard" comment on static `etlStores` list** (`budget/test/export.test.ts:503`, confidence: high)

Two reviewers flagged the same root issue: `etlStores` is a hardcoded static list, but the comment described it as a "generic guard" that would automatically catch a forgotten store in `export.ts`. It doesn't — a new ETL store added to `export.ts` that is omitted from this list produces no test failure. Replaced the comment with accurate text: the list covers ETL-emitted stores from `export.go` and must be updated manually; `statementItems`, `reconciliationNotes`, and `reconciliationEvents` are excluded as app-only stores not populated in `setupMocks`. _(commit 91764565)_

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

### Security

No attack surface issues found. The diff adds a pure data-transformation function (`weeklyAggregateToRawJson`) that maps already-validated IDB records to plain JSON-serializable scalars. No new injection points, auth surfaces, Firestore rule changes, or credential exposure. CodeQL: 0 open alerts on this PR ref.

PR #1343, opened 2026-06-11
budget: honor mid-period overrides in bulk balance functions

### LINKED ISSUE #1264: budget: mid-period budget overrides are honored by the per-transaction balance but ignored by the bulk balance functions
## Context

`computeBudgetBalance` (`src/balance.ts:203-239`) resolves the period containing an override date and honors it. `computePeriodBalances` / `computeAllBudgetBalances` (`src/balance.ts:284-294,326-336`) use `findLatestOverride` (date <= periodStart) then check `date >= periodStart && date < periodEnd`.

## Defect

`findLatestOverride` only returns overrides with `date <= periodStart`, so the within-period check passes only when `date === periodStart` exactly. The UI's "Add override" defaults to today at midnight UTC (`src/pages/budgets-hydrate.ts:533`) — almost never a period boundary — so a typical override does nothing in the period-balance chart and bulk balances, while the transaction-detail balance reflects it. Two views of the same budget disagree.

## Acceptance criteria

- [ ] Bulk and per-transaction balance functions apply a mid-period override identically.
- [ ] A shared period-start resolver is used by all three functions, with a test asserting agreement.

natb1:
Closes #1264

## Problem

`budget/src/balance.ts` computes budget balances three ways: per-transaction
(`computeBudgetBalance`) and two bulk functions (`computePeriodBalances`,
`computeAllBudgetBalances`). A *mid-period budget override* — a manual balance
reset whose date falls strictly inside a period — was honored by the
per-transaction function but **silently ignored by both bulk functions**, so the
transaction-detail balance and the period-balance chart disagreed for the same
budget.

Root cause: both bulk functions resolved the override with
`findLatestOverride(budget.overrides, periodStartMs)`, which returns only
overrides with `date <= periodStartMs`, then gated on
`override.date >= periodStartMs && override.date < periodEndMs`. Those two
conditions can both hold only when `date === periodStartMs` *exactly*. The UI's
"Add override" defaults to today at midnight UTC — almost never a period
boundary — so a typical override did nothing in the bulk paths.
`computeBudgetBalance` avoided the bug by locating the override's *own* period,
honoring any in-period date.

Every pre-existing override test used `date === periodStart` (Jan 13 = w2 start),
which is why the bug went uncaught.

## Fix

Introduce one shared, period-scoped resolver and route all three functions
through it:

```ts
findOverrideInPeriod(overrides, periodStartMs, periodEndMs)
```

It returns the latest override whose date falls within the half-open period
`[periodStartMs, periodEndMs)`, or null. The reset rule becomes identical and
period-local in all three functions: if an override falls inside this period,
`running = override.balance`; else apply normal rollover.

`computeBudgetBalance` was rewritten from its `findLatestOverride` + `findIndex`
+ `startIdx` seeding to a per-period reset loop using the shared resolver. The
rewrite is behavior-preserving — a later in-period override unconditionally
overwrites `running`, the same effect the old seeding produced — and the full
existing override suite stays green, unchanged, as the equivalence proof. With
no remaining callers, `findLatestOverride` is removed.

## Changes

- Add exported `findOverrideInPeriod`; route `computePeriodBalances` and
  `computeAllBudgetBalances` through it.
- Rewrite `computeBudgetBalance` to use the shared resolver (per-period reset
  loop).
- Remove the now-dead `findLatestOverride`, its test, and a stale doc-comment
  reference in `entities/budget.ts` (now naming `findOverrideInPeriod`).
- Add a regression test (`mid-period override agreement (#1264)`) with a
  strictly mid-period override (Jan 16, inside w2 `[Jan13, Jan20)`) asserting all
  three functions agree (`txn-w2` = 20, `txn-w3` = 95). Against the pre-fix bulk
  code w2 would be 130 (override ignored); only the corrected resolver passes.

## Acceptance criteria

- [x] Bulk and per-transaction balance functions apply a mid-period override
  identically.
- [x] A shared period-start resolver (`findOverrideInPeriod`) is used by all
  three functions, with a test asserting agreement.

## Verification

- `npx vitest run --root budget test/balance.test.ts` — 162 tests pass,
  including the unchanged `computeBudgetBalance` override suite (equivalence) and
  the new mid-period agreement test.
- `grep -rn findLatestOverride budget` — returns nothing.

### DIFF
--- a/budget/src/balance.ts
+++ b/budget/src/balance.ts
@@ -146,15 +146,6 @@ function periodsForBudget(periods: BudgetPeriod[], budgetId: BudgetId): BudgetPe
     .sort((a, b) => a.periodStart.toMillis() - b.periodStart.toMillis());
 }
 
-/** Return the override with the greatest date <= beforeMs, or null. Requires overrides sorted by date ascending. */
-export function findLatestOverride(overrides: BudgetOverride[], beforeMs: number): BudgetOverride | null {
-  let result: BudgetOverride | null = null;
-  for (const o of overrides) {
-    if (o.date.toMillis() <= beforeMs) result = o;
-    else break;
-  }
-  return result;
-}
 
 /**
  * Latest override whose date falls within the half-open period
--- a/budget/src/balance.ts
+++ b/budget/src/balance.ts
@@ -217,43 +217,30 @@ export function computeBudgetBalance(
 
   const targetPeriod = periods[targetPeriodIndex];
 
-  // Check for override that applies at or before the target period start
-  const targetStartMs = targetPeriod.periodStart.toMillis();
-  const override = findLatestOverride(budget.overrides, targetStartMs);
-
-  // Determine which period to start accumulating from
-  let startIdx = 0;
+  // Accumulate through prior periods, resetting to any override that falls
+  // inside the period (else applying normal rollover), then subtracting the
+  // period total. A later in-period override overwrites all prior accumulation.
   let running = 0;
-
-  // Find the period containing the override date (if any)
-  const overridePeriodIdx = override ? periods.findIndex(
-    (p) => {
-      const overrideMs = override.date.toMillis();
-      return p.periodStart.toMillis() <= overrideMs && overrideMs < p.periodEnd.toMillis();
-    },
-  ) : -1;
-
-  if (override && overridePeriodIdx !== -1 && overridePeriodIdx <= targetPeriodIndex) {
-    // Start from the override: set balance to override value, subtract the override period's total, then continue
-    startIdx = overridePeriodIdx + 1;
-    running = override.balance - periods[overridePeriodIdx].total;
-  }
-
-  // Accumulate through prior periods (from startIdx, which is 0 when no override)
-  for (let i = startIdx; i < targetPeriodIndex; i++) {
+  for (let i = 0; i < targetPeriodIndex; i++) {
+    const periodStart = periods[i].periodStart.toMillis();
+    const periodEnd = periods[i].periodEnd.toMillis();
+    const override = findOverrideInPeriod(budget.overrides, periodStart, periodEnd);
     const prevMs = i > 0 ? periods[i - 1].periodStart.toMillis() : null;
-    const allow = periodAllowance(budget.allowance, budget.allowancePeriod, prevMs, periods[i].periodStart.toMillis());
-    running = applyRollover(running, allow, budget.rollover);
+    const allow = periodAllowance(budget.allowance, budget.allowancePeriod, prevMs, periodStart);
+    running = override ? override.balance : applyRollover(running, allow, budget.rollover);
     running -= periods[i].total;
   }
 
-  // Apply rollover entering the target period (unless override is in this period)
-  if (override && overridePeriodIdx === targetPeriodIndex) {
-    // Override is in the target period: running starts at override balance
-    running = override.balance;
+  // Enter the target period: an in-period override replaces the rollover and no
+  // total is subtracted (the same-period txn walk below subtracts within-period spend).
+  const targetStartMs = targetPeriod.periodStart.toMillis();
+  const targetEndMs = targetPeriod.periodEnd.toMillis();
+  const targetOverride = findOverrideInPeriod(budget.overrides, targetStartMs, targetEndMs);
+  if (targetOverride) {
+    running = targetOverride.balance;
   } else {
     const prevMs = targetPeriodIndex > 0 ? periods[targetPeriodIndex - 1].periodStart.toMillis() : null;
-    const allow = periodAllowance(budget.allowance, budget.allowancePeriod, prevMs, targetPeriod.periodStart.toMillis());
+    const allow = periodAllowance(budget.allowance, budget.allowancePeriod, prevMs, targetStartMs);
     running = applyRollover(running, allow, budget.rollover);
   }
 
--- a/budget/src/balance.ts
+++ b/budget/src/balance.ts
@@ -156,6 +156,25 @@ export function findLatestOverride(overrides: BudgetOverride[], beforeMs: number
   return result;
 }
 
+/**
+ * Latest override whose date falls within the half-open period
+ * [periodStartMs, periodEndMs), or null. Requires overrides sorted by
+ * date ascending.
+ */
+export function findOverrideInPeriod(
+  overrides: BudgetOverride[],
+  periodStartMs: number,
+  periodEndMs: number,
+): BudgetOverride | null {
+  let result: BudgetOverride | null = null;
+  for (const o of overrides) {
+    const ms = o.date.toMillis();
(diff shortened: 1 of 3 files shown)


natb1:
## QA Summary — #1264: mid-period budget overrides

**PR:** #1343 | **Branch:** `1264-budget-mid-period-budget-ov`

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests pass | ✅ PASS |
| 2 | Budget app TypeScript compiles clean | ✅ PASS |
| 3 | Budgets page loads in browser | ⏭ SKIP |
| 4 | Period balance chart reflects mid-period overrides | 👤 Deferred |
| 5 | Transaction balance agrees with period chart | 👤 Deferred |

**PASS: 2 | SKIP: 1 | FAIL: 0 | Bugs found: 0**

---

### Machine-verifiable checks

**Unit tests (162/162):** All pass, including the new mid-period agreement regression test that directly guards the bug fix. Full budget suite: 963/970 (7 pre-existing skips unrelated to this change).

**TypeScript:** `npx tsc --noEmit --project budget` — clean.

**Dead code removed:** `findLatestOverride` has zero references in `budget/` (confirmed via grep).

### Browser walkthrough — SKIP

The QA server started successfully (Vite port 37981, Firestore emulator 40503), but macOS Chrome could not reach it. SSH port forwarding is needed first:

```
ssh -N -L 37981:localhost:37981 -L 40503:localhost:40503 nixos
```

Playwright acceptance tests also cannot run locally (known Nix environment constraint; CI's `acceptance` job is authoritative).

### Deferred to office-hours

The following items need a human with live override data:

1. **Period balance chart reflects mid-period overrides** — View a budget that has an override; confirm the period chart's bar/value for the overridden period shows the post-override balance (not the pre-fix accumulated value, which would be ~30% higher).

2. **Transaction balance agrees with period chart** — Compare the per-transaction running balance for a transaction inside an overridden period vs the period chart's running balance. These should now agree (pre-fix they disagreed — the core bug).

To exercise these, either use real data with a mid-period override or add one via the "Add override" button (defaults to today at midnight UTC, which is mid-period by design).

natb1:
## QA Summary — #1264: mid-period budget overrides

**PR:** #1343 | **Branch:** `1264-budget-mid-period-budget-ov`

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests pass | ✅ PASS |
| 2 | Budget app loads without JS errors | ✅ PASS |
| 3 | Budgets page loads and shows period balance chart | ✅ PASS |
| 4 | Period balance chart reflects mid-period overrides | 👤 Deferred |
| 5 | Transaction balance agrees with period chart | 👤 Deferred |

**PASS: 3 | FAIL: 0 | Bugs found: 0**

---

### Machine-verifiable checks

**Unit tests (162/162):** All pass, including the new mid-period agreement regression test that directly guards the bug fix. Full budget suite: 963/970 (7 pre-existing skips unrelated to this change).

**TypeScript:** `npx tsc --noEmit --project budget` — clean.

**Dead code removed:** `findLatestOverride` has zero references in `budget/` (confirmed via grep).

**Browser (macOS Chrome via SSH tunnel):** App loaded at `http://localhost:41207` — Vite connected, analytics-disabled debug message, zero JS errors. Period balance chart rendered with all 4 budget lines (Food, Housing, Vacation, Other) and transaction chart. No console errors throughout the walkthrough.

> Note: Playwright acceptance tests cannot run locally (known Nix environment constraint; CI's `acceptance` job is authoritative).

### Deferred to office-hours

The seed data has no overrides by design (`BudgetSeedData` omits `overrides`, defaulting to `[]`). Verifying the visual effect of the fix requires a human with real data or a manually-added override:

1. **Period balance chart reflects mid-period overrides** — View a budget that has an override; confirm the period chart's value for the overridden period shows the post-override balance (not the pre-fix accumulated value, which would be noticeably higher).

2. **Transaction balance agrees with period chart** — Compare the per-transaction running balance for a transaction inside an overridden period vs the period chart's running balance for that same period. These should now agree (pre-fix they disagreed — the core bug).

To exercise: use real data with a mid-period override, or add one via the "Add override" button (defaults to today at midnight UTC, which is mid-period by design).

natb1:
## Review

### Fixed

- **`findOverrideInPeriod` has no dedicated unit tests** (`budget/test/balance.test.ts`) — Added a `describe('findOverrideInPeriod')` block covering 7 boundary edge cases: empty overrides → null; all overrides before `periodStart` → null; override exactly at `periodStart` → returned; override exactly at `periodEnd − 1ms` → returned; override at `periodEnd` → null; multiple overrides in period → latest returned; override after period → null. (commit `3f4230fc`)
- **Test title/comments describe old `findLatestOverride` algorithm** (`budget/test/balance.test.ts:1425`) — Updated test title to `'multiple overrides in separate prior periods: each applied in its own period'` and replaced old per-latest-override comment with a step-by-step trace using the new per-period semantics. (commit `3f4230fc`)
- **Comment references removed variable `overridePeriodIdx`** (`budget/test/balance.test.ts:1471`) — Replaced stale comment with one using `findOverrideInPeriod` semantics: override at Jan 15 falls in no period boundary — `findOverrideInPeriod` returns null for each period. (commit `3f4230fc`)
- **Comment uses old `findLatestOverride` terminology** (`budget/test/balance.test.ts:1444`) — Updated to: `// Override at Jan 27 >= w2.end (Jan 20): findOverrideInPeriod returns null for w2 → normal rollover.` (commit `3f4230fc`)

### Informational

- **Unconditional `periodAllowance()` call in prior-period loop** (`budget/src/balance.ts:219–221`) — `prevMs` and `allow` are computed on every iteration, but when `findOverrideInPeriod` returns an override the result is discarded. The cost is minimal (a couple of arithmetic operations per period) and extracting conditionally would complicate the loop without measurable benefit.
- **Regression test: `txn1` is before override date** (`budget/test/balance.test.ts:1566–1568`) — `txn1` has `timestamp: ts('2025-01-15')` (Jan 15) while the override date is Jan 16. This intentionally validates that the override applies to transactions *on or after* its date, not those before it in the same period. No change needed — the test is correct by design.

### Out-of-scope (security, pre-existing)

- **`idbToBudget` deserializes `BudgetOverride.balance` without validation** (`budget/src/entities/budget.ts:197–210`) — `o.balance` is mapped directly from IndexedDB rows. A corrupted or tampered local IDB store could supply `NaN`, `Infinity`, or an injected string, propagating to balance computations. Pre-existing code, not touched by this diff.
- **`findOverrideInPeriod` relies on overrides being sorted ascending** (`budget/src/balance.ts:155–167`) — The early-break `else if (ms >= periodEndMs) break` is only correct when overrides are sorted. If an upstream caller supplies an unsorted slice the break fires prematurely. The sort invariant is documented on `BudgetOverride` but not enforced at runtime. Pre-existing assumption, not introduced by this diff.

### Security

No attack surface changes to review beyond the pre-existing items noted above. No CodeQL alerts. No dependency changes.

PR #1347, opened 2026-06-11
budget: encrypted exports download as .benc; Sankey hydration stops leaking a document listener

natb1:
Closes #1267

Fixes two independent defects in the budget app.

## Encrypted export mislabeled

`budget/src/main.ts`: the export click handler builds a BENC binary blob
(`application/octet-stream`) when `importPassword` is set, but always named the
download `budget-<group>-<date>.json`. The artifact was mislabeled for any
non-budget tool and contradicted the `.benc` accept filter on the upload input.
Re-upload still worked via magic-byte sniffing, but the extension was wrong.

The download extension is now derived from `importPassword` — the same flag that
already selects the binary vs. JSON blob — so encrypted exports download as
`.benc` and unencrypted exports keep `.json`. The encrypted e2e test
(`budget/e2e/export.spec.ts`) now also asserts the suggested filename ends in
`.benc`; the unencrypted test continues to pin `…-<today>.json` unchanged.

## Sankey listener leak

`budget/src/pages/home-chart.ts`: every `hydrateCategorySankey` call registered a
permanent `document.addEventListener(TRANSACTIONS_APPENDED_EVENT, …)` that was
never removed. A MutationObserver re-hydrates after every navigation back to
`/transactions`, so stale listeners accumulated — each kept a dead container
alive and re-ran `filterTable` with stale closure state on every scroll-append.

The handler is now stored in a module-level `appendedListener` handle, and each
hydration removes the prior handler before registering its own (remove-before-add),
so at most one listener is ever active. The handler body is unchanged — the
`container.isConnected`-guarded chart-update block and the unconditional
table re-filter block (#578 behavior) are preserved verbatim. A new
`home-chart.test.ts` test hydrates twice and asserts the second hydration removes
the first listener.

## Verification

- `npm run test --prefix budget` — unit suites green, including the new
  listener-accumulation test.
- `budget/e2e/export.spec.ts` — unencrypted test asserts `…-<today>.json`,
  encrypted test now also asserts the `.benc` suffix.

### DIFF
--- a/budget/e2e/export.spec.ts
+++ b/budget/e2e/export.spec.ts
@@ -82,7 +82,8 @@ test.describe("export", () => {
 
     const download = await triggerExportDownload(page);
 
-    expect(download.suggestedFilename()).toMatch(/\.benc$/);
+    const today = new Date().toISOString().slice(0, 10);
+    expect(download.suggestedFilename()).toBe(`budget-Test Household-${today}.benc`);
 
     const content = await (await download.createReadStream()).toArray();
     const buf = Buffer.concat(content);
--- a/budget/e2e/export.spec.ts
+++ b/budget/e2e/export.spec.ts
@@ -82,6 +82,8 @@ test.describe("export", () => {
 
     const download = await triggerExportDownload(page);
 
+    expect(download.suggestedFilename()).toMatch(/\.benc$/);
+
     const content = await (await download.createReadStream()).toArray();
     const buf = Buffer.concat(content);
     expect(buf.subarray(0, 4).toString()).toBe("BENC");
--- a/budget/src/main.ts
+++ b/budget/src/main.ts
@@ -407,7 +407,8 @@ exportButton.addEventListener("click", async () => {
     a.href = url;
     const date = new Date().toISOString().slice(0, 10);
     const groupName = state.source === "local" ? state.groupName : "budget";
-    a.download = `budget-${groupName}-${date}.json`;
+    const ext = importPassword ? "benc" : "json";
+    a.download = `budget-${groupName}-${date}.${ext}`;
     document.body.appendChild(a);
     a.click();
     document.body.removeChild(a);
--- a/budget/src/pages/home-chart.ts
+++ b/budget/src/pages/home-chart.ts
@@ -216,6 +216,15 @@ export function hydrateCategorySankey(container: HTMLElement): void {
   }
   const parsed: unknown = JSON.parse(scriptEl.textContent);
   assertChartTransactions(parsed);
+
+  // Remove any prior run's document listener before this run can early-return,
+  // so an empty-data re-hydration cannot leave a stale listener firing with the
+  // previous run's closed-over state (#1267).
+  if (appendedListener) {
+    document.removeEventListener(TRANSACTIONS_APPENDED_EVENT, appendedListener);
+    appendedListener = null;
+  }
+
   const allTxns = parsed;
   if (allTxns.length === 0) {
     container.textContent = "No transaction data to chart.";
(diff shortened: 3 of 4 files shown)


natb1:
## QA — #1267 budget: encrypted exports as .benc; Sankey listener fix

**Result: All machine-verifiable items PASS — no bugs found.**

| # | Item | Result |
|---|------|--------|
| 1 | Unencrypted export downloads as `.json` | ✅ PASS |
| 2 | Encrypted export downloads as `.benc` | ✅ PASS |
| 3 | Sankey hydration — no listener accumulation across navigations | ✅ PASS |
| 4 | All existing tests remain green | ✅ PASS |

**Counts:** 4 walked · 4 PASS · 0 FAIL · 0 SKIP · 0 needs-human-judgment · 0 bugs

### Evidence

- **Item 1 & 2 (export filenames):** CI `acceptance` job passed. The existing e2e test (`export.spec.ts`) asserts `…-${today}.json` for unencrypted exports and continues to pass; the new assertion `expect(download.suggestedFilename()).toMatch(/\.benc$/)` for encrypted exports passed in CI.
- **Item 3 (Sankey listener):** Local unit tests passed (982/982). The new `home-chart.test.ts` test verifies remove-before-add: the second `hydrateCategorySankey` call removes the first handler before registering its own; a `TRANSACTIONS_APPENDED_EVENT` after two hydrations fires the side-effect exactly once.
- **Item 4 (regression):** All CI checks green — `unit-tests`, `lint`, `typecheck`, `acceptance`, `hook-tests`, `playwright-version-sync`.

### Notes

Local Playwright acceptance check skipped (Nix env / `apt-get` unavailable — known limitation); CI `acceptance` is authoritative and passed. Chrome browser walkthrough skipped: Vite server at WSL port 33929 not reachable from Windows Chrome via WSL2 NAT (emulators bind `0.0.0.0`; Vite port does not). All items verified via CI + local unit test evidence.

natb1:
## Review findings

### Fixed

- **`budget/src/pages/home-chart.ts:220-228`** — Early-return paths in `hydrateCategorySankey` (empty `allTxns` at line 220, empty `weeks` at line 226) bypassed the remove-before-add cleanup, leaving a stale `appendedListener` active after empty-data re-hydrations. Moved the `removeEventListener` / null-reset block above those early returns so every exit path clears the prior listener. (a3a70b4a)

- **`budget/e2e/export.spec.ts:85`** — Encrypted export e2e test asserted only `toMatch(/\.benc$/)` (bare extension regex) rather than pinning the full filename structure. Changed to `toBe(\`budget-Test Household-${today}.benc\`)`, consistent with the pattern the unencrypted test already uses on line 28–29. (a3a70b4a)

- **`budget/test/pages/home-chart.test.ts:842-877`** — Listener-accumulation test's `addSpy.mockRestore()` / `removeSpy.mockRestore()` calls were in the test body without a guard, so a mid-test assertion failure would leave spies unrestored and the second hydration's document listener still attached. Wrapped the spy setup, assertions, and `removeEventListener` cleanup in a `try/finally` block so restoration is unconditional. (b0f5456b)

### Informational

- **`budget/src/pages/home-chart.ts:18`** — The module-level `appendedListener` variable persists across all tests in the same vitest module run. Because vitest reuses the module between tests unless `vi.resetModules()` is called, a listener left active by one test can interfere with a later test. The `try/finally` cleanup applied above mitigates the most acute case; complete isolation would require resetting the module or exporting a reset helper. No code change — tracking only.

- **`budget/src/pages/home-chart.ts:220-229`** — Red-team flagged the early-return paths as leaving stale listeners. This duplicates the code-review finding above (code-review-0), which was fixed. The red-team framing (a security / denial-of-service concern) is not accurate — the issue is a test-isolation and resource-leak bug, not an exploitable attack surface. Dismissed as a false-positive framing of the already-fixed issue.

PR #1356, opened 2026-06-11
qa: pin Vite dev port to a fixed 8-slot collision-safe pool

### LINKED ISSUE #1349: qa: pin the QA-server Vite dev port to a fixed 8-slot pool (collision-safe) so Chrome is approved once
## Context

`run-qa-server.sh` allocates **every** port — the Vite dev server *and* the
Firebase emulators — through `find_available_ports()` (`lib.sh`), which binds to
port `0` and takes whatever ephemeral port the OS returns (typically 32768+). So
the Vite dev server lands on a fresh random port every run.

The `claude-in-chrome` extension gates `navigate` / `tabs_create_mcp` /
`computer` / `javascript_tool` / `gif_creator` **per origin**
(`scheme+host+port`) — see `.claude/rules/chrome-extension.md`. A fresh random
Vite port every run is a fresh origin every run, so the operator must re-approve
the extension popup on every single QA session.

**Key narrowing insight:** only the **Vite port** is an origin the extension
navigates to. The emulator ports are reached by the *page's own JS*
(fetch / Firestore WebChannel), never by the extension, so they never trigger an
approval prompt. Pinning **only** the Vite port to a known pool gets
approve-once; emulator ports stay ephemeral (and should, to avoid collisions
between concurrent QA servers).

## Design

A fixed pool of 8 Vite ports (e.g. `5170`–`5177`). A QA server claims the first
free slot; the operator approves those 8 origins once in Chrome and never again.
8 slots = headroom for up to 8 concurrent QA servers across worktrees.

### Collision-safety is a hard requirement (up to 8 concurrent workers)

A naive "probe with a throwaway bind, close, then let Vite bind" has a
TOCTOU race: two workers starting concurrently can both observe `5170` free,
both pick it, and the loser dies on Vite's `--strictPort`. With up to 8 workers
racing across the 8-slot pool this is a real failure, not a corner case.

Make **the lock, not the bind, the claim.** Each worker walks the pool and, for
each candidate port, acquires a non-blocking `flock` on a per-port lockfile
(e.g. `${TMPDIR:-/tmp}/qa-vite-port-<P>.lock`). The first port whose lock it
wins becomes its slot; the lock fd is held open for the whole server lifetime
(released on exit). Two concurrent workers can never select the same slot
because `flock` arbitrates atomically — closing the probe→bind window entirely
for worker-vs-worker contention. A foreign (non-QA) process squatting a pool
port is handled by a bindability check after the lock is won (release the lock,
try the next slot) with `--strictPort` as the loud backstop.

Because the lock must outlive a command substitution, the claim runs in the main
shell (opening the fd in the caller), not inside `VITE_PORT=$(...)`:

~~~bash
# lib.sh
QA_VITE_PORT_POOL=(5170 5171 5172 5173 5174 5175 5176 5177)
QA_VITE_PORT_LOCK_DIR="${TMPDIR:-/tmp}"

# Sets VITE_PORT and holds an flock on fd 200 for the script's lifetime so
# concurrent QA workers never select the same pool slot. Errors (does NOT fall
# back to a random port — that would re-trigger the Chrome approval prompt) when
# all 8 slots are held.
claim_fixed_vite_port() {
  local p lockfile
  for p in "${QA_VITE_PORT_POOL[@]}"; do
    lockfile="${QA_VITE_PORT_LOCK_DIR}/qa-vite-port-${p}.lock"
    exec 200>"$lockfile" || continue
    if flock -n 200; then
      if node -e "const net=require('net');const s=net.createServer();s.once('error',()=>process.exit(1));s.listen(${p},'0.0.0.0',()=>s.close(()=>process.exit(0)));" 2>/dev/null; then
        VITE_PORT="$p"; return 0      # fd 200 stays open in caller, holding the lock
      fi
      flock -u 200                    # foreign process owns the port; try next slot
    fi
    exec 200>&-
  done
  echo "ERROR: all ${#QA_VITE_PORT_POOL[@]} QA Vite ports (${QA_VITE_PORT_POOL[*]}) are in use" >&2
  return 1
}
~~~

`run-qa-server.sh` — call `claim_fixed_vite_port` (not in a subshell) for the
Vite port; keep emulator ports ephemeral:

~~~bash
# was: read -r VITE_PORT EXTRA_PORTS <<< "$(find_available_ports "$PORT_COUNT")"
claim_fixed_vite_port
echo "Vite dev server will use port $VITE_PORT (fixed pool — approve once in Chrome)"
EMU_PORT_COUNT=$((PORT_COUNT - 1))
EXTRA_PORTS=""
[ "$EMU_PORT_COUNT" -gt 0 ] && EXTRA_PORTS="$(find_available_ports "$EMU_PORT_COUNT")"
~~~

Vite already launches with `--strictPort`, so a race lost to a foreign process
fails loudly rather than drifting to an unapproved port.

`run-acceptance-tests.sh` is **out of scope** — Playwright runs inside WSL with
no extension, so its `find_available_ports` allocation is fine as-is.

## Acceptance criteria

- [ ] `lib.sh` defines an 8-member fixed Vite-port pool and a
      `claim_fixed_vite_port` helper that claims the first free slot via a
      `flock` held for the server's lifetime.
- [ ] `run-qa-server.sh` uses the fixed-pool claim for the Vite port; emulator
      ports remain ephemeral via `find_available_ports`.
- [ ] **No collisions for up to 8 concurrent workers:** 8 QA servers started
      concurrently each get a distinct pool port and all 8 start successfully.
- [ ] A 9th concurrent worker (all 8 slots held) exits non-zero with a clear
      "all QA Vite ports in use" diagnostic — it does **not** fall back to a
      random port.
- [ ] A foreign process squatting one pool port is skipped (the worker claims
      the next free slot) rather than dying on `--strictPort`.
- [ ] `test-dispatch-scripts.sh` covers: single claim returns a pool member;
      8 concurrent claims yield 8 distinct ports; the 9th errors; a squatted
      pool port is skipped.
- [ ] `run-acceptance-tests.sh` is unchanged.

## Notes

The macOS-browser SSH tunnel (separate issue) benefits from this — the Vite
port it forwards becomes one of 8 stable values — but the two are independent:
the Windows default needs no port pinning at all.

natb1:
Closes #1349

## Summary

`run-qa-server.sh` allocated **every** port — the Vite dev server and the
Firebase emulators — through `find_available_ports()`, which binds to port `0`
and takes whatever ephemeral port the OS hands back. So the Vite dev server
landed on a fresh random port every run.

The `claude-in-chrome` extension gates `navigate` / `tabs_create_mcp` /
`computer` / `javascript_tool` / `gif_creator` **per origin**
(`scheme+host+port`). A fresh random Vite port every run is a fresh origin every
run, so the operator had to re-approve the extension popup on every QA session.

Only the **Vite port** is an origin the extension navigates to. Emulator ports
are reached by the page's own JS (fetch / Firestore WebChannel), never by the
extension, so they never trigger an approval prompt. This PR pins **only** the
Vite port to a known 8-slot pool (`5170`–`5177`): the operator approves those 8
origins in Chrome once and never again. Emulator ports stay ephemeral so
concurrent QA servers across worktrees never collide.

## How

- **`lib.sh`** — adds the fixed 8-member pool (`QA_VITE_PORT_POOL`), a lock
  directory (`QA_VITE_PORT_LOCK_DIR`), and `claim_fixed_vite_port`. The helper
  walks the pool and, for each candidate, acquires a non-blocking `flock` on a
  per-port lockfile held on fd 200 for the script's lifetime. **The lock, not
  the bind, is the claim** — `flock` arbitrates atomically, so two concurrent
  workers can never select the same slot, closing the TOCTOU race for up to 8
  concurrent workers. A post-lock bindability check skips a pool port squatted
  by a foreign process. When all 8 slots are held it errors loudly and does
  **not** fall back to a random port (which would re-trigger the Chrome prompt).

- **`run-qa-server.sh`** — calls `claim_fixed_vite_port` in the main shell (not a
  subshell, so the `flock`'d fd survives the script's lifetime) for the Vite
  port, and keeps the emulator ports ephemeral via `find_available_ports`. Under
  `set -euo pipefail` an all-slots-held failure aborts loudly. Vite still
  launches with `--strictPort` as the backstop for a slot lost to a foreign
  process between the bindability check and Vite's own bind.

- **`test-dispatch-scripts.sh`** — covers the four acceptance criteria: a single
  claim returns a pool member; 8 concurrent claims yield 8 distinct ports; a 9th
  claim (all slots held) exits non-zero with the "all … in use" diagnostic and
  no random-port fallback; a foreign squatter on a pool port is skipped. Each
  case isolates via a fresh lock-dir and a known-free pool discovered through
  `find_available_ports`, so the suite never touches the real `5170`–`5177` pool.

`run-acceptance-tests.sh` is intentionally **unchanged** — Playwright runs inside
WSL with no extension, so its `find_available_ports` allocation is fine.

## Verification

  full suite green (2115/2115), confirmed stable across 5 consecutive runs.
- `bash -n` clean on `lib.sh`, `run-qa-server.sh`, and `test-dispatch-scripts.sh`.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -751,6 +751,38 @@ find_available_port() {
   find_available_ports 1
 }
 
+# Fixed pool of Vite dev-server ports for QA servers. The claude-in-chrome
+# extension gates navigation per origin (scheme+host+port), so pinning the Vite
+# port to a known 8-slot pool lets the operator approve those 8 origins in Chrome
+# once instead of re-approving a fresh random port every QA session. Emulator
+# ports stay ephemeral (find_available_ports) — the page's own JS reaches them,
+# never the extension, so they never trigger an approval prompt.
+#
+# File-scope globals (not local/readonly) so tests can override them.
+QA_VITE_PORT_POOL=(5170 5171 5172 5173 5174 5175 5176 5177)
+QA_VITE_PORT_LOCK_DIR="${TMPDIR:-/tmp}"
+
+# Sets VITE_PORT and holds an flock on fd 200 for the script's lifetime so
+# concurrent QA workers never select the same pool slot. Errors (does NOT fall
+# back to a random port — that would re-trigger the Chrome approval prompt)
+# when all 8 slots are held.
+claim_fixed_vite_port() {
+  local p lockfile
+  for p in "${QA_VITE_PORT_POOL[@]}"; do
+    lockfile="${QA_VITE_PORT_LOCK_DIR}/qa-vite-port-${p}.lock"
+    exec 200>"$lockfile" || continue
+    if flock -n 200; then
+      if node -e "const net=require('net');const s=net.createServer();s.once('error',()=>process.exit(1));s.listen(${p},'0.0.0.0',()=>s.close(()=>process.exit(0)));" 2>/dev/null; then
+        VITE_PORT="$p"; return 0      # fd 200 stays open in caller, holding the lock
+      fi
+      flock -u 200                    # foreign process owns the port; try next slot
+    fi
+    exec 200>&-
+  done
+  echo "ERROR: all ${#QA_VITE_PORT_POOL[@]} QA Vite ports (${QA_VITE_PORT_POOL[*]}) are in use" >&2
+  return 1
+}
+
 # Print the Remote access block for QA-server startup: the universal localhost
 # URL plus a copy-paste `ssh -L` command forwarding the Vite port and every
 # allocated emulator port to a remote client's localhost, so the served origin
--- a/.claude/skills/dispatch-propagate/scripts/run-qa-server.sh
+++ b/.claude/skills/dispatch-propagate/scripts/run-qa-server.sh
@@ -29,8 +29,16 @@ if [ "$USES_AUTH" = true ]; then PORT_COUNT=$((PORT_COUNT + 1)); fi
 if [ "$USES_STORAGE" = true ]; then PORT_COUNT=$((PORT_COUNT + 1)); fi
 if [ "$USES_FUNCTIONS" = true ]; then PORT_COUNT=$((PORT_COUNT + 1)); fi
 
-read -r VITE_PORT EXTRA_PORTS <<< "$(find_available_ports "$PORT_COUNT")"
-echo "Vite dev server will use port $VITE_PORT"
+# Claim the Vite port from a fixed pool (collision-safe via flock, held for the
+# script's lifetime) so the claude-in-chrome extension approves the origin once.
+# Called in the main shell — not a subshell — so the flock'd fd 200 survives.
+claim_fixed_vite_port
+echo "Vite dev server will use port $VITE_PORT (fixed pool — approve once in Chrome)"
+# Emulator ports stay ephemeral — the page's own JS reaches them, never the
+# extension, so they never trigger an approval prompt.
+EMU_PORT_COUNT=$((PORT_COUNT - 1))
+EXTRA_PORTS=""
+[ "$EMU_PORT_COUNT" -gt 0 ] && EXTRA_PORTS="$(find_available_ports "$EMU_PORT_COUNT")"
 
 NAMESPACE=""
 if [ "$USES_FIRESTORE" = true ]; then
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -22101,15 +22101,17 @@ rm -rf "$cvp_lockdir3" "$cvp_results3" "$cvp_stderr3"
 # free slot (post-lock bindability check), not die on the squatted port.
 cvp_lockdir4=$(mktemp -d)
 cvp_squat_port="${CVP_POOL_ARR[0]}"
-node -e "require('net').createServer().listen(${cvp_squat_port},'0.0.0.0');" &
+# The squatter retries on error (so a transient collision can't kill it) and
+# announces readiness by printing READY once bound. We wait for that line rather
+# than bind-probing the port ourselves — a probe would momentarily hold the port
+# and race the squatter's own listen, which is exactly the flake to avoid.
+cvp_squat_out=$(mktemp)
+node -e "const net=require('net');function bind(){const s=net.createServer();s.on('error',()=>setTimeout(bind,50));s.listen(${cvp_squat_port},'0.0.0.0',()=>console.log('READY'));}bind();setInterval(()=>{},1e9);" >"$cvp_squat_out" 2>/dev/null &
 cvp_squat_pid=$!
-# Wait until the squatter has actually bound (the port is no longer bindable).
-for _ in $(seq 1 80); do
-  if node -e "const net=require('net');const s=net.createServer();s.once('error',()=>process.exit(1));s.listen(${cvp_squat_port},'0.0.0.0',()=>s.close(()=>process.exit(0)));" 2>/dev/null; then
-    sleep 0.1
-  else
-    break
-  fi
+# Wait until the squatter reports it has bound the port.
+for _ in $(seq 1 100); do
+  grep -q READY "$cvp_squat_out" 2>/dev/null && break
+  sleep 0.1
 done
 cvp_out4=$(
   source "$SCRIPT_DIR/lib.sh"
@@ -22126,7 +22128,7 @@ else
 fi
 kill "$cvp_squat_pid" 2>/dev/null || true
 wait "$cvp_squat_pid" 2>/dev/null || true
-rm -rf "$cvp_lockdir4"
+rm -rf "$cvp_lockdir4" "$cvp_squat_out"
 
 # ============================================================================
 # dispatch-reconcile-ready tests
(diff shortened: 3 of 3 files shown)


natb1:
## QA Summary — #1349 (PR #1356)

**Non-browser path** — all changes are script-only (lib.sh, run-qa-server.sh, test-dispatch-scripts.sh).

| # | Item | Result |
|---|------|--------|
| 1 | Syntax check: lib.sh + run-qa-server.sh | PASS |
| 2 | 8-member fixed Vite-port pool `(5170–5177)` defined in lib.sh | PASS |
| 3 | `claim_fixed_vite_port`: flock-based claim, sets `VITE_PORT`, errors on exhaustion | PASS |
| 4 | run-qa-server.sh: Vite uses fixed-pool claim (main shell, not subshell); emulator ports ephemeral via `find_available_ports` | PASS |
| 5 | run-acceptance-tests.sh unchanged | PASS |
| 6 | Test suite: all 4 new cases pass (single claim, 8 concurrent distinct ports, 9th errors, squatter skipped) | PASS |

**Totals:** 6 passed, 0 failed, 0 skipped, 0 deferred to office-hours.

**Test output (claim_fixed_vite_port section):**
```
PASS: claim: single claim succeeds (exit 0)
PASS: single claim's VITE_PORT is a pool member
PASS: claim: 8 concurrent claims all succeed
PASS: claim: 8 concurrent claims yield 8 distinct ports
PASS: claim: 9th claim (all slots held) exits non-zero
PASS: claim: 9th claim leaves VITE_PORT unset (no random-port fallback)
PASS: 9th claim prints the 'all … in use' diagnostic
PASS: claim: squatted pool port → claim succeeds on next slot (exit 0)
PASS: claim skipped the squatted port, took next free slot
```

Overall test suite: **2115/2115 passed, 0 failed**.

natb1:
## Review — #1349: pin Vite dev port to a fixed 8-slot collision-safe pool

No security vulnerabilities or code defects found. No fixes applied.

---

### Fixed
_None._

### Required (security)
_None._

### Informational

- **`lib.sh:773`** — `exec 200>"$lockfile" || continue` silently skips all 8 pool ports when `QA_VITE_PORT_LOCK_DIR` is unwritable or nonexistent; the caller gets the "all ports in use" error message rather than a more specific I/O error. Acceptable for a dev-tooling script. _(code-review)_
- **`lib.sh:769`** — `claim_fixed_vite_port` hardcodes fd 200. If a caller ever holds fd 200 open, the `exec 200>` silently clobbers it. Documented tradeoff: the function must hold the fd in the caller's process so the lock survives past the function scope; fd 200 is conventional for this pattern in bash. _(code-review)_
- **`lib.sh:778`** — `flock -u 200` before `exec 200>&-` is redundant; closing the fd implicitly releases all flocks. Harmless defensive code. _(review)_
- **`test-dispatch-scripts.sh:22013`** — Case 1 runs `claim_fixed_vite_port` inside a command-substitution subshell (`cvp_out1=$(...)`), which means fd 200 is closed when the subshell exits and the lock is not held in the parent. The test observes the result but does not verify lock persistence; it still validates the single-claim happy path. _(code-review)_
- **`test-dispatch-scripts.sh:22007`** — The test pool is discovered once via `find_available_ports 8`; there is a narrow TOCTOU window between discovery and the concurrent workers binding those ports. In practice this is the same ephemeral-port race the feature is designed to eliminate for the production pool, but acceptable in a test harness that uses its own temp lock-dir. _(code-review, review)_
- **`test-dispatch-scripts.sh:22069`** — Case 3's polling loop waits up to 8 × 0.1 s per iteration for all 8 holder subshells to write their result files. Under heavy load this could time out before holders finish, causing a spurious FAIL. _(code-review)_
- **`test-dispatch-scripts.sh:22069–22072`** — Same Case 3 polling window; the 80-iteration cap is a practical bound, not a guarantee. Noted as a known test-harness limitation. _(review)_
- **`test-dispatch-scripts.sh:22109`** — Case 4 squatter readiness is polled with a `node` ping loop; if the squatter server never becomes ready the poll exits after its own cap and the test continues, which could produce a false PASS rather than a false FAIL. _(red-team)_

### Dismissed

- **`test-dispatch-scripts.sh:22069–22072`** — Flagged as a reliability concern for the 8-second timeout in Case 3; the wait bound is intentionally generous and already matches the pattern used elsewhere in the suite. _Dismissed: correct design, not a bug._ _(review)_
- **`lib.sh:773`** — Flagged by red-team as a potential symlink-follow issue on a shared host. The `exec 200>"$lockfile"` call follows symlinks, but `QA_VITE_PORT_LOCK_DIR` defaults to `$TMPDIR` which is per-user on any modern system; on a shared `/tmp` the sticky bit prevents cross-user symlink attacks on existing files. _Dismissed: not a realistic threat in this context._ _(red-team, review)_

### False-positive (security)

- **`lib.sh:775`** — `${p}` interpolated into inline Node.js source flagged as injection risk. `p` is drawn exclusively from the integer literal array `QA_VITE_PORT_POOL=(5170…5177)` — no external input. _False-positive._ _(input-validation)_
- **`test-dispatch-scripts.sh:22109`** — `${cvp_squat_port}` interpolated into Node.js source flagged similarly. Derived from `find_available_ports` output (a port number integer). _False-positive._ _(input-validation)_
- **`lib.sh:763,772`** — `QA_VITE_PORT_LOCK_DIR` derived from `$TMPDIR` with no sanitization flagged. `TMPDIR` is a POSIX-standard environment variable controlled by the OS/user; sanitizing it would add noise with no security benefit in this dev-tooling context. _False-positive._ _(input-validation)_

### Out-of-scope (security)

- **`lib.sh:763`** — `QA_VITE_PORT_LOCK_DIR` defaults to `/tmp` on a shared host (world-writable, sticky bit). A local attacker could pre-create lockfiles to claim all pool slots. Pre-existing environment constraint; not introduced by this PR. _(red-team)_
- **`lib.sh:773`** — `exec 200>"$lockfile"` follows symlinks (no `O_NOFOLLOW`). On a system where `/tmp` is genuinely world-writable and shared, a symlink attack could redirect the lockfile. Same pre-existing environment constraint. _(red-team)_

PR #1365, opened 2026-06-11
budget-etl: mark derived monthly anchors Virtual so they never shadow real statements

natb1:
Closes #1274

## Problem

`deriveMonthlyStatements` generates intermediate monthly balance anchors for an
account whose single statement file has a balance snapshot but transactions
spanning multiple months. Each anchor got `StatementID = inst + "-" + acct +
"-" + period` — identical to the ID a *real* statement of the same
account/period produces. In `mergeStatements`, dir statements (which include
these anchors) overrode input statements by `StatementID`, so if the snapshot
already held a real statement for a month a new long-span file also covered
(e.g. a year-to-date export landing after months of normal imports), the
derived **estimate** silently replaced the real balance. Anchors carried no
marker, so in later runs they were indistinguishable from real statements and
persisted.

## Fix

Mark derived anchors `Virtual` and restructure the merge so they gap-fill only
and never shadow a real statement.

- **Add `Virtual bool` to `budget.StatementData`** (defaults to `false`, so
  every existing literal stays valid).
- **Propagate it** through `buildExportStatements` into the JSON export.
- **Set `Virtual: true`** on the derived anchors in `deriveMonthlyStatements`.
- **Restructure `mergeStatements`** to enforce a three-way priority — real dir
  > real input > derived/virtual anchor:
  - Seed the result with the non-virtual dir statements (recording covered IDs).
  - Keep each real input statement, skipping virtual inputs and those overridden
    by a real dir statement of the same ID (preserves current behavior).
  - Append a derived/virtual anchor only when no real statement (dir or input)
    already covers its period.

The existing input-loop guard already skips virtual input statements, so
prior-snapshot anchors are skipped and recomputed on read-back — existing
`.benc` snapshots self-heal without migration.

## Tests

- `TestMergeStatementsPreservesRealOverDerivedAnchor` — a real input statement
  ($99) wins over a `Virtual` derived anchor ($10 estimate) with the same
  `StatementID`; a derived anchor for a period with no real statement is still
  emitted (gap-fill preserved); exactly one statement per ID.
- `TestDeriveMonthlyStatementsVirtual` — every derived anchor carries
  `Virtual == true`.

## Verification

`go build ./...`, `go test ./...`, and `go vet ./...` all pass in `budget-etl`.

## Out of scope

The frontend already handles virtual statements correctly (no balance/net-worth
path filters virtual; the all-virtual-account badge cannot misfire here, since
the one real statement file remains non-virtual). Existing snapshots self-heal,
so no migration is needed.

### DIFF
--- a/budget-etl/internal/budget/budget.go
+++ b/budget-etl/internal/budget/budget.go
@@ -21,6 +21,10 @@ type StatementData struct {
 	// statement file this data was parsed from; empty for ETL-synthesized
 	// statements (e.g. derived balance anchors).
 	SourceFile string
+	// Virtual marks an ETL-synthesized statement (e.g. a derived monthly balance
+	// anchor) so it can be distinguished from a real parsed statement. Real
+	// statements leave this false.
+	Virtual bool
 }
 
 // StatementDocID generates a deterministic document ID from a statement ID
--- a/budget-etl/main.go
+++ b/budget-etl/main.go
@@ -1193,7 +1193,11 @@ func runMerge(input fileOpts, dir, groupName string, disc parse.DiscoverOpts, ou
 }
 
 // mergeStatements merges dir-parsed statements with input statements.
-// Dir statements override input by statementID; input-only statements are retained.
+// Merge priority by statementID: real dir > real input > derived/virtual anchor.
+// A real dir statement overrides the stale snapshot copy of the same ID. A
+// derived/virtual anchor never shadows a real statement (dir or input); it
+// gap-fills only — emitted solely for a period with no real statement. Virtual
+// input statements (stale prior-snapshot anchors) are skipped and recomputed.
 // Uses maxDates to set LastTransactionDate on all statements (dir and input-only).
 func mergeStatements(dirStmts []budget.StatementData, inputStmts []export.Statement, maxDates map[string]*time.Time) []export.Statement {
 	for i := range dirStmts {
@@ -1202,14 +1206,33 @@ func mergeStatements(dirStmts []budget.StatementData, inputStmts []export.Statem
 	}
 	dirExport := buildExportStatements(dirStmts)
 
-	dirByStmtID := make(map[string]bool, len(dirExport))
+	// realDirByStmtID tracks the IDs of non-virtual (real) dir statements so a
+	// stale input copy of the same ID is dropped in favor of the real dir one.
+	realDirByStmtID := make(map[string]bool, len(dirExport))
 	for _, s := range dirExport {
-		dirByStmtID[s.StatementID] = true
+		if !s.Virtual {
+			realDirByStmtID[s.StatementID] = true
+		}
 	}
 
-	result := dirExport
+	// covered tracks IDs already emitted by a real statement (dir or input), so
+	// a derived anchor only fills periods with no real statement.
+	covered := make(map[string]bool, len(dirExport))
+
+	// Seed result with the real (non-virtual) dir statements.
+	var result []export.Statement
+	for _, s := range dirExport {
+		if s.Virtual {
+			continue
+		}
+		result = append(result, s)
+		covered[s.StatementID] = true
+	}
+
+	// Real input statements: skip virtual ones and those overridden by a real
+	// dir statement; otherwise retain and update LastTransactionDate.
 	for _, s := range inputStmts {
-		if dirByStmtID[s.StatementID] || s.Virtual {
+		if s.Virtual || realDirByStmtID[s.StatementID] {
 			continue
 		}
 		// Update input-only statement's LastTransactionDate from merged transactions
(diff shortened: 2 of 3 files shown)


natb1:
## QA Summary — #1274 (PR #1365)

**Non-browser path** — all changes are in `budget-etl/` (Go backend only).

### Items

| # | Title | Result |
|---|---|---|
| 1 | Build compiles without errors | ✅ PASS |
| 2 | All tests pass | ✅ PASS |
| 3 | Derived anchors marked Virtual | ✅ PASS (covered by TestDeriveMonthlyStatements) |
| 4 | Real statement preserved over derived anchor | ✅ PASS (covered by new mergeStatements test) |
| 5 | Gap-fill preserved (derived anchor emitted when no real statement) | ✅ PASS |
| 6 | `go vet` passes | ✅ PASS |

**6 PASS / 0 FAIL / 0 SKIP**

### Deferred to office-hours
None — no needs-human-judgment items.

### Bugs found
None.

natb1:
## Review findings — #1274 (Virtual derived monthly anchors)

### Fixed

- **`budget-etl/main.go:438` — `runInputJSON` dropped all virtual statements, including derived monthly anchors** (`code-review`): The filter `if s.Virtual { continue }` was intended to discard only Synchrony virtual statements but also silently dropped derived monthly anchors, which cannot be re-derived in the `--input-json` path. Narrowed the filter to `if s.Virtual && strings.HasPrefix(s.StatementID, "synchrony-virtual-") { continue }` so derived anchors survive the round-trip. Fix commit: `84278448`.

- **`budget-etl/main_test.go` — Missing round-trip test for the `runInputJSON` path** (`code-review`): No test covered the scenario where derived anchors produced by `runMerge` are subsequently fed into `runInputJSON`. Added `TestRunInputJSONPreservesDerivedAnchors`, which calls `runMerge` with multi-month transaction data to generate derived anchors, then feeds the output into `runInputJSON` and asserts the anchors survive. Fix commit: `84278448`.

### Informational

- **`budget-etl/main_test.go:1129` — `TestMergeStatementsPreservesRealOverDerivedAnchor` doesn't exercise the `LastTransactionDate` update branch** (`code-review`): The test passes an empty `maxDates` map, so the date-update path at `main.go:1240-1243` is never reached. Not a correctness concern for this PR; `TestRunMerge` exercises the date path via full integration.

- **`budget-etl/main_test.go:513` — `TestRunMerge` doesn't assert the `Virtual` field** (`code-review`): The existing round-trip test checks `StatementID`, `Balance`, `Institution`, and `Period` but not `Virtual`. Low priority; the new dedicated tests cover the `Virtual` invariant directly.

- **`budget-etl/main_test.go:1075` — `TestMergeStatementsPreservesRealOverDerivedAnchor` doesn't cover real-dir-overrides-real-input case** (`review`): The test exercises real-input-beats-virtual-dir but not the existing real-dir-beats-real-input behavior. `TestRunMerge` already covers this path implicitly.

- **`budget-etl/main.go:1195-1259` — Doc comment doesn't mention the self-healing snapshot property** (`review`): The updated comment describes the priority order correctly but omits that existing `.benc` snapshot anchors are skipped on read-back and recomputed. Noted; the PR body covers this, and it would also be reasonable in a follow-up comment update.

### Dismissed

- **`budget-etl/main.go:1255` — Defensive `covered[s.StatementID] = true` after each virtual anchor append** (`review`): Reviewer flagged as possibly redundant. The derived-anchor `StatementID` space is disjoint from real statement IDs within a single dir parse, so no collision can occur in practice. The defensive write is cheap and harmless; no change needed.

- **`budget-etl/main.go:1220` — `covered` map pre-allocated with `len(dirExport)` including virtual entries** (`review`): The capacity hint is a slight over-estimate since only non-virtual dir entries are seeded, but the excess is bounded and correct; Go maps grow automatically. No change needed.

### False-positive (security)

- **`budget-etl/main.go:1234-1235` — Tampering with snapshot file to suppress a real balance** (`red-team`): The `Virtual` flag is parsed from JSON written by a prior ETL run. An attacker who can write the snapshot file can corrupt it arbitrarily regardless of this PR. This is a local-file-access CLI tool with no untrusted-input boundary; not a meaningful threat model.

### Out-of-scope (security)

- **`budget-etl/main.go:1353` and `parse.go:45` — StatementID collision via crafted institution/account names** (`red-team`): The separator `-` in `inst + "-" + acct + "-" + period` allows a crafted institution name to collide with a different account's ID. Pre-existing concern; no change in this PR's logic.

- **`budget-etl/main.go:1213-1215` and `:1419` — Fabricated derived anchor permanently promoted to real statement** (`red-team`): A `Virtual: true` anchor written into the snapshot is skipped on read-back by the existing `s.Virtual` guard, so it cannot be "promoted." The concern is a pre-existing one about snapshot file integrity generally, not introduced by this PR.

PR #1366, opened 2026-06-11
budget-etl: reject category-filter fields on categorization rules

natb1:
Closes #1276

## Problem

A *categorization* rule that sets `matchCategory`, `category`, or
`excludeCategory` can never behave as its author intends, yet nothing rejected
those fields.

`ApplyCategorization` only visits transactions whose `Category == ""` and passes
that empty string to `Rule.Match`. Against an empty category:

- `MatchCategory != ""` → the rule can **never fire**.
- `Category != ""` → the rule can **never fire**.
- `ExcludeCategory != ""` → the exclusion is a **silent no-op**.

These three fields are meaningful only for `budget_assignment` rules, which run
against transactions that already carry a category. A rule author who set them on
a categorization rule got a silently never-matching (or silently inert) rule with
no feedback.

## Change

`convertExportRules` now rejects a `categorization` rule that carries any
category-filter field with a clear, rule-ID-prefixed error:

```
rule <id>: category-filter fields (matchCategory/excludeCategory/category) are only valid on budget_assignment rules, not categorization rules
```

`convertExportRules` is the right gate: it runs only on the general rule set
(after `splitRules` removes transaction-specific rules), so every rule it sees is
either `categorization` or `budget_assignment`. The check is scoped specifically
to `Type == "categorization"`, not a blanket non-`budget_assignment` check, so it
does not over-reject any future rule type.

Runtime semantics (`Rule.Match` / `ApplyCategorization`) are unchanged — the
defect was the missing author-facing validation, not the matching behavior.

## Test

`TestConvertExportRulesRejectsCategoryFilterOnCategorization` is table-driven over
the three fields set individually on a `categorization` rule (each must error) and
on a `budget_assignment` rule (each must convert without error, guarding against
an over-broad check).

### DIFF
--- a/budget-etl/main.go
+++ b/budget-etl/main.go
@@ -513,6 +513,9 @@ func buildRule(r rules.Rule, minAmountDollars, maxAmountDollars *float64) (rules
 func convertExportRules(exportRules []export.Rule) ([]rules.Rule, error) {
 	ruleSet := make([]rules.Rule, len(exportRules))
 	for i, r := range exportRules {
+		if r.Type == "categorization" && (r.MatchCategory != "" || r.ExcludeCategory != "" || r.Category != "") {
+			return nil, fmt.Errorf("rule %s: category-filter fields (matchCategory/excludeCategory/category) are only valid on budget_assignment rules, not categorization rules", r.ID)
+		}
 		built, err := buildRule(rules.Rule{
 			ID:              r.ID,
 			Type:            r.Type,
--- a/budget-etl/main_test.go
+++ b/budget-etl/main_test.go
@@ -152,6 +152,92 @@ func TestConvertExportRulesEmpty(t *testing.T) {
 	}
 }
 
+func TestConvertExportRulesRejectsCategoryFilterOnCategorization(t *testing.T) {
+	tests := []struct {
+		name    string
+		rule    export.Rule
+		wantErr bool
+	}{
+		{
+			name: "categorization with MatchCategory errors",
+			rule: export.Rule{
+				ID:            "r1",
+				Type:          "categorization",
+				Pattern:       "coffee",
+				Target:        "Food:Coffee",
+				MatchCategory: "Food",
+			},
+			wantErr: true,
+		},
+		{
+			name: "categorization with ExcludeCategory errors",
+			rule: export.Rule{
+				ID:              "r2",
+				Type:            "categorization",
+				Pattern:         "coffee",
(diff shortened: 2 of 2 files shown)


natb1:
## QA Pass — #1276 (PR #1366)

**Path:** Non-browser (pure Go backend change — `budget-etl/main.go`, `budget-etl/main_test.go`)

### Items walked: 6 | PASS: 6 | FAIL: 0 | SKIP: 0

| # | Item | Result |
|---|------|--------|
| 1 | Go test suite (`go test ./...`) | ✓ PASS |
| 2 | `categorization` + `MatchCategory` → error | ✓ PASS |
| 3 | `categorization` + `ExcludeCategory` → error | ✓ PASS |
| 4 | `categorization` + `Category` → error | ✓ PASS |
| 5 | `budget_assignment` + category-filter fields → no error | ✓ PASS |
| 6 | Binary build (`go build -buildvcs=false ./...`) | ✓ PASS |

### Bugs found
None.

### Deferred to office-hours

natb1:
## Review

### Fixed

- **R01** `budget-etl/main_test.go:232` — Changed `t.Errorf` to `t.Fatalf` for the primary `wantErr` assertion in `TestConvertExportRulesRejectsCategoryFilterOnCategorization`, consistent with the `t.Fatalf` pattern used in `TestConvertExportRules`. (commit `79a0815a`)

### Informational

- **CR-001+R02** `budget-etl/main_test.go:229-238` — The new test does not assert that the error message contains the rule ID. Adding `strings.Contains(err.Error(), tc.rule.ID)` inside the error branch would pin the rule-ID prefix convention against regressions, but the acceptance criteria pass without it.

### Deferred

- **CR-002** `budget-etl/main.go:547-568` — Transaction-specific rules are split off by `splitRules` before `convertExportRules` runs, so the new category-filter validation never sees them. `MatchCategory`, `ExcludeCategory`, or `Category` set on a transaction-specific categorization rule remain silently inert — the same class of author confusion #1276 fixed for general rules. Filed as #1371.

- **R03** `budget-etl/main.go:520-521, 561` — The string literal `"categorization"` is repeated at three sites; a named constant (`ruleTypeCategorization`) would consolidate them. Pre-existing pattern not introduced by this PR; no issue filed (low priority).

### Security

No security findings. CodeQL: no open alerts on this PR. No `package.json`/`package-lock.json` changes — dependency audit skipped.

PR #1376, opened 2026-06-11
print: guard image-archive navigation against stale renders

### LINKED ISSUE #1279: print: image-archive navigation has no staleness guards (wrong page, stacked spreads, leaked observer)
## Context

`shell.ts` does not serialize navigation — each event independently calls `goNext().catch(...)`. Unlike `pdf.ts` (generation counter, and #1162 for its spread path), `image-archive.ts` has no superseded-render check.

## Defects

- **goToPage race** (`src/viewer/image-archive.ts:211-221`): two rapid `next()` calls race their blob fetches; if page N resolves after N+1, `imgEl.src` shows N while the page label/alt/persisted position say N+1.
- **renderPageInto stacks images** (`src/viewer/image-archive.ts:200-209` with `spread-controller.ts:138-145`): the `appendChild` lands *after* the await, so a render that was superseded by a clear+repopulate adds a second `` to the spread slot.
- **init past destroy** (`src/viewer/image-archive.ts:189-197`): no `if (destroyed) return;` after the first `getObjectUrl` await, so teardown during init still appends an `` and creates/observes a leaked `ResizeObserver`.

## Acceptance criteria

- [ ] A generation/staleness token (matching `pdf.ts`) guards `goToPage`, `renderPageInto`, and `init`.
- [ ] Rapid navigation never displays a stale page or stacks images; teardown during init leaks nothing.

natb1:
Closes #1279

## Summary

The image-archive (zip-of-images) renderer in the print viewer had no superseded-render check, unlike the sibling `pdf.ts` renderer which serializes overlapping renders with a `renderGen` generation counter and appends its page wrapper synchronously before any await. Navigation events are not serialized — `shell.ts` wires each tap/click/key to `goNext().catch(...)` / `goPrev().catch(...)` without awaiting — so rapid input produced overlapping async renders that raced their blob fetches.

This PR ports the `pdf.ts` staleness discipline to the image-archive path and to the shared `SpreadController.render()`.

## Defects fixed

- **goToPage race** — two rapid `next()` calls each reached `goToPage`, which set `_currentPage`/`alt` synchronously but assigned `imgEl.src` after `await getObjectUrl(...)`. If page N's blob resolved after N+1's, `imgEl.src` ended on N while `_currentPage`/`alt`/persisted position said N+1.
- **renderPageInto stacked images** — `renderPageInto` did `target.appendChild(img)` *after* its await, so a render superseded by a `render()` clear+repopulate appended a second `` into the slot. `render()` also had no generation guard, so a superseded `render()` still proceeded to its second (`right`) call, stacking the right slot.
- **init past destroy** — a `if (destroyed)` check existed *before* the first `getObjectUrl` await but not after it. Teardown during init still ran `container.appendChild(imgEl)` and created/observed a `ResizeObserver` — a leaked observer and a stray ``.

## Changes

**`print/src/viewer/image-archive.ts`** (Unit 1):

- Added a `renderGen` closure counter alongside `destroyed`.
- `goToPage` claims `const gen = ++renderGen;` synchronously after the range check; the post-await guard became `if (destroyed || gen !== renderGen) return;`, so a superseded blob fetch never assigns a stale `imgEl.src` or runs `prefetchNextPage`.
- `renderPageInto` now creates, sets `alt`, and appends the `` *before* the `await getObjectUrl(...)`. After the await it guards with `if (!target.contains(img)) return;` (supersession — `render()`'s `innerHTML=""` detached it) and removes the img on teardown before assigning src.
- `init` captures the url then guards `if (destroyed) return;` before `imgEl.src`, `appendChild`, prefetch, and the `ResizeObserver` block.

**`print/src/viewer/spread-controller.ts`** (Unit 2):

- Added `private renderGen = 0;`. `render()` claims `const gen = ++this.renderGen;` synchronously and bails with `if (gen !== this.renderGen) return;` after each `renderPageInto`, stopping a superseded render before its second (`right`) call. This is renderer-agnostic, so it also hardens the pdf spread path that flows through the same `render()`.

## Tests

Added staleness tests mirroring the never-resolving-blob pattern in `search.test.ts`:

- `image-archive.test.ts`: goToPage race (stale fetch resolving last never overwrites the fresh src/page/alt), renderPageInto no-double-append (superseded render leaves exactly one ``), init teardown during blob fetch (no img appended, no `ResizeObserver` observed).
- `spread-controller.test.ts`: concurrent render no-stack (each slot holds exactly one child, the stale render's guard bails before its right call, position reflects the latest spread index).

## Verification

- `npx vitest run --root print` — 354 passed, 1 skipped.
- `npx tsc --noEmit --project print` — clean.
- `npm run build --prefix print` — succeeds.

## Acceptance criteria

- [x] A generation/staleness token (matching `pdf.ts`) guards `goToPage`, `renderPageInto`, and `init`.
- [x] Rapid navigation never displays a stale page or stacks images; teardown during init leaks nothing.


natb1:
## QA Summary — #1279

**Items walked:** 3/5 (machine-verifiable pre-browser checks)  
**PASS:** 3 | **FAIL:** 0 | **SKIP:** 0  
**Deferred to office-hours:** 2

---

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests (355/355) | ✅ PASS |
| 2 | Type-check (`npx tsc --noEmit --project print`) | ✅ PASS |
| 3 | Build (`npm run build --prefix print`) | ✅ PASS |
| 4 | PDF viewer spread renders without stacking | ⏸ Deferred — Chrome extension permission denied for `localhost:5171` |
| 5 | PDF next/prev page navigation | ⏸ Deferred — Chrome extension permission denied for `localhost:5171` |

---

### Deferred to office-hours

**Item 4 — PDF spread renders correctly**: Chrome extension navigation to `localhost:5171` was denied (two attempts). The extension needs the origin permitted in Chrome before the browser walkthrough can run. This exercises the `spread-controller.ts` render-generation guard added by this PR.

**Item 5 — PDF next/prev navigation**: Same blocker as item 4.

**Item 6 — Image-archive rapid navigation visual check**: `test-image-archive` seed is `testOnly: true`; not seeded in the public QA server. Covered by new unit tests (goToPage-race, renderPageInto-no-stack, init-teardown, spread-concurrent-render — all passing).

---

### Pre-QA acceptance tests

Playwright could not run locally (known env limitation: Nix provides chromium 1217, project needs 1208). CI runs acceptance tests in the `acceptance` job.

---

**Blocker:** Chrome extension permission for `localhost:5171` needs to be granted in the Windows Chrome browser before the browser walkthrough items can complete.

natb1:
## QA Summary — #1279

**Items walked:** 5/5  
**PASS:** 5 | **FAIL:** 0 | **SKIP:** 0  
**Deferred to office-hours:** 0

---

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests (355/355) | ✅ PASS |
| 2 | Type-check (`npx tsc --noEmit --project print`) | ✅ PASS |
| 3 | Build (`npm run build --prefix print`) | ✅ PASS |
| 4 | PDF spread renders without stacking (Republic, spread mode) | ✅ PASS |
| 5 | PDF next/prev navigation (Page 1/3 → Pages 2-3/3 → Page 1/3) | ✅ PASS |

---

### Item 4 detail — spread renders without stacking

Opened Republic in spread view (`aria-pressed: true`). DOM check: 2 `pdf-page-wrapper` elements, each with exactly 1 canvas — no stacking on initial load or after toggle.

### Item 5 detail — navigation

- Next: "Page 1 / 3" → "Pages 2-3 / 3" — counter updated, Republic Book I page 2 content rendered, 2 wrappers × 1 canvas each throughout
- Prev: "Pages 2-3 / 3" → "Page 1 / 3" — correct return to first spread, no stacking, no console errors

### Image-archive note

`test-image-archive` seed is `testOnly: true` — not seeded in the public QA server. Image-archive race condition fixes (goToPage, renderPageInto, init) are covered by the new unit tests (355/355 passing), which include the goToPage-race, renderPageInto-no-stack, init-teardown, and spread-concurrent-render tests.

---

**GIF:** `qa-fix-walkthrough-1279.gif` (downloaded to browser)

natb1:
## Review summary

### Fixed

- **print/src/viewer/image-archive.ts:213–214** — `renderPageInto` now synchronously appends an `` before the blob fetch; if `getObjectUrl` rejects the element was left as an src-less orphan. Wrapped the post-await block in `try/catch` to call `img.remove()` on failure before rethrowing, restoring the clean-on-error invariant. (commit adb08b18)
- **print/test/viewer/image-archive.test.ts:942–964** — Renamed the destroyed-during-blob-fetch test from "returns without appending if destroyed during blob fetch" to "removes the synchronously-appended img if destroyed during blob fetch", added an explanatory comment and a pre-await assertion confirming the img is briefly appended then cleaned up. (commit adb08b18)

### Informational

- **print/src/viewer/image-archive.ts:219** — `target.contains(img)` returns `true` for containment within a _detached_ subtree (a node appended to a detached root is still "contained" in it). In the current flow `innerHTML = ""` detaches the img's parent slot from the DOM, making `target.contains(img)` reliably `false` as expected — but this is a subtle containment contract worth noting for future readers.

### False-positive (security)

- **print/src/viewer/image-archive.ts:214** — Synchronous `` append before blob fetch raised as a potential flicker/accessibility issue. The img has no `src` at append time but carries the correct `alt` and is never visible until `img.src` is set; dismissed as not a real concern.
- **print/src/viewer/image-archive.ts:219–221** — Guard order (`target.contains(img)` before `destroyed`) flagged. Both guards are equivalent here: `destroyed` implies the slot was cleared, so `target.contains(img)` is already false. Order does not affect correctness.
- **print/src/viewer/image-archive.ts:100–112** — `getObjectUrl` returning `""` when destroyed at `blob().then()` time. The empty-string url is assigned to a src-less img that is then removed by the `destroyed` guard; the cached empty-string urlPromise is unreachable after `destroy()` clears state. Not a real concern.
- **print/src/viewer/spread-controller.ts:139** — `renderGen` never reset (monotonically increasing). The guard only tests `gen !== this.renderGen` — a stale `gen` is always less than the current value, so the check is correct regardless of magnitude.

### Deferred

- **print/src/viewer/spread-controller.ts:132–155** — `leave()` does not bump `renderGen`, so an in-flight `render()` interrupted by `leave()` passes the gen check and calls `renderPageInto(right, rightEl)` on the already-detached `rightEl`. Blob decoding still runs unnecessarily. Pre-existing gap, not worsened by this diff. Tracked as follow-up #1383 (blocked by #1279).

PR #1397, opened 2026-06-11
print: flush pending position save on teardown; folder-scope local reading-position keys

natb1:
Closes #1283

Fixes two independent reading-position defects in the print viewer.

## Unit 1 — Flush pending position save on viewer teardown

The viewer cleanup closure cancelled the 500ms-debounced save timer with
`clearTimeout` but never flushed it, so a page turn within the last 500ms
before "Back to Library" (or any unmount) was lost.

- Extracted the debounced save body into a `persistPosition()` helper (read
  `getSpreadPosition()`, dedup against `lastSavedPosition`, write via
  `saveReadingPosition` authed / `saveLocalPosition` unauthed).
- `scheduleSave()` keeps its debounce; its timer callback now just nulls the
  timer and calls `persistPosition()`.
- New `flushSave()` cancels any pending timer and calls `persistPosition()`
  synchronously. Teardown now calls `flushSave()` first (before
  renderer/controller teardown can disturb `getSpreadPosition()`) instead of
  discarding the timer.
- Inverted the existing "cleanup cancels pending save timer" test to assert
  cleanup **flushes** the pending position.

## Unit 2 — Folder-scope the local-folder item id

A local-folder item's identity was filename-only (`id = LOCAL_ID_PREFIX +
name`), so `Notes.pdf` in folder A and `Notes.pdf` in folder B shared one
`reading-position:` key and one `spread-mode:` key.

- `fileToLocalItem` takes a required `folderId` and mints
  `id = LOCAL_ID_PREFIX + folderId + "/" + name`.
- `createLocalSource` derives `folderId = directory.name` and binds it via
  closure; the `mediautil` `toItem` signature `(file, name)` is unchanged, so
  `mediautil/src/local-folder.ts` is untouched.
- All id-derived keys (URL, `reading-position:`, `spread-mode:`) become
  folder-scoped uniformly with no `shell.ts` change. A folder-scoped id like
  `local:books/Notes.pdf` encodes its `/` to `%2F`, staying a single URL path
  segment that the router round-trips.
- Folder identity uses `directory.name` — intrinsic, survives browser restarts
  and re-picking the same folder (a minted nonce would not). Residual
  acceptable limitation: two *distinct* folders that share the same name still
  collide; the acceptance criterion (same-named files in *different* folders)
  is met because differently-named folders no longer collide.

One-time effect: existing local reading positions saved under the old
filename-only keys orphan once (new keys are folder-scoped) — an acceptable
low-stakes reset for local items; no migration.

## Verification

- `npx tsc --noEmit --project print` — clean.
- `npx vitest run --root print` — 356 passed, 1 skipped.
- `npx vitest run --root mediautil` — 16 passed (toItem signature unchanged).
- `npm run build --prefix print` — production build succeeds.


natb1:
# Fix-checks summary — PR #1397

## Iteration 1

- **Failed checks:** `acceptance` (Playwright e2e).
- **Outcome:** fixed
- **Reproduced:** no (Playwright cannot run locally — Nix chromium 1217 vs project 1208 mismatch; CI is authoritative). Root cause is deterministic and confirmed directly from the CI failure log.
- **Reproduce command:** `.claude/skills/dispatch-propagate/scripts/run-acceptance-tests.sh print` (CI-only).
- **Failure excerpt:** `print/e2e/viewer.spec.ts:488` — "image archive loads from cache on second view @cache @testonly": `Expected substring: "1 / 5"` / `Received string: "Page 2 / 5"` (all three projects: mobile, tablet, desktop).
- **Why not caught:** This PR's Unit 1 change (viewer teardown now flushes the pending debounced reading-position save instead of discarding it) changes a behavior the test incidentally relied on. The test turns to page 2, then clicks "Back to Library" within the 500ms debounce window. Before Unit 1 the pending save was discarded, so the second visit reset to page 1 and the test asserted "1 / 5". After Unit 1 the save is flushed, so the second visit correctly restores to page 2 ("Page 2 / 5"). The test was written against the pre-fix behavior; CI ran the new acceptance suite for the first time on this PR and surfaced the now-stale assertion.
- **Fix:** Updated the second-visit block of the cache test (`print/e2e/viewer.spec.ts`) to expect the restored page-2 position (which itself proves the lazy-loaded page 2 is served from cache while Firebase Storage is blocked), then navigate back to page 1 to verify page 1 is cached too. Commit `06656810`.

natb1:
## QA Summary — PR #1397 (Issue #1283)

**Items walked: 4** | **PASS: 4** | **FAIL: 0** | **SKIP: 0** | **Deferred: 1**

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Type-check (`npx tsc --noEmit --project print`) | ✅ PASS |
| 2 | Unit tests (`npx vitest run --root print`) — 358/358 | ✅ PASS |
| 3 | Production build (`npm run build --prefix print`) | ✅ PASS |
| 4 | Viewer teardown flush — turn page, navigate away before debounce fires, verify position saved | ✅ PASS |
| 5 | Folder-scoped local reading-position keys | ⏭ Deferred to office-hours |

### Teardown flush verification (Item 4)

Tested via Chrome extension against QA server (`http://localhost:5172`):

1. Navigated to `/view/plato-republic` (Page 1 / 3) with cleared localStorage.
2. Clicked next-page; a `MutationObserver` fired the moment the position label updated to "Page 2 / 3" (i.e., immediately after `scheduleSave()` set the 500ms timer).
3. Clicked "Back to Library" within milliseconds — well before the 500ms debounce could fire.
4. `cleanupView()` → `flushSave()` ran synchronously; `localStorage["reading-position:plato-republic"]` = `"2"`.
5. Re-opened the viewer — restored to Page 2 / 3. ✅

No console errors.

### Deferred to office-hours

**Item 5 — Folder-scoped local reading-position keys**: Verifying this end-to-end requires opening two local folders (each containing a same-named PDF) and inspecting `localStorage` keys. The `FileSystemDirectoryHandle` picker requires a user gesture that cannot be automated.

Unit tests cover the key logic fully:
- `fileToLocalItem(file, "Notes.pdf", "folderA")` → `id = "local:folderA/Notes.pdf"`
- `fileToLocalItem(file, "Notes.pdf", "folderB")` → `id = "local:folderB/Notes.pdf"`
- Distinct ids → distinct `localStorage` keys `reading-position:local:<folder>/Notes.pdf`

**Manual walkthrough for office-hours**: open two differently-named folders each containing a same-named PDF, set distinct reading positions, confirm `localStorage` shows distinct `reading-position:local:<folderA>/…` and `reading-position:local:<folderB>/…` keys.

### Pre-QA acceptance check note

Playwright headless Chromium binary (`chromium_headless_shell-1217`) cannot execute in this WSL/Nix environment. This is a known persistent environment constraint — CI is authoritative for acceptance tests. The `@testonly`-tagged acceptance test for teardown flush will run in CI.

### Bugs found

None.

natb1:
## Review findings — PR #1397 (#1283)

Fix commit: f6f7d564

### Fixed

- **rv-1** `print/src/viewer/shell.ts:194-200` — `flushSave()` called `persistPosition()` unconditionally. Before `renderer.init` resolves, `_currentPage` is 0 so `getSpreadPosition()` returns `'0'`; `lastSavedPosition` has already been set to the loaded page (e.g. `'5'`). An unconditional flush would write `'0' != '5'` and overwrite the real saved position when the user navigates away early. Moved `persistPosition()` inside the `if (saveTimer)` guard — flushSave now only persists when there is an actual pending debounced save, which can only be scheduled after `updateNav()` fires post-init. _(commit f6f7d564)_

### Informational

- **rv-3** `print/src/library.ts:103` — Same-named folders still collide on `directory.name`. Accepted residual limitation documented in the issue. A stored folder nonce in IndexedDB would be needed for full disambiguation; out of scope for this PR.
- **rv-4** `print/e2e/viewer.spec.ts:486-497` — e2e test updated correctly to assert page-2 position is restored on second visit after teardown flush. No change needed.
- **rv-5** `print/src/pages/home.ts` — Local item links use `encodeURIComponent` correctly; cloud item ids never contain slashes. URL round-trip is safe.

### Dismissed

- **cr-2** `print/src/viewer/shell.ts:194-200` — Suggestion to document the unconditional `persistPosition()` call with a comment. Superseded by the rv-1 fix making the call conditional; no comment needed.

### Deferred

- **rv-2+cr-1** `print/test/viewer/shell.test.ts:328-345` — The `cleanup flushes pending save timer` test only covers the authenticated path (`uid='uid1'`, checks `saveReadingPosition`). The unauthenticated path (`uid=null`, writes to `localStorage` via `saveLocalPosition`) has no corresponding test. Filed as #1410 (blocked_by #1283).

### Security

No open CodeQL alerts. No dependency changes. No security findings.

PR #1399, opened 2026-06-11
blog: client OG update preserves prerendered title/og:url for posts without previewDescription

natb1:
Closes #1286

## Summary

The client `updateOgMeta` (`blog/src/og-meta.ts`) gated its non-post branch on
`if (!post?.previewDescription)`, treating a post that merely *lacks*
`previewDescription` as "not a post page". Because `afterRender` runs
`updateOgMeta` on the initial load of a prerendered post page, such a post had
its correct prerendered `<title>` and `og:url` actively overwritten with site
defaults (or stripped) on hydration. The bug is latent today — every seeded post
has a `previewDescription` — but is a correctness trap.

This change makes the client path mirror `postOgEntries` and the prerender path:
a post page always emits title/`og:title`/`og:url`/`og:type`/`twitter:*`, and
only the description/image entries stay conditional on
`previewDescription`/`previewImage`.

## Changes

- **`blog/src/og-meta.ts`** — gate the non-post branch on `!post` so it fires
  only for genuine non-post navigation (home / SPA). A `previewDescription`-less
  post now falls through to the post branch. Added a symmetric
  stale-description cleanup (mirroring the existing `previewImage` cleanup) that
  removes `og:description`/`description`/`twitter:description` when a post lacks
  `previewDescription`, so tags from a previously-viewed post-with-description do
  not linger. Reworded the now-stale comment above `postOgEntries`.
- **`blog/test/og-meta.test.ts`** — rewrote the test that asserted the defective
  "removes all OG tags when post has no previewDescription" behavior into one
  asserting the title/`og:url`/`og:type`/`twitter:title` are preserved while
  description tags stay null. Added a stale-description-cleanup test symmetric
  with the existing image-cleanup test.

## Verification

- `npx vitest run --root blog og-meta` — 33/33 pass.
- `npx tsc --noEmit --project blog` — net unchanged (the blog `tsconfig`'s `lib`
  omits `dom`, so all DOM usage flags project-wide; the four new lines follow the
  identical pre-existing pattern already present throughout `og-meta.ts`). The
  production build runs through Vite, not this `tsc` invocation.


natb1:
## QA pass — #1286 client OG update preserves prerendered title/og:url

Non-browser path (changes are in `blog/src/og-meta.ts` and `blog/test/og-meta.test.ts` only).

### Items walked: 4 | PASS: 4 | FAIL: 0 | SKIP: 0

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests pass for rewritten + new OG-meta tests | ✅ PASS — 33/33 tests |
| 2 | No regression in full blog unit suite | ✅ PASS — 353/353 tests |
| 3 | Type-check | ✅ PASS (pre-existing noise, no new categories) |
| 4 | Non-post gate is `!post`; stale-description cleanup present | ✅ PASS |

### Notes

**Item 3 detail:** `npx tsc --noEmit --project blog` produces 144 errors (vs 140 on the baseline) — the +4 are all TS2584 "Cannot find name 'document'" proportional to the 3 new `document.querySelector` calls in the stale-description cleanup. Same pre-existing misconfiguration (missing `lib: dom` in `blog/tsconfig.json`). No new error category.

**Item 4 detail:** `updateOgMeta` line 135 gate confirmed as `if (!post)`. Stale-description cleanup (lines 148–152) removes `og:description`, `description`, `twitter:description` when `previewDescription` is absent — symmetric with the existing image cleanup. Comment updated to accurately reflect that `og:title`/`og:url`/`og:type`/`twitter:card`/`twitter:title` apply unconditionally on both paths.

### Deferred to office-hours

None.

### Bugs found

None.

natb1:
## Review

No code changes applied. 3 findings from code-review and review reviewers.

### Fixed

_None._

### Required (security)

_None._

### Informational

- **`blog/test/og-meta.test.ts:96`** (code-review): The test 'preserves post title/og:url/og:type when post has no previewDescription' asserts that `og:description` and `description` are null, but omits an assertion that `twitter:description` is also null. The source at `og-meta.ts:151` removes `twitter:description` in the same cleanup block, so the behavior is correct — but the test does not verify this third tag. Suggestion: add `expect(getNameContent("twitter:description")).toBeNull();` after the `description` assertion at line 96.

- **`blog/test/og-meta.test.ts:88`** (review): Same gap from a different angle — the rewritten test does not assert `twitter:description` is null when `previewDescription` is absent, leaving a blind spot: a regression that re-added `twitter:description` in the no-description case would not be caught by this test alone.

### Dismissed

- **`blog/test/og-meta.test.ts:89`** (code-review): The test uses `{ ...basePost, previewDescription: undefined }`, which still carries `previewImage`, but does not assert that `og:image` and `twitter:image` remain present. _Rationale: the acceptance criteria focus on title/og:url not being clobbered; image-presence assertions are out of scope for this issue, and the existing image-cleanup test covers the `!previewImage` branch._

### False-positive (security)

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

Issue #1391, opened 2026-06-11
dispatch: introduce dispatch-launch-worker and detach it from the router instead of spawning a model worker


natb1:
## Context

With the plan-phase relevance review moved into `/plan-issue` (sub-issue A), every
`dispatch-route` directive is either mechanical or a single-skill invocation — so
phase initiation no longer needs a model session. This issue introduces the
launcher script and switches the router to detach it instead of spawning
`/dispatch-worker`.

Parent epic explains why a script suffices: phase determination is a pure function
of GitHub state, provisioning is already a script, and parallelism + per-target
failure isolation come from process boundaries, not model sessions.

## Scope

**New `dispatch-launch-worker <N> <wt>`** — a per-target script, launched detached
(`setsid nohup … &`) so it owns an independent lifetime:

1. `dispatch-provision-worktree "$wt"` — the `origin/main` merge + `direnv`/`npm`.
2. Route on the merged tree (`dispatch-route` / `dispatch-phase`).
3. Handle every mechanical disposition with no model:
   - `done` / `waiting` / `wrong-worktree` → release reservation + exit;
   - `provision-failed` → write the office-hours reason + park + exit;
   - `push-stranded` → `git push origin HEAD` + `dispatch-mark-deviation` + park;
   - `conflict` → route to `fix-conflicts`.
4. For a real phase, exec `claude --bg "/<phase-skill> <N> <wt>" --cwd "$wt"`.

**Router wiring** — `dispatch-spawn-worker` / `dispatch-materialize-spawn` detach
the launcher instead of spawning `claude --bg "/dispatch-worker"`. The reservation
marker is still written **pre-detach, under the lock** (unchanged ordering), so a
launcher that dies before spawning leaves a sweepable orphan. The lock is released
after the launcher is detached, without waiting for provisioning.

The merged-tree guarantee (#1047) holds: the launcher provisions before exec'ing
the phase skill, so the model session is born in the merged worktree.

## Acceptance criteria

- [ ] `dispatch-launch-worker` exists, is executable, and runs provision → route →
      mechanical-disposition → spawn-phase-skill in one detached process.
- [ ] Mechanical dispositions (`done`, `waiting`, `wrong-worktree`,
      `provision-failed`, `push-stranded`) are handled in the launcher with no
      `claude` session spawned.
- [ ] A real-phase disposition execs `claude --bg "/<phase-skill> <N> <wt>"` with
      cwd set to the worktree; the phase skill is born in the provisioned, merged
[...]

PR #1402, opened 2026-06-12
dispatch: single-pass --top selection collapses the fan-out's per-slot selector calls

natb1:
Closes #1317

## Problem

The autonomous dispatch chain targets up to `max_concurrent_workers` (default 8)
live workers, but live concurrency sat at ~2–3 and troughed to 0–1 even when the
weekly pace gate was wide open and the queue held 100+ startable items. The
limiter was **spawn throughput**, not pacing or depth.

`dispatch-materialize-spawn`'s fan-out loop filled `GAP` slots by calling
`dispatch-select-target` **once per slot, serialized, holding the selection lock
for the whole run**. Each selector call is ~131s (`gh pr list`, `gh issue list
--limit 300`, a batched `blockedBy` GraphQL query, per-issue leaf traces).
Filling `gap=5` took ~9 minutes, during which early-spawned workers drained and
self-closed, so net concurrency never accumulated. The held lock also dropped the
per-worker Stop-hook ticks that should compound concurrency.

## Fix

`dispatch-select-target` already scanned the whole queue and built ranked
per-`(priority, topic, phase)` bucket maps, then discarded everything except the
single top item. This PR has it emit the **top-N distinct targets in one pass**
via a new `--top N` flag, and has the fan-out loop consume that single list
instead of re-invoking the selector once per slot. A `gap=N` fan-out now performs
**O(1)** selector scans, not O(N) — one ~131s scan instead of N — letting
concurrency accumulate toward target.

### Unit 1 — `dispatch-select-target --top N`

The bucket maps (`FIRST_PR` / `FIRST_ISSUE`) became capped, oldest-first newline
lists. The PR scan appends up to `N` per bucket; the issue scan descends each
help-wanted root past not-ready/excluded leaves (a growing local exclude set
within the one scan) to enumerate multiple startable leaves. Emission branches on
`TOP`: `--top 1` (the default) keeps the existing echo-first-and-exit loop
**verbatim** — byte-identical to the prior no-flag behavior for every existing
caller — while `--top N>1` walks the same priority-axis → topic-category → phase
nesting and drains each bucket oldest-first into a results list. `--top` is
default-mode only: rejected (exit 1) alongside `--qa` / `--health-only` /
`--main-broken-sha` / `--priority-only`. Every existing skip filter
(worktree-claim, `dispatch-ci-ready`, `done`, office-hours, `--exclude`,
open-blocker) is preserved unchanged.

### Unit 2 — fan-out consumes one `--top` scan

The fan-out driver processes the seed `N` first, then issues exactly one
`dispatch-select-target --top (GAP - spawned) --exclude <seed>` scan and consumes
its newline-delimited target list. The single static seed `--exclude` plus the
seed's reservation marker close the seed re-selection race; the dropped
intra-loop growing `--exclude` thread is subsumed by the selector's internal
one-scan dedup, with a defensive `SEEN` dedup on consumption as the backstop. The
terminal tokens (`propagate` / `drain`), the `spawned K of gap G [(stop_reason)]`
summary, and the single end-of-run lock release are preserved verbatim.

### Unit 3 — tests

New `--top` selector coverage (N distinct ranked lines in priority order,
`--exclude` honored, fewer-than-N with no padding, multi-leaf descent in one scan
with live/reserved/office-hours leaves excluded, `--top 1` byte-parity, and the
four rejection combinations) and the fan-out tests rewritten to the single-scan
model: the fake selector emits a multi-line `--top` list in one call, and the
invocation assertion flips from `>=2` to **exactly 1 selector scan per fan-out
run** (the acceptance single-scan assertion). The #1062 dedup-mechanism cases
relocated to the selector `--top` tests where the dedup now lives.

A `drain_bucket` bug surfaced by the multi-target tests is also fixed: its
trailing arithmetic test returned 1 when the results array was still under `TOP`,
which under `set -e` aborted the whole scan at the bare call site — making
`--top N>1` non-functional. An explicit `return 0` keeps a not-yet-full scan
going.

## Coordination with #1389

This issue owns **selection only**. It does not touch the spawn primitive
(`claude --bg "/dispatch-worker"` stays — #1391 swaps it for a detached launcher
later) and keeps the reservation markers (they guard the cross-tick race, which
single-pass selection does not address). Lands before #1391.

## Verification

- Full dispatch suite: **2202/2202 pass**.
- `bash -n` clean on both edited scripts.
- `--top 1` output byte-identical to the no-flag call (tested).
- `--top` rejection (exit 1, order-independent) and bad-value validation confirmed.
- Single-scan acceptance assertion: a `--gap N` fan-out issues exactly one
  selector scan (`wc -l logs/select-target.log == 1`).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-materialize-spawn
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-materialize-spawn
@@ -18,14 +18,16 @@
 # supporting detail (a path, a blocker list, the CI line) precedes the token.
 #
 # With `--gap <N>` and `queue` mode, the script FANS OUT: it spawns up to N
-# workers in one held-lock run, selecting a distinct next target each iteration
-# via dispatch-select-target. The first target is the passed-in <issue-N>;
-# targets 2..N come from successive dispatch-select-target calls. The worker no
-# longer registers synchronously (#1048), so the loop threads its SEEN set as
-# `--exclude` to dispatch-select-target AND the just-written reservation marker
-# makes dispatch-select-target's reserved-skip (#1046) skip the reserved
-# worktree — so the next selection yields a distinct target even though the prior
-# worker has not yet registered.
+# workers in one held-lock run. The seed (target 1) is the passed-in <issue-N>;
+# targets 2..GAP come from ONE `dispatch-select-target --top <GAP-after-seed>`
+# scan (#1317) — a single queue pass that returns up to that many distinct
+# ranked targets, instead of one serialized selector call per slot. The seed's
+# just-written reservation marker (dispatch-select-target's reserved-skip, #1046)
+# plus a single static `--exclude <issue-N>` on that one scan close the seed
+# re-selection race; selection dedups the remaining targets internally within
+# that one scan, so no per-slot SEEN threading is needed. The worker no longer
+# registers synchronously (#1048); the reservation marker, not registration,
+# closes the spawn-gap race for each spawned slot.
 # `explicit` mode and `--gap 1` process a single target (backward-compatible).
 # Fan-out (`--gap N` with N>1) is autonomous-only; a manual run arrives as
 # `--gap 1` and takes the single-target path.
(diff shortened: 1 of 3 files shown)


natb1:
## QA Summary — #1317 dispatch: single-pass `--top` selection

**PR #1402** · non-browser path (scripts only)

### Items walked: 4 · PASS: 4 · FAIL: 0 · SKIP: 0

| # | Item | Result |
|---|------|--------|
| 1 | Syntax check both edited scripts (`bash -n`) | PASS |
| 2 | `--top` rejected with `--qa`, `--priority-only`, `--health-only`, `--main-broken-sha` | PASS |
| 3 | `--top 1` byte-identical to no-flag call (via test suite assertion) | PASS |
| 4 | Full test suite (`test-dispatch-scripts.sh`) | **PASS — 2202/2202** |

### Bugs found
None.

### Deferred to office-hours

natb1:
## Review: dispatch fan-out -- single-pass `--top` selection (#1317)

### Fixed

**CR-001 — Dead variable `EXCLUDE_ARGS` removed** (`dispatch-select-target:204-209`)
`EXCLUDE_ARGS` was built at lines 208-209 with a comment describing it as a leaf-trace passthrough, but was never referenced after construction. The multi-leaf descent loop uses `root_excl_args` (rebuilt per iteration), making `EXCLUDE_ARGS` dead code introduced by the refactor. Removed the variable and its misleading comment block.
Fix commit: 9c894767

---

### Deferred

**CR-002 — `root_recorded` increments on `bucket_append` no-op, can short-circuit mixed-phase root descent** (`dispatch-select-target:892-894`)
`bucket_append` silently drops an entry when the phase bucket is already at TOP capacity. `root_recorded` is incremented unconditionally regardless of whether the leaf was stored or dropped. For TOP=1 this is harmless. For TOP>1, a root with leaves in both implement and plan phases can have its plan-phase descent short-circuited if the plan bucket was pre-filled by a prior root. Edge case not covered by current tests; does not affect TOP=1 behavior.
Filed as follow-up: #1403

---

### Out-of-scope

**CR-003 — `|| true` masks selector hard-failures as queue-exhausted** (`dispatch-materialize-spawn:503`)
The `|| true` on the selector call suppresses any non-zero exit, so a GitHub API credential error or GraphQL failure leaves `SELECTED` empty and the fan-out silently reports queue-exhausted. Byte-identical to the pre-existing per-slot call pattern — not a regression introduced by this diff.

---

### Dismissed

**CR-004 — `bucket_count` called twice per `bucket_append`, minor subshell overhead** (`dispatch-select-target:674`)
`bucket_count` spawns a subshell each call; it is called inside `bucket_append` and again at the loop's coarse-skip and inner break conditions. For TOP≤8 and typical issue counts the overhead is negligible. No correctness issue.

---

### Informational

**RV-002** (`test-dispatch-scripts.sh:18566`) — Fake test selector emits a trailing `empty` after partial results; the real selector does not. Consumer handles both paths correctly, so acceptance criteria are unaffected. The divergence makes the selector/consumer contract boundary slightly harder to reason about.

**RV-003** (`dispatch-materialize-spawn:503`) — Header comment says "targets 2..GAP" but the `--top` argument is `GAP - spawned`, which equals `GAP` when the seed does not spawn. Intentional: compensates for a non-spawning seed so a single pass can still fill the full gap. Comment is slightly misleading in that case.

**RV-004** (`dispatch-materialize-spawn:498-533`) — Skipped targets within the single scan are not replaced; per-run fill may be under-target when selected targets skip. Acknowledged design trade-off documented in the issue and code comments.

**RV-005** (`dispatch-select-target:972-978`) — `drain_bucket`'s explicit `return 0` correctly prevents `set -e` from aborting when the results array is under-full at loop exit. Implementation is correct; comment explains the mechanism clearly.

**RV-006** (`test-dispatch-scripts.sh:19262-19271`) — All acceptance criteria satisfied. O(1) scan assertion validated directly by `wc -l logs/select-target.log == 1`. `--top` tests cover ordering, exclusion, fewer-than-N, live-session skip, reservation-marker skip, office-hours-parked skip, `--top 1` parity, and mode-rejection. Fan-out tests cover list consumption, PR branch-prefix keying, defensive SEEN dedup, empty frontier, deferred-selection stop, and unexpected-line hard error.

---

### False-positive (security)

**SEC-002** — `cur_n` extracted from selector output without integer validation. Selector outputs are derived exclusively from GitHub API integers; a non-numeric value passed to `gh issue view` would fail with non-zero exit, triggering `release_lock; exit 2` — not silent data corruption. No exploitation path.

**SEC-003** — Unfiltered `$line` echoed to stderr on unexpected selector output. Impact limited to log pollution; no `eval` or downstream parsing; script exits immediately with code 2. Same pattern as pre-existing code.

PR #1411, opened 2026-06-12
blog: drop unparseable third-party feed dates at the parse boundary

natb1:
Closes #1288

## Problem

A single third-party blog feed publishing an unparseable `<pubDate>` /
`<published>` date degraded the **entire** blogroll, not just that one entry.

`formatUtcDate` (`blog/src/date.ts`) throws on any string `new Date(...)` cannot
parse. Third-party feed dates reached it **unvalidated** through two independent
feed-parse boundaries:

- **Runtime (browser):** `parseAtomFeed` / `parseRssFeed`
  (`blog/src/blog-roll/parse-feed.ts`) copied the feed date straight into
  `LatestPost.publishedAt`, which was then passed to `formatUtcDate` inside the
  per-entry hydration loop in `info-panel.ts`. One bad date threw mid-loop,
  skipping every remaining entry and the `sortBlogrollByDate` call.
- **Build-time (Node):** `parseAtomFeedXml` / `parseRssFeedXml`
  (`blog/src/blog-roll/vite-plugin-feed-fetch.ts`) build the `buildTimeFeeds` map
  consumed during prerender. A single bad external date threw and failed the
  whole build.

## Fix

Validate feed dates at the **feed-parse boundary**, where untrusted third-party
data enters, and drop unparseable values to `undefined`:

- New shared helper `isParseableDate` in `blog/src/date.ts`, mirroring the
  `isNaN(new Date(...).getTime())` check already inside `formatUtcDate`. It uses
  only `new Date` / `isNaN`, so it is safe in both the browser runtime path and
  the Node build-time path.
- Both runtime parsers (`parse-feed.ts`) and both build-time parsers
  (`vite-plugin-feed-fetch.ts`) now drop the date when it is missing or
  unparseable before returning the entry.

`formatUtcDate`'s throw is **unchanged** — first-party post dates come from
validated `PublishedPost` data and should still surface a clear error on
corruption (`.claude/rules/code-style.md`). This change stops feeding
`formatUtcDate` untrusted input rather than weakening it. The render-side
empty-string fallbacks and `sortBlogrollByDate` in `info-panel.ts` already
tolerate an absent `publishedAt`, so no render-side change is needed.

## Tests

- `date.test.ts` — `isParseableDate` returns `true` for ISO and RFC-2822 dates,
  `false` for `"not-a-date"` and `""`.
- `parse-feed.test.ts` — an Atom feed and an RSS feed with an unparseable date
  parse successfully with `publishedAt` dropped to `undefined`, `title`/`url`
  preserved.
- `vite-plugin-feed-fetch.test.ts` — the matching build-time cases yield a
  non-null result with `publishedAt === undefined`.

Full blog suite: 360 passed. `npx tsc --noEmit --project blog` introduces no new
errors (the remaining `Document`/`DOMParser` DOM-lib errors pre-exist on `main`).

## Acceptance criteria

- An unparseable feed date degrades only that entry, not the rest of the
  blogroll — both parse boundaries drop the bad date to `undefined`; the
  hydration loop and prerender no longer throw mid-iteration.
- Validation happens at the feed-parse boundary, not at render time — the guard
  lives in `parse-feed.ts` and `vite-plugin-feed-fetch.ts`; `formatUtcDate` and
  the render path are unchanged.

### DIFF
--- a/blog/src/blog-roll/parse-feed.ts
+++ b/blog/src/blog-roll/parse-feed.ts
@@ -7,13 +7,18 @@ function parseAtomFeed(doc: Document): LatestPost | null {
   const title = entry.querySelector("title")?.textContent ?? "";
   const linkEl = entry.querySelector('link[rel="alternate"][href]') ?? entry.querySelector("link[href]");
   const url = linkEl?.getAttribute("href") ?? "";
-  const published =
-    entry.querySelector("published")?.textContent ??
-    entry.querySelector("updated")?.textContent ??
-    undefined;
+  const rawPublished = entry.querySelector("published")?.textContent;
+  const rawUpdated = entry.querySelector("updated")?.textContent;
   if (!title || !url) return null;
   if (!url.startsWith("http://") && !url.startsWith("https://")) return null;
-  const publishedAt = published && isParseableDate(published) ? published : undefined;
+  // Prefer the first parseable source: when <published> is present but
+  // unparseable, fall back to <updated> instead of dropping the date entirely.
+  const publishedAt =
+    rawPublished && isParseableDate(rawPublished)
+      ? rawPublished
+      : rawUpdated && isParseableDate(rawUpdated)
+        ? rawUpdated
+        : undefined;
   return { title, url, publishedAt };
 }
 
--- a/blog/src/blog-roll/parse-feed.ts
+++ b/blog/src/blog-roll/parse-feed.ts
@@ -1,4 +1,5 @@
 import type { LatestPost } from "./types.ts";
+import { isParseableDate } from "../date.ts";
 
 function parseAtomFeed(doc: Document): LatestPost | null {
   const entry = doc.querySelector("feed > entry");
@@ -12,7 +13,8 @@ function parseAtomFeed(doc: Document): LatestPost | null {
     undefined;
   if (!title || !url) return null;
   if (!url.startsWith("http://") && !url.startsWith("https://")) return null;
-  return { title, url, publishedAt: published };
+  const publishedAt = published && isParseableDate(published) ? published : undefined;
+  return { title, url, publishedAt };
 }
 
 function parseRssFeed(doc: Document): LatestPost | null {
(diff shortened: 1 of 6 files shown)


natb1:
## QA Pass — #1288

**Non-browser path** (changes are to `@commons-systems/blog` library package — `blog/src/`, no standalone frontend app).

### Items walked: 3 / PASS: 3 / FAIL: 0 / SKIP: 0

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests (24 files, 360 tests) | ✅ PASS |
| 2 | Type-check (`npx tsc --noEmit --project blog`) | ✅ PASS — 3 pre-existing DOM errors unchanged; no new errors from this PR |
| 3 | Parse-boundary code inspection | ✅ PASS |

### Item 3 detail
- `parse-feed.ts`: `isParseableDate` guard on `publishedAt` in both `parseAtomFeed` and `parseRssFeed`
- `vite-plugin-feed-fetch.ts`: same guard in `parseAtomFeedXml` and `parseRssFeedXml`
- `date.ts`: `isParseableDate` added; `formatUtcDate` throw preserved (first-party data still surfaces errors clearly)
- Render path (`info-panel.ts`) untouched — existing `publishedAt` truthiness guards already handle `undefined`

### Acceptance criteria
- [x] Unparseable feed date degrades only that entry — parse boundaries drop bad date to `undefined`; loop and prerender no longer throw mid-iteration
- [x] Validation at the feed-parse boundary — guards in `parse-feed.ts` and `vite-plugin-feed-fetch.ts`; render path unchanged

### Deferred to office-hours
None.

natb1:
## Review findings — #1411

### Fixed (499fd414)

- **parse-feed.ts:10-16** — `parseAtomFeed` silently blocked the `<updated>` fallback when `<published>` was present but unparseable. The `??`-chaining resolved both fields together, so an invalid `<published>` caused the whole expression to yield a non-parseable string rather than falling through to `<updated>`. Fixed: read `<published>` and `<updated>` separately and pick the first parseable one, so an invalid `<published>` degrades to `<updated>` instead of dropping the date entirely. Applied the identical pattern in `vite-plugin-feed-fetch.ts`. _(source: code-review)_

- **parse-feed.test.ts:116** — No test covered the case where `<published>` is present but unparseable while `<updated>` holds a valid date. Added an Atom test asserting `publishedAt` falls through to the `<updated>` value. Added the matching case for `parseAtomFeedXml` in `vite-plugin-feed-fetch.test.ts`. _(source: code-review)_

### False-positive (security)

- **date.ts:3-5** — `isParseableDate` uses `new Date(value).getTime()`, flagged for accepting V8-specific non-standard date formats that differ between Node (build-time) and browsers (runtime). The function's contract is "will `formatUtcDate` accept this?" — both environments ultimately call the same `new Date()` path, so the inconsistency would only matter for V8-non-standard formats that one environment accepts and the other rejects. Third-party feeds in practice use RFC 2822 or ISO 8601; the risk is theoretical. _(source: red-team)_

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1412, opened 2026-06-12
router: only hijack same-origin clicks that hit a real route; honor download/target and skip protocol-relative hrefs

natb1:
Closes #1291

## Problem

The router's document-level click handler (`router/src/index.ts`) hijacked
**every** same-origin absolute-path anchor click for client-side navigation.
Two defects fell out of that over-broad claim:

- **Non-SPA links hijacked.** `matchRoute` fell back to `routes[0]` for any
  unmatched path, so the handler couldn't tell a real route from an unknown
  path. A link to a real hosted file was swallowed and the home route rendered
  at that URL. Reachable today: the blog info panel renders the RSS
  `<a href="/feed.xml">` (fellspiral passes `/feed.xml`); clicking it pushed
  `/feed.xml` into history and rendered home instead of opening the feed. There
  was also no opt-out — `download` and any non-`_blank` `target` were ignored.
- **Protocol-relative crash.** `href.startsWith("/")` also matched
  `//host/path`. The handler called `preventDefault()` then
  `history.pushState({}, "", href)`, which throws an uncaught `SecurityError`
  for a cross-origin URL. Content-driven anchors (Firestore post bodies) make
  this reachable.

## Change

The router now intercepts a click **only** when the href is a same-origin
absolute path that hits a real (non-fallback) route and carries no
native-navigation opt-out; everything else falls through to the browser.

- `matchRoute` now returns `Route | undefined` — the `?? routes[0]` fallback is
  dropped from the function and moved to its sole call site inside `navigate`,
  so unknown *visited* paths still render the home route exactly as before.
- `onClick` now: skips anchors with a `download` attribute, skips any non-empty
  `target` (replacing the `=== "_blank"` check; still covers `_blank`), skips
  protocol-relative `//host/...` hrefs (fixing the `SecurityError`), and parses
  the path (mirroring `parsePath`'s trailing-slash normalization) to require a
  non-fallback `matchRoute` hit before calling `pushState`. `/feed.xml` now
  falls through to native navigation.

## Tests

Five cases added in `router/test/router.test.ts`, asserting native
fall-through via a `history.pushState` spy (the outlet alone can't distinguish
"intercepted to home" from "left on home"): `download` attribute, `target="_self"`,
protocol-relative `//external.com/path` (regression guard for the
`SecurityError`), unknown same-origin `/feed.xml` (the RSS-hijack fix), and a
positive control confirming a real `/about` route is still intercepted. All
46 router tests pass; the package type-checks with no new errors.

### DIFF
--- a/router/src/index.ts
+++ b/router/src/index.ts
@@ -119,7 +119,9 @@ export function createHistoryRouter(
     if (e.metaKey || e.ctrlKey || e.shiftKey || e.altKey) return;
     const href = anchor.getAttribute("href");
     if (!href || !href.startsWith("/") || href.startsWith("//")) return;
-    const path = new URL(href, location.origin).pathname.replace(/\/$/, "") || "/";
+    const url = new URL(href, location.origin);
+    if (url.origin !== location.origin) return;
+    const path = url.pathname.replace(/\/$/, "") || "/";
     if (!matchRoute(routes, path)) return;
     e.preventDefault();
     history.pushState({}, "", href);
--- a/router/src/index.ts
+++ b/router/src/index.ts
@@ -25,10 +25,10 @@ export interface Router {
   showTerminalError(html: string): void;
 }
 
-function matchRoute(routes: [Route, ...Route[]], path: string): Route {
+function matchRoute(routes: [Route, ...Route[]], path: string): Route | undefined {
   return routes.find((r) =>
     typeof r.path === "string" ? r.path === path : r.path.test(path),
-  ) ?? routes[0];
+  );
 }
 
 /**
@@ -54,7 +54,7 @@ function createNavigator(
     } catch (e) {
       if (!deferProgrammerError(e)) reportError(e);
     }
-    const route = matchRoute(routes, path);
+    const route = matchRoute(routes, path) ?? routes[0];
     try {
       const html = await route.render(path);
       if (id === navigationId) {
@@ -114,10 +114,13 @@ export function createHistoryRouter(
     if (e.button !== 0) return;
     const anchor = (e.target as Element).closest("a");
     if (!anchor) return;
-    if (anchor.getAttribute("target") === "_blank") return;
+    if (anchor.hasAttribute("download")) return;
+    if (anchor.getAttribute("target")) return;
     if (e.metaKey || e.ctrlKey || e.shiftKey || e.altKey) return;
     const href = anchor.getAttribute("href");
-    if (!href || !href.startsWith("/")) return;
+    if (!href || !href.startsWith("/") || href.startsWith("//")) return;
+    const path = new URL(href, location.origin).pathname.replace(/\/$/, "") || "/";
+    if (!matchRoute(routes, path)) return;
     e.preventDefault();
     history.pushState({}, "", href);
     void nav.navigate();
--- a/router/test/router.test.ts
+++ b/router/test/router.test.ts
@@ -549,6 +549,31 @@ describe("createHistoryRouter", () => {
     }
   });
 
+  it("does not intercept backslash-prefixed href that resolves cross-origin", async () => {
+    router = createHistoryRouter(outlet, routes);
+    await vi.waitFor(() => {
+      expect(outlet.innerHTML).toBe("<h2>Home</h2>");
+    });
+
+    const pushStateSpy = vi.spyOn(history, "pushState");
+    const anchor = document.createElement("a");
+    // "/\evil.com/about" — passes startsWith("/") && !startsWith("//")
+    // string guards but the WHATWG URL parser normalizes the backslash to a
+    // slash, resolving cross-origin to //evil.com/about.
+    anchor.setAttribute("href", "/" + "\\" + "evil.com/about");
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — #1291 router click-handler fix

**Non-browser path** — changes are in `router/` only (library, no frontend app).

| Item | Result |
|---|---|
| Unit tests (46 total, 4 new cases) | ✅ PASS |
| TypeScript type-check (no new errors) | ✅ PASS |
| Existing interception behavior preserved | ✅ PASS |
| RSS-link hijack fix (unknown paths fall through) | ✅ PASS |

### New test cases added

- `does not intercept anchor with download attribute` — native download falls through
- `does not intercept anchor with target=_self` — any non-empty target falls through
- `does not intercept protocol-relative href and does not throw` — `//host/path` no longer reaches `pushState`, no `SecurityError`
- `does not intercept unknown same-origin path with no matching route` — `/feed.xml` (no route) falls through to native nav

### Preserved existing behavior

All prior tests still pass: same-origin anchor interception, `_blank` fall-through, external links, hash-only links, unknown-path fallback route.

**0 bugs found. 0 judgment-call items. Clean autonomous pass.**

natb1:
## Review findings — PR #1412

### Fixed

- **`router/test/router.test.ts`** — Added a regression test for the backslash-prefix bypass: an anchor with `href="/\evil.com/about"` passes the `startsWith("/") && \!startsWith("//")` string guards but the WHATWG URL parser resolves it cross-origin, which would cause `history.pushState` to throw a `SecurityError`. The test asserts `pushState` is not called and the outlet is unchanged, guarding against any future weakening of the same-origin guard. (fix commit: cf8b9b28)

- **`router/src/index.ts:121-125`** — Backslash-normalization bypass: `/\evil.com/about` passes the string guards and the WHATWG URL parser resolves it to `https://evil.com`. The code path exists. Both skeptics refuted: every attacker-reachable anchor (markdown-rendered post bodies) is unconditionally wrapped with `target="_blank"` by `marked-config.ts:59` before the click handler sees it, and the handler returns at the non-empty `target` check before reaching the URL resolution. There is no user-controlled content path that injects a bare anchor without `target`. The residual risk is a latent defect, not a currently exploitable vulnerability. Not fixed this PR.

### Informational

- **`router/src/index.ts:118`** — `anchor.getAttribute("target")` returns `null` for a missing attribute (caught, guard works) but returns `""` for `<a target="">` (empty string, which is falsy in JS), treating it the same as no `target`. An explicit `target=""` is unusual and not a standard-compliant override; this edge case is documented here for awareness.

- **`router/test/router.test.ts:100-118`** — The pre-existing `target=_blank` test asserts outlet content to confirm non-interception. Now that the route-gate (`matchRoute`) also guards interception, a `pushState` spy would make the assertion more precise (distinguishing "returned at target check" from "returned at route-gate"). No change required — the outlet assertion is sufficient for the original test's intent.

### Dismissed

- **`router/src/index.ts:122-125`** — The click handler resolves the href to a normalized pathname for route matching (`new URL(href, location.origin).pathname`) but passes the raw attribute string to `history.pushState`. Not an issue: the guard already ensures `href` is a same-origin absolute path (`startsWith("/") && \!startsWith("//")`) before this point, so the raw string and the resolved pathname are identical for well-formed hrefs. No functional difference.

### Deferred

- **`router/src/index.ts:29-31`** — RegExp routes with the `g` or `y` flag are stateful: `RegExp.prototype.test` advances `lastIndex` on each call, and the click handler now calls `matchRoute` as a second code path (in addition to the `navigate` call site), doubling call frequency on matched routes. Pre-existing latent issue; new call site increases exposure. Follow-up filed: #1419.

natb1:
# Fix-checks summary — PR #1412

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `npx vitest run --root router`
- **Failure excerpt:** `router > test/router.test.ts > createHistoryRouter > does not intercept backslash-prefixed href that resolves cross-origin` — `AssertionError: expected [Function] to not throw an error but 'SecurityError: Failed to execute 'pushState' on 'History': A history state object with URL '/\evil.com/about' cannot be created in a document with origin 'https://external.com'' was thrown`.
- **Why not caught:** The implementation's click-interception predicate guarded same-origin only with cheap string prefix checks (`href.startsWith("/") && !href.startsWith("//")`). A backslash-prefixed href `/\evil.com/about` passes both, but the WHATWG URL parser normalizes the backslash to a slash, resolving it cross-origin (host `evil.com`, pathname `/about`). `/about` matches a real route, so the handler intercepted and called `history.pushState(..., href)` with the raw href, throwing an uncaught `SecurityError`. The PR's own added test encoded this contract but the implementation predicate missed the backslash case — the string guards alone cannot detect cross-origin resolution.
- **Fix:** After the string guards, parse the URL once and verify it is actually same-origin before deriving the path: `const url = new URL(href, location.origin); if (url.origin !== location.origin) return;`. This makes backslash-prefixed (and any other) hrefs that resolve cross-origin fall through to native navigation, so `pushState` is never reached. Commit `01df5b94`.

PR #1413, opened 2026-06-12
idbutil: close on versionchange, stop onblocked leak, surface onUpgrade errors

natb1:
Closes #1292

## Summary

`createDbConnection` in `idbutil/src/connection.ts` memoizes a versioned
IndexedDB connection. Its open-request handlers had three defects that broke
multi-tab schema upgrades. This PR fixes all three and covers them with tests.

## Defects fixed

1. **No `onversionchange` handler.** An open connection never closed when
   another tab opened the same DB at a bumped `version`, so after a deploy that
   bumps the version (budget is already at v5), an old-code tab permanently
   blocked the new tab's upgrade. Now `request.onsuccess` wires
   `db.onversionchange = () => { db.close(); dbPromise = null; }` — the
   connection releases itself and clears the cached promise. (`close()` does not
   fire `onclose`, so the promise is nulled explicitly.)

2. **`onblocked` leak.** `blocked` does not mean the open failed — the request
   stays pending and later fires `onsuccess`. The old handler rejected and
   nulled `dbPromise`, so the eventual `resolve` was a no-op and the now-open
   `IDBDatabase` leaked, itself a future blocker. `onblocked` now only
   `console.warn`s the transient block and settles nothing; the request settles
   once the other tabs' `onversionchange` handlers close their connections.

3. **`onUpgrade` errors masked.** When `config.onUpgrade` threw, the
   versionchange transaction aborted and `onerror` fired with a generic
   `AbortError`. The thrown cause is now captured in `onupgradeneeded`, the
   transaction is aborted explicitly, and `onerror` rejects with the original
   cause.

## Tests

- `onversionchange` closes the connection, clears the cache, and lets another
  tab upgrade (v1 → v2).
- `onblocked` does not reject an open that later succeeds.
- Strengthened the existing upgrade-throws test to assert the original
  `"upgrade failed"` message rather than a generic abort error.

`npx vitest run --project idbutil` — 31 passed; type-check clean. The public
`{ openDb, closeDb }` shape is unchanged, so no consumer edits are needed.

### DIFF
--- a/idbutil/src/connection.ts
+++ b/idbutil/src/connection.ts
@@ -24,19 +24,28 @@ export function createDbConnection(config: DbConnectionConfig): {
   function openDb(): Promise<IDBDatabase> {
     if (!dbPromise) {
       dbPromise = new Promise((resolve, reject) => {
+        let upgradeError: unknown;
         const request = indexedDB.open(config.name, config.version);
         request.onupgradeneeded = (event) => {
-          config.onUpgrade(request.result, event.oldVersion);
+          try {
+            config.onUpgrade(request.result, event.oldVersion);
+          } catch (err) {
+            upgradeError = err;
+            request.transaction?.abort();
+          }
         };
         request.onsuccess = () => {
-          request.result.onclose = () => { dbPromise = null; };
-          resolve(request.result);
+          const db = request.result;
+          db.onclose = () => { dbPromise = null; };
+          db.onversionchange = () => { db.close(); dbPromise = null; };
+          resolve(db);
         };
         request.onblocked = () => {
-          dbPromise = null;
-          reject(new Error("Database upgrade blocked. Close other tabs using this app and try again."));
+          console.warn(
+            `IndexedDB upgrade of "${config.name}" is blocked by another connection; waiting for it to close.`,
+          );
         };
-        request.onerror = () => { dbPromise = null; reject(request.error); };
+        request.onerror = () => { dbPromise = null; reject(upgradeError ?? request.error); };
       });
     }
     return dbPromise;
--- a/idbutil/test/connection.test.ts
+++ b/idbutil/test/connection.test.ts
@@ -88,6 +88,91 @@ describe("openDb", () => {
     expect(db2).not.toBe(db1);
     await closeDb();
   });
+
+  it("onversionchange closes the connection, clears the cache, and lets another tab upgrade", async () => {
+    const name = uniqueDbName();
+    // Tab 1: open at v1 and hold the connection.
+    const tab1 = createDbConnection({
+      name,
+      version: 1,
+      onUpgrade(db) { db.createObjectStore("s"); },
+    });
+    const db1 = await tab1.openDb();
+    expect(typeof db1.onversionchange).toBe("function");
+    // Wrap the wired handler so we can observe it fired while keeping its
+    // close-and-clear-cache behavior.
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — #1292

**PR**: #1413 — idbutil: close on versionchange, stop onblocked leak, surface onUpgrade errors
**Path**: non-browser (pure `idbutil/` changes)

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests pass (existing suite — 31 tests, 2 files) | ✅ PASS |
| 2 | `onversionchange` closes connection, clears cache, lets other tab upgrade | ✅ PASS |
| 3 | `onblocked` does not reject an open that later succeeds | ✅ PASS |
| 4 | `onUpgrade` error surfaces as original error, not generic `AbortError` | ✅ PASS |
| 5 | TypeScript compiles cleanly (via vitest TS transform — no standalone tsconfig) | ✅ PASS |

Deferred to office-hours: none
Bugs found: none

natb1:
## Review — PR #1413

_No code changes applied (no Required/Fixed findings)._

---

### Informational

- **`idbutil/test/connection.test.ts:143`** — The "onblocked does not reject an open that later succeeds" test does not spy on `console.warn`, so it cannot verify that the `onblocked` code path actually executed. The test is structurally sound (it asserts no rejection), but a `console.warn` spy would make intent more explicit.
- **`idbutil/test/connection.test.ts:154-157`** — The `onblocked` test's manual `onversionchange` override calls `setTimeout(() => db1.close(), 0)` but does not null `tab1`'s `dbPromise`. This is intentional for the test (it is the mock holder, not the subject), but worth noting for clarity.
- **`idbutil/test/connection.test.ts:104-108`** — The versionchange spy wrapper uses a non-null assertion (`db1.onversionchange!`) immediately after asserting `typeof db1.onversionchange === "function"`. Redundant assertion; harmless.

---

### Dismissed

- **`idbutil/src/connection.ts:48`** — `upgradeError` closure variable initialized to `undefined`, set only inside `onupgradeneeded`. Raised as a concern that `onerror` could fire before the handler runs; dismissed — if `onerror` fires without `onupgradeneeded` ever running, `upgradeError` is `undefined` and `request.error` (the real IDB error) is used as the rejection reason, which is correct behavior.

---

### False-positive

- **`idbutil/src/connection.ts:40`** — Flagged: `db.onversionchange` closes the connection immediately on any same-origin version bump. This is the correct and intended behavior (the handler is wired to release the connection so another tab can upgrade). Not a bug.
- **`idbutil/src/connection.ts:27-34`** — Flagged: `upgradeError` capture and propagation could expose internal errors to callers. This is the fix for acceptance criterion #3 (surface `onUpgrade` errors); surfacing the real error to the caller is correct.
- **`idbutil/src/connection.ts:44`** — Flagged: `console.warn` interpolates `config.name`. `config.name` is a developer-supplied DB name string (not user data); no data exposure.
- **`idbutil/src/connection.ts:48`** — Flagged: `onerror` rejects with `upgradeError ?? request.error`. Both are Error objects from trusted internal sources; no data exposure.

---

### Deferred

- **`idbutil/src/connection.ts:43-47`** — The `onblocked` handler emits a `console.warn` and leaves `openDb()` permanently pending if the blocking connection never closes (e.g., a tab without a `versionchange` handler). All callers that `await openDb()` will hang with no timeout or cancellation path. Tracked in #1420 (sub-issue of #1292).

PR #1414, opened 2026-06-12
idbutil: reject oversized blobs and evict on the replace delta

natb1:
Closes #1294

## Problem

`idbutil/src/lru-blob-cache.ts` — the shared IndexedDB LRU blob cache used by `print` and `audio` — had two accounting defects in its eviction path:

1. **An oversized blob wiped the entire cache.** When `data.byteLength > maxBytes`, the eviction target `totalSize + incomingSize - maxBytes` exceeded `totalSize`, so the eviction cursor deleted *every* existing entry. `putEntry` then wrote the oversized blob anyway, leaving `totalBytes > maxBytes`. One un-storable item destroyed the whole cache.

2. **Replacing a key over-evicted.** `putEntry` passed the gross incoming size to `evictIfNeeded` instead of the net delta (`newSize - oldSize`), so a replace evicted as if the whole incoming blob were new — discarding more entries than the update required.

## Fix

- **Reject oversized blobs.** `putEntry` now throws as its first statement when `data.byteLength > maxBytes` (strict `>`, so a blob exactly at the cap still fits), before opening the DB or evicting — the cache is untouched on rejection. All call sites already `.catch()` the best-effort write, so the rejection surfaces to `reportError`/`logError` cleanly (per `.claude/rules/code-style.md`: prefer clear errors over silent fallbacks).

- **Evict on the net delta.** `evictIfNeeded` now takes the `key`, reads both the `__total__` sentinel and the key's current size in one readonly transaction (`getTotalAndOldSize`), and computes `delta = incomingSize - oldSize`. The early-return guard and eviction target both use `totalSize + delta`. The eviction cursor also skips the replaced `key` — required so the put's later re-read of `oldSize` stays consistent and the total can't overshoot `maxBytes`.

`getTotalSize` was removed: `evictIfNeeded` was its only caller, and `getStats` has its own inline `__total__` read.

## Tests

Two new cases in `idbutil/test/lru-blob-cache.test.ts` (both fail against the old code):

- *rejects an oversized entry without evicting the cache* — verifies the existing entry survives, the oversized blob is not stored, and `totalBytes`/`entryCount` are unchanged.
- *replace evicts on the size delta, not gross incoming size* — cap 250, `b`=100 + `a`=100, then replace `a`→120 (delta 20, fits): `b` survives, total ends at 220.

Full suite: 31 passed. `tsc --noEmit` clean for both `audio` and `print` (idbutil is a source-export package with no own tsconfig; its consumers type-check `lru-blob-cache.ts`).

### DIFF
--- a/idbutil/src/lru-blob-cache.ts
+++ b/idbutil/src/lru-blob-cache.ts
@@ -87,15 +87,24 @@ export function createLruBlobCache(config: LruBlobCacheConfig): LruBlobCache {
     };
   }
 
-  async function getTotalSize(db: IDBDatabase): Promise<number> {
+  async function getTotalAndOldSize(
+    db: IDBDatabase,
+    key: string,
+  ): Promise<{ totalSize: number; oldSize: number }> {
     return new Promise((resolve, reject) => {
       const tx = db.transaction(META_STORE, "readonly");
-      const req = tx.objectStore(META_STORE).get(TOTAL_KEY);
-      req.onsuccess = () => {
-        const entry = req.result as MetaEntry | undefined;
-        resolve(entry ? entry.size : 0);
+      const store = tx.objectStore(META_STORE);
+      const totalReq = store.get(TOTAL_KEY);
+      const oldReq = store.get(key);
+      tx.oncomplete = () => {
+        const totalEntry = totalReq.result as MetaEntry | undefined;
+        const oldEntry = oldReq.result as MetaEntry | undefined;
+        resolve({
+          totalSize: totalEntry ? totalEntry.size : 0,
+          oldSize: oldEntry ? oldEntry.size : 0,
+        });
       };
-      req.onerror = () => reject(req.error);
+      tx.onerror = () => reject(tx.error);
     });
   }
 
@@ -121,10 +130,11 @@ export function createLruBlobCache(config: LruBlobCacheConfig): LruBlobCache {
    * the subsequent put in the caller (putEntry); concurrent writes may briefly
    * exceed the cap, which is acceptable for a best-effort cache.
    */
-  async function evictIfNeeded(db: IDBDatabase, incomingSize: number): Promise<void> {
-    const totalSize = await getTotalSize(db);
+  async function evictIfNeeded(db: IDBDatabase, key: string, incomingSize: number): Promise<void> {
+    const { totalSize, oldSize } = await getTotalAndOldSize(db, key);
+    const delta = incomingSize - oldSize;
 
-    if (totalSize + incomingSize <= maxBytes) return;
+    if (totalSize + delta <= maxBytes) return;
 
     const evictTx = db.transaction([DATA_STORE, META_STORE], "readwrite");
     const metaStore = evictTx.objectStore(META_STORE);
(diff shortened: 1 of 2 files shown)


natb1:
## QA Summary — #1294

**PR:** #1414  
**Path:** non-browser (idbutil only)

### Items walked

| # | Item | Result |
|---|------|--------|
| 1 | Rejects an oversized entry without evicting the cache | ✅ PASS |
| 2 | Replace evicts on the size delta, not gross incoming size | ✅ PASS |
| 3 | Existing eviction tests still pass | ✅ PASS |
| 4 | Type-check | N/A (idbutil has no tsconfig.json — pre-existing; TypeScript validated by tests) |

**Totals:** 3 PASS · 0 FAIL · 0 SKIP

### Deferred to office-hours
None.

### Bugs found
None.

Verified by running `npx vitest run --project idbutil`: 2 test files, 31 tests, all green.

natb1:
## Review: #1294 — idbutil: reject oversized blobs and evict on the replace delta

No code changes applied — all findings resolved as informational, false-positive, or deferred.

---

### Fixed

_None._

### Required (security)

_None._

### Informational

- **`idbutil/src/lru-blob-cache.ts:128-132`** — The existing TOCTOU comment correctly notes that `evictIfNeeded`'s readonly size read and `putEntry`'s readwrite transaction are two separate IDB transactions; a concurrent writer can interleave between them. The new delta-eviction logic does not worsen this pre-existing race. No change needed; noted for awareness. _(source: review)_

- **`idbutil/src/lru-blob-cache.ts:156-159`** — The eviction cursor skips `entry.key === key` (the replaced entry) to preserve delta-accounting consistency. If `key` happens to equal `__total__` (the sentinel), the sentinel skip fires first (`entry.key === TOTAL_KEY`), so the two conditions do not interact. Structurally sound. _(source: red-team)_

### Dismissed

_None._

### False-positive (security)

- **`idbutil/src/lru-blob-cache.ts:90-108`** — `getTotalAndOldSize` calls `store.get(key)` where `key` is caller-supplied. If `key === '__total__'`, the read returns the sentinel metadata object, not a blob descriptor — `oldSize` would be `NaN` or an unexpected value. However, `__total__` is an internal sentinel key that is never passed as a user-visible cache key; all call sites supply media-derived keys. Not an exploitable path. _(source: red-team)_

- **`idbutil/src/lru-blob-cache.ts:191-195`** — The error message includes `data.byteLength` and `maxBytes` verbatim. These values are integer byte counts from an in-process IndexedDB cache operating entirely in the browser's local storage. No sensitive data is disclosed; the error surfaces in the browser console via `reportError`/`logError`. Not a meaningful information disclosure risk. _(source: red-team, data-exposure)_

### Deferred

- **`idbutil/test/lru-blob-cache.test.ts:179`** — The replace-delta test only covers the positive-delta (size grows, no eviction needed) case. The negative-delta-with-eviction path and the behavior of the `entry.key === key` cursor skip under that scenario are untested. Tracked in #1417 (blocked by #1294). _(source: code-review)_

- **`idbutil/src/lru-blob-cache.ts:56`** — `maxBytes` is not validated at construction time; `maxBytes: 0` silently disables the cache and `maxBytes: -1` bypasses the oversized-blob guard. Tracked in #1418 (blocked by #1294). _(source: input-validation)_

### Out-of-scope (security)

- **`idbutil/src/lru-blob-cache.ts:133-137`** — Two concurrent `putEntry` calls for the same key can both read `oldSize` before either write commits, causing both to evict on the full incoming size rather than the delta, potentially over-evicting. This is a pre-existing TOCTOU race that the diff does not introduce or worsen; no fix in scope here. _(source: red-team)_

PR #1415, opened 2026-06-12
audio: gate playTrack async resolves behind a generation token

natb1:
Closes #1302

## Problem

`audio/src/player.ts`'s `playTrack` was fire-and-forget async with no
generation/cancellation guard. `resolveAudioSource` on a cache miss does a
`getDownloadURL` plus a full-file fetch (seconds of latency), so a single
`playTrack` call has a long pending window during which queue state can change.
Three defects resulted:

- **Play-after-remove** — queue a track (fetch starts), uncheck it before the
  fetch resolves → `remove` → `stop()` clears `src`, then the pending `.then`
  fired anyway, set `audioEl.src`, and called `play()`. The removed track
  audibly played while the UI showed it unchecked.
- **Out-of-order resolves + blob leak** — two overlapping `playTrack` calls each
  assigned `currentObjectUrl = url`; the first was overwritten without
  `URL.revokeObjectURL`, permanently leaking a blob holding a whole audio file,
  and `audioEl.src` could settle on a different track than the highlighted row.
- **Stale index** — on resolve failure, `advanceOrStop(index)` used the index
  captured at call time against a queue that may have been mutated mid-flight.

## Fix

Mirror the router's `navigationId` pattern (`router/src/index.ts`): a
closure-scoped `playGeneration` counter, captured per `playTrack` call as
`const generation = ++playGeneration`, gates every post-await side effect.

- The `.then` handler bails when `generation !== playGeneration`, revoking the
  just-resolved blob so it never leaks, before touching `currentObjectUrl`,
  `audioEl.src`, or `play()`.
- The `.catch` handler still logs the real failure, but only advances when the
  generation is live — and advances from the **live** `currentIndex`, not the
  stale captured `index` (matching `onEnded`).
- `stop()` bumps the counter, so any resolve still pending when playback stops
  is invalidated. This is what makes the play-after-remove path inert.

Only the most-recent live request can resurrect playback; stopped or superseded
resolves become inert and clean up their own blob. No public API or signature
changes.

## Tests

Three new cases in `audio/test/player.test.ts` drive the defects with a
deferred-promise mock for controllable, interleaved resolution:

- **play-after-remove** — stale resolve does not call `play()`, does not set
  `src`, and revokes the blob.
- **out-of-order resolve does not desync** — `src` ends on the gen-2 blob; the
  gen-1 blob is revoked, never assigned.
- **stale failed resolve does not advance** — a superseded rejection is inert; a
  live rejection advances from the correct live index.

All 24 cases pass (21 existing + 3 new); `tsc --noEmit` and lint are clean.

### DIFF
--- a/audio/src/player.ts
+++ b/audio/src/player.ts
@@ -28,6 +28,7 @@ export function initPlayer(
   const queue: PlayRequest[] = [];
   let currentIndex = -1;
   let currentObjectUrl: string | null = null;
+  let playGeneration = 0;
 
   function renderPlaylist(): void {
     if (queue.length === 0) {
@@ -62,11 +63,16 @@ export function initPlayer(
   function playTrack(index: number): void {
     const item = queue[index];
     if (!item) throw new Error(`playTrack: index ${index} out of range (queue length ${queue.length})`);
+    const generation = ++playGeneration;
     currentIndex = index;
     renderPlaylist();
     revokeCurrentObjectUrl();
     resolveAudioSource(item.storagePath)
       .then((url) => {
+        if (generation !== playGeneration) {
+          URL.revokeObjectURL(url);
+          return;
+        }
         currentObjectUrl = url;
         audioEl.src = url;
         audioEl.play().catch((err) => {
@@ -77,11 +83,13 @@ export function initPlayer(
       .catch((err) => {
         if (deferProgrammerError(err)) return;
         logError(err, { operation: "audio-source-resolve", storagePath: item.storagePath });
-        advanceOrStop(index);
+        if (generation !== playGeneration) return;
+        advanceOrStop(currentIndex);
       });
   }
 
   function stop(): void {
+    playGeneration++;
     currentIndex = -1;
     revokeCurrentObjectUrl();
     audioEl.pause();
--- a/audio/test/player.test.ts
+++ b/audio/test/player.test.ts
@@ -313,12 +313,12 @@ describe("initPlayer", () => {
 
       // now resolve the pending promise — should be treated as stale
       deferred[0]!.resolve("blob:http://localhost/t1-blob");
-      await Promise.resolve();
-      await Promise.resolve();
 
+      await vi.waitFor(() => {
+        expect(revokeObjectURLSpy).toHaveBeenCalledWith("blob:http://localhost/t1-blob");
+      });
       expect(audioEl.play).not.toHaveBeenCalled();
       expect(audioEl.getAttribute("src")).toBeNull();
-      expect(revokeObjectURLSpy).toHaveBeenCalledWith("blob:http://localhost/t1-blob");
     });
 
     it("out-of-order resolve does not desync: gen-1 blob is revoked and src ends on gen-2 blob", async () => {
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — #1302 (PR #1415)

Non-browser path (changed files: `audio/src/player.ts`, `audio/test/player.test.ts`).

### Results: 3 PASS · 0 FAIL · 0 SKIP

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests (`npx vitest run --root audio test/player.test.ts`) | ✅ PASS — 24/24 passed |
| 2 | Type check (`npx tsc --noEmit --project audio`) | ✅ PASS — zero errors |
| 3 | Lint (`run-lint.sh --app audio`) | ✅ PASS — zero ESLint errors |

### New test coverage

The 24-test suite includes 3 new deferred-mock cases that pin the three race conditions fixed in #1302:

- **Play-after-remove**: stale resolve after `remove()` does not start playback and revokes the blob
- **Out-of-order resolve**: the later generation wins; the earlier blob is revoked without being assigned
- **Stale failed resolve**: a superseded catch branch does not call `advanceOrStop`

### Deferred to office-hours

### Bugs found

None.

natb1:
## Review — #1302

**Changed surface:** `audio/src/player.ts`, `audio/test/player.test.ts` (code, no deps)
**CodeQL:** no open alerts
**npm audit:** not applicable (no dependency changes)

---

### Fixed (commit 3e9e7027)

- **`audio/test/player.test.ts:316`** — Three generation-token tests drained promise chains with bare `await Promise.resolve(); await Promise.resolve()` pairs, which is fragile under variable microtask queue depth. Replaced with `vi.waitFor()` blocks matching the existing pattern in the test file. _(review)_
- **`audio/test/player.test.ts:371`** — The "stale failed resolve does not advance" test did not pin the catch-path generation guard: a bug that incorrectly advanced the queue would grow `deferred` to 3 but nothing asserted on that. Added `expect(deferred).toHaveLength(2)` after the gen-1 reject to catch that failure mode. _(code-review)_

---

### Informational

- **`audio/test/player.test.ts:372`** — `playCallsBefore` captures `mock.calls.length` immediately after `player.remove()` triggers a new `playTrack` call for track2; the timing is fine because the deferred-mock harness holds gen-2 unresolved while gen-1 is rejected, but the variable name implies a "before" snapshot taken after a side effect. No change needed. _(review)_

---

### False-positive (security)

- **`audio/src/player.ts:83-85`** — `logError` is called unconditionally in `.catch` before the generation guard. Flagged as a potential double-log on stale resolves. Not an issue: the error is real regardless of staleness and should be logged; only the advance (line 87) is guarded. _(red-team)_
- **`audio/src/player.ts:66,92`** — `playGeneration` uses prefix `++` in `playTrack` and postfix `++` in `stop()`. Flagged as a potential numeric-overflow or off-by-one concern. Not an issue: both forms correctly advance the counter before the guard check, and JS numbers cannot overflow in a single session. _(red-team)_

---

### Out-of-scope (security)

- **`audio/src/player.ts:86-87`** — In the `.catch` guard, `advanceOrStop(currentIndex)` reads `currentIndex` at the time the catch fires, not at the time the promise was initiated. A concurrent `playTrack` between reject and the guard passing could advance from the wrong position. This is a valid concern but exists in pre-existing `onEnded` logic and is independent of this PR's generation-token fix; it warrants its own issue. _(red-team)_

---

_No deferred follow-ups. No security follow-ups. Deviation: none._

PR #1416, opened 2026-06-12
utils: fix media ordering, app-check init flag, nav double-define, redirect cancel code

natb1:
Closes #1295

Four low-severity correctness fixes in the shared utility packages, each
independent and each a latent trap rather than an active user-facing bug. Each
site is brought into line with its siblings / its own documented intent so the
next consumer does not inherit a surprise.

## Changes

- **`firestoreutil` — `getUserMedia` ordering.** `getUserMedia` returned raw
  Firestore document-ID order while its siblings `getPublicMedia` and
  `getAllAccessibleMedia` sort by `addedAt` descending. It now sorts with the
  same inline comparator, consistent with the siblings it sits beside (and is
  re-exported by audio/print).

- **`firebaseutil` — `initAppCheck` premature flag.** The deferred-init closure
  set `initialized = true` before the fallible `doInitAppCheck()`. A thrown
  programmer error left the guard tripped, so every later call was a silent
  no-op and `getAppCheckHeaders` returned `{}` forever. The flag is now set only
  after a successful init, so a later call retries. `doInitAppCheck()` is
  synchronous, so reentrancy safety is unchanged.

- **`components` — `customElements` double-define.** The bare
  `customElements.define("app-nav", …)` threw a DOMException on a second module
  evaluation (HMR / two entry chunks). It is now guarded with
  `if (!customElements.get("app-nav"))`, making the define idempotent.

- **`authutil` — redirect-cancel error code.** The redirect-only flow suppressed
  `auth/popup-closed-by-user` — a popup-flow code that never fires in a redirect
  flow. The redirect user-cancel code is `auth/user-cancelled`; a backed-out
  consent now is silently dismissed instead of toasting + logging an error. The
  matching JSDoc sentence was updated to agree.

## Verification

Per-package vitest suites, lint, and type-check remain the verification path for
each touched package. Each edit is a single-site change consistent with existing
siblings / documented intent.

### DIFF
--- a/authutil/src/firebase-auth.ts
+++ b/authutil/src/firebase-auth.ts
@@ -74,7 +74,7 @@ function showAuthError(message: string): void {
  * page load (user returning from GitHub OAuth or emulator picker).
  * Auth errors display a dismissible toast rather than throwing, to
  * prevent unhandled rejections from blocking app initialization.
- * The auth/popup-closed-by-user error is intentionally dismissed without
+ * The auth/user-cancelled error is intentionally dismissed without
  * showing a toast, since it indicates normal user cancellation.
  */
 export function createFirebaseAuth(app: FirebaseApp, options?: FirebaseAuthOptions): AppAuth {
@@ -85,7 +85,7 @@ export function createFirebaseAuth(app: FirebaseApp, options?: FirebaseAuthOptio
   }
 
   getRedirectResult(auth).catch((error) => {
-    if (isAuthError(error) && error.code === "auth/popup-closed-by-user") {
+    if (isAuthError(error) && error.code === "auth/user-cancelled") {
       console.debug("Auth redirect cancelled by user");
       return;
     }
--- a/authutil/test/firebase-auth.test.ts
+++ b/authutil/test/firebase-auth.test.ts
@@ -136,9 +136,9 @@ describe("createFirebaseAuth", () => {
     expect(toast?.textContent).toContain("Network error");
   });
 
-  it("dismisses popup-closed-by-user without showing toast", async () => {
-    const popupError = Object.assign(new Error("popup closed"), {
-      code: "auth/popup-closed-by-user",
+  it("dismisses user-cancelled without showing toast", async () => {
+    const cancelError = Object.assign(new Error("user cancelled"), {
+      code: "auth/user-cancelled",
     });
     let rejectFn!: (error: unknown) => void;
     mockGetRedirectResult.mockReturnValue(
@@ -149,7 +149,7 @@ describe("createFirebaseAuth", () => {
     const debugSpy = vi.spyOn(console, "debug").mockImplementation(() => {});
 
     createFirebaseAuth(mockApp);
-    rejectFn(popupError);
+    rejectFn(cancelError);
     await new Promise((r) => setTimeout(r, 0));
 
     expect(debugSpy).toHaveBeenCalledWith("Auth redirect cancelled by user");
--- a/components/src/nav.ts
+++ b/components/src/nav.ts
@@ -109,6 +109,8 @@ class AppNavElement extends HTMLElement {
   }
 }
 
-customElements.define("app-nav", AppNavElement);
+if (!customElements.get("app-nav")) {
+  customElements.define("app-nav", AppNavElement);
+}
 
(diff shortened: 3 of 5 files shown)


natb1:
# Fix-checks summary — PR #1416

## Iteration 1

- **Failed checks:** `unit-tests`
- **Outcome:** `fixed`
- **Reproduced:** yes
- **Reproduce command:** `.claude/skills/dispatch-propagate/scripts/run-unit-tests.sh` (target suite: `npm run test --prefix authutil`)
- **Failure excerpt:**
  ```
  FAIL authutil test/firebase-auth.test.ts > createFirebaseAuth > dismisses popup-closed-by-user without showing toast
  AssertionError: expected "debug" to be called with arguments: [ 'Auth redirect cancelled by user' ]
  Number of calls: 0
  stderr: Unhandled Firebase auth error code: auth/popup-closed-by-user
  ```
- **Why not caught:** Unit 4 of #1295 changed the source (`authutil/src/firebase-auth.ts:87-94`) to suppress the redirect-flow cancel code `auth/user-cancelled` (previously `auth/popup-closed-by-user`), but the existing test still constructed an error with `code: "auth/popup-closed-by-user"` and asserted the `console.debug` suppression fired. The source no longer matched that code, so the debug call never happened and the test failed. The test was not updated alongside the source change during the implement phase.
- **Fix:** Updated `authutil/test/firebase-auth.test.ts` (the `dismisses … without showing toast` test) to feed `code: "auth/user-cancelled"` — the redirect-flow code the source now suppresses — and renamed the test description, error variable (`popupError` → `cancelError`), and error message string to match. Commit `81254349`. Verified: `authutil` suite green (46 tests pass); `tsc --noEmit --project authutil` shows only the pre-existing baseline DOM/`ImportMeta.env` lib errors (same on `origin/main`), none introduced.

natb1:
## QA Summary — #1295 utils: minor correctness fixes

**Path:** Non-browser (all changes in utility packages — no frontend app dirs touched)

| Item | Check | Result |
|---|---|---|
| 1. `getUserMedia` sorts by `addedAt` desc | `vitest run --project firestoreutil` (60 tests) | ✅ PASS |
| 2. `initAppCheck` sets flag after init | `vitest run --project firebaseutil` (85 tests) | ✅ PASS |
| 3. `customElements.define` double-registration guard | `vitest run --project components` (66 tests) | ✅ PASS |
| 4. Redirect-cancel uses `auth/user-cancelled` | `vitest run --project authutil` (46 tests) | ✅ PASS |
| 5. Lint — all four packages | `eslint src/` per package | ✅ PASS |
| 6. Type-check — all four packages | `run-typecheck.sh` (regression-only) | ✅ PASS |

**Totals:** 6 PASS · 0 FAIL · 0 SKIP  
**Needs human judgment:** none  
**Bugs found:** none

`firebaseutil` and `authutil` have pre-existing type errors on `origin/main` (Vite-specific globals not in `tsconfig.lib.json`); the regression check skips them — no new regressions introduced.

natb1:
## Review

No code changes required — all four fixes are correct as implemented.

**Fixed**

_None._

**Required (security)**

_None._

_None._

**Informational**

- `firebaseutil/src/app-context.ts:194-197` — `initAppCheck` closure is declared `async` but has no `await` between the reentrancy guard and `initialized = true`. After the fix, if `doInitAppCheck()` throws, `initialized` stays `false` and the next call retries — reentrancy-safe because there are no `await` points between guard and assignment.
- `firestoreutil/src/media-queries.ts:25-26` — `getUserMedia` now sorts by `addedAt` desc using `localeCompare`, consistent with `getPublicMedia` and `getAllAccessibleMedia`.
- `components/src/nav.ts:112-114` — `customElements.define("app-nav", AppNavElement)` is guarded with `if (!customElements.get("app-nav"))`, preventing a `DOMException` on second module evaluation (HMR / two entry chunks).
- `authutil/src/firebase-auth.ts:88` — Suppressed error code updated from `auth/popup-closed-by-user` (popup flow) to `auth/user-cancelled` (redirect flow). JSDoc updated to match.

**Dismissed**

_None._

**False-positive (security)**

_None._

**Deferred**

_None._

**Out-of-scope (security)**

_None._

PR #1421, opened 2026-06-12
audio: handle media-element error events — log, evict the corrupt cache entry, advance the queue

natb1:
Closes #1303

## Problem

`audio/src/player.ts` attached only an `ended` listener to the `<audio>` element,
and `audioEl.play().catch(...)` covered only the play-promise rejection. A decode
error mid-stream — a corrupt cached `ArrayBuffer`, or a file whose extension
doesn't match its codec (`storage.ts` infers MIME from the extension alone) —
fires an `error` event on the element. That event never fires `ended` and was
handled nowhere: `advanceOrStop` was never called, so playback halted, the row
stayed `aria-current`, the queue never advanced, and no `logError` telemetry was
emitted.

## Fix

A symmetric `error` listener on the media element, plus the cache-eviction
primitive it needs, built in three commits:

1. **`idbutil`: add `deleteEntry(key)` to the shared LRU blob cache.** The
   primitive had `clearCache()` (nukes everything) but no single-entry delete.
   `deleteEntry` removes one entry in one `readwrite` transaction and decrements
   the `__total__` size sentinel by exactly the removed size. A missing key is a
   no-op that never touches the sentinel (so accounting can't drift). Strictly
   additive — `print` also consumes this primitive and is unaffected.

2. **`audio`: surface `removeFile(storagePath)`.** A thin wrapper over
   `cache.deleteEntry` in `audio-cache.ts`, re-exported from the `storage.ts`
   facade so the player reaches it through its single `./storage.js` import
   surface.

3. **`audio`: handle media-element `error` events in the player.** A new
   `onError` handler, structurally symmetric with `onEnded`:
   - Guards `if (currentIndex < 0) return;` first, so the spurious `error` that
     `stop()`'s `audioEl.load()` can fire after teardown is ignored (no wrong
     advance or evict).
   - Logs via `logError` with the storage path and `audioEl.error.code` under a
     distinct `operation: "audio-element-error"` label, separating decode
     failures from the existing play-promise and resolve-failure telemetry. No
     `deferProgrammerError` wrapping — an element `error` event is operational
     telemetry, not a thrown JS programmer error.
   - Evicts the failed track's cached buffer fire-and-forget
     (`removeFile(...).catch(logError)`), so a corrupt buffer doesn't poison a
     re-add — a re-add re-fetches from storage instead of re-reading the corrupt
     cached buffer.
   - Calls `advanceOrStop` synchronously (not awaiting eviction), so the queue
     advances immediately even if IndexedDB is slow or fails.
   The listener is registered next to `ended` and torn down alongside it in
   `destroy()`.

## Testing

- `idbutil/test/lru-blob-cache.test.ts` — `deleteEntry` correctness: survives a
  sibling entry, drops `totalBytes` by exactly the removed size, no-op on a
  never-inserted key, and correct accounting on delete-then-reput (drift guard).
  Existing eviction/stats tests still pass (protects `print`).
- `audio/test/audio-cache.test.ts` — `removeFile` round-trip plus stats checks.
- `audio/test/player.test.ts` — new "error event" describe: advance-on-error,
  `logError` carrying the `MediaError` code under `audio-element-error`,
  `removeFile` called with the failed track's path, clean stop at end of queue,
  and the `currentIndex < 0` guard no-op.
- Full `idbutil` and `audio` vitest suites pass; `tsc --noEmit` clean for both
  workspaces.


natb1:
## QA pass — #1303 (PR #1421)

**Path:** non-browser (pure-library: `audio/`, `idbutil/`)

| # | Item | Result |
|---|------|--------|
| 1 | `idbutil` `deleteEntry` unit tests | ✅ PASS |
| 2 | `audio` `removeFile` unit tests | ✅ PASS |
| 3 | Player `error` event unit tests (advance, log, evict, guard) | ✅ PASS |
| 4 | `idbutil` type-check | ⏭ SKIP — no standalone tsconfig; types validated via audio build |
| 5 | `audio` type-check (`tsc --noEmit`) | ✅ PASS |
| 6 | `audio` build (`tsc && vite build`) | ✅ PASS |

**Counts:** 5 PASS · 0 FAIL · 1 SKIP  
**Tests:** 143 passing across 10 test files

No bugs found. No needs-human-judgment items. No deferred items for office-hours.

natb1:
## Review: #1303 — audio: no error listener on the media element

### Fixed

- **`audio/test/player.test.ts`** — Added `removes error event listener` test in the `destroy` describe block: dispatches `error` after `player.destroy()` and asserts `mockLogError` and `mockRemoveFile` are not called, mirroring the existing `removes ended event listener` test. (Fix commit: 67ff0deb)

### Informational

- **`audio/src/player.ts:114`** — The `removeFile` eviction in `onError` races with the fire-and-forget `putFile` issued inside `resolveAudioSource`. If a new fetch completes after the eviction, the newly-cached buffer is unaffected. In practice the error event fires before any re-fetch, so the window is tight and the consequence is a redundant IDB write — not data loss. Documented for awareness.

- **`audio/src/player.ts:114`** — `removeFile(item.storagePath)` is called unconditionally on any audio-element error, including transient ones (e.g. network interruptions) unrelated to the cached file. The existing design is acceptable: evicting on a transient error causes a re-fetch on next play, which is cheap and correct. No change required.

- **`idbutil/src/lru-blob-cache.ts:238`** — In `deleteEntry`, when the meta-store lookup finds no entry for the key, the `onsuccess` callback returns early, leaving the transaction to commit with no write ops. The promise still resolves via `tx.oncomplete`. This is correct and safe; noted for future readers.

- **`idbutil/src/lru-blob-cache.ts:234`** — `existingReq.onsuccess` has no `existingReq.onerror` handler. IDB request errors bubble to the transaction error handler, so `tx.onerror → reject(tx.error)` still fires. No issue.

### False-positive (security)

- **`audio/src/player.ts:106`** (red-team) — `onError` reads `queue[currentIndex]` without a generation guard. The guard against stale invocations is `currentIndex < 0`, which is set by `stop()` on teardown. No reachable path exists where `currentIndex` is positive but the queue slot is mismatched at the time `onError` fires synchronously. Not a real issue.

- **`audio/src/player.ts:109-113`** (data-exposure) — `storagePath` and `MediaError.code` are written to `logError`. `storagePath` is a Firebase Storage path (no PII), and `MediaError.code` is a numeric constant from the HTML spec. No sensitive data exposure.

- **`audio/src/player.ts:114-116`** (data-exposure) — `storagePath` in the cache-evict `logError` catch handler. Same as above — no PII.

### Dismissed

- **`audio/src/storage.ts:5`** — `removeFile` is imported into `storage.ts` solely to be re-exported, creating an indirect import path from `player.ts`. This is by design: `player.ts` imports only from `./storage.js` (its single storage facade). Keeping the facade pattern is intentional and correct.

PR #1422, opened 2026-06-12
ci: dedupe prod-deploy trigger, pin diff base to the merge range, add a concurrency group

natb1:
Closes #1306

## Problem

`.github/workflows/prod-deploy.yml` had two defects that made production deploys
race and silently skip apps:

1. **Double trigger.** A merged PR touching app code fired **both** a `push` to
   `main` and a `pull_request: closed` (merged) event. `deploy-and-smoke`'s `if`
   admitted both, so two identical deploy+smoke runs raced — double Firebase
   Hosting deploys, double Cloudflare purges, interleaved smoke tests against the
   same prod URLs.

2. **Nondeterministic diff base.** The job checked out `ref: main` (the branch
   tip *at run time*) and the smoke script diffed `--base HEAD~1`. When two merges
   landed close together, both runs checked out `main` at the *later* tip and both
   computed the *later* merge's diff — so apps changed only by the earlier merge
   were never deployed. This repo merges in tight bursts, so this was a live
   failure mode.

## Fix

- **Drop `pull_request` from `deploy-and-smoke`'s `if`** so it runs once per merge
  (on `push`). The top-level `on: pull_request: [closed]` trigger stays —
  `cleanup-preview` still keys off `github.event.pull_request.merged`.
- **Check out `github.sha`** (the merge commit) instead of the live `main` tip, and
  pass `DIFF_BASE=github.event.before` to the smoke script, which now diffs
  `before...sha`. Each run is pinned to *its own* merge range, independent of
  execution order. `get-changed-apps.sh --base <ref>` already runs
  `git diff --name-only "$BASE"...HEAD`; after the `github.sha` checkout
  `HEAD == github.sha`, so this yields exactly `github.event.before...github.sha`
  — no change to `get-changed-apps.sh`.
- **Add a `concurrency` group `prod-deploy`** (no cancel) so successive deploy runs
  don't execute simultaneously.

The smoke script's `else` branch now guards `DIFF_BASE` with a `:?` expansion
(clear error over a fallback, per `.claude/rules/code-style.md`) — matching the
idiom already in `run-prod-deploy.sh` and `cloudflare-purge.sh`.

## Known limitation — `concurrency` with no-cancel

GitHub Actions `concurrency` with `cancel-in-progress: false` does **not** queue
every pending run. The effective pending-queue depth is **1**: when a newer run
enters a group that already has one running plus one pending, the **older pending
run is superseded (cancelled)**. So under 3+ rapid successive merges (this repo's
normal burst pattern), an intermediate merge's deploy run can be dropped:

```
merge1 runs → merge2 pending → merge3 arrives → merge2 dropped → merge3 runs
```

The thing that actually guarantees per-merge correctness is the **diff-base fix**
(`before...sha`), not the concurrency group; the group only prevents *interleaved*
deploys/purges. This trade-off (no interleaving, at the cost of a possibly-dropped
intermediate deploy under bursts) is accepted because the issue prescribes exactly
`concurrency: { group: prod-deploy }` (no cancel). A dropped intermediate deploy is
recoverable by re-running the workflow via `workflow_dispatch`, which sets
`DEPLOY_ALL=1` and deploys every app.

## Deviation — job-level (not workflow-level) concurrency

The acceptance criterion reads as a workflow-level `concurrency`, but this PR adds
it at **job level** on `deploy-and-smoke` only. Workflow-level would place
`cleanup-preview` (a `pull_request` run) in the same `prod-deploy` group, where
under bursts a pending `cleanup-preview` could be superseded and dropped by an
incoming push run. Job-level scopes the group to `deploy-and-smoke` — exactly the
job whose races (double hosting deploys, double Cloudflare purges, interleaved
smoke) the issue names — and leaves `cleanup-preview` free to run concurrently.
`cancel-in-progress: false` is written explicitly so "no cancel" is
self-documenting.

## Scope boundary — `cloudflare-purge.sh` left unchanged

`.github/scripts/cloudflare-purge.sh` independently diffs
`git diff --name-only HEAD~1 HEAD` to pick non-hashed URLs to purge. It runs inside
`run-prod-deploy.sh`, downstream of the smoke script, and is not in this issue's
named scope. The checkout fix already removes its nondeterminism for the normal
case: a merge commit's first parent is the prior `main` tip, so after checking out
`github.sha`, `HEAD~1` equals `github.event.before` for both squash-merges and
`--no-ff` merge commits. The two diverge only on a direct multi-commit push to
`main` (rare on a protected branch), where `cloudflare-purge.sh` would see only the
tip commit's non-hashed files. This residual is left unchanged.

## Verification

CI-workflow change that can't fully run locally (it needs the `push`/`pull_request`
event payloads). Verified in layers:

- `prod-deploy.yml`: `deploy-and-smoke.if` no longer mentions `pull_request`;
  checkout uses `ref: ${{ github.sha }}`; `DIFF_BASE` is in the deploy step env;
  job-level `concurrency.group: prod-deploy` with `cancel-in-progress: false`;
  `cleanup-preview` untouched (still `ref: main`).
- Smoke script: the `DIFF_BASE` guard exits non-zero with a clear message when the
  var is absent in the push branch; with a real base, `get-changed-apps.sh --base`
  resolves changed apps as before.
- End-to-end is observable post-merge: exactly **one**
  `Production Deploy / deploy-and-smoke` run should appear for the merge, checking
  out the merge SHA and deploying only that merge's changed apps.


natb1:
## QA Summary — Issue #1306

**PR**: #1422 · ci: dedupe prod-deploy trigger, pin diff base to the merge range, add a concurrency group

**Path**: Non-browser (CI workflow + script changes only)

### Results

| # | Item | Result |
|---|------|--------|
| 1 | YAML validity (`prod-deploy.yml` parses) | ✅ PASS |
| 2 | `deploy-and-smoke.if` drops `pull_request` clause | ✅ PASS |
| 3 | Checkout `github.sha` + `fetch-depth: 0` | ✅ PASS |
| 4 | `DIFF_BASE: ${{ github.event.before }}` in deploy step env | ✅ PASS |
| 5 | Job-level `concurrency: { group: prod-deploy, cancel-in-progress: false }` | ✅ PASS |
| 6 | `cleanup-preview` unchanged (`ref: main`, original `if:`) | ✅ PASS |
| 7 | Smoke script exits non-zero with clear message on missing `DIFF_BASE` | ✅ PASS |
| 8 | `get-changed-apps.sh --base <ref>` works with explicit base | ✅ PASS |
| 9 | `shellcheck` on smoke script | ⏭️ SKIP (not installed) |

**8 PASS · 0 FAIL · 1 SKIP · 0 needs-human-judgment**

### Notes

- `cleanup-preview` retains `ref: main` — it cleans up by PR number and doesn't diff, so the SHA pinning doesn't apply.
- Concurrency is job-level (not workflow-level), scoping the group to `deploy-and-smoke` only so `cleanup-preview` runs freely.
- End-to-end verification (exactly one `deploy-and-smoke` run per merge) is only observable post-merge via the Actions tab.

No bugs found. No judgment-call items. Clean autonomous pass.

natb1:
## Review — PR #1422 (`1306-ci-prod-deploy-fires-twice`)

No code changes applied (no Required security findings; no Fixed code-review findings).

---

### Informational

- **`.claude/skills/dispatch-propagate/scripts/run-all-prod-deploy-smoke.sh:17`** — The `:?` guard on `DIFF_BASE` is bypassed when `DEPLOY_ALL=1` short-circuits the `else`-branch for `workflow_dispatch` events. This is intentional: `workflow_dispatch` deploys all apps unconditionally and does not use `DIFF_BASE`.

- **`.claude/skills/dispatch-propagate/scripts/run-all-prod-deploy-smoke.sh:17`** — On the very first push to `main` (repository initialization), GitHub sets `github.event.before` to the null SHA `0000000000000000000000000000000000000000`. The `:?` guard would pass (the variable is non-empty), and `git diff --name-only 0000...HEAD` would list all files — `DEPLOY_ALL=1` via `workflow_dispatch` is the recovery path. Acceptable for a protected `main` branch.

---

### False-positive (security)

- **`.github/workflows/prod-deploy.yml:55`** — `${{ github.event.before }}` expanded into `DIFF_BASE` env var: not an injection vector; GitHub Actions env vars set via `env:` are not re-expanded in shell command strings.

- **`.github/workflows/prod-deploy.yml:27-29`** — Concurrency run-order concerns (3 entries): no cancelation of in-progress runs; `workflow_dispatch` vs push competition; `pull_request` trigger retained for `cleanup-preview`. All are either correct by design or non-issues.

- **`.claude/skills/dispatch-propagate/scripts/run-all-prod-deploy-smoke.sh:17-18`** — `DIFF_BASE` injection via `get-changed-apps.sh --base`: `DIFF_BASE` is a git ref, validated by git itself; not a shell injection surface.

---

### Deferred

- **`.github/workflows/prod-deploy.yml:27-29`** — Concurrency group drops intermediate merges in burst scenarios: in a 3-merge burst, B's pending run is evicted by C, and `B.sha...C.sha` doesn't cover B's changes. Fix: anchor `DIFF_BASE` to a `last-prod-deploy` tag rather than `github.event.before`. Filed as #1446 (blocked by #1306).

---

_No security findings. No dependency or CodeQL alerts._

PR #1423, opened 2026-06-12
audio: scope home cache listener, drop dead stopPropagation, clear cache on sign-out

natb1:
Fixes three lower-severity audio bugs in the Home view and its cache layer.

## 1. Home-view listener outlives the view (false-positive telemetry)

`afterRenderHome` registers a `document`-level `CACHE_UPDATED_EVENT` listener
that is aborted only at the *next* `afterRenderHome`. Navigating Home→About
never re-runs `afterRenderHome` (About has no `afterRender`), so the document
listener stays live. A fire-and-forget cache write from a still-playing track
then dispatches `CACHE_UPDATED_EVENT`, the stale listener calls
`refreshCacheStats(outlet)`, and `outlet.querySelector("#cache-stats")` returns
`null` because About is now mounted — logging a false-positive
`"#cache-stats element not found in outlet"` error on a normal flow.

**Fix:** scope the listener with the codebase's staleness-guard pattern
(`isOutletCurrent` from `@commons-systems/router/hydrate`, as used by
blog/landing/fellspiral). The callback captures `#cache-info` at mount and
returns early when the outlet has been re-rendered by the router. The existing
`logError` is kept — with the guard, a missing `#cache-stats` is now reachable
only when Home *is* mounted, so it is a genuine programmer error worth logging.
A new regression test in `audio/test/pages/home.test.ts` asserts the error is
not logged when the event fires after navigation away.

We use the staleness guard rather than a router per-route unmount hook because
the router has no such hook and the codebase has standardized on this pattern;
the acceptance criterion explicitly accepts "guard before logging the
missing-element error."

## 2. Dead `stopPropagation` removed

The delegated click handler's `e.stopPropagation()` (with a comment claiming it
prevents the parent `<details>` from toggling) is dead: `stopPropagation` does
not cancel a default action (the summary's toggle is a default action, cancelled
only by `preventDefault`), and the listener is delegated on `outlet` — an
ancestor of the summary — so the event has already passed the summary by the
time it runs. The only ancestor listener it could affect is the router's
`document`-level click handler, which no-ops on non-anchor targets. Removal is a
true behavioral no-op; the misleading comment is removed with it.

## 3. Private audio cleared on sign-out

Nothing cleared the IndexedDB audio cache on sign-out, so a member-gated track's
bytes persisted on the device until the user clicked "Clear audio cache" —
leaking private audio to the next user on a shared device.

**Fix:** `main.ts`'s `onAuthStateChanged` now clears the cache on a real
signed-in→signed-out transition. The clear is awaited *before* `router.navigate()`
so the re-rendered Home reads the already-cleared cache; failures are logged, not
swallowed. Public-domain bytes cached in the session are cleared too (they
re-fetch — privacy outweighs the cache-hit cost).

**Scope boundary (decided, not missed):** this clears the *persisted IndexedDB
cache only*. An already-playing member track keeps its in-memory blob URL (built
via `new Blob([...])`, independent of IndexedDB) and the player queue retains its
track metadata. Stopping in-session playback / flushing the queue is intentionally
out of scope for this issue.

## Verification

- `npx tsc --noEmit --project audio` — clean.
- `npx vitest run --root audio` — 98 passed, 1 skipped (includes the new
  regression test).
- Manual browser QA (bug 1 listener suppression, bug 2 checkbox toggle
  behavior, bug 3 cache cleared on sign-out) is covered in the issue's
  verification plan.


natb1:
## QA Summary — #1305 audio: home-view listeners / dead stopPropagation / sign-out cache clear

**PR:** #1423  
**Path:** Non-browser (audio package — unit tests + source inspection)

### Items walked

| # | Item | Result |
|---|------|--------|
| 1 | Staleness guard: `CACHE_UPDATED_EVENT` suppressed after Home→About nav | ✅ PASS |
| 2 | Dead `stopPropagation` and misleading comment removed | ✅ PASS |
| 3 | Sign-out clears IndexedDB audio cache (`clearCache` called on signed-in→signed-out only) | ✅ PASS |
| 4 | Type-check: `npx tsc --noEmit --project audio` | ✅ PASS |
| 5 | Full unit suite green (all existing tests unaffected) | ✅ PASS (98 passed, 1 skipped) |

**5 PASS · 0 FAIL · 0 SKIP**

### Details

**Item 1 — Staleness guard:**  
Unit test suite ran `npx vitest run --root audio` — 7 test files, 98 tests pass. The new regression test (verifies `mockLogError` is not called when `CACHE_UPDATED_EVENT` fires on a stale outlet) is included in this green run.

**Item 2 — Dead stopPropagation:**  
`grep stopPropagation audio/src/pages/home.ts` → not found.  
`grep "prevents the click from toggling" audio/src/pages/home.ts` → not found.

**Item 3 — Sign-out cache clear:**  
`audio/src/main.ts:66-76` — `clearCache()` is imported, called only when `wasSignedIn && user === null`, awaited before `router.navigate()`, errors passed to `logError`. Scope boundary (in-session playback / player queue not cleared) noted in PR body.

**Item 4 — Type-check:**  
`npx tsc --noEmit --project audio` → zero errors.

### Deferred to office-hours

### Bugs found

None.

natb1:
## Review pass — #1305

### Fixed

**`code-review-0` · `audio/src/pages/home.ts:154`** (code-review, review)
Staleness guard `if (cacheInfo && !isOutletCurrent(...))` silently skips the guard when `cacheInfo` is null — an outlet with no `#cache-info` section would bypass the early-return and attempt `refreshCacheStats` anyway. Changed to `if (!cacheInfo || !isOutletCurrent(outlet, cacheInfo)) return;` so both the null-anchor and detached-anchor cases return early, consistent with the pattern in `blog/src/pages/home.ts`.

**`code-review-3` + `review-2` · `audio/test/pages/home.test.ts:317–334`** (code-review, review)
New staleness-guard test used a mid-test `mockLogError.mockClear()` to wipe errors from prior `afterRenderHome` calls, making it order-sensitive and prone to masking failures in later tests. Added `beforeEach(vi.clearAllMocks)` to the `afterRenderHome` describe block (matching the pattern in `renderHome`), removed the mid-test clear, and extended coverage with two additional tests: one asserting the null-anchor guard path (no `#cache-info` in outlet), one asserting that a second `afterRenderHome` call aborts the previous document listener before registering a new one.

Fix commit: 025f3423

---

### Informational

**`code-review-1` · `audio/src/main.ts:71`** (code-review)
`router.navigate()` is delayed until after `await clearCache()` completes. During the IndexedDB transaction, the nav bar already shows the signed-out state (`navEl.user = user` runs before the await) but the private library DOM remains visible. This is an acceptable UX trade-off — the cache must be cleared before rendering a fresh Home view — and aligns with the PR's stated design rationale.

**`code-review-2` · `audio/src/pages/home.ts:117`** (code-review)
`e.stopPropagation()` removal is confirmed correct. The comment ("prevents the parent `<details>` from toggling") was wrong on two counts: `stopPropagation` cannot cancel a default action, and the listener is on `outlet` — an ancestor of `<summary>` — so the event has already propagated past the summary. The removal is a behavioral no-op and the misleading comment is gone.

**`review-0` · `audio/src/main.ts:76`** (review)
Same async-window observation as `code-review-1` — the private library DOM is visible for the duration of the IndexedDB clear. Informational only; matches PR's documented scope boundary.

---

### Refuted (adversarial verify dropped)

**`red-team-1` · `audio/src/main.ts:72-74`** (firebase, red-team) — *Required → Refuted, not fixed*
Claimed: silently-swallowed `clearCache()` failure allows private blobs to persist, inducible by an attacker filling IndexedDB quota.

Skeptics refuted on two independent grounds:
1. **Attack vector is technically impossible.** `clearCache()` calls `IDBObjectStore.clear()` — a delete operation that frees storage. `QuotaExceededError` is only thrown by storage-growing writes (`putEntry`). The named trigger mechanism does not hold; there is no attacker-inducible path to cause this code path to fail.
2. **No in-app disclosure primitive after sign-out.** The post-signout view calls only `getPublicMedia()`; private storagePaths are never surfaced in the UI. Reaching residual blobs requires DevTools access, which presupposes local interactive access to the browser profile — generic local-device access, not an application-layer exploit.

Finding remains accurately described as a robustness/UX gap (a failed clear is possible under spontaneous IDB faults), but does not meet the bar for a Required security deviation. Not fixed.

---

### False-positive (security)

**`red-team-2` · `audio/src/main.ts:65-77`** (red-team)
Claimed: async window between `navEl.user = user` (immediate) and `await clearCache()` completing creates a UI state where the user appears signed out but private content is still reachable. The signed-out re-render is blocked by `router.navigate()` after the await, so the concern is the nav bar showing signed-out state while the Home DOM still shows private tracks — this is the intentional ordering in the PR and the window is sub-second under normal IDB operation. No attacker-exploitable primitive exists here; it is a cosmetic flicker, not an information disclosure.

---

### Out-of-scope (security)

**`red-team-0` · `audio/src/main.ts:65-77`** (red-team)
Sign-out cache-clear guard fires only when the tab is live and observes the auth-state transition. A tab closed before sign-out (or a tab restored after sign-out) does not run `clearCache()`. Pre-existing design; not touched by this diff.

**`data-exposure-0` · `audio/src/main.ts:73`** (data-exposure)
`logError(err, { operation: 'signout-clear-cache' })` passes the raw IndexedDB error to the Firestore error sink, which may durably write the error object's message. Not introduced by this diff's intent; the error logging pattern is used throughout the codebase.

**`data-exposure-1` · `audio/src/main.ts:72-74`** (data-exposure)
Catch block logs unconditionally regardless of error kind (e.g. `QuotaExceededError`, `SecurityError`). Pre-existing pattern; not introduced or worsened by this diff.

No deferred follow-up issues filed (no in-scope deferred findings). No security follow-up issues filed (out-of-scope findings below the meaningfulness threshold for independent tracking).

PR #1431, opened 2026-06-12
scaffolding: realign the app template deps, errors rule, and install hint with real apps

natb1:
Closes #1308

## Summary

The Firebase app scaffolder (`scaffolding/firebase/`) generates a new app
package from templates. Its `package.json.tmpl` and the `create.go` post-render
output had drifted from every real app, so a freshly scaffolded app was subtly
broken. This PR realigns the template, the inserted Firestore rules, and the
printed next step with the conventions every real app already follows, and adds
a Go drift test that fails if the template diverges from the root again.

## Changes

**Template `package.json.tmpl` (Unit 1)**
- Every `@commons-systems/*` specifier changed from `file:../<pkg>` to `"*"`
  (npm-workspace resolution, as all real apps use).
- Added `"@commons-systems/errorutil": "*"` — every real app depends on it.
- Deleted the per-app `devDependencies` block. The shared toolchain
  (`@playwright/test`, `eslint`, `vite`, `vitest`, `happy-dom`, `typescript`,
  …) stays hoisted to the root `package.json`. Its pins were stale
  (`@playwright/test 1.57.0` vs root `1.59.1`, `vitest ^3.0.5` vs `^4.1.8`,
  `happy-dom ^17.4.4` vs `^20.8.9`) — a scaffolded app re-introducing a second,
  older `@playwright/test` is exactly the failure the `playwright-version-sync`
  job guards against. The dropped `@commons-systems/config` is not part of the
  default app baseline.

**`create.go` (Unit 2)**
- `InsertFirestoreRules` now inserts a `/<app>/{env}/errors/{errorId}` rule
  block (create-only, env-pinned, schema-bounded — identical to every real
  app), referencing the global `isKnownErrorEnv`/`isValidErrorLog` helpers. A
  scaffolded app adopting the error-telemetry pattern would otherwise hit
  permission-denied on error writes.
- The printed next step changed from `cd <app> && npm install` (documented in
  `.claude/rules/build.md` as failing with E404 in this workspace — per-package
  installs cannot resolve the unpublished `@commons-systems/*` deps) to a
  repo-root `npm install`.

**Drift test (Unit 3)**
- New `template_deps_test.go` reads the real template and root `package.json`
  and asserts: shared-dep versions are byte-equal (the firebase-drift guard —
  the root set is read as the union of root `dependencies` + `devDependencies`,
  since the root has no `dependencies` key), the template carries no
  `devDependencies`, every `@commons-systems/*` dep is `"*"`, and `errorutil`
  is present. Proven non-vacuous: bumping the template's `firebase` turns the
  shared-version assertion red.

## Verification

`cd scaffolding/firebase && go test ./...` passes, including the brittle
byte-exact round-trip `TestInsertThenRemoveFirestoreRules` (the new errors
block round-trips cleanly) and the new drift test. No app build, emulator, or
browser QA is needed — this change only affects generated-app scaffolding
inputs and a Go unit test; it ships no runtime app code.


natb1:
## QA Summary — Issue #1308

**PR #1431** — scaffolding: realign the app template deps, errors rule, and install hint with real apps

No browser component detected (all changes in `scaffolding/firebase/` — Go source and template only). All items verified via shell commands.

### Results: 7 PASS / 0 FAIL / 0 SKIP

| # | Item | Result |
|---|------|--------|
| 1 | Full scaffolding Go test suite | PASS |
| 2 | Drift test `TestTemplateDepsMatchRoot` | PASS |
| 3 | Rules insertion test `TestInsertFirestoreRules` (includes errors block) | PASS |
| 4 | Template `package.json.tmpl` is valid JSON | PASS |
| 5 | No stale `file:../` specifiers remain in template | PASS |
| 6 | Errors rule helper functions (`isKnownErrorEnv`, `isValidErrorLog`) defined in `firestore.rules` | PASS |
| 7 | Install hint uses root-level `npm install` | PASS |

### Deferred to office-hours

None.

### Bugs found

None.

natb1:
## Review findings — PR #1431 (`1308-scaffolding-the-app-templat`)

### Fixed

- **`template_deps_test.go:48` — assertion numbering mismatch** (`code-review`): Renumbered in-code assertion labels from (2)–(5) to (1)–(4) to match the four doc-comment bullets in `TestTemplateDepsMatchRoot`, including the (2)→(1) reference in the `rootUnion` comment. Fix commit: ef8c91e3.

---

### Informational

- **`create.go:218` — generated rule references repo-specific helpers** (`code-review`): `InsertFirestoreRules` emits `allow create: if isKnownErrorEnv(env) && isValidErrorLog();`, which depends on two helper functions that must exist in the target app's Firestore rules. Scaffolded apps are expected to include the global rules that define these helpers — this is a known coupling, not a defect.

- **`template_deps_test.go:50` — `missing.css` not covered by shared-dep version check** (`code-review`): The shared-dep version assertion only catches packages present in both the template's `dependencies` and the root union. `missing.css` lives in the template but not in the root union, so a version drift there would go undetected by this test. Acceptable scope limitation; the test comments document this.

- **`template_deps_test.go:48–66` — shared-dep assertion silently passes for template-only packages** (`review`): Same as above — packages in the template but not the root union are not version-checked. The `sharedSeen == 0` guard prevents vacuous success, but the broader class of template-only packages is out of scope for this test.

- **`rules_test.go:58–60` — new assertion does not cover round-trip removal** (`review`): `TestInsertFirestoreRules` now checks that the errors rule block is inserted. The round-trip test (`TestInsertThenRemoveFirestoreRules`) tests that insert+remove returns the original file, which indirectly covers the new block. The scope of the new assertion is intentionally narrow.

---

### Dismissed

- **`template_deps_test.go:26` — five nested `filepath.Dir()` calls for repo root** (`review`): The comment accurately explains the path: strip the filename, then climb four parent directories. The call chain matches the filesystem layout. _No bug; dismissed._

---

### False-positive (security)

- **`template_deps_test.go:26` — repo root resolution via `runtime.Caller(0)`** (`red-team`): The concern was that the path could be spoofed. `runtime.Caller(0)` returns the source file path baked in at compile time — it is not influenced by test input or environment variables. _Not a real risk._

- **`rules_test.go:28–38` — test fixture has no `isKnownErrorEnv`/`isValidErrorLog`** (`auth`): The test fixture is minimal by design; `InsertFirestoreRules` only inserts the rule text string — it does not validate that helpers exist in the surrounding rules file. The helper existence check is a deploy-time concern, not a unit-test concern. _Not a defect._

- **`package.json.tmpl:17` — template declares `errorutil` without verifying helper alignment** (`auth`): The `errorutil` dep and the error Firestore rule are introduced together, matching every real app's baseline. _Not a security issue._

- **`template_deps_test.go:99–106` — `t.Fatalf` logs absolute filesystem paths** (`data-exposure`): Test helper paths in test output are a development-time convenience. Test binaries do not run in production. _Not a meaningful exposure._

---

### Out-of-scope (security)

These findings are real design considerations in the existing app pattern, not issues introduced by this PR's diff:

- **Unauthenticated error-log creates / no rate limiting** (`red-team`, `data-exposure`): The scaffolded errors rule has no `request.auth != null` guard by design — error telemetry is written before auth state is confirmed. Rate limiting and per-document ID structure are governed by the existing `isValidErrorLog()` implementation, which is pre-existing. Not introduced by this diff.

- **`isKnownErrorEnv`/`isValidErrorLog` must exist as global helpers** (`red-team`, `auth`): These functions are required to be present in the app's global Firestore rules. The scaffolder emits a rule that references them; the helper definitions are authored once, repo-wide. Pre-existing pattern, not introduced here.

- **`*` wildcard specifiers and workspace resolution** (`red-team`): Changing from `file:../pkg` to `"*"` uses npm workspace resolution, which resolves to the local workspace package. This is the established pattern for every real app and is not a supply-chain regression.

- **`errorId` fully client-controlled** (`red-team`): Document IDs are client-chosen in all Firestore patterns. Access control is enforced by the `allow create` predicate, not by ID structure. Pre-existing design.

- **`extras` field accepts large opaque strings** (`data-exposure`): Documented in `firebaseutil`. Pre-existing behavior, not changed by this PR.

_No follow-up issues filed (no CodeQL alerts; no qualifying npm advisories; no deferred code-review findings)._

PR #1448, opened 2026-06-13
dispatch: CI-cadence reseed for CI-pending priority work at target=0

natb1:
Closes #1444

## Problem

When dispatch is above the weekly usage curve, `dispatch-target-workers` returns
`0` (the pace gate). At `target=0`, a **priority** issue whose phase just finished
fails to drain promptly: it stalls until the pace window reopens (up to ~an hour)
instead of resuming the moment its CI concludes. Priority work is meant to bypass
pacing entirely, but today it only bypasses the *spawn* gate, not the *wakeup
timing*.

Root cause: at the at-cap branch, `dispatch-select-tick` probes
`dispatch-select-target --priority-only`. A lone priority PR whose CI is still
pending is skipped by the readiness gate (`dispatch-ci-ready` → not ready), so
`--priority-only` returns `empty`. `empty` routes to the **pace** reseed
(`dispatch-schedule-reseed`), which arms the wakeup at the pace-curve reopen time,
not at CI cadence. The defect is that `--priority-only` conflated two states behind
one `empty`: "no priority work exists" (deserves the pace reseed) vs. "priority
work exists but is CI-pending" (needs a CI-cadence wakeup).

The fix is scoped to `target=0`. At `target>0` the at-cap state has busy workers
that re-probe priority on every baton-pass, so the pace reseed is fine —
non-priority work proceeds while priority CI runs.

## Change

**Unit 1 — `dispatch-select-target`:** split the `--priority-only` `empty` output
into `empty` vs `waiting <N>`. A CI-pending priority PR that is skipped *solely* by
the readiness gate — not office-hours-parked, not blocked, not excluded, not
claimed by a live worktree — is now reported as `waiting <N>`. The office-hours and
blocked checks were factored into two helpers (`pr_issue_office_hours_parked`,
`pr_blocked_by_open`) so the waiting branch and the existing skip branches share one
definition; default/`--qa` mode behavior is byte-identical.

**Unit 2 — `dispatch-select-tick`:** the at-cap probe gains a `waiting <N>` case
arm. At `target=0` it arms a CI-cadence reseed via the existing
`dispatch-schedule-target-reseed <N>` machinery (re-runs `dispatch-tick <N>` on a
short delay; CI concluded → spawns gate-exempt; still pending → re-arms, bounded by
the ci-wait attempt counter → office-hours escalation). At `target>0` it falls
through to the existing pace reseed. The lock is released in both branches, exactly
as the `empty` arm does today; no new lock-holding sleep is introduced.

## Tests

Added to `test-dispatch-scripts.sh` (suite stays green at 2286/2286):

- `--priority-only` emits `waiting <N>` for a lone CI-pending priority PR; `empty`
  when no priority item exists; `empty` when the CI-pending item is also
  office-hours-parked or blocked; a selectable priority item is never preempted by
  `waiting`; and a claimed CI-pending priority PR stays `empty`.
- `dispatch-select-tick` at `target=0` on `waiting <N>` calls
  `dispatch-schedule-target-reseed <N>` (asserting the correct issue number `<N>`
  propagates) and not the pace reseed; at `target>0` it falls through to the pace
  reseed.

## Out of scope

The plain `empty` path at `target>0` under cap with an empty queue releases the lock
with no reseed, so a CI-pending-priority-only queue with no other work stalls until
the #725 heartbeat. Separate finding; not addressed here.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -735,8 +735,23 @@ while IFS=$'\t' read -r pr_num _created pr_branch pr_closing; do
     # CI is the sole reason this item is unselectable. PR_SORTED is oldest-first,
     # so the first recorded candidate is the oldest.
     if [[ "$PRIORITY_ONLY" == "true" && -z "$WAITING_NUM" ]]; then
-      local_issue="${pr_branch%%-*}"
-      if [[ -n "$local_issue" && -z "${EXCLUDED[$local_issue]:-}" ]] \
+      # The reseed target must be the canonical priority issue, NOT the PR
+      # branch prefix (#1444). A branch may diverge from its closing-issue
+      # number (e.g. `100-fix` closing priority issue `200`), and a branch with
+      # no `<N>-` prefix yields a non-numeric value the reseed sink rejects —
+      # silently stranding the very priority item this path exists to rescue.
+      # Resolve the priority-labeled closing issue from `pr_closing` (already in
+      # scope), reusing ISSUE_LABELS_JSON exactly as pr_combined_labels does. The
+      # line-712 gate guarantees at least one closing issue carries `priority` in
+      # --priority-only mode, and dispatch-find-pr resolves issue→PR via the
+      # issue's closedByPullRequestsReferences fallback, so the priority issue
+      # number is the resolvable, correct target.
+      local_issue=$(printf '%s' "$pr_closing" \
+        | jq -r --argjson map "$ISSUE_LABELS_JSON" \
+          'map(.number | tostring
+               | select((($map[.] // []) | any(.name == "priority"))))
+           | .[0] // empty')
+      if [[ "$local_issue" =~ ^[0-9]+$ && -z "${EXCLUDED[$local_issue]:-}" ]] \
          && ! pr_issue_office_hours_parked "$local_issue" \
          && ! pr_blocked_by_open "$pr_closing"; then
         WAITING_NUM="$local_issue"
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -72,6 +72,8 @@
 #   main-broken <sha>           — origin/main's HEAD CI has a failing check
 #   pr <num> <branch> <phase>   — the highest-priority priority=1 PR
 #   issue <num>                 — the highest-priority priority=1 help-wanted issue
+#   waiting <issue#>            — a priority item exists but its draft PR's CI is
+#                                 still pending (skipped solely by dispatch-ci-ready)
 #   empty                       — no priority=1 item eligible
 #
 # Selection is three-tier: the `priority` label is the *outermost* axis, topic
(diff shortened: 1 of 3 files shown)


natb1:
# QA Summary — Issue #1444 / PR #1448

**Change:** `dispatch-select-target` splits `--priority-only` `empty` into `empty` vs `waiting <N>` for CI-pending priority items; `dispatch-select-tick` arms a CI-cadence reseed at `target=0` when the probe returns `waiting <N>`.

**Browser component:** none (pure shell-script change — no server started, no browser used)

## Results

| # | Item | Result |
|---|------|--------|
| 1 | Test suite (2286 tests including new AC1–AC5 + guard cases) | ✓ PASS |
| 2 | AC1: `--priority-only` emits `waiting <N>` for CI-pending priority PR | ✓ PASS |
| 3 | AC2: `--priority-only` emits `empty` when no priority item exists | ✓ PASS |
| 4 | AC3: CI-pending + office-hours-parked or blocked → `empty` (not `waiting`) | ✓ PASS |
| 5 | AC4: `dispatch-select-tick` at `target=0` on `waiting <N>` → CI-cadence reseed, lock released, `concurrency-cap` | ✓ PASS |
| 6 | AC5: `dispatch-select-tick` at `target>0` at cap on `waiting` → pace reseed (not target reseed) | ✓ PASS |
| 7 | Unexpected probe output surfaces a clear error (catch-all arm confirmed in script) | ✓ PASS |
| 8 | `waiting <N>` output is a single parseable line; `WAIT_NUM="${PRIO#waiting }"` yields a bare integer | ✓ PASS |

**Totals:** 8 PASS / 0 FAIL / 0 SKIP

## Deferred to office-hours

None.

## Bugs found

None.

natb1:
## Review — #1444 dispatch: priority work fails to drain at CI-pending target=0

### Fixed

- **`dispatch-select-target:738` — branch-prefix `local_issue` used as reseed target** (`code-review`, `red-team`)
  The `waiting` detection block derived the reseed target from `${pr_branch%%-*}` (the PR branch prefix). A branch whose name diverges from its closing-issue number (e.g. `100-fix` closing priority issue `200`) silently reseeds the wrong issue, and a branch without a numeric prefix yields a value that `dispatch-schedule-target-reseed` rejects. Fixed to resolve the priority-labeled closing issue from `pr_closing` via `ISSUE_LABELS_JSON` (mirroring how `pr_combined_labels`/`has_priority_in_labels` already use it), with a `^[0-9]+$` guard that skips the capture on any non-numeric result. Fix commit: `7b360d1b`.

### Informational

- **`dispatch-select-target:738` — `local_issue` assigned without `declare`/`local`** (`review`)
  The variable is assigned inside the `while` loop body (not inside a function), so `declare`/`local` is not applicable here. The assignment is loop-scoped by the surrounding `while` construct and the variable is consumed only within the block. No change needed; noted for awareness.

### False-positive

- **`dispatch-select-tick:268` — `WAIT_NUM` derived from `dispatch-select-target` stdout** (`red-team`)
  The concern was that multiline stdout could corrupt `WAIT_NUM="${PRIO#waiting }"`. `dispatch-select-target` emits exactly one line on the `waiting` path (the single `echo "waiting $WAITING_NUM"` at line ~972), so the strip is correct and the value is already proven to be a single numeric token by the `^[0-9]+$` guard now applied upstream. Dismissed as not an actual risk.

### Dismissed

- **`dispatch-select-target:739` — `EXCLUDED[$local_issue]` key was the branch prefix** (`red-team`)
  `EXCLUDED` is keyed by issue number throughout the loop. With the fix applied in the Fixed section above, `local_issue` is now the canonical priority issue number resolved from `pr_closing`, so the `EXCLUDED` lookup is consistent with every other site. This finding is resolved as a consequence of the Fixed item and is no longer applicable.

---

_No deferred filings. No security follow-ups. No out-of-scope advisories. CI passing._

Issue #920, opened 2026-05-29
dispatch: dispatch-select-target issue queue re-selects an open issue whose PR is already ready


natb1:
## Context / motivation

`dispatch-select-target`'s issue-queue path
(`.claude/skills/dispatch-propagate/scripts/dispatch-select-target`, ~491–527)
selects any open `help wanted` issue, gated only by `dispatch:office-hours` and
worktree-liveness (`issue_has_live_worktree`). It never checks whether the issue
already has a PR. The PR ladder drops a non-draft ("done") PR entirely, but the
underlying issue stays open + `help wanted` until the PR merges and GitHub
auto-closes it. In the window between "PR flipped to ready (security phase's
terminal action)" and "PR merged / issue closed," the dispatch chain has
terminated — no worker runs in the worktree — so `issue_has_live_worktree`
returns false and the issue is freely re-selectable. A router then selects it,
spawns a worker, which derives phase `done` and stops, wasting a tick.

## Reproduction (observed live)

Issue #896 was selected by a background `/dispatch` router while its PR #898 was
already non-draft (ready). PR #898 merged at 2026-05-29T22:59:47Z; #896
auto-closed at 22:59:48Z. The worker spawned into the #896 worktree derived
phase `done` and stopped. Per-worktree dedup at `dispatch-spawn-worker` did not
catch it — the chain had ended, so there was no competing worker to dedup
against.

- **Expected:** an open issue whose PR is already ready (non-draft) is not
  eligible for the issue queue.
- **Actual:** it remains eligible until the PR merges and the issue closes.

## Root cause

The issue-queue loop has no PR-status gate; protection relies entirely on
worktree-liveness, which has a gap once the chain reaches a ready PR (no live
session, issue still open). The ladder makes the exposure broader than just the
ready case: within a bucket `FIRST_ISSUE` ranks above a `qa`-phase PR, and
`waiting`/`done` PRs are dropped from the ladder, so for `qa`/`waiting`/`done`
phases only liveness timing prevents an issue's own re-selection.

## Desired gate

Exclude an open issue from the issue queue when it has a **non-draft (ready)
PR** — not merely any PR. A *draft* PR must **not** gate the issue: draft PRs are
the normal in-flight dispatch state, and an orphaned/stalled draft should leave
the issue selectable so the chain can resume by recycling the worktree. The gate
keys on `isDraft == false`.

Computable with no extra API call: `PR_LIST` (dispatch-select-target ~347)
[...]

PR #1449, opened 2026-06-13
dispatch: exclude done-ready targets from selection + backstop headless-tick continuation

natb1:
Closes #1445

## Problem

The autonomous dispatch chain propagates only via a spawned worker's Stop-hook or
an armed `dispatch-reseed-*` timer. A headless tick that spawns no worker **and**
arms no reseed ends the chain silently.

At `target=0`, when the top priority item is an issue whose closing PR is already
**done/ready** (non-draft, merge-pending — not dispatchable), the tick selects it
(`issue <N>`), `dispatch-materialize-spawn` re-derives its phase as `done` and
drains it as `drain target-done`, spawning no worker and arming no reseed. The
chain dies with real dispatchable priority work still waiting. Observed idle
window in the incident: Jun 12 21:12 → Jun 13 09:43.

Two independent root defects, both fixed here:

1. **Selection / materialization disagree.** `dispatch-select-target`'s leaf-trace
   loop gates the leaf on `dispatch-ci-ready` (which *passes* a non-draft/done PR)
   but never checks the done phase, and the existing `READY_PR_CLOSED_ISSUES`
   exclusion keys only on the **root** help-wanted issue, not the traced leaf.
2. **No backstop on the headless path.** The #1010 "park-if-no-continuation"
   invariant lives in `dispatch-self-close`, which the headless `dispatch-tick`
   deliberately never calls. A `drain target-done` (or any no-spawn drain) on the
   autonomous path therefore has no safety net.

### 1. Exclude done-ready leaves from selection (primary fix)

`dispatch-select-target` now precomputes `READY_PR_BRANCH_ISSUES` — issue numbers
whose own `<N>-*` branch carries a **non-draft** PR — keyed on the `headRefName`
branch prefix (the traced leaf), the same way `dispatch-ci-ready`/`dispatch-phase`
resolve a PR. A non-draft PR is `done` per `dispatch-phase`'s non-draft→done rule,
so this set is provably equivalent to `done` over the open-PR domain, built with
**zero extra `gh` calls** from the already-fetched `PR_LIST`.

In the leaf-trace loop, after the `dispatch-ci-ready` gate and before the leaf is
recorded, a done leaf is excluded (`root_excl["$leaf"]=1; continue`) — mirroring
`dispatch-materialize-spawn`'s spawn-boundary done re-check at selection time, and
continuing the descent to the next startable leaf. This runs in the shared scan,
covering both default and `--priority-only` selection. The root-keyed
`READY_PR_CLOSED_ISSUES` skip is retained as a cheap pre-trace skip for a done
root; the new branch-keyed set is the authoritative leaf check.

### 2. Backstop the headless tick on a no-spawn drain (defense-in-depth)

`dispatch-tick`'s `run_materialize` token-routing `case` is split: `propagate` /
`resolve merge-conflict` (a worker/resolver was spawned → its Stop-hook carries
the chain) and `notify *` (the one-shot explicit path) keep a bare `exit 0`, while
`drain` / `drain *` (no worker spawned) now invoke `dispatch-tick-recover`
(best-effort) before `exit 0`. `dispatch-tick-recover` (#1150) is the
budget-independent continuation-guarantor: it arms a bounded-backoff reseed timer
when no continuation exists, self-limits, and escalates past its cap via a
`dispatch:chain-stalled` issue. This closes the residual selection→materialize
race (a PR flipping non-draft in the worker-boot gap) and any future
no-continuation drain.

### 3. Tests

`test-dispatch-scripts.sh` gains: three `dispatch-select-target` cases (priority-only
returns `empty` for a done-ready leaf; a leaf-specific variant proving the
branch-keyed set catches a done **sub-issue** leaf the root-keyed set would miss;
a default-mode skip-and-select-next case), and five `dispatch-tick` cases
(extending the existing harness with a fake `dispatch-tick-recover`) asserting
`drain`/`drain target-done` invoke recover while `propagate`/`notify *`/`resolve
merge-conflict` do not.

## Scope boundary

This PR closes #1445 only. Explicitly out of scope: the `target>0, only-done-ready`
benign idle, and the plain-`empty`-path-releases-lock-with-no-reseed finding — that
is the #1444 sibling's separate path.

## Verification

**2285/2285 pass**, no existing test regressed (including the `--priority-only`,
#920 ready-PR-closed, and materialize done-re-check cases).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -886,20 +886,25 @@ while IFS=$'\t' read -r issue_num issue_labels; do
       # which precedes the issue within a bucket and recycles the orphan worktree
       # via `enter` just as the issue path would. dispatch-ci-ready resolves the
       # all-digit leaf to its PR by headRefName prefix. Reuses the PR list above.
-      if ! DISPATCH_PR_LIST="$PR_LIST" "$SCRIPT_DIR/dispatch-ci-ready" "$leaf" >/dev/null; then
+      #
+      # The done-ready exclusion runs FIRST, before dispatch-ci-ready, to skip the
+      # subprocess fork for a done-ready leaf. A non-draft (ready) PR on the leaf's
+      # own <leaf>-* branch makes the leaf phase `done` (dispatch-phase's
+      # non-draft→done rule). dispatch-materialize-spawn would re-derive that done
+      # phase at the spawn boundary and drain it, spawning no worker — killing the
+      # chain. Exclude the done leaf here at selection time, mirroring that
+      # spawn-boundary done re-check, and continue descending the same root to the
+      # next startable leaf — exactly like the not-ready skip below (#1445). The
+      # root-keyed READY_PR_CLOSED_ISSUES skip above only catches a done *root*;
+      # this branch-keyed set is the authoritative check on the traced leaf.
+      # READY_PR_BRANCH_ISSUES membership implies the leaf already has a non-draft
+      # PR, for which dispatch-ci-ready would unconditionally pass — so skipping the
+      # call here changes nothing but the saved fork.
+      if [[ -n "${READY_PR_BRANCH_ISSUES[$leaf]:-}" ]]; then
         root_excl["$leaf"]=1
         continue
       fi
-      # A non-draft (ready) PR on the leaf's own <leaf>-* branch makes the leaf
-      # phase `done` (dispatch-phase's non-draft→done rule). dispatch-materialize-spawn
-      # would re-derive that done phase at the spawn boundary and drain it,
-      # spawning no worker — killing the chain. Exclude the done leaf here at
-      # selection time, mirroring that spawn-boundary done re-check, and continue
-      # descending the same root to the next startable leaf — exactly like the
-      # not-ready skip above (#1445). The root-keyed READY_PR_CLOSED_ISSUES skip
-      # above only catches a done *root*; this branch-keyed set is the
-      # authoritative check on the traced leaf.
-      if [[ -n "${READY_PR_BRANCH_ISSUES[$leaf]:-}" ]]; then
+      if ! DISPATCH_PR_LIST="$PR_LIST" "$SCRIPT_DIR/dispatch-ci-ready" "$leaf" >/dev/null; then
         root_excl["$leaf"]=1
         continue
       fi
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -792,6 +792,22 @@ done < <(printf '%s' "$PR_LIST" | jq -r '
   .[] | select(.isDraft == false)
   | (.closingIssuesReferences // [])[].number')
 
+# Issue numbers whose own <N>-* branch carries a non-draft (ready) PR. Unlike
+# READY_PR_CLOSED_ISSUES (which keys on closingIssuesReferences, i.e. the root
+# help-wanted issue), this set keys on the headRefName branch-prefix issue number
+# — the traced leaf — exactly how dispatch-ci-ready/dispatch-phase resolve a PR
+# (by <N>- prefix). A non-draft PR is phase `done` per dispatch-phase's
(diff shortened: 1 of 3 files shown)


natb1:
## QA Summary — #1445

**PR:** #1449 | **CI:** passing | **Date:** 2026-06-13

### Items Executed

| 1 | Done-ready priority leaf is excluded from selection | script-verifiable | ✅ PASS |
| 2 | Headless tick arms a continuation on a no-spawn drain | script-verifiable | ✅ PASS |
| 3 | Acceptance-criteria coverage is genuine, not just green | needs-human-judgment | ⏸ Deferred |

**PASS: 2 / FAIL: 0 / SKIP: 0 / Deferred: 1**

### Deferred to office-hours

**Item 3 — Acceptance-criteria coverage is genuine, not just green**

Human judgment is needed to confirm that the new test cases for #1445 assert on observable, substantive outcomes (selection result `empty` / excluded leaf; presence vs. absence of `dispatch-tick-recover` invocation per token type) rather than tautological stubs, and that all four acceptance criteria map to the new cases. This cannot be decided by a green suite run alone.

### Script-verifiable results

Both script-verifiable items ran via the full dispatch unit test suite:

```
→ 2285/2285 passed, 0 failed
```

All 8 new test cases (3 `dispatch-select-target` + 5 `dispatch-tick`) are included in this count. No existing test regressed.

### Bugs found

None.

natb1:
## Review — PR #1449 (#1445 dispatch: done-ready priority drain kills chain)

**Fix commit:** e85a4591

### Fixed

- **dispatch-select-target:902** — `READY_PR_BRANCH_ISSUES` check was running after `dispatch-ci-ready`, incurring a subprocess fork on every done-ready leaf even though `READY_PR_BRANCH_ISSUES` membership implies `dispatch-ci-ready` would unconditionally pass. Moved the done-ready exclusion before the `dispatch-ci-ready` call so done-ready leaves exit without spawning the subprocess. _(code-review)_

- **test-dispatch-scripts.sh:3369** — The three original #1445 done-ready tests covered (a) a top-level issue as its own leaf, (b) a sub-issue leaf via `--priority-only`, and (c) default-mode skip to the next issue — but none proved the skip _continues descent within the same root_. Added test 41-bis: root 55 with sub-issues 5500 (non-draft PR on `5500-foo` → done-ready) and 5501 (open, no PR, startable), using empty `closingIssuesReferences` so only `READY_PR_BRANCH_ISSUES` fires. Asserts result is `issue 5501`, catching a regression where done-ready skip terminated the root rather than falling through to the sibling. _(code-review)_

### Informational

- **dispatch-tick:328** — The new `drain` arm invokes `dispatch-tick-recover` unconditionally, but explicit dispatch (`dispatch <N>`) also routes through `run_materialize "explicit"` and can emit a `drain` token. On that path `dispatch-tick-recover` will run even for an explicit one-shot call. This is benign (recover self-limits and no-ops when a continuation already exists), but callers using explicit dispatch may see recover's side effects. _(code-review)_

- **test-dispatch-scripts.sh:19367** — The five new `dispatch-tick` #1445 tests only cover `drain target-done`, bare `drain`, `propagate`, `notify target-blocked`, and `resolve merge-conflict`. Pre-existing terminal variants (`drain worktree-conflict`, `drain ci-reseeded`, `drain ci-waiting`, etc.) are not covered by the new recover-invocation assertions. These drain variants would also invoke recover under the new split, but the test harness doesn't assert it. _(review)_

- **dispatch-tick:328** — The drain arm uses a soft failure: `dispatch-tick-recover || echo ... >&2; exit 0`. A `dispatch-tick-recover` failure is logged to stderr but the tick exits 0 regardless — the continuation is not guaranteed. This matches the "best-effort" design intent documented in the PR, but a silent non-guarantee on the recover path could be surprising. _(review)_

- **dispatch-select-target:903** — The done-leaf exclusion sets `root_excl["$leaf"]=1` but does not add the done leaf to the global `EXCLUDED` map. This mirrors the existing not-ready skip, so it is intentional, but it means a done leaf is not globally excluded across root iterations — it could be revisited if it appears as a sub-issue under multiple roots. In practice done-ready leaves have a fixed branch state and will be skipped each time, so this is not a functional regression. _(review)_

### Security

No attack surface changed — all modifications are shell scripts in the dispatch automation harness (no application code, no Firestore rules, no credentials). No CodeQL alerts open on this PR.

PR #1455, opened 2026-06-13
dispatch: cap open-PR fetches with an explicit --limit and loud truncation guard

natb1:
Closes #1312

## Problem

`gh pr list` defaults to **30 results** when no `--limit` is given. Across the
dispatch pipeline, every *open-PR* snapshot fetched with no `--limit`, so once
more than 30 PRs are open the list silently truncated at 30. An autonomous
fan-out can quietly cross 30 open PRs, and the truncation corrupts every
predicate built on the snapshot:

- `dispatch-find-pr` falsely reports "no PR" → a worker opens a duplicate PR.
- `dispatch-select-target`'s ready-PR / closed-issue gate misses ready PRs.
- `dispatch-phase` / `dispatch-ci-ready` mis-derive a PR's phase.
- `dispatch-route` and `dispatch-stop.sh` share a truncated `DISPATCH_PR_LIST`
  with their child predicates.
- `dispatch-reconcile-ready` skips truncated-away ready PRs.

The codebase already caps the *issue* fetch at `--limit 300` with a comment that
gh's default of 30 would silently truncate (`dispatch-select-target`). This PR
extends that discipline to every open-PR fetch.

## Change

**Unit 1 — one shared guarded helper, all call sites migrated.** Adds
`pr_list_open` to `lib.sh`: it fetches with an explicit
`--limit "$DISPATCH_PR_LIST_LIMIT"` (default 300, mirroring the open-issue cap)
and makes truncation **loud** — when the result length equals the limit it errors
to stderr (`… likely truncated …`) and returns non-zero, so the failure can never
recur silently. Every open-PR fetch site now routes through it:

- `dispatch-phase`, `dispatch-ci-ready`, `office-hours-select-target`,
  `dispatch-select-target`, `dispatch-reconcile-ready` (already sourced `lib.sh`).
- `dispatch-find-pr` (both self-fetches), `dispatch-route`, and the
  `dispatch-stop.sh` hook now source `lib.sh` and call the helper.
  `dispatch-stop.sh` keeps its deliberate `2>/dev/null || DISPATCH_PR_LIST=""`
  failure-tolerance — on truncation children self-fetch and surface the loud
  error themselves.

Helper + every call-site migration + every test-stub update land in **one
commit**: the harness's `gh` stub matches the exact arg string, so a partial
migration would silently fall through to `[]` (green-but-wrong tests).

**Unit 2 — tests that lock in the fix.**

- *No truncation past 30:* a 35-PR fixture where `dispatch-find-pr 33` must
  resolve to `33` — under the old default-30 behavior PR 33 would be truncated
  away and the lookup would falsely return empty.
- *Loud guard at the limit:* with `DISPATCH_PR_LIST_LIMIT=5` and exactly 5 open
  PRs, `pr_list_open` exits non-zero and writes `likely truncated` to stderr.

## Out of scope (deliberate)

- `dispatch-sweep` (`--state all --limit 200`, no equals-limit guard) — not an
  open-PR snapshot; possible follow-up.
- `dispatch-drift-scan` (`--state merged --limit 100`, already guarded).
- `digest/SKILL.md` (a `--state merged` fetch in a markdown skill body).

## Verification

`test-dispatch-scripts.sh` — 2275/2275 passed; `grep -rn 'gh pr list --state
open'` returns only the helper definition.

### DIFF
--- a/.claude/hooks/dispatch-stop.sh
+++ b/.claude/hooks/dispatch-stop.sh
@@ -106,6 +106,7 @@ ISSUE_NUM="${JOB_NAME%%-*}"
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)" || exit 0
 SCRIPTS="$SCRIPT_DIR/../skills/dispatch-propagate/scripts"
+source "$SCRIPTS/lib.sh"
 
 # Resolve the why-comment reason. The #826 enrichment hook may write a
 # context-specific reason to $CLAUDE_JOB_DIR/office-hours-reason; when present
@@ -169,8 +170,7 @@ PR_NUM=$("$SCRIPTS/dispatch-find-pr" "$ISSUE_NUM" 2>/dev/null) || PR_NUM=""
 # phase derivation below via DISPATCH_PR_LIST, avoiding a redundant `gh pr list`
 # per predicate. On fetch failure DISPATCH_PR_LIST stays empty and each script
 # falls back to its own self-fetch.
-DISPATCH_PR_LIST=$(gh pr list --state open \
-  --json number,headRefName,isDraft,statusCheckRollup,labels,mergeable 2>/dev/null) \
+DISPATCH_PR_LIST=$(pr_list_open "number,headRefName,isDraft,statusCheckRollup,labels,mergeable" 2>/dev/null) \
   || DISPATCH_PR_LIST=""
 export DISPATCH_PR_LIST
 
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-ci-ready
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-ci-ready
@@ -39,7 +39,7 @@ fi
 if [[ -n "${DISPATCH_PR_LIST:-}" ]]; then
   PR_LIST="$DISPATCH_PR_LIST"
 else
-  PR_LIST=$(gh pr list --state open --json number,headRefName,isDraft,statusCheckRollup,labels,mergeable)
+  PR_LIST=$(pr_list_open "number,headRefName,isDraft,statusCheckRollup,labels,mergeable")
 fi
 
 # Find the matching PR.
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-find-pr
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-find-pr
@@ -34,6 +34,9 @@
 #     (default: 1). Set to 0 in tests to avoid sleeps.
 set -euo pipefail
 
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+source "$SCRIPT_DIR/lib.sh"
+
 ARG="${1:-}"
 if [[ -z "$ARG" || ! "$ARG" =~ ^[0-9]+$ ]]; then
   echo "error: usage: dispatch-find-pr <issue-num>" >&2
@@ -50,7 +53,7 @@ prefix_match() {
 if [[ -n "${DISPATCH_PR_LIST:-}" ]]; then
   PR_LIST="$DISPATCH_PR_LIST"
 else
-  PR_LIST=$(gh pr list --state open --json number,headRefName)
+  PR_LIST=$(pr_list_open "number,headRefName")
 fi
 
 RESULT=$(prefix_match "$PR_LIST")
@@ -59,7 +62,7 @@ if [[ -n "$RESULT" ]]; then echo "$RESULT"; exit 0; fi
 # Retry-on-empty (self-fetch path only).
 if [[ -z "${DISPATCH_PR_LIST:-}" ]]; then
   sleep "${DISPATCH_FIND_PR_RETRY_DELAY:-1}"
-  PR_LIST=$(gh pr list --state open --json number,headRefName)
+  PR_LIST=$(pr_list_open "number,headRefName")
   RESULT=$(prefix_match "$PR_LIST")
   if [[ -n "$RESULT" ]]; then echo "$RESULT"; exit 0; fi
 fi
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-find-pr
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-find-pr
@@ -19,7 +19,9 @@
 #   2. Cross-check via closedByPullRequestsReferences: fetches the issue's
 #      own list of closing PRs and returns any OPEN reference. This catches
 #      the branch-renamed-away case and provides a second authority on the
-#      "no PR" question regardless of which path resolved PR_LIST.
+#      "no PR" question regardless of which path successfully resolved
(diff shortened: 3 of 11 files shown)


natb1:
## QA Summary — #1312 dispatch: gh pr list without --limit truncates the open-PR snapshot at 30 across the pipeline

**PR:** #1455
**QA run:** 2026-06-13

### Results

| Item | Title | Result |
|------|-------|--------|
| 1 | `pr_list_open` helper exists and is sourceable from `lib.sh` | ✅ PASS |
| 2 | Helper passes explicit `--limit` with default 300 | ✅ PASS |
| 3 | No call site retains a bare unlimited `gh pr list --state open` | ✅ PASS |
| 4 | All 8 named call sites invoke the helper | ✅ PASS |
| 5 | Full dispatch script test suite (2312/2312 passed, incl. 2 new #1312 tests) | ✅ PASS |
| 6 | Loud truncation guard fires at the limit (`likely truncated` + non-zero exit) | ✅ PASS |
| 7 | Truncation guard behavior is appropriate for the dispatch pipeline | ⏸ Deferred to office-hours |

**6 PASS / 0 FAIL / 0 SKIP / 1 deferred**

### Script-verifiable checks

**Item 1** — `pr_list_open` is a function in `lib.sh`, sourceable with `type pr_list_open`. ✅

**Item 2** — `lib.sh:117` sets `DISPATCH_PR_LIST_LIMIT="${DISPATCH_PR_LIST_LIMIT:-300}"`, and `lib.sh:128` passes `--limit "$DISPATCH_PR_LIST_LIMIT"`. ✅

**Item 3** — `grep -rn 'gh pr list --state open'` returns exactly one hit: `lib.sh:128`, inside the helper definition itself. No call site bypasses the helper. ✅

**Item 4** — All 8 migrated scripts (`dispatch-phase`, `dispatch-ci-ready`, `office-hours-select-target`, `dispatch-select-target`, `dispatch-reconcile-ready`, `dispatch-find-pr`, `dispatch-route`, `dispatch-stop.sh`) reference `pr_list_open`. ✅

**Item 5** — `test-dispatch-scripts.sh` ran 2312/2312 tests passing (PR cited 2275; count grew as other PRs added tests since). The two new #1312 tests ran: the 35-PR no-truncation test (`dispatch-find-pr 33` resolves to `33`) and the loud-guard test (`DISPATCH_PR_LIST_LIMIT=5` → non-zero exit + `likely truncated` on stderr). ✅

**Item 6** — `lib.sh:133–134` writes the "likely truncated" error to stderr and returns 1 when `len -eq DISPATCH_PR_LIST_LIMIT`. ✅

### Deferred to office-hours

**Item 7 — Truncation guard behavior is appropriate for the dispatch pipeline** (needs-human-judgment)

Success depends on a judgment call about operational behavior: whether loud-fail-at-limit (non-zero exit) is the right contract across all migrated call sites, and whether the 300 default provides adequate headroom for this repo's open-PR volume. The mechanical correctness is verified; this item asks whether the design choice is appropriate given operational context.

natb1:
## QA follow-up — item 7 resolved (clean pass)

The one deferred item from the prior QA comment (the operational-behavior judgment call) has been resolved by the QA operator as **PASS**. QA is now a clean pass: **7/7 items resolved, 0 bugs.**

### Item 7 — Is the truncation guard's behavior appropriate for the dispatch pipeline?

**Verdict: PASS.**

**Loud-fail-at-limit is correctly handled at all 8 call sites** — no site silently swallows truncation in a way that re-creates the original bug:

- **6 sites run `set -euo pipefail`** — a failing `VAR=$(pr_list_open ...)` command substitution aborts the script with a non-zero exit (confirmed empirically), so truncation aborts loudly: `dispatch-phase`, `dispatch-ci-ready`, `office-hours-select-target`, `dispatch-select-target`, `dispatch-route`, `dispatch-find-pr`.
- **`dispatch-reconcile-ready`** (deliberately not `set -e`, so a per-PR error doesn't abort its loop) captures the rc via `|| PR_LIST_RC=$?` and then `exit 1`s with its own stderr message.
- **`dispatch-stop.sh` hook** deliberately tolerates failure (`2>/dev/null || DISPATCH_PR_LIST=""`) — correct for a Stop hook that must not crash. On truncation, children self-fetch through the same guarded helper under `set -e` and surface the loud error themselves. Truncation is **deferred, not hidden**.

**300 is comfortable headroom:**

- Current open PRs: **9** (~33× headroom). Highest all-time PR number: 1475.
- Mirrors the existing open-issue cap (`--limit 300`) for consistency.
- Overridable via `DISPATCH_PR_LIST_LIMIT`.
- A genuine approach to 300 open PRs would itself be an anomaly worth the loud failure the guard provides.

Applying `dispatch:qa-done`.

natb1:
## Review pass — #1312

### Fixed

- **dispatch-route:122** — `pr_list_open` return code was unchecked; with `set -euo pipefail` active a gh failure or truncation guard would abort silently with no directive, leaving the chain in an inconsistent state. Now captures `PR_LIST_RC`, writes an `office-hours-reason`, and emits `STOP pr-list-failed` so the Stop hook escalates to office-hours rather than aborting unguarded. (commit 6354d973)

- **dispatch-find-pr:56** — On the self-fetch path, a `pr_list_open` failure (gh error or truncation guard) would also fall through implicitly via `set -e`. Made the abort explicit with `|| exit "$?"` and added a comment documenting the deliberate no-retry / no-fall-through-on-truncation behavior: a second fetch at the same limit hits the same ceiling, and falling through to the cross-check would silently resolve to "no PR" — the original #1312 bug. (commit 6354d973)

### Informational

- **dispatch-stop.sh:173** — The `pr_list_open` call here is wrapped in `2>/dev/null || DISPATCH_PR_LIST=""`, which silences the truncation guard's stderr diagnostic. This is deliberate (the hook's failure-tolerance contract: children self-fetch and surface the loud error themselves), but callers who inspect `DISPATCH_PR_LIST` after the hook runs will see an empty string with no stderr explanation.

- **lib.sh:117** — `DISPATCH_PR_LIST_LIMIT` is assigned at module-load time (not `export`, not `local`). Any script that sources `lib.sh` before setting its own `DISPATCH_PR_LIST_LIMIT` will bind the default at source time; overriding it after is fine but the window for confusion exists.

### Dismissed

- **test-dispatch-scripts.sh:1241** — Test 12 sources `$TMPDIR_TEST/lib.sh` (a copy of the real `lib.sh`) via `bash -c`. The stub `gh` is on `PATH`, so `pr_list_open` resolves to the stub-backed call correctly. No test isolation issue.

- **lib.sh:117-128** — `DISPATCH_PR_LIST_LIMIT` is passed to `gh pr list --limit` without integer validation. `gh` will reject a non-integer at runtime with a clear error; dispatch scripts run in controlled environments where the override is only set in tests. Defensive validation here would add noise without defensive value.

### Security

CodeQL scan: 0 open alerts. No dependency changes (`deps=false`). No security review surface (`app_or_rules=false` — shell scripts and tests only).

Issue #1454, opened 2026-06-13
dispatch: high abnormal worker death rate (~185 dead-session-stranded reclaims)


natb1:
## Context

Sub-issue of #1450. The reservation ledger journal shows **185
`dead-session-stranded` reclaims vs 336 clean `live-worker-redundant`** — about
a third of spawned workers exit abnormally (the reserving session died without
becoming/then leaving a recognized completed worker) rather than completing
their phase. Observed example: the #1311–#1314 batch spawned at 11:02:31 on
2026-06-13 was gone ~1 minute later.

## Defect

Workers that die shortly after spawn bleed headcount faster than the refill loop
restores it, capping the sustainable fleet size independent of tick speed. The
cause is not yet characterized — candidates include immediate API/boot errors,
silent parking on an unexpected prompt (see #1311: the input-block hook is dead
code, so a worker hitting a prompt parks silently with no baton pass), or
premature stop.

## Acceptance criteria

- [ ] The dominant cause(s) of `dead-session-stranded` reclaims are identified
      from worker transcripts (sample the recent stranded sessions).
- [ ] Abnormal early-exit workers either recover/retry or surface the failure,
      rather than silently stranding.
- [ ] The stranded-reclaim fraction is measurably reduced.

## References

- Parent epic: #1450. Likely related: #1311 (silent input-block park).

PR #1476, opened 2026-06-13
dispatch: converge to target worker count via end-of-tick convergence reseed

natb1:
Closes #1453

## Problem

The autonomous dispatch fleet refills workers only at tick boundaries, and a
tick computes its spawn budget `GAP = TARGET_N − LIVE_COUNT` once, fills up to
`GAP` slots in one batch, then waits for the next worker event. There is no
standing heartbeat, so `effective_live` (busy workers + outstanding
reservations) is never actively driven toward `TARGET_N` and headcount sawtooths
far below target.

Two dynamics push the count below target with nothing armed to recover:

1. **Partial fill.** The selector scan returns fewer ready targets than `GAP`
   (some CI-pending or transiently worktree-conflicted), so the run spawns
   `N < GAP` and exits — no follow-up armed.
2. **Drain during a tick.** A tick takes minutes. While it runs, busy workers
   complete; each completion's Stop-hook calls `dispatch-spawn-tick`, but that
   uses a fixed unit name, so concurrent calls collide and dedup. When the
   in-flight tick finishes — even after reporting "gap filled" — the workers that
   drained leave the fleet below target with no pending tick. A `spawned < GAP`
   trigger does not catch this (the tick's own `GAP` slots were all filled); only
   a fresh `effective_live` re-count at tick end does.

## Change

Drive `effective_live` toward `TARGET_N` by arming a short-delay **convergence
reseed** at the end of any autonomous tick that reached a normal materialize
decision but may have ended below target — modeled on the existing `ci-waiting`
reseed precedent rather than on a periodic heartbeat.

**Unit 1 — new `dispatch-schedule-convergence-reseed`.** Re-counts effective-live
workers (busy + reservations, daemon-UNKNOWN → 0 fail-toward-arming) and arms a
transient no-arg `dispatch-tick` timer only when below target and not gated:

- `LIVE >= TARGET_N` → `converged` no-op (this also covers the pace-gate-closed
  case for free: a pace pause makes `dispatch-target-workers` return `0`, so
  `LIVE >= 0` always holds).
- `--exhausted` reports `exhausted` → `exhausted` no-op (the cap-path pace reseed
  owns resumption).
- otherwise arms `dispatch-converge-<epoch>` at `NOW + DELAY` (default 120s),
  carrying the same `systemd-run` contract as the sibling reseed scripts
  (`--collect`, `OnFailure=dispatch-tick-recover.service`, `KillMode=process`,
  `Persistent=true`, `--setenv=PATH`), with `@<epoch>` idempotency (`already
  exists` collision → no-op). Every value reaching an arithmetic context is
  integer-validated (the siblings' RCE-hardening posture).

**Unit 2 — wire `dispatch-tick`.** On the `propagate` terminal token only, and on
the autonomous queue path only (`[[ -z "$MANUAL" && -z "$ARG" ]]` — explicit and
bare-manual `/dispatch` are gate-exempt single-target spawns), call the new
script best-effort: a failure logs one stderr note and the tick still exits 0.
Placement on `propagate` is what makes the trigger a fresh `effective_live`
recount rather than the stale `spawned < GAP` signal, catching both the
partial-fill and drain-during-tick cases. A paced/exhausted tick returns
`concurrency-cap` and never reaches this branch.

## Tests

Full `test-dispatch-scripts.sh` suite green (2336/2336). New coverage:

- `dispatch-schedule-convergence-reseed`: below-target arm (full `systemd-run`
  argv incl. no-trailing-arg `ExecStart`), busy+reservation sum, at-target
  `converged`, pace-gate-closed `converged`, token-exhausted `exhausted`,
  idempotent already-exists re-arm.
- `dispatch-tick` wiring: autonomous `propagate` (issue + pr) calls the script
  once; explicit and `--manual` `propagate` do not call it; `concurrency-cap`,
  `drain`, and `drain target-done` do not call it; a convergence-arm failure
  still exits 0.

## Alternatives not chosen

A standing periodic heartbeat timer (the issue's second option) would also
converge the fleet, but runs against the architecture's deliberate move away
from a coarse static heartbeat toward event/reset-driven reseeds, and is better
understood as addressing the fully-drained absorbing dead-state (#1445) — a
separate concern. The convergence reseed subsumes it for this issue's criteria
while firing only when actually below target. A global bounded attempt counter
was considered and left out: termination is already natural (a follow-up tick
that cannot fill returns `empty`/`concurrency-cap`, neither of which arms a
convergence reseed), so a counter is at most optional hardening against
degenerate fast-churn (#1454 territory).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-schedule-convergence-reseed
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-schedule-convergence-reseed
@@ -1,0 +1,252 @@
+#!/usr/bin/env bash
+# dispatch-schedule-convergence-reseed — arm a short-delay convergence reseed
+# when an autonomous dispatch tick ends below the worker target.
+#
+# The autonomous fleet refills workers only at tick boundaries: a tick computes
+# a spawn budget GAP = TARGET_N − LIVE_COUNT once, fills up to GAP slots, then
+# waits for the next worker-completion event to tick again. effective_live
+# (busy workers + reservations) is never actively driven back UP to TARGET_N, so
+# two dynamics can leave the fleet below target with nothing armed to recover:
+#
+#   1. Partial fill — the selector returns fewer ready targets than GAP, so the
+#      tick spawns fewer than GAP workers and ends below target.
+#   2. Drain during a tick — workers complete mid-tick, their Stop-hook ticks
+#      collide / dedup, and when the in-flight tick finishes below target there
+#      is no pending tick to refill.
+#
+# This script is the convergence safety net: called at the END of an autonomous
+# tick that spawned workers but ended below target, it re-counts effective-live
+# workers and, if the fleet is still below target and not gated, arms a
+# short-delay transient systemd.user timer that re-runs the headless NO-ARG
+# `dispatch-tick` (re-evaluate the whole queue). The fleet thus converges back
+# to TARGET_N event-driven, without a periodic heartbeat.
+#
+# It is the convergence analog of dispatch-schedule-target-reseed (single-target,
+# CI-keyed) and dispatch-schedule-reseed (pace/cap-reset), but NO-ARG (whole-queue
+# re-evaluation) and on a short FIXED delay.
+#
+# Behavior, in order:
+#   1. Source lib.sh + lib-claude-agents.sh + lib-reservation-ledger.sh.
+#   2. TARGET_N = dispatch-target-workers (default count mode).
+#   3. Recount effective-live workers:
+#        BUSY = claude_agents_count_busy_workers (daemon UNKNOWN → 0, fail toward
+#               arming — mirrors dispatch-tick-recover's no-continuation posture).
+#        RESV = reservation_count.
+#        LIVE = BUSY + RESV.
+#   4. Suppression gates (in order):
+#        LIVE >= TARGET_N → `converged`, exit 0. This also covers the pace-gate-
+#          closed case for free: a pace pause makes dispatch-target-workers return
+#          0, so LIVE >= 0 always holds → no-op. Checking this BEFORE --exhausted
+#          means a TARGET=0 pace pause never even queries exhaustion.
+#        --exhausted == exhausted → `exhausted`, exit 0 (genuine token exhaustion;
+#          a reseed would spawn nothing).
+#   5. Otherwise arm the timer at FIRE = NOW + DELAY. Unit dispatch-converge-<FIRE>,
+#      ExecStart the NO-ARG dispatch-tick in the main worktree. The @<FIRE> epoch
+#      suffix gives idempotency: a same-second re-arm collides on the unit name and
+#      systemd refuses to recreate — treated as a no-op (exit 0).
+#
+# Stdout protocol (exactly one line):
+#   scheduled <unit> at <FIRE>   timer was armed.
+#   converged                    at/above target (incl. pace-gate-closed) no-op.
+#   exhausted                    token-exhaustion no-op.
+#
+# Exit codes:
+#   0  on every action / no-op path (arm, converged, exhausted, idempotent
+#      already-exists collision).
+#   2  on a bad env-integer override.
+#   non-zero — `systemd-run` failed for a reason other than the already-exists
+#              collision; the exit code is passed through.
+#
+# Environment overrides for testability:
+#
+#   DISPATCH_CONVERGE_RESEED_NOW
+#     Override the current epoch seconds. Default: date +%s.
+#
+#   DISPATCH_CONVERGE_RESEED_DELAY
+#     Override the reseed delay in seconds. Default: 120.
+#
+#   DISPATCH_CONVERGE_RESEED_SYSTEMD_RUN_CMD
+#     Override the `systemd-run` binary. Default: `systemd-run`.
+#
+#   DISPATCH_CONVERGE_RESEED_MAIN_WORKTREE
+#     Override the main worktree path. When set, the script does NOT require a
+#     git repo. Default: <project-root>/worktrees/main, where project-root is
+#     the parent of git --git-common-dir. Mirrors dispatch-schedule-reseed.
+#
+#   DISPATCH_CONVERGE_RESEED_TARGET_WORKERS_CMD
+#     Override the dispatch-target-workers binary (consulted for both the default
+#     target and the --exhausted query). Default: $SCRIPT_DIR/dispatch-target-workers.
+#
+#   CLAUDE_AGENTS_CMD
+#     Inherited from lib-claude-agents.sh: replaces the `claude` invocation in
+#     claude_agents_count_busy_workers. Default: `claude`.
+#
+#   DISPATCH_RESERVATION_DIR
+#     Inherited from lib-reservation-ledger.sh: override the ledger directory.
+#
+# Sandbox: `systemd-run --user` talks to the user's systemd over D-Bus, and the
+# recount reads `claude agents --json` over a Unix socket; this script is not
+# sandbox-safe and callers must invoke it with `dangerouslyDisableSandbox: true`.
+# See `.claude/rules/sandbox.md`.
+#
+# NOT set -e: exits are handled explicitly (matching the sibling reseed scripts).
+set -uo pipefail
+
+# ---- Step 0: defaults & config ----------------------------------------------
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=lib.sh
+source "$SCRIPT_DIR/lib.sh"
+# shellcheck source=lib-claude-agents.sh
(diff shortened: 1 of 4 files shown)


natb1:
## QA Summary — #1453 dispatch: converge to target worker count

**PR #1476** · CI: passing · No browser component (pure shell/script change)

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Full dispatch test suite (criterion 4) | ✅ PASS |
| 2 | Below-target arms no-arg dispatch-tick reseed (criterion 1 proxy) | ✅ PASS |
| 3 | Suppression: at-target, pace-gate-closed, exhausted (criterion 3) | ✅ PASS |
| 4 | dispatch-tick autonomous-only propagate wiring (criterion 2) | ✅ PASS |
| 5 | Fleet actually reaches/holds ~8 live workers at runtime (criterion 1, end-to-end) | ⏸ Deferred to office-hours |

**PASS: 4 / FAIL: 0 / SKIP: 0 / Deferred: 1**

### Script-verifiable results

**Item 1 — Full suite:** `test-dispatch-scripts.sh` — **2336/2336 passed, 0 failed**

**Item 2 — Below-target arm:**
- `PASS: below target stdout names the convergence unit` (`scheduled dispatch-converge-10120 at 10120`)
- `PASS: below target systemd-run argv (unit + calendar + collect + OnFailure + KillMode + cwd + setenv + no-arg exec)`

**Item 3 — Suppression gates:**
- `PASS: at target; no systemd-run invocation`
- `PASS: pace gate closed; no systemd-run invocation`
- `PASS: exhausted; no systemd-run invocation`

**Item 4 — dispatch-tick wiring (8 assertions):**
- `PASS: converge-issue: convergence script called exactly once` (autonomous issue + propagate)
- `PASS: converge-pr: convergence script called exactly once` (autonomous pr + propagate)
- `PASS: converge-explicit: convergence script NOT called (ARG set)` (explicit dispatch)
- `PASS: converge-manual: convergence script NOT called (MANUAL set)` (--manual)
- `PASS: converge-besteffort: exit 0 despite arm failure` (failure never wedges tick)
- `PASS: converge-cap: convergence script NOT called (no propagate branch)`
- `PASS: converge-drain: convergence script NOT called`
- `PASS: converge-drain-td: convergence script NOT called`

### Deferred to office-hours

**Item 5 — Live fleet holds ~8 workers at runtime (criterion 1, end-to-end):**
Requires observing a running systemd-user environment with a live `claude` daemon and real worker spawns to confirm `effective_live` converges to TARGET_N and holds there. The test suite verifies the scripts *arm* the timer correctly; whether that results in the fleet holding 8 workers requires a live fleet observation. Deferred for human judgment.

natb1:
## QA Summary — #1453 dispatch: converge to target worker count

**PR #1476** · CI: passing · No browser component (pure shell/script change)

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Full dispatch test suite (criterion 4) | ✅ PASS |
| 2 | Below-target arms no-arg dispatch-tick reseed (criterion 1 proxy) | ✅ PASS |
| 3 | Suppression: at-target, pace-gate-closed, exhausted (criterion 3) | ✅ PASS |
| 4 | dispatch-tick autonomous-only propagate wiring (criterion 2) | ✅ PASS |
| 5 | Fleet actually reaches/holds ~8 live workers at runtime (criterion 1, end-to-end) | ⏭ SKIP — verify on main |

**PASS: 4 / FAIL: 0 / SKIP: 1**

### Script-verifiable results

**Item 1 — Full suite:** `test-dispatch-scripts.sh` — **2336/2336 passed, 0 failed**

**Item 2 — Below-target arm:**
- `PASS: below target stdout names the convergence unit` (`scheduled dispatch-converge-10120 at 10120`)
- `PASS: below target systemd-run argv (unit + calendar + collect + OnFailure + KillMode + cwd + setenv + no-arg exec)`

**Item 3 — Suppression gates:**
- `PASS: at target; no systemd-run invocation`
- `PASS: pace gate closed; no systemd-run invocation`
- `PASS: exhausted; no systemd-run invocation`

**Item 4 — dispatch-tick wiring (8 assertions):**
- `PASS: converge-issue: convergence script called exactly once` (autonomous issue + propagate)
- `PASS: converge-pr: convergence script called exactly once` (autonomous pr + propagate)
- `PASS: converge-explicit: convergence script NOT called (ARG set)` (explicit dispatch)
- `PASS: converge-manual: convergence script NOT called (MANUAL set)` (--manual)
- `PASS: converge-besteffort: exit 0 despite arm failure` (failure never wedges tick)
- `PASS: converge-cap: convergence script NOT called (no propagate branch)`
- `PASS: converge-drain: convergence script NOT called`
- `PASS: converge-drain-td: convergence script NOT called`

### Skipped

**Item 5 — Live fleet holds ~8 workers at runtime (criterion 1, end-to-end):**
Requires a running systemd-user environment with a live `claude` daemon and real worker spawns. Cannot be machine-verified pre-merge; will be verified on main.

natb1:
## Review — #1453 dispatch: converge to target worker count

### Fixed

- **Missing test: `resolve merge-conflict` does not arm convergence reseed** (`test-dispatch-scripts.sh:19915`)
  The terminal-token routing loop asserted every recognized token (including `resolve merge-conflict`) exits 0, but did not assert the convergence reseed was absent on that token. Added a dedicated test mirroring the `converge-drain`/`converge-drain-td` pattern. (commit `2e85bde5`)

- **Missing `DISPATCH_DAEMON_*` stubs in `sr_setup` / `cr_setup`** (`test-dispatch-scripts.sh:10649`)
  `ensure_daemon_service` (called from `ensure_recover_unit`) consults `DISPATCH_DAEMON_UNIT_DIR`, `DISPATCH_DAEMON_SYSTEMCTL_CMD`, and `DISPATCH_DAEMON_CLAUDE_CMD` at setup time. Without stubs those fall through to real systemd and `claude` paths, making tests non-hermetic. Added the missing stubs and unsets to `sr_setup`/`sr_teardown` (cr_setup already had them in the working tree). (commit `2e85bde5`)

### Dismissed

- **DELAY `^[0-9]+$` accepts 0** (`dispatch-schedule-convergence-reseed:175-178`)
  A `DELAY=0` produces `FIRE=NOW`, and with `Persistent=true` systemd fires immediately on the next bus sync — effectively a very-short delay, not a silent loop. The sibling reseed scripts use the same validation posture. Not a real defect.

### Informational

- **BUSY vs RESV failure asymmetry** (`dispatch-schedule-convergence-reseed:129-141`)
  A `claude_agents` daemon failure silently defaults BUSY to 0 (fail toward arming), while `reservation_count` failure exits 2 (hard error). The asymmetry is intentional — daemon UNKNOWN is a transient reachability issue; a corrupt reservation ledger is a misconfiguration — but worth noting for future reference.

- **Best-effort stderr note not asserted** (`test-dispatch-scripts.sh:19878-19884`)
  The `TICK_CONVERGE_RC=1` test asserts exit 0 and that `converge.log` has one line, but does not check that dispatch-tick emitted the `convergence reseed arm failed` note to stderr. Minor coverage gap; adding a stderr assert would make the contract explicit.

- **Convergence reseed stdout not captured in `dispatch-tick`** (`dispatch-schedule-convergence-reseed:241-258`)
  The convergence reseed script's stdout (`converged`, `exhausted`, `scheduled dispatch-converge-X at X`) flows to the tick's stdout. This is consistent with the sibling reseed scripts' behavior and is generally harmless in headless runs, but callers that care about clean stdout would need to redirect it explicitly.

### False-positive (security)

- **Env-var injection via `DISPATCH_CONVERGE_RESEED_*` overrides** (`dispatch-schedule-convergence-reseed:105-106`)
  These overrides are testability hooks, not an attack surface — all values reaching arithmetic context are integer-validated (the `^[0-9]+$` guards reject non-numeric input), and the systemd-run command is not constructed from them via shell expansion. Not a real vulnerability.

- **Attacker-controlled worktree path via `DISPATCH_CONVERGE_RESEED_MAIN_WORKTREE`** (`dispatch-schedule-convergence-reseed:188-194,233`)
  An attacker who can set env vars in the dispatch process already has arbitrary-code execution. The env override is a test seam, not a privilege boundary.

- **Suppressing convergence reseed silently via best-effort guard** (`dispatch-tick:326-329`)
  The `|| echo ... >&2` guard is auditable — every suppression writes a line to stderr. A suppression does not produce silent divergence; at worst the fleet misses one convergence cycle and re-arms on the next worker-completion event.

- **Spurious reseeds via daemon BUSY→0 fallback** (`dispatch-schedule-convergence-reseed:129-135`)
  A spurious reseed fires a single no-arg `dispatch-tick` after `DELAY` seconds. The tick's own pace, concurrency, and exhaustion gates prevent over-spawning. Redundant ticks are harmless and the design explicitly accepts this posture.

- **Spoofing `already exists` to suppress scheduling** (`dispatch-schedule-convergence-reseed:241-247`)
  An attacker who can inject into systemd's stderr or replace the systemd-run binary already has full systemd control. The substring match is a defensive dedup, not a trust boundary.

Issue #1473, opened 2026-06-13
dispatch: router deprioritizes enhancement issues below all non-enhancement work


natb1:
Part of #1470. Blocked by the foundation sub-issue.

## Context

`dispatch-select-target` ignores the `enhancement` label today; enhancement issues
rank by topic like any other work. With `enhancement` repurposed as a deferral
marker, the router must drain all non-enhancement work before any enhancement
issue, then order enhancement issues by topic.

## Approach

1. **Add `has_enhancement_in_labels`** mirroring the existing
   `has_priority_in_labels` helper.
2. **Insert an enhancement axis** nested as `priority ⊃ enhancement ⊃ topic ⊃
   phase`. Priority stays outermost (a human can escalate one enhancement; the
   `--priority-only` at-cap bypass needs no special-casing). Within each priority
   tier, non-enhancement items drain before enhancement items; within the
   enhancement tier the existing topic ladder orders them.
3. **Scope: deprioritize enhancement *issue selection*, not in-flight PRs**
   (recommended). An enhancement issue waits behind all non-enhancement work, but an
   enhancement PR already in flight (review/qa) is not stranded — it completes
   normally. The simpler fallback (enhancement bit on all work items, extending the
   bucket key to `<category>|<pri>|<enh>|<phase>`) sinks enhancement PRs too; the
   implementer chooses. Either way the change touches all four emission paths
   (`TOP==1`, `TOP>1`, `--qa`, `--priority-only`).
4. **`bug`/`security` stay in `TOPIC_CATEGORIES` unchanged** — within the
   non-enhancement tier they still rank bugs/security above user-filed features; in
   the enhancement tier those buckets are always empty.
5. **One-time backlog sweep** (rollout step): strip `enhancement` from all
   currently-open issues so the legacy default-labeled backlog is not deprioritized
   en masse. New follow-ups re-acquire the label going forward.
6. **Tests** in `test-dispatch-scripts.sh` cover the new ordering across the
   affected emission paths.

## Acceptance criteria

- [ ] `dispatch-select-target` selects every eligible non-enhancement issue before
      any enhancement issue, in every emission path.
- [ ] Within the enhancement tier, issues are ordered by the existing topic ladder.
- [ ] `priority` remains the outermost axis (a `priority` enhancement outranks
      non-priority non-enhancement work).
- [ ] `bug`/`security` ranking within the non-enhancement tier is unchanged.
- [ ] A one-time sweep removes the legacy `enhancement` label from currently-open
[...]

PR #1477, opened 2026-06-13
dispatch: gate /file-issue enhancement label on --follow-up provenance flag

natb1:
Closes #1471

## What

Repurpose the `enhancement` type label from a meaningless "not a bug" default
into a **provenance + deferral marker**: it now means *a phase skill auto-filed
this as a follow-up, and the issue is neither a bug nor a security issue*.

Surface 1 of epic #1470. Siblings #1472 (phase skills pass `--follow-up`) and
#1473 (router enhancement axis) are out of scope here.

## How

Documentation-only change to two skill files — `/file-issue` and the
`ref-issue-labels` reference it uses. The type invariant relaxes from "exactly
one of `bug`/`enhancement`" to the three-valued `{bug} | {enhancement, follow-up
only} | {none}`, gated on an explicit `--follow-up` provenance flag.

Authoritative truth table:

| Source | `enhancement`? |
|---|---|
| User files interactively (no `--follow-up`) | No |
| Phase skill auto-files, non-bug & non-security (`--follow-up`) | **Yes** |
| Any `bug` (incl. auto-filed) | No |
| Any `security` topic (incl. auto-filed) | No |

Classification order is **defect/security first**: a structural defect is `bug`;
a `security`-topic non-defect gets the `security` topic and no type label; only a
non-bug, non-security issue reaches the provenance gate, where `--follow-up`
decides `enhancement` vs. no type label.

`file-issue/SKILL.md`:
- **Step 1** — new "Parse provenance flag" subsection strips a leading
  `--follow-up` token and sets `$FOLLOW_UP` before mode detection (so
  `--follow-up <text>` doesn't fail the `^#?[0-9]+$` match). Documented as
  meaningful only on creation paths.
- **Step 3b** — compliance wording made provenance-aware.
- **Step 5** — both decomposition-recursion sites forward `--follow-up` to
  sub-issue calls, gated on the parent's `$FOLLOW_UP`.
- **Step 6** — type classification rewritten to the defect/security-first
  three-valued rule; `enhancement` is no longer a default; the atomic type-swap
  is extended to strip a stale `enhancement` when the new classification yields
  no type label. Bare issue-number-mode re-filing (flag always absent) therefore
  strips a stale `enhancement` — this complements, and does not re-implement, the
  epic's one-time backlog sweep (#1473).
- **Step 6 apply template** — drop `--add-label "<type>"` when no type matched,
  mirroring the existing topic-drop idiom; the `--remove-label` swap may fire
  alone.

`ref-issue-labels/SKILL.md` mirrors the same three-valued invariant,
provenance-marker definition, extended swap, and empty-type apply template, kept
byte-consistent with `/file-issue`. The untrusted-input guards in both files are
unchanged.

## Verification

This change is to skill instruction documents — verification is by review of the
edited prose for internal consistency. Confirmed:
- No surviving "default when the issue is not a bug" or "exactly one of
  `bug`/`enhancement`" phrasing in either file (grep, no matches).
- The two apply-template comment blocks are byte-identical.
- The untrusted-input guards are unchanged.


natb1:
## QA Summary — Issue #1471 / PR #1477

**Documentation-only change** to `.claude/skills/file-issue/SKILL.md` and `.claude/skills/ref-issue-labels/SKILL.md`.
No browser component. Shell-command lane only.

### Results

| 1. `--follow-up` flag stripped before mode detection | script-verifiable | ✅ PASS |
| 2. `--follow-up` + non-bug/non-security → `enhancement` | script-verifiable | ✅ PASS |
| 3. No `--follow-up` → no type label for non-bug spec | script-verifiable | ✅ PASS |
| 4. Bug spec → `bug`, never `enhancement` | script-verifiable | ✅ PASS |
| 5. Security-topic spec → never `enhancement` | script-verifiable | ✅ PASS |
| 6. Decomposition gate forwards `--follow-up` (both recursion sites) | script-verifiable | ✅ PASS |
| 7. `ref-issue-labels` documents relaxed invariant; no old phrasing | script-verifiable | ✅ PASS |
| 8. Apply-template comment blocks byte-identical across both skills | script-verifiable | ✅ PASS |
| 9. `$FOLLOW_UP` forced off in issue-number mode (prose consistency) | needs-human-judgment | ⏸ Deferred |
| 10. No residual "not-a-bug → enhancement default" phrasing | needs-human-judgment | ⏸ Deferred |

**Counts:** 8 PASS, 0 FAIL, 0 SKIP, 2 deferred to office-hours.

### Deferred items (office-hours)

**Item 9 — `$FOLLOW_UP` by-construction guarantee:**
Verify the Step 1 "force `$FOLLOW_UP` off once mode resolves to bare issue number" prose is internally consistent with the Step 6 stale-`enhancement` strip's stated reliance on it. This is a prose-logic consistency check that requires human judgment.

**Item 10 — Residual sentence in `bug` bullet:**
Both `file-issue/SKILL.md:393` and `ref-issue-labels/SKILL.md:28` still contain:
> "A request whose body lacks structural defect signals is `enhancement` even if it mentions bug-flavored keywords."

Under the new three-valued rule, a non-bug, non-security, non-`--follow-up` issue should carry **no type label**, not `enhancement`. This sentence reads as an unconditional "not-a-bug → `enhancement`" default — which contradicts the provenance-gated rule at `file-issue:410` ("else NO type label"). Needs human to determine if this is a genuine inconsistency requiring a fix, or an acceptable bug-classification discriminator whose `enhancement` reference is understood as "falls into the non-bug bucket."

No QA server was started (documentation-only change, no browser component).

natb1:
## Review (dispatch:review-fix)

### Security

### Fixed (578ea52c)

- **`ref-issue-labels/SKILL.md:2`** (code-review): Frontmatter `description:` field still implied the old exactly-one-type invariant ("choose a bug/enhancement type"). Updated to reflect the three-valued contract and document that type may be absent for user-filed non-bug non-security issues.
- **`ref-issue-labels/SKILL.md:33`** (code-review): Skill stated the enhancement-vs-none decision is "provenance-gated and caller-signaled" without defining how the caller signals provenance, leaving it impossible for a model invoking the skill to know what to pass. Clarified that enhancement-vs-none is resolved by `file-issue` Step 6 before invoking `ref-issue-labels`; the skill classifies only the defect/security dimension.
- **`ref-issue-labels/SKILL.md:39`** (code-review): Security-topic tie-break ("security-topic issues never carry `enhancement`") referenced the topic label as already computed, creating a forward reference — the type step runs before topic classification. Added proactive security-signal heuristics to the type step so the suppression fires at type-classification time without depending on a later label.
- **`file-issue/SKILL.md:430`** (review): Topic classification step said "Invoke ref-issue-labels via the Skill tool to classify `<type>` and an optional `<topic>`", implying the skill determines type. Reworded to "classify an optional `<topic>`", making it explicit that type is already determined by the Type classification subsection and only the topic is delegated to ref-issue-labels.
- **`file-issue/SKILL.md:407`** (review): Type classification rule 2 ("Non-defect with the `security` topic → no type label") was label-based, forward-referencing the topic computed in a later step. Rewrote rule 2 to be content-signal-based ("Non-defect whose body carries a security signal — vulnerability, CVE, advisory, CodeQL alert, security hardening → no type label"), making the rule self-contained.
- **`file-issue/SKILL.md:433-441`** (review): Apply template lacked a skip guard for the all-clauses-dropped case (no type, no topic, no stale label to remove), which would emit a bare `gh issue edit <N>` with no `--add-label` or `--remove-label`. Added a guard comment to skip the `gh issue edit` call entirely in that case. Same addition applied to the matching template block in `ref-issue-labels/SKILL.md`.

### Informational

- **`file-issue/SKILL.md:193`** (code-review): Compliance check (Step 3b) now says "`enhancement` applies only to a non-bug, non-security issue filed with `--follow-up`." In issue-number mode `$FOLLOW_UP` is always unset, so this wording could confuse a model re-filing an issue that already carries `enhancement`. No change needed — the Step 6 stale-enhancement strip handles this case correctly and Step 1 already forces `$FOLLOW_UP` off in issue-number mode; the compliance check is just slightly under-specified for that path.

PR #1479, opened 2026-06-13
dispatch: bound and cache the per-tick selection scan

natb1:
Closes #1452

## Why

A dispatch tick (`dispatch-select-tick` gate + `dispatch-materialize-spawn`
fan-out) was taking ~8 minutes wall-clock — I/O-bound on hundreds of serial `gh`
and `claude agents --json` round-trips. Because tick latency exceeded worker
lifetime, headcount sawtoothed far below target. This bounds and caches the
per-tick selection scan so a tick drops from minutes to seconds, with
**byte-identical selection output**.

Three independent waste sources are removed, each spanning process boundaries:

1. Every machine-wide `claude agents --json` call re-queried the daemon with no
   caching — firing from the gate's `reservation_sweep` + busy-worker count, then
   once per PR and per traced subtree node.
2. `dispatch-trace-leaf`'s `issue-blocking`/`issue-sub-issues` lookups were
   un-memoized and re-fetched per node on every per-root re-invocation.
3. The issue walk traced every help-wanted root even after the emittable buckets
   that could win were full.

## What

Built as six units (one commit each):

- **Unit 1 — `lib-claude-agents.sh`:** the three machine-wide functions
  (`claude_agents_list_all`, `claude_sessions_with_name`,
  `claude_agents_count_busy_workers`) read a single cached array via
  `DISPATCH_AGENTS_SNAPSHOT` when set+readable, via a private `_claude_agents_raw`
  seam; added `claude_agents_snapshot_capture <path>`. The two freshness-sensitive
  functions (`claude_sessions_under`, `verify_agent_registered_under`, the
  registration-race poll) are deliberately excluded. The UNKNOWN fail-safe is
  preserved bit-for-bit when the snapshot is unset/unreadable/empty.

- **Unit 2 — `dispatch-tick`:** captures one snapshot before the selection call
  and exports `DISPATCH_AGENTS_SNAPSHOT` + a tick-scoped `DISPATCH_TRACE_CACHE_DIR`
  so both sibling subprocesses inherit them. The sentinel cleanup and the new
  workspace cleanup are consolidated into one EXIT/INT/TERM trap (bash traps
  overwrite, not stack). A capture failure removes the partial file, leaves the
  snapshot var unset, and degrades to live per-call reads without aborting the
  tick. The workspace is created unconditionally so manual and explicit paths
  benefit too.

- **Unit 2b — `dispatch-select-tick`:** comment-only clarification that the
  autonomous gate now adds no live `claude agents --json` round-trips.

- **Unit 3 — `dispatch-trace-leaf`:** `open_children` memoizes the
  `issue-blocking`/`issue-sub-issues` sibling calls per `(script, num)` under
  `DISPATCH_TRACE_CACHE_DIR` via `cached_child_lookup`. The gh-failure hard-error
  contract is preserved: a non-zero sibling exit propagates as before and caches
  nothing (never cache a failed result); an empty-but-successful result is
  cacheable, distinguished by exit code not output emptiness. Unset dir → direct
  sibling call, no caching.

- **Unit 4 — `dispatch-select-target`:** the help-wanted issue walk processes
  roots in `(priority desc, category-rank asc, createdAt asc)` order — the same
  rank order the emission loops drain — with `pri`/`cat-rank`/`cat-name`
  precomputed in the `ISSUE_SORTED` jq (reusing the exact `category_of_labels` /
  `has_priority_in_labels` expressions). A `group_emittable_count` helper and a
  group-boundary tracker break the walk once strictly-higher-rank
  `(pri, category)` groups already hold TOP emittable entries combined
  (`FIRST_PR` + `FIRST_ISSUE`), so it stops tracing once no remaining lower-rank
  root can change the output. The reorder is sound because `EXCLUDED` is frozen
  before the walk (each group's buckets are fed only by its own roots, within-group
  createdAt order preserved) and emission drains groups in the same rank order.

- **Unit 5 — `test-dispatch-scripts.sh`:** call-logging on the
  `claude`/`issue-blocking`/`issue-sub-issues` fakes plus assertions: snapshot
  consumed → 0 live `claude agents` calls (with a live-path contrast); one capture
  per tick; `(script, num)` memoized once under the cache dir (with an uncached
  ≥2 contrast); an injected `gh` failure still hard-fails under caching and writes
  no cache file; the early-exit yields the same emitted set and never traces a
  lower-rank root once a higher tier fills TOP=1. Both env vars default off so
  existing tests keep the live/uncached path.

## Testing

`./.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` —
**2328/2328 passed, 0 failed** (the pre-change suite was 2309; +19 new
assertions). The existing selection tests guard byte-identical output and stayed
green throughout.

The scope note from the plan holds: the ≤1-`claude agents`-per-tick bound applies
to the selection phase; the detached launcher's `verify_agent_registered_under`
legitimately keeps a fresh post-selection query and is not folded onto the
snapshot.


natb1:
## QA Summary — #1452 / PR #1479

### Items Executed

| 1 | Full dispatch test suite (2328/2328 passed) | script-verifiable | ✅ PASS |
| 2 | `lib-claude-agents.sh` snapshot logic present | script-verifiable | ✅ PASS |
| 3 | `dispatch-trace-leaf` + `dispatch-tick` wiring present | script-verifiable | ✅ PASS |
| 4 | `dispatch-select-target` early-exit group logic present | script-verifiable | ✅ PASS |
| 5 | `teardown()` unsets `DISPATCH_AGENTS_SNAPSHOT` + `DISPATCH_TRACE_CACHE_DIR` | script-verifiable | ✅ PASS |
| 6 | Representative tick completes in seconds over current backlog | needs-human-judgment | ⏸ Deferred |

### Counts
- PASS: 5 / FAIL: 0 / SKIP: 0
- Deferred to office-hours: 1

### Deferred to Office-Hours
**Item 6 — Criterion 4 (live tick timing):** Verify a representative `dispatch-tick` over the current backlog completes in seconds (down from ~8 min baseline). Requires a live daemon, real backlog state, and human judgment of what constitutes "representative" and "in seconds." Not machine-verifiable against fixtures.

### Bugs Found
None.

### No Walkthrough
All changes are dispatch shell scripts — no frontend or browser component detected.

natb1:
## Review findings — PR #1479

### Fixed

- **`claude_agents_snapshot_capture` leaks daemon stderr** (`lib-claude-agents.sh:179`): `claude agents --json` stderr was not suppressed; added `2>/dev/null` so daemon noise does not surface during snapshot capture. `dispatch-tick` already logs a clear degradation message on failure, so the raw stderr adds nothing. — commit `15443ce8`

### Informational

- **`_claude_agents_raw` `return 0` after `cat`** (`lib-claude-agents.sh:163-164`): `return 0` executes unconditionally even if `cat "$DISPATCH_AGENTS_SNAPSHOT"` fails (e.g., the file disappears between the `-r` check and the `cat`). In practice the race is extremely narrow and the caller already falls back on UNKNOWN, but the exit code from `cat` is silently swallowed. Left as-is since defensive handling here would add complexity with no real-world benefit.

- **`cached_child_lookup` non-atomic cache write** (`dispatch-trace-leaf:215`): `printf '%s' "$out" >"$cache_file"` is a non-atomic write — a concurrent reader could see a partial file mid-write. Per-tick parallelism in the current dispatch architecture is not high enough to make this a real concern, and the cache is tick-scoped (removed on exit), so the worst case is a stale read within a single tick.

- **`DISPATCH_AGENTS_SNAPSHOT` / `DISPATCH_TRACE_CACHE_DIR` exported before `dispatch-materialize-spawn`** (`dispatch-tick:302`): the exported env vars are inherited by the detached launcher chain. The launchers do not call `claude_agents_list_all` or `cached_child_lookup`, so the snapshot and cache vars are inert in those processes — no correctness concern.

- **Early-exit group-boundary tracker updates `walk_prev_cat` / `walk_prev_pri` on every group change** (`dispatch-select-target:925-934`): these variables are updated redundantly inside the boundary block. The logic is correct; it is slightly verbose but not a bug.

- **Test D memo assertion coverage** (`test-dispatch-scripts.sh:24486`): Test D asserts `issue-blocking-calls.log` and `issue-sub-issues-calls.log` each have exactly 1 entry for issue 830. The test does not assert on the emitted leaf (brittle with duplicate bucket entries), which is fine — the assertion on lookup count is the meaningful invariant.

- **`_dispatch_tick_cleanup` trap `rm -rf` expansion timing** (`dispatch-tick:138`): the trap string `'_dispatch_tick_cleanup'` is a function call, so `$DISPATCH_TICK_WORKSPACE` is expanded at invocation time, not at `trap` definition time. This is correct behavior — the variable is set before the trap fires — but worth noting as a non-obvious bash trap idiom.

### False-positive (security)

- **`_claude_agents_raw` reads `DISPATCH_AGENTS_SNAPSHOT` without path sanitization** (`lib-claude-agents.sh:162`): flagged as a potential path traversal. `DISPATCH_AGENTS_SNAPSHOT` is set exclusively by `dispatch-tick` (internal, same process group) to a `mktemp -d`-rooted path; it is never sourced from user input or external HTTP. No sanitization needed in this internal coordination variable.

- **`cached_child_lookup` builds cache path from env var** (`dispatch-trace-leaf:206`): flagged as potential path injection via `$DISPATCH_TRACE_CACHE_DIR/$prefix-$num`. Both `prefix` (literal strings `blocking` / `sub-issues`) and `num` (issue number from jq-parsed GitHub API JSON) are controlled inputs from the same process group; `DISPATCH_TRACE_CACHE_DIR` is set exclusively by `dispatch-tick`. No injection risk.

### Deferred

- **`ISSUE_SORTED` jq duplicates `category_of_labels` / `has_priority_in_labels` logic** (`dispatch-select-target:895-906`): the jq filter inlines the category-resolution and priority-detection expressions that the shell helpers implement. A divergence would silently mis-sort roots, undermining the early-exit soundness argument. Filed as #1490 (blocked by #1452, the implementing issue).

PR #1496, opened 2026-06-13
dispatch-phase: gh_retry the no-PR label fetch, fail clear instead of mapping to plan

natb1:
Closes #1314

## Problem

`dispatch-phase` answers "which workflow phase skill runs for an issue/branch."
For a no-PR target it derives the phase from whether the issue carries the
`dispatch:planned` label — present → `implement`, absent → `plan` — learned via
`gh issue view <num> --json labels`.

The label fetch swallowed any failure and emitted `plan`:

```bash
ISSUE_LABELS=$(gh issue view "$ISSUE_NUM" --json labels | jq -r '.labels[].name' 2>/dev/null) || {
    echo "plan"
    exit 0
}
```

So a transient `gh`/network blip was indistinguishable from "no
`dispatch:planned` label" and yielded `plan`. A genuinely-planned issue then
routed to relevance-review / re-planning instead of `/implement` — wasting a
tick or duplicating the plan. This is the "defensive fallback that buries
errors" anti-pattern called out in `.claude/rules/code-style.md`; the rest of
`dispatch-phase` already fails clear under `set -euo pipefail`.

## Fix

Wrap the fetch in `gh_retry` (`lib.sh:54`, already sourced) so a *transient*
failure self-heals via retry, and fail clear (exit non-zero) on a real failure
instead of swallowing it. Because `gh_retry` wraps a single command (not a
pipeline), the former `gh … | jq …` one-liner is split into a `gh_retry` fetch
and a separate `jq` parse, each failing clear:

```bash
ISSUE_JSON=$(gh_retry gh issue view "$ISSUE_NUM" --json labels) || {
  echo "dispatch-phase: gh issue view #$ISSUE_NUM (labels) failed; cannot derive phase" >&2
  exit 1
}
ISSUE_LABELS=$(jq -r '.labels[].name' <<<"$ISSUE_JSON") || {
  echo "dispatch-phase: failed to parse labels JSON for #$ISSUE_NUM" >&2
  exit 1
}
```

`dispatch-route` runs `PHASE=$(... dispatch-phase "$N")` under `set -euo
pipefail`, so a non-zero exit propagates as a route failure — the desired
fail-clear behavior. No caller change is needed.

**Empty-labels behavior is preserved:** a successful fetch of an unlabeled issue
yields `{"labels":[]}` → `jq` prints nothing → the `dispatch:planned` check
misses → `plan`. Legitimately-unplanned issues still route to `plan`; only
*failures* now exit non-zero. The trailing `2>/dev/null` (which had masked `jq`
parse errors) is dropped.

**Out of scope:** the bare `gh pr list` at `dispatch-phase:52` is not part of
this fix — it has no silent fallback and already fails clear under `set -e`.

## Tests

`test-dispatch-scripts.sh` gains a per-issue fail-count stub sentinel
(`issue-view-fail-<num>`) that emits a transient `HTTP 503` error and decrements
on each call — absent by default, so every existing test is unaffected. Two new
`dispatch-phase` tests:

- **transient-exhausted → exit 1, not `plan`** — fail count above the retry
  budget; asserts non-zero exit and empty stdout (the #1314 bug).
- **transient-then-succeed → `implement`** — fail twice within budget, then
  serve a `dispatch:planned` label; proves the retry fires and derives the
  correct phase.

Full suite: `2394/2394 passed, 0 failed`.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-phase
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-phase
@@ -26,6 +26,7 @@
 # For a no-PR target, the phase is `plan` unless the issue carries the
 # `dispatch:planned` label (then `implement`). This is determined by fetching
 # the issue's labels via `gh issue view`.
+# A gh/network failure in this fetch exits non-zero (it is no longer silently mapped to plan); transient blips self-heal via gh_retry.
 #
 # Environment:
 #   DISPATCH_PR_LIST — optional. A JSON array of open PRs supplied by the
@@ -78,9 +79,13 @@ if [[ -z "$PR" || "$PR" == "null" ]]; then
       exit 1
     fi
   fi
-  ISSUE_LABELS=$(gh issue view "$ISSUE_NUM" --json labels | jq -r '.labels[].name' 2>/dev/null) || {
-    echo "plan"
-    exit 0
+  ISSUE_JSON=$(gh_retry gh issue view "$ISSUE_NUM" --json labels) || {
+    echo "dispatch-phase: gh issue view #$ISSUE_NUM (labels) failed; cannot derive phase" >&2
+    exit 1
+  }
+  ISSUE_LABELS=$(jq -r '.labels[].name' <<<"$ISSUE_JSON") || {
+    echo "dispatch-phase: failed to parse labels JSON for #$ISSUE_NUM" >&2
+    exit 1
   }
   if grep -qxF "dispatch:planned" <<<"$ISSUE_LABELS"; then
     echo "implement"
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -397,6 +397,17 @@ case "$args" in
     # $STUB_DIR/issue-labels-<num>.json supplies the labels object; absence means
     # the issue carries no labels.
     num=$(echo "$args" | awk '{print $3}')
+    # #1314: a per-issue fail-count sentinel makes the stub emit a transient
+    # (HTTP 503) error and decrement the count, so gh_retry retries. Absent by
+    # default → every existing test is unaffected; exhausted → serve labels.
+    if [[ -f "$STUB_DIR/issue-view-fail-${num}" ]]; then
+      remaining=$(cat "$STUB_DIR/issue-view-fail-${num}")
+      if [[ "$remaining" -gt 0 ]]; then
+        echo $((remaining - 1)) > "$STUB_DIR/issue-view-fail-${num}"
+        echo "HTTP 503: Service Unavailable" >&2
+        exit 1
+      fi
+    fi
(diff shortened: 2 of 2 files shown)


natb1:
## QA summary — issue #1314 / PR #1496

**Pass: 3 | Fail: 0 | Skip: 0 | Deferred: 1**

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | gh failure exits non-zero (not `plan`) | shell-command | PASS |
| 2 | Transient blip self-heals via `gh_retry`, derives correct phase | shell-command | PASS |
| 3 | Exit-code contract consistent with callers | — | deferred |
| 4 | JSON-parse failure is a clear error, not silent `plan` | shell-command | PASS |

### Test suite

`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — **2394/2394 passed, 0 failed**

Covers test 1d (persistently-transient → exit 1, not `plan`, the #1314 bug) and test 1e (transient-then-succeed → `implement`, proves retry fires).

### Code review (item 4)

`dispatch-phase:87` — `jq` parse failure exits 1 before `grep dispatch:planned` at line 90. A malformed JSON payload cannot fall through to `echo "plan"`.

### Deferred to office-hours

**Item 3 — Exit-code contract consistent with callers.** Requires manual review of whether any `dispatch-phase` caller masks a generic non-zero exit (e.g. `|| echo plan`, `2>/dev/null`, `|| true`) that would re-introduce the silent-plan defect downstream. Callers are expected to be clean under `set -euo pipefail`, but this warrants a human look before applying `dispatch:qa-done`.

### Bugs found

None.

natb1:
## Review findings — #1314

### Fixed

- **`test-dispatch-scripts.sh:803`** (code-review) — test 1d pinned the fail-count to 5 to exhaust `GH_RETRY_ATTEMPTS` but never explicitly exported `GH_RETRY_ATTEMPTS=4`, making the assertion rely silently on the lib.sh default. Added `export GH_RETRY_ATTEMPTS=4` before the fail-count line and `unset GH_RETRY_ATTEMPTS` in the matching teardown. (commit `71506ec6`)

### Informational

- **`test-dispatch-scripts.sh:797`** (review) — the new tests cover the transient-exhausted (exit 1) and transient-then-recover (implement) paths, but there is no dispatch-phase-level test asserting that a successful fetch of an issue with no labels still routes to `plan`. The behavior is preserved by the implementation and covered indirectly by existing tests (1a–1c), but an explicit empty-labels case would make the coverage self-documenting.

### False-positive

- **`test-dispatch-scripts.sh:399`** (input-validation) — `num` is extracted from `$args` via `awk '{print $3}'` inside the `issue view * --json labels` stub; the test framework controls `$args` directly, so no injection surface exists.
- **`dispatch-phase:86`** (input-validation) — `ISSUE_LABELS` is populated from `jq -r '.labels[].name'` over GitHub API JSON; GitHub label names are controlled by the repository owner, not user input, so no injection surface exists.

PR #1498, opened 2026-06-13
error-sink: word-boundary 'dob' match so 'adobe' keys aren't dropped

natb1:
Closes #1359

## Problem

`SENSITIVE_KEY_RE` in `firebaseutil/src/error-sink.ts` strips sensitive-named
keys from the error sink's `extras` field by **case-insensitive substring**
match. The `dob` alternative (added so date-of-birth fields get dropped) matches
inside the word `adobe` (the sequence a**dob**e), so any context key containing
`adobe` — `adobeId`, `adobeVersion`, `adobeFontId` — was silently dropped with a
`console.warn`, losing non-PII diagnostic data.

## Fix

`dob` moves out of the case-insensitive `SENSITIVE_KEY_RE` into its own
**case-sensitive, word-boundary** regex, `SENSITIVE_DOB_RE`:

```js
const SENSITIVE_DOB_RE = /(?:^|[^A-Za-z])(?:dob|Dob|DOB)(?![a-z])|[A-Za-z](?:Dob|DOB)(?![a-z])/;
```

The matching loop now drops a key when `SENSITIVE_KEY_RE` **or**
`SENSITIVE_DOB_RE` matches.

The `i` flag is the reason `dob` cannot stay in `SENSITIVE_KEY_RE`: detecting a
camelCase hump (`userDob`) requires distinguishing lowercase `r` from uppercase
`D`, but under `i` the character classes are case-folded and that information is
gone. So the `dob` rule must be case-sensitive.

### Why not the issue's suggested `\bdob\b`?

`\b` is a `\w`/`\W` transition, so `\bdob\b` does **not** match the camelCase
hump in `userDob` (the char before `D` is the word char `r`), and treats `_` as
a word char so it misses snake_case `user_dob`/`dob_field` too. It would keep
those keys in `extras` (violating the "still drop dob tokens" requirement) and
break the existing `userDob` drop-test. `SENSITIVE_DOB_RE` was verified in Node
against 19 cases:

- **DROP:** `dob`, `DOB`, `Dob`, `userDob`, `userDOB`, `dobField`, `user_dob`,
  `dob_field`, `DOBField`, `myDob`, `dob123`, `x-dob`.
- **KEEP:** `adobe`, `adobeId`, `adobeVersion`, `adobeFontId`, `dateOfBirth`,
  `photobooth`, `doberman`.

## Tests

- New retention test: `adobeId` survives in `doc.extras` (the bug repro).
- Existing `userDob` drop-test left in place as the regression guard proving
  `dob`-token keys still drop.
- Full `firebaseutil` suite passes (86 tests); lint clean.

## Follow-up (out of scope)

The other alternatives in `SENSITIVE_KEY_RE` have the same class of substring
false-positive — `key` matches `monkey`/`keyboard`, `ssn` likewise. They are
left as plain case-insensitive substrings here (far less likely to false-match
common diagnostic identifiers, and the comment now warns future readers not to
naively word-bound them). A separate issue can weigh whether they warrant the
same boundary treatment.

### DIFF
--- a/firebaseutil/src/error-sink.ts
+++ b/firebaseutil/src/error-sink.ts
@@ -33,10 +33,21 @@ const RESERVED_KEYS = new Set(["operation", "kind", "message", "stack", "code", 
 
 // Strips obviously-sensitive keys from extras by name before serialization.
 // Matches substrings case-insensitively: token, secret, password, auth, key,
-// credential, email, phone, ssn, dob, address. Note the g flag is intentionally
+// credential, email, phone, ssn, address. Note the g flag is intentionally
 // absent — .test() is stateless. Cross-reference: see the RESERVED_KEYS comment
 // above for the residual risk of values stored under non-matching key names.
-const SENSITIVE_KEY_RE = /token|secret|password|auth|key|credential|email|phone|ssn|dob|address/i;
+//
+// `dob` (date of birth) is NOT included here because the bare substring `dob`
+// false-matches the word `adobe` (a-dob-e) under the `i` flag (#1359).
+// Instead it is handled by the separate case-sensitive SENSITIVE_DOB_RE below,
+// which matches `dob` only as a standalone identifier token — as a standalone
+// word, a camelCase hump (e.g. userDob, myDOB), or after a snake/other separator
+// (e.g. user_dob) — without matching compound words like `adobe` or `photobooth`.
+// The other alternatives (token, key, ssn, etc.) are longer or far less likely
+// to collide with common identifiers, so they remain as plain case-insensitive
+// substrings; word-bounding them would be over-engineering for little gain.
+const SENSITIVE_KEY_RE = /token|secret|password|auth|key|credential|email|phone|ssn|address/i;
+const SENSITIVE_DOB_RE = /(?:^|[^A-Za-z])(?:dob|Dob|DOB)(?![a-z])|[A-Za-z](?:Dob|DOB)(?![a-z])/;
 
 // Size caps (UTF-16 code units) — must mirror the firestore.rules isValidErrorLog() caps exactly.
 const CAPS = {
@@ -109,7 +120,7 @@ export function createFirestoreErrorSink(options: ErrorSinkOptions): ErrorSink {
     const droppedSensitiveKeys: string[] = [];
     for (const [key, value] of Object.entries(context)) {
       if (RESERVED_KEYS.has(key)) continue;
-      if (SENSITIVE_KEY_RE.test(key)) {
+      if (SENSITIVE_KEY_RE.test(key) || SENSITIVE_DOB_RE.test(key)) {
         droppedSensitiveKeys.push(key);
         continue;
       }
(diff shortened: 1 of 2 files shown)


natb1:
## QA Summary — #1359 (error-sink: dob substring false-match)

**Result: PASS** — 4/4 items passed, 0 failed, 0 skipped.

No browser walkthrough needed (pure `firebaseutil` package change).

### Items

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | adobeId is retained in doc.extras (false-positive fix) | shell-command | PASS |
| 2 | userDob is still dropped from doc.extras (regression guard) | shell-command | PASS |
| 3 | Full firebaseutil suite passes (no collateral breakage) | shell-command | PASS |
| 4 | Source regex literal and comment reflect the fix | shell-command | PASS |

### Details

- **Item 1 (adobeId retention):** `npx vitest run --project firebaseutil error-sink` — 35 tests passed including the new `retains a key containing 'adobe' (adobeId does not false-match 'dob')` test. `extras.adobeId === "font-42"` confirmed.
- **Item 2 (userDob still drops):** Same run — existing `strips a key with a PII substring (userDob matches 'dob')` test still passes. Regression guard intact.
- **Item 3 (full suite):** `npx vitest run --project firebaseutil` — 86 tests, 4 files, all passed. No collateral breakage.
- **Item 4 (source reflects fix):** Grep confirmed: `SENSITIVE_KEY_RE` at line 49 contains no `dob`; `SENSITIVE_DOB_RE` defined at line 50; loop at line 123 tests both; comment at line 41 documents `adobe` false-match and #1359.

### Counts
- PASS: 4
- FAIL: 0
- SKIP: 0
- Deferred to office-hours: 0
- Bugs found: none

natb1:
## Review — PR #1498

### Fixed (commit d1d83fe7)

- **`firebaseutil/src/error-sink.ts:50`** — `SENSITIVE_DOB_RE` second branch used `[A-Za-z]` before `(?:Dob|DOB)`, which caused all-uppercase keys like `ADOBE` to false-match the `DOB` substring. Narrowed to `[a-z]` so only a lowercase camelCase hump prefix triggers the drop.
- **`firebaseutil/test/error-sink.test.ts:439`** — Existing test only covered the `adobeId` (lowercase camelCase) retention case. Added a parallel `ADOBE` (all-caps) test case to catch the second-branch bug.

### Informational

- **`firebaseutil/src/error-sink.ts:50`** — `SENSITIVE_DOB_RE` first branch `(?:^|[^A-Za-z])(?:dob|Dob|DOB)(?![a-z])` correctly matches lowercase `dob` only when preceded by a non-alpha character (start-of-string, digit, separator). No change needed.
- **`firebaseutil/src/error-sink.ts:50`** — The regex correctly excludes digits and separators (e.g., `dob123`, `x-dob`) from the camelCase hump path and delegates them to the first branch. No change needed.
- **`firebaseutil/test/error-sink.test.ts:439`** — The `adobeId` retention test suppresses `console.warn` but does not assert that warn was never called with `adobeId` as the dropped key. This would strengthen the test but is out of scope for this fix.

### False-positive (security)

- **`firebaseutil/src/error-sink.ts:50`** — Flagged that `SENSITIVE_DOB_RE` does not match `dob` at the start of a key when followed by a lowercase letter (e.g., `dobrze`). This is correct: the `(?![a-z])` negative lookahead is intentional — `dob` must be a complete identifier token, not a prefix of an unrelated word. Not a vulnerability.

PR #1500, opened 2026-06-13
dispatch: bound the lock's daemon probe + flock, validate TARGET_N

natb1:
Closes #1315

## What

Two robustness defects let a wedged (not dead) Claude daemon — or a broken
environment — stall dispatch routing. This bounds both hazards.

### 1. Unbounded daemon query inside the flock (`dispatch-acquire-lock`)

`resolve_holder_state` shelled out to `claude agents --json` with no timeout
while the `try_acquire` subshell held `flock 9`. A wedged daemon blocked the
in-section acquirer indefinitely, and every other router blocked behind
`flock 9` → routing deadlocked until the daemon recovered.

- The daemon probe is now wrapped in `timeout -k 5 "$PROBE_TIMEOUT"`
  (`DISPATCH_LOCK_PROBE_TIMEOUT`, default `10`). The `-k 5` SIGKILL backstop is
  required: a wedged client that ignores SIGTERM would otherwise hang `timeout`
  itself. A timed-out probe exits non-zero, so the existing fail-safe folds it to
  LIVE → the acquirer stays BUSY → clean `busy`. The `9>&-` fd-close stays on the
  `timeout` simple command, preserving the existing semantics.
- The in-section `flock` now waits with `flock -w "$FLOCK_TIMEOUT"`
  (`DISPATCH_LOCK_FLOCK_TIMEOUT`, default `15`, deliberately > probe timeout +
  margin). On wait-timeout the subshell emits a distinct `FLOCK_TIMEOUT` sentinel
  and exits **without** running the read-check-write — `flock -w` alone would fall
  through unlocked and could clobber a live holder. The parent decodes the
  sentinel to contended (busy) with a `(flock wait timeout)` diagnostic.

`--release` and the headless-holder branch are unchanged.

### 2. Unguarded `TARGET_N` (`dispatch-select-tick`)

`TARGET_N=$(dispatch-target-workers)` had no failure guard. An empty/non-numeric
value made the arithmetic gate throw "operand expected" and silently skip the
worker-cap gate (fail-open). `dispatch-target-workers` always prints a number in
count mode (its own fallback to `1`), so a non-numeric value here means a broken
environment. Per the repo's clear-error-over-defensive-fallback rule, the script
now validates `TARGET_N` against `^[0-9]+$` (which allows `0`), releases the lock,
and exits `2` with a clear message — no second fallback.

## Tests

`test-dispatch-scripts.sh` gains coverage for each acceptance criterion:

- Wedged daemon (SIGTERM-ignoring fake under an external `timeout`) → `busy`
  within budget, foreign holder untouched, proving the `-k` backstop.
- In-section flock-wait timeout (held lock + stale recorded holder) → `busy`,
  lock file **not** clobbered — the test that catches the fall-through bug.
- Empty and non-numeric `TARGET_N` → `exit 2`, lock released, clear stderr; plus
  a `TARGET_N=0`-still-passes-the-guard regression.

Full suite: 2403/2403 passing.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-acquire-lock
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-acquire-lock
@@ -98,13 +98,27 @@
 #                               Default: 300.
 #   DISPATCH_LOCK_WAIT_INTERVAL --wait: poll gap in seconds (fractional
 #                               allowed). Default: 5.
+#   DISPATCH_LOCK_PROBE_TIMEOUT Max seconds the `claude agents --json` liveness
+#                               probe may run before it is treated as a wedged
+#                               daemon (timeout-bounded, with a SIGKILL
+#                               backstop). A healthy probe is sub-second.
+#                               Default: 10.
+#   DISPATCH_LOCK_FLOCK_TIMEOUT Max seconds to wait for the in-section flock
+#                               before abandoning the attempt as contended.
+#                               MUST exceed DISPATCH_LOCK_PROBE_TIMEOUT plus
+#                               margin: the probe is the only slow op inside the
+#                               critical section, so a waiter must not abandon a
+#                               holder that is legitimately mid-probe.
+#                               Default: 15.
 #
 # Stdout protocol (exactly one word):
 #   acquired  this dispatch tick holds the lock — proceed with the tick.
 #   busy      another dispatch tick holds the lock — stop cleanly, no side effects.
-#             Under --wait this means the wait timeout elapsed; a one-line
-#             diagnostic (elapsed time + holding sessionId) is written to
-#             stderr.
+#             Also covers the case where the in-section flock wait timed out (a
+#             wedged holder pinned the critical section): the attempt resolves to
+#             contended rather than deadlocking. Under --wait, busy means the
+#             wait timeout elapsed; a one-line diagnostic (elapsed time + holding
+#             sessionId, or "(flock wait timeout)") is written to stderr.
 #   released  --release: our own recorded sessionId was cleared from the lock
 #             file.
 #   noop      --release: nothing to release — the lock was unheld or held by
@@ -140,6 +154,8 @@ fi
 
 WAIT_TIMEOUT="${DISPATCH_LOCK_WAIT_TIMEOUT:-300}"
 WAIT_INTERVAL="${DISPATCH_LOCK_WAIT_INTERVAL:-5}"
+PROBE_TIMEOUT="${DISPATCH_LOCK_PROBE_TIMEOUT:-10}"
+FLOCK_TIMEOUT="${DISPATCH_LOCK_FLOCK_TIMEOUT:-15}"
 
 # resolve_holder_state <sessionId> — query the daemon's live-session registry
 # for <sessionId> and report both liveness and cwd. Covers the same failure
@@ -159,12 +175,22 @@ WAIT_INTERVAL="${DISPATCH_LOCK_WAIT_INTERVAL:-5}"
 # Return: 0 = live (or opaque failure → fail-safe live)
 #         1 = definitely not live (reclaim)
 #
+# The daemon probe is `timeout`-bounded (DISPATCH_LOCK_PROBE_TIMEOUT, with a
+# `-k 5` SIGKILL backstop): a wedged daemon that never returns must not block
+# the probe — and so the in-section flock — indefinitely. The KILL backstop is
+# required because plain `timeout` only sends SIGTERM and then waits for the
+# child, so a client that ignores SIGTERM would make `timeout` itself hang,
+# reproducing the deadlock. A timed-out probe exits non-zero, which the
+# `if ! out=$(...)` below folds to fail-safe LIVE → the caller stays BUSY.
+#
 # The `9>&-` redirection closes fd 9 (the flock fd, opened by the caller's
 # subshell) in the `claude` child. flock is on the open file description and
 # is released only when the LAST referencing fd closes; without the close,
 # any background helper `claude` ever spawns would inherit fd 9 and hold the
 # flock past our subshell's exit. Today `claude agents --json` is a query-and-
 # exit client, but the cost of defending against a future fork is one redir.
+# The redirections stay on the `timeout` simple command, which execs `claude`
+# with fd 9 already closed and stderr already silenced.
 resolve_holder_state() {
   local sid="$1" out rows match
   # Headless holders (the systemd-launched dispatch-tick's synthetic
@@ -184,7 +210,7 @@ resolve_holder_state() {
     kill -0 "$pid" 2>/dev/null || return 1
     return 0
   fi
-  if ! out=$("${CLAUDE_AGENTS_CMD:-claude}" agents --json 2>/dev/null 9>&-); then
+  if ! out=$(timeout -k 5 "$PROBE_TIMEOUT" "${CLAUDE_AGENTS_CMD:-claude}" agents --json 2>/dev/null 9>&-); then
     return 0
   fi
   if [[ -z "${out//[[:space:]]/}" ]]; then
@@ -265,7 +291,16 @@ try_acquire() {            # 0 = acquired, 1 = contended; sets CONTENDED_SID
   local result
   result=$(
     exec 9>>"$LOCK_FILE"
-    flock 9                # blocking — a short mutex over the read-check-write
+    # Bounded mutex over the read-check-write. The only slow op inside the
+    # critical section is the timeout-bounded daemon probe, so a wait that
+    # exceeds FLOCK_TIMEOUT (> PROBE_TIMEOUT + margin) means the holder is
+    # wedged, not legitimately mid-probe. On timeout we leave the subshell
+    # WITHOUT running the read-check-write (which would be a racy unlocked
+    # read + clobbering write); the parent maps the sentinel to contended.
+    if ! flock -w "$FLOCK_TIMEOUT" 9; then
+      printf 'FLOCK_TIMEOUT'
+      exit 0
+    fi
     local recorded="" cwd="" live=0
     IFS= read -r recorded <"$LOCK_FILE" 2>/dev/null || true
     # Busy iff a different sessionId is recorded AND it is still a live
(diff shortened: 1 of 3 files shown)


natb1:
## QA Summary — #1315 (dispatch: bound the lock's daemon query and guard TARGET\_N arithmetic)

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Full dispatch-scripts suite passes locally | shell-command | ✅ PASS |
| 2 | FLOCK\_TIMEOUT > PROBE\_TIMEOUT invariant is documented but not enforced | — | ⏸ Deferred to office-hours |
| 3 | Default timeout values (probe 10s, flock 15s) are operationally appropriate | — | ⏸ Deferred to office-hours |

**Counts:** 1 PASS · 0 FAIL · 0 SKIP · 2 deferred

### Machine-verifiable acceptance criteria

Both acceptance criteria verified by the self-contained dispatch-scripts test suite (2403/2403 passing):

- **AC1 — Wedged daemon → clean busy:** test 5b (probe timeout, SIGTERM-ignoring fake, external `timeout` bound) and test 5c (in-section flock-wait timeout, lock not clobbered) both pass.
- **AC2 — TARGET\_N validated:** non-numeric TARGET\_N and empty TARGET\_N tests both exit 2 with a clear error and release the lock; TARGET\_N=0 regression guard passes.

### Deferred to office-hours

**Item 2 — FLOCK\_TIMEOUT > PROBE\_TIMEOUT invariant is documented but not enforced**

The doc block states FLOCK\_TIMEOUT must exceed PROBE\_TIMEOUT plus margin, but no startup guard enforces this. An operator who raises `DISPATCH_LOCK_PROBE_TIMEOUT` above `DISPATCH_LOCK_FLOCK_TIMEOUT` would silently re-introduce the original bug (a waiter abandons a holder that is mid-probe). The Opus triage agent notes this does not block the PR; it is a design consideration. Judgment call: add a startup guard, or leave as a documented operator contract.

**Item 3 — Default timeout values are operationally appropriate**

Probe 10s, flock 15s, with a `-k 5` SIGKILL backstop giving a hard probe ceiling of ~15s. Judgment call: confirm these defaults suit the real fleet cadence and tick interval.

### No bugs found

No walkthrough needed (pure shell-script PR, no frontend component). No QA server started.

natb1:
## QA update — the two deferred judgment items were resolved in-session

Following up on the prior QA comment (which deferred two `needs-human-judgment`
items to office-hours): both have now been resolved directly.

### Item 2 — enforce `FLOCK_TIMEOUT > PROBE_TIMEOUT` → **resolved with a code change**

The header documented that `DISPATCH_LOCK_FLOCK_TIMEOUT` must exceed
`DISPATCH_LOCK_PROBE_TIMEOUT`, but nothing enforced it. This is a real footgun:

- `DISPATCH_LOCK_PROBE_TIMEOUT=0` — verified that GNU `timeout 0 <cmd>`
  **disables** the bound and runs the probe unbounded, silently re-opening the
  exact wedged-daemon deadlock this PR fixes. `^[0-9]+$` alone would accept `0`.
- `DISPATCH_LOCK_FLOCK_TIMEOUT <= PROBE_TIMEOUT` — a waiter abandons a holder
  that is legitimately mid-probe and reports a false `busy`.

`dispatch-acquire-lock` now validates both knobs at startup on the acquire/wait
path and exits `2` with a clear message (clear-error-over-fallback rule).
`--release` consumes neither knob, so it is left unguarded and unchanged.

Threshold note: the guard uses strict `FLOCK > PROBE`, **not** `> PROBE + 5`.
The `-k 5` SIGKILL grace only bites a SIGTERM-ignoring probe — and in that
pathological case the holder is genuinely wedged, so abandoning-to-`busy` is the
fail-safe outcome. A `> PROBE + 5` guard would also wrongly reject the shipped
`15`/`10` defaults (15 vs 10+5=15).

### Item 3 — default timeout values (probe 10s, flock 15s) → **assessed, no change**

A healthy probe is sub-second; the `-k 5` backstop caps a wedged probe at ~15s;
the `--wait` budget (`DISPATCH_LOCK_WAIT_TIMEOUT=300`, 5s poll) bounds the overall
poll. A worst-case ~15s critical-section stall only on a genuinely wedged daemon
is an acceptable liveness/false-busy trade-off for the fleet. Defaults kept.

### Tests

`test-dispatch-scripts.sh` gains: `FLOCK <= PROBE` → exit 2; `PROBE=0` → exit 2;
shipped `10`/`15` defaults pass the guard (regression for the strict-`>`
decision); `--release` unaffected by a bad knob. The existing flock-wait-timeout
test now uses `PROBE=1`/`FLOCK=2` (smallest valid pair). **Full suite: 2409/2409
passing.**

Pushed as commit `e0d27bb7` on this branch.

> Note: this added new implementation code during the QA phase. Flagging so it
> can get a review pass before merge if desired.

natb1:
## Review: PR #1500 (issue #1315)

**Surface:** shell scripts (code) · **CodeQL alerts:** 0 · **Dependency changes:** none · **Security findings:** all false-positive · **Code changes applied:** none

---

### Informational

- **M-002** `test-dispatch-scripts.sh` (missing); `dispatch-acquire-lock:360,364` — No test coverage for non-numeric `FLOCK_TIMEOUT` or `PROBE_TIMEOUT` values hitting the `^[0-9]+$` format check. Test 5d covers `PROBE_TIMEOUT=0` (fails the `-lt 1` integer comparison) and `FLOCK_TIMEOUT <= PROBE_TIMEOUT` (inequality check), but no test uses `'abc'` or `'1.5'`. These are distinct code paths with distinct error messages. The validation logic is correct; this is test coverage only. Suggested: add a test setting `DISPATCH_LOCK_FLOCK_TIMEOUT='abc'` (and a parallel `PROBE_TIMEOUT='abc'` case) asserting exit 2 with the 'non-negative integer' message.

- **F02** `dispatch-acquire-lock:300` — The inline comment reads "a wait that exceeds FLOCK_TIMEOUT (> PROBE_TIMEOUT + margin)". The parenthetical describes the shipped defaults (15 > 10 + 5) but could be read as the enforcement constraint being `FLOCK > PROBE + margin`. The actual enforced constraint (correctly documented at lines 368–378) is `FLOCK > PROBE` (strict, no margin term). No behavioral impact; the validation block is authoritative.

- **F-002** `test-dispatch-scripts.sh:6953-6955` — Test 5c relies on `sleep 0.5` for the flock holder to grab the lock before the acquirer contends. On a heavily loaded system this is a timing dependency with no explicit handshake. Low confidence of actual flakiness given the 0.5 s margin; no fix required for this PR.

- **F-006** `dispatch-select-tick:207-212` — `dispatch-target-workers`'s exit code is silently discarded; only stdout is checked. `dispatch-target-workers` is documented to always print a numeric fallback before exiting non-zero, so the stdout guard covers the primary failure modes and a separate exit-code check would add no safety. Informational only.

- **SEC-006** `dispatch-acquire-lock:217` — Worst-case probe latency is `PROBE_TIMEOUT + 5 s` (the `-k 5` SIGKILL grace period). With shipped defaults that is 15 s. This is bounded, intentional, and correctly documented. The `FLOCK_TIMEOUT > PROBE_TIMEOUT` constraint accounts for it.

- **Design affirmed (positive findings):** Test 5b correctly discriminates missing `-k` flag via the SIGTERM-ignoring fake daemon + external `timeout 12` (F04). Test 5c's dual assertion (busy output AND lock not clobbered) directly catches the clobber-on-timeout bug (F05). The `FLOCK > PROBE` constraint and its justification (not `FLOCK > PROBE + 5`) are correct (F06). The `^[0-9]+$` regex correctly admits `TARGET_N=0` and test 5g verifies it (F07). Both teardown helpers now unset the new knob vars, preventing cross-test env pollution (F08).

---

### Dismissed

- **M-001** `dispatch-acquire-lock:364-366` — `FLOCK_TIMEOUT=0` passes `^[0-9]+$` then fails the `-le PROBE_TIMEOUT` inequality check, so the user sees "must exceed PROBE_TIMEOUT" rather than a "must be positive" message. The error is still actionable; `FLOCK_TIMEOUT=0` is always rejected with exit 2. Cosmetic imprecision, no correctness or security impact.

- **F-005** `dispatch-acquire-lock:281,350-379` — Knob validation runs after `mkdir -p $(dirname $LOCK_FILE)`, so a misconfigured knob creates the lock-file directory before exit 2. Directory creation is idempotent and harmless.

---

### False-positive (security)

- **SEC-001** `dispatch-acquire-lock:161-162,217,304` — Env var injection via `DISPATCH_LOCK_PROBE_TIMEOUT` / `DISPATCH_LOCK_FLOCK_TIMEOUT`. Both values are gated through anchored `^[0-9]+$` regexes that reject any non-digit content before the value reaches the command line, and double-quoted at every use site. _(A03:2021 Injection / Tampering)_

- **SEC-002** `dispatch-acquire-lock:356-379` — `PROBE_TIMEOUT=0` would disable GNU `timeout`'s bound. The startup guard at lines 360–362 explicitly rejects `PROBE_TIMEOUT=0` (`-lt 1`) with a descriptive error and exit 2 before `resolve_holder_state` is ever reached. _(A05:2021 Security Misconfiguration / Elevation of Privilege)_

- **SEC-003** `dispatch-acquire-lock:217` — fd 9 (the flock fd) is closed via `9>&-` on the `timeout` simple command before exec; `claude` starts with fd 9 already closed. The `-k 5` SIGKILL backstop eliminates the SIGTERM-ignoring-child hang. No fd leak path. _(A06:2021 / Denial of Service)_

- **SEC-004** `dispatch-acquire-lock:295-347` — Race condition / clobber on flock timeout. On `flock -w` timeout, the `if ! flock ...` branch executes `printf 'FLOCK_TIMEOUT'; exit 0` and the subshell terminates without reaching the read-check-write. Append-mode open (`exec 9>>"$LOCK_FILE"`) prevents truncation. Test 5c directly verifies the no-clobber property. _(A04:2021 Insecure Design / Tampering)_

- **SEC-007** `dispatch-select-tick:208-212` — `TARGET_N` guard: anchored `^[0-9]+$` regex handles empty string (fails), zero (passes — correct), and shell-special characters. `release_lock` is called before exit 2, preserving the lock invariant. _(A03:2021 Injection / Tampering)_

---

### Out-of-scope (pre-existing, not touched by this diff)

- **F-004** `dispatch-acquire-lock:421` — `--release` path uses an unbounded `flock 9`. The PR explicitly scopes the new timeout knobs to acquire/wait modes. Pre-existing behavior.

- **SEC-005** `dispatch-acquire-lock:217` — `${CLAUDE_AGENTS_CMD:-claude}` executes an operator-controlled binary. Pre-existing pattern; this diff only prepended `timeout -k 5 "$PROBE_TIMEOUT"` in front of it (a hardening improvement). _(A03:2021 Injection / Elevation of Privilege)_

PR #1501, opened 2026-06-13
budget-etl: bound CSV duplicate-ID suffix search with a per-ID counter

natb1:
Closes #1375

## Summary

`parseCSV` in `budget-etl/internal/parse/csv.go` assigns collision-free
synthetic suffixes (`<id>-2`, `<id>-3`, …) to duplicate transaction IDs. The
previous suffix search restarted at `n=2` for every duplicate row and
incremented until it found a free candidate. Under adversarial input where every
data row shares one ID, the k-th duplicate scanned `n=2..k`, giving O(M²) total
work — a ~200k-row CSV is ~2×10¹⁰ iterations, minutes of CPU, with no file-size
guard or iteration cap anywhere in the call chain.

This PR bounds the search with a per-ID counter (`nextSuffix`) keyed by the
**original** ID. Each duplicate's search starts from the stored counter instead
of from `n=2`, and the counter only ever advances, so total work collapses to
O(M) amortized. Because `used` only grows and `original` is fixed for the file, a
candidate skipped once never becomes free again — so starting from the stored
counter produces **identical output** to the previous restart-at-2 code. This is
the performance hardening that PR #1367 (the collision-free correctness fix) did
not deliver.

## Changes

- **`internal/parse/csv.go`** — Replace the unbounded inner suffix loop with a
  `nextSuffix map[string]int` that records the next index to try per original ID.
  The counter is keyed by `origID` captured *before* `txnID` is reassigned, and
  the candidate-free check still tests both `used` and `original` so a synthetic
  suffix never collides with a real later ID.
- **`internal/parse/csv_test.go`** — Add
  `TestParseCSV_AllDuplicateIDsPerformance`: a 50,000-row CSV all sharing one ID
  must parse in well under 2s (the unfixed code took minutes), with
  `len(Transactions) == 50000` and correct boundary suffixes (`X`, `X-2`, …,
  `X-50000`).

## Verification

- `go test ./...` — all packages pass, including the existing
  `TestParseCSV_DuplicateIDs` and `TestParseCSV_DuplicateIDCollision`
  (output unchanged) and the new performance test (~0.05s).
- `go vet ./...` — clean.

### DIFF
--- a/budget-etl/internal/parse/csv.go
+++ b/budget-etl/internal/parse/csv.go
@@ -72,6 +72,7 @@ func parseCSV(path string) (ParseResult, error) {
 
 	var txns []Transaction
 	used := make(map[string]struct{})
+	nextSuffix := make(map[string]int)
 	for i, row := range records[1:] {
 		if len(row) < 6 {
 			return ParseResult{}, fmt.Errorf("%s: line %d: expected 6 fields, got %d", path, i+2, len(row))
@@ -102,15 +103,22 @@ func parseCSV(path string) (ParseResult, error) {
 		}
 
 		if _, taken := used[txnID]; taken {
-			for n := 2; ; n++ {
-				candidate := fmt.Sprintf("%s-%d", txnID, n)
+			origID := txnID
+			n := nextSuffix[origID]
+			if n < 2 {
+				n = 2
+			}
+			for {
+				candidate := fmt.Sprintf("%s-%d", origID, n)
 				_, inUsed := used[candidate]
 				_, inOriginal := original[candidate]
+				n++
 				if !inUsed && !inOriginal {
 					txnID = candidate
 					break
 				}
 			}
+			nextSuffix[origID] = n
 		}
 		used[txnID] = struct{}{}
 
--- a/budget-etl/internal/parse/csv.go
+++ b/budget-etl/internal/parse/csv.go
@@ -105,6 +105,7 @@ func parseCSV(path string) (ParseResult, error) {
 		if _, taken := used[txnID]; taken {
 			origID := txnID
 			n := nextSuffix[origID]
+			// zero value means first duplicate for this ID; suffixes start at 2
 			if n < 2 {
 				n = 2
 			}
--- a/budget-etl/internal/parse/csv_test.go
+++ b/budget-etl/internal/parse/csv_test.go
@@ -3,6 +3,7 @@ package parse
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 )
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — #1375 (budget-etl all-duplicate-ID O(M²) fix)

**PR #1501** | No browser component (pure Go change)

### Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Performance under all-duplicate-ID input (AC #1 + AC #3) | shell-command | ✅ PASS |
| 2 | Existing duplicate-ID tests unchanged (AC #2) | shell-command | ✅ PASS |
| 3 | Static analysis of changed package | shell-command | ✅ PASS |
| 4 | Wall-clock timing reliability of performance assertion | — | ⏸ Deferred to office-hours |

**PASS: 3 / FAIL: 0 / SKIP: 0 / Deferred: 1**

### Details

**Item 1 — Performance test** (`TestParseCSV_AllDuplicateIDsPerformance`): 50,000-row all-duplicate CSV parsed in **0.04s** (well under the 2s bound). `len(Transactions) == 50000`, boundary suffixes `X` / `X-2` / `X-50000` correct. PASS.

**Item 2 — Existing correctness tests**: `TestParseCSV_DuplicateIDs` and `TestParseCSV_DuplicateIDCollision` both PASS — no regression in suffix assignment or collision avoidance.

**Item 3 — `go vet ./internal/parse/`**: Clean, no issues.

### Deferred to office-hours

**Item 4 — Wall-clock timing reliability**: Human judgment needed to confirm the ~0.04s observed result vs the 2s ceiling gives sufficient margin for CI under load. The observed 50× headroom (0.04s vs 2s) appears comfortable, but a human should confirm whether a pure wall-clock assertion is robust enough or whether a complexity-based guard would be more appropriate.

natb1:
## Review findings — #1375

### Fixed

- **`budget-etl/internal/parse/csv.go:108`** — Zero-value map semantics undocumented: the `if n < 2 { n = 2 }` guard relies on Go's map zero-value (absent key → 0) to detect the first duplicate for a given ID, but this is not obvious to a reader. Added an inline comment documenting the intent and the reliance on zero-value map semantics. (commit b847f1c6)

### Informational

- **`budget-etl/internal/parse/csv_test.go:172`** — Performance test uses a wall-clock deadline (`elapsed > 2*time.Second`) to distinguish O(M) from O(M²). Wall-clock timing is inherently environment-dependent (slow CI runner, resource contention). The 2 s bound provides a wide margin over the actual O(M) runtime (~50 ms), so flakiness is unlikely in practice — but the constraint is worth noting if the test ever becomes a flakiness source.

### Security

No attack surface changes detected (Go parse logic, no auth, network, or storage paths affected). CodeQL: no open alerts on this PR ref.

PR #1502, opened 2026-06-13
budget-etl: O(1) per-duplicate CSV suffix assignment via per-base counter

natb1:
Closes #1374

## Problem

`parseCSV` in `budget-etl/internal/parse/csv.go` deduplicates repeated
transaction IDs by appending a `-N` suffix. The previous inner loop restarted its
search at `n := 2` on every duplicate and had no upper bound. For K duplicate
rows of ID `X` plus N pre-existing suffixed IDs (`X-2` … `X-(N+1)`) in the file,
each duplicate rescanned from `X-2` past every already-taken slot, giving
O(K² + KN) total work. Because `csv.NewReader.ReadAll()` already loads the whole
file into memory, an adversarial statement file in the input directory could
drive quadratic CPU with no I/O relief.

## Fix

Replace the unbounded restart-from-2 inner loop with a per-base
next-free-suffix counter (`nextSuffix map[string]int`, keyed by the base
`txnID`). Each base's counter is monotonic and never rewinds, so every `X-k`
slot is passed at most once across that base's lifetime. Per-duplicate suffix
assignment becomes amortized O(1) — overall O(K + N) linear.

Three correctness invariants are preserved:

1. **Both** the `used[candidate]` and `original[candidate]` checks stay inside
   the loop, so a synthetic suffix can never collide with a real later ID
   (`TestParseCSV_DuplicateIDCollision`).
2. A zero-value guard clamps the counter up to `2` before the loop, since Go's
   `map[string]int` returns 0 for an unseen base.
3. `nextSuffix[base]` is recorded after the loop with the already-incremented
   `n`, so the counter advances monotonically.

The first-pass `original` collection, field parsing, and the `ParseResult` shape
are unchanged.

## Test

Adds `TestParseCSV_DuplicateIDStress` in the adversarial K + N shape: N = 10000
pre-seeded suffixed IDs followed by K = 10000 duplicate rows. It asserts
`parseCSV` succeeds, returns `K + N` distinct IDs, and completes under a 1-second
bound. At K = N = 10000 the old O(K² + KN) code does ~2×10⁸ map lookups (multiple
seconds → fails the bound) while the new O(K + N) code does ~4×10⁴ ops
(microseconds → passes; it ran in ~15 ms locally). So the test is both the
acceptance criteria's documentation **and** a real regression guard — reverting
the fix to the quadratic loop would turn it red. The two existing dedup tests
(`TestParseCSV_DuplicateIDs`, `TestParseCSV_DuplicateIDCollision`) stay green.

### DIFF
--- a/budget-etl/internal/parse/csv.go
+++ b/budget-etl/internal/parse/csv.go
@@ -106,13 +106,13 @@ func parseCSV(path string) (ParseResult, error) {
 			base := txnID
 			n := nextSuffix[base]
 			if n < 2 {
-				n = 2 // map zero-value guard: first collision starts at X-2, never X-0
+				n = 2 // map zero-value guard: first collision starts at X-2; stored values are always >= 3, so n is never 0 or 1
 			}
 			for {
 				candidate := fmt.Sprintf("%s-%d", base, n)
 				_, inUsed := used[candidate]
 				_, inOriginal := original[candidate]
-				n++
+				n++ // advance before acceptance check so nextSuffix stores the slot after the one just taken
 				if !inUsed && !inOriginal {
 					txnID = candidate
 					break
--- a/budget-etl/internal/parse/csv.go
+++ b/budget-etl/internal/parse/csv.go
@@ -72,6 +72,7 @@ func parseCSV(path string) (ParseResult, error) {
 
 	var txns []Transaction
 	used := make(map[string]struct{})
+	nextSuffix := make(map[string]int)
 	for i, row := range records[1:] {
 		if len(row) < 6 {
 			return ParseResult{}, fmt.Errorf("%s: line %d: expected 6 fields, got %d", path, i+2, len(row))
@@ -102,15 +103,22 @@ func parseCSV(path string) (ParseResult, error) {
 		}
 
 		if _, taken := used[txnID]; taken {
-			for n := 2; ; n++ {
-				candidate := fmt.Sprintf("%s-%d", txnID, n)
+			base := txnID
+			n := nextSuffix[base]
+			if n < 2 {
+				n = 2 // map zero-value guard: first collision starts at X-2, never X-0
+			}
+			for {
+				candidate := fmt.Sprintf("%s-%d", base, n)
 				_, inUsed := used[candidate]
 				_, inOriginal := original[candidate]
+				n++
 				if !inUsed && !inOriginal {
 					txnID = candidate
 					break
 				}
 			}
+			nextSuffix[base] = n // record next slot to try; counter never rewinds
 		}
 		used[txnID] = struct{}{}
 
--- a/budget-etl/internal/parse/csv_test.go
+++ b/budget-etl/internal/parse/csv_test.go
@@ -1,8 +1,10 @@
 package parse
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 )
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — #1374 budget-etl duplicate-ID suffix O(1) fix

**Result: 4 PASS · 0 FAIL · 0 SKIP · 1 deferred to office-hours**

No browser component (pure Go change). All machine-verifiable items ran via the shell-command lane.

### Items executed

| # | Title | Result |
|---|-------|--------|
| 1 | Existing dedup tests stay green (`TestParseCSV_DuplicateIDs`, `TestParseCSV_DuplicateIDCollision`) | ✅ PASS |
| 2 | Adversarial K+N stress test — sub-second runtime, K+N distinct IDs | ✅ PASS (14.5 ms for K=N=10000) |
| 3 | Stress test under 30s wall-clock timeout (regression guard) | ✅ PASS |
| 4 | Full `internal/parse` package suite | ✅ PASS |
| 5 | Suffix-assignment invariants under mixed real-vs-synthetic collisions | ⏳ deferred to office-hours |

### Deferred to office-hours

**Item 5 — Suffix-assignment invariants (needs-human-judgment)**

Requires human code inspection to confirm all three correctness invariants are preserved for *any* ordering of duplicates and pre-seeded suffix IDs, not just the test shapes:
1. Both `used[candidate]` and `original[candidate]` checked inside the loop before accepting a candidate.
2. Zero-value guard clamps `n` to `2` (first synthetic suffix is `X-2`, never `X-0`).
3. `nextSuffix[base]` recorded with already-incremented `n` — counter never rewinds.

Relevant files: `budget-etl/internal/parse/csv.go` lines 103–123, `csv_test.go` `TestParseCSV_DuplicateIDCollision` and `TestParseCSV_DuplicateIDStress`.

### Bugs found

None.

natb1:
## QA Summary (updated) — #1374 budget-etl duplicate-ID suffix O(1) fix

Supersedes the prior comment's "1 deferred to office-hours". The deferred human-judgment item (Item 5) was resolved on follow-up.

**Result: 5 PASS · 0 FAIL · 0 SKIP · 0 deferred · 0 bugs — clean pass.**

### Items

| # | Title | Result |
|---|-------|--------|
| 1 | Existing dedup tests stay green (`TestParseCSV_DuplicateIDs`, `TestParseCSV_DuplicateIDCollision`) | ✅ PASS |
| 2 | Adversarial K+N stress test — sub-second, K+N distinct IDs | ✅ PASS (14.5 ms at K=N=10000) |
| 3 | Stress test under 30s wall-clock timeout (regression guard) | ✅ PASS |
| 4 | Full `internal/parse` package suite | ✅ PASS |
| 5 | Suffix-assignment invariants under mixed real/synthetic collisions | ✅ PASS (resolved on follow-up) |

### Item 5 resolution

Confirmed by code inspection **and** an adversarial ordering test (run locally, not committed) that all emitted `TransactionID`s are globally unique for any interleaving of duplicates and pre-seeded suffix IDs:

| Input (file order) | Emitted IDs |
|---|---|
| `X, X, X-2` | `X, X-3, X-2` |
| `X, X, X, X-2, X-3` | `X, X-4, X-5, X-2, X-3` |
| `X, X, X-3` | `X, X-2, X-3` |
| `X-2, X-2, X` | `X-2, X-2-2, X` |
| `A, A, A, A-2, A-4` | `A, A-3, A-5, A-2, A-4` |
| `X, X-2, X, X-3, X` | `X, X-2, X-4, X-3, X-5` |

The monotonic `nextSuffix` counter is **provably equivalent** to the old restart-from-2 loop for output: `used`/`original` never shrink, so every slot between `2` and `nextSuffix[base]` is permanently taken — the old loop rescanning from 2 lands on the identical first-free slot. So the fix is a pure performance change with no behavioral drift in which suffix a duplicate receives (also pinned by `TestParseCSV_DuplicateIDCollision`).

### Bugs found

None.

natb1:
## Review (dispatch:review-fix)

### Fixed

- **`budget-etl/internal/parse/csv.go:108`** — Clarified the `n < 2` zero-value guard comment: the stored value after a collision is always `>= 3`, so `n` is never 0 or 1 (previously the comment only mentioned "never X-0"). (commit `c5ae9d25`)
- **`budget-etl/internal/parse/csv.go:115`** — Added inline comment on `n++` to make clear that the increment happens before the acceptance check, so `nextSuffix` stores the slot after the one just taken. (commit `c5ae9d25`)

### Informational

- **`budget-etl/internal/parse/csv_test.go:195`** — `TestParseCSV_DuplicateIDStress` asserts `elapsed > time.Second` as a hard failure. The test comment explains why K=N=10000 is chosen (the old quadratic code does ~2×10⁸ lookups vs ~4×10⁴ for the new code), making the 1-second bound very conservative, but technically the bound is host-dependent. No change warranted given the documented asymmetry.
- **`budget-etl/internal/parse/csv_test.go:192–195`** — Same wall-clock concern from the review pass: on a heavily loaded or very slow CI machine the 1-second bound could be a false alarm. The documented asymmetry (~10,000× difference) makes this extremely unlikely in practice.
- **`budget-etl/internal/parse/csv_test.go:165–213`** — The stress test verifies all 20,000 returned IDs are distinct but does not check which specific IDs the K duplicate rows were assigned. The existing `TestParseCSV_DuplicateIDs` and `TestParseCSV_DuplicateIDCollision` cover specific-assignment correctness; the stress test's role is regression-guarding the complexity bound.

- **`budget-etl/internal/parse/csv.go:107–121`** — Claim: `nextSuffix` skips inOriginal slots and might leave gaps. Not a real issue — the monotonic counter is intentional; skipping consumed slots is correct behavior, not a vulnerability.
- **`budget-etl/internal/parse/csv.go:111–119`** — Claim: first duplicate of a base ID with N pre-seeded originals does O(N) lookups on its first pass. This is the designed behavior — the counter advances from where it left off on subsequent duplicates, giving amortized O(1) per duplicate across all K rows. The adversarial test demonstrates this holds.
- **`budget-etl/internal/parse/csv_test.go:187–195`** — Claim: wall-clock timeout is environment-dependent and could cause flakiness. As above — the ~10,000× asymmetry makes this a theoretical concern, not a practical regression risk.

### Dismissed

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._ (No CodeQL alerts; no dependency changes.)

PR #1503, opened 2026-06-13
print: bump renderGen in leave() to cancel in-flight render()

natb1:
Closes #1383

## Problem

`SpreadController.leave()` tears down spread mode — it empties `this.spreads`,
removes the `.spread-page` slot elements from the DOM, and disconnects the resize
observer — but it never bumps `renderGen`. A `render()` call already past its
`spreads.length === 0` early-return guard and awaiting the left
`renderPageInto(spread.left, leftEl)` will, when it resumes, pass the
`gen !== this.renderGen` check (because `leave()` left `renderGen` untouched) and
proceed to `renderPageInto(spread.right, rightEl)` on the now-detached `rightEl`.
The appended `` lands on a detached node and is garbage-collected, but the
blob-decode work runs needlessly.

`render()` mirrors `pdf.ts`'s `renderGen` discipline: each `render()` claims a
generation via `++this.renderGen`, and any later caller that bumps the counter
marks an in-flight render stale so it bails at its next post-await gen check.
`goNext`/`goPrev` get this for free because they call `render()`. `leave()` is
the one teardown path that invalidates the in-flight render's target without
bumping the generation. This gap predates #1376 (which added the generation guard
to `render()`); #1376 neither caused nor fixed it.

## Fix

Increment `renderGen` as the first statement of `leave()`, so any in-flight
`render()` bails at its next post-await gen check — matching `render()`'s existing
`renderGen` discipline.

## Test

Adds one leave-during-render test to `spread-controller.test.ts`, modeled on the
existing #1279 concurrent-render test. It starts a `render()` that hangs after the
left slot, calls `leave()`, releases the stale render last, and asserts the right
slot's `renderPageInto` was never called (`toHaveBeenCalledTimes(1)`). With the
fix removed the test fails (received count 2), confirming it drives the change.

## Verification

- `npx vitest run --root print test/viewer/spread-controller.test.ts` — 12/12 pass.
- Sanity check: reverting the one-line fix makes the new test fail (count 2),
  restoring it passes again.
- `npm run build --prefix print` — type-check and build succeed.

### DIFF
--- a/print/src/viewer/spread-controller.ts
+++ b/print/src/viewer/spread-controller.ts
@@ -97,6 +97,8 @@ export class SpreadController {
   }
 
   leave(): number {
+    // Bump the generation counter to cancel any in-flight render() calls.
+    this.renderGen++;
     const spread = this.spreads.length > 0 ? this.spreads[this.spreadIndex]! : null;
     const currentPage = spread !== null
       && this.preSpreadPage >= spread.left
--- a/print/test/viewer/spread-controller.test.ts
+++ b/print/test/viewer/spread-controller.test.ts
@@ -233,7 +233,7 @@ describe("SpreadController", () => {
 
     const { controller } = makeController({ renderPageInto });
 
-    // enter(2) => index 1, two-page spread {left:2,right:3}; exercises both slots.
+    // enter(2) => index 1, two-page spread {left:2,right:3}; only the left slot is reached before leave() cancels the render.
     controller.enter(2);
 
     // Start a render that appends one left child then hangs on left renderPageInto.
--- a/print/test/viewer/spread-controller.test.ts
+++ b/print/test/viewer/spread-controller.test.ts
@@ -214,6 +214,43 @@ describe("SpreadController", () => {
     expect(controller.position).toBe("2");
   });
 
+  it("leave() during render() cancels the in-flight render before the right slot (#1383)", async () => {
+    // leave() bumps renderGen, so an in-flight render() that is awaiting its
+    // left renderPageInto must bail before calling the right renderPageInto.
+    let resolveFirstLeft: (() => void) | null = null;
+    const renderPageInto = vi
+      .fn()
+      .mockImplementationOnce((_page: number, target: HTMLElement) => {
+        // First (stale) left render: append synchronously, then hang.
+        target.appendChild(document.createElement("img"));
+        return new Promise<void>(resolve => {
+          resolveFirstLeft = resolve;
+        });
+      })
+      .mockImplementation(async (_page: number, target: HTMLElement) => {
+        target.appendChild(document.createElement("img"));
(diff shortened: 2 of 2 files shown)


natb1:
# QA Summary — #1383 / PR #1503

All 3 QA items are script-verifiable (shell-command lane). No browser walkthrough needed — the fix is an internal rendering-state change with no user-visible UI delta.

## Items

| # | Title | Result |
|---|-------|--------|
| 1 | Targeted spread-controller unit suite (12 tests) | ✅ PASS |
| 2 | `leave()` increments `renderGen` first | ✅ PASS |
| 3 | Full print vitest suite (362 tests, 24 files) | ✅ PASS |

## Detail

**Item 1** — `npx vitest run --root print test/viewer/spread-controller.test.ts` → 12/12 pass, including the new leave-during-render test asserting `renderPageInto` is called exactly once (right slot never reached after `leave()`).

**Item 2** — `awk '/leave\(\)/{…}' print/src/viewer/spread-controller.ts` confirms `this.renderGen++` is the first statement of `leave()`, before any DOM teardown.

**Item 3** — `npx vitest run --root print` → 362/362 pass. The blob-URL / DNS `DOMException` lines in output are pre-existing warnings from `epub.test.ts`'s happy-dom environment; no tests failed.

## Counts

- PASS: 3
- FAIL: 0
- SKIP: 0
- Deferred to office-hours: 0
- Bugs found: 0

natb1:
## Review — #1383

### Fixed

- **print/test/viewer/spread-controller.test.ts:236** — The comment `// enter(2) => index 1, two-page spread {left:2,right:3}; exercises both slots.` was copied verbatim from the concurrent-render test (#1279, line 191) where both slots are actually exercised. In this leave-during-render test the right slot is intentionally never reached, so "exercises both slots" was inaccurate. Replaced with `// enter(2) => index 1, two-page spread {left:2,right:3}; only the left slot is reached before leave() cancels the render.` (commit bfb257c7).

### Out-of-scope (security — pre-existing, diff did not touch)

- **print/src/viewer/spread-controller.ts:152** — After `leave()` bumps `renderGen` and removes spread slot elements, a render whose left `renderPageInto` call is already in flight will still complete that left call before the generation check fires. The blob-decode work for the left slot therefore runs to completion even on a stale render. This is a pre-existing efficiency gap in how `renderGen` discipline interacts with already-in-progress async work; it is not introduced by this diff.
- **print/src/viewer/spread-controller.ts:145–148** — `render()` queries `.spread-left` and `.spread-right` and casts to `HTMLElement` without null-checking before accessing `.innerHTML`. If the elements were removed from the DOM between the `querySelector` call and the assignment, this would throw. Pre-existing; the diff does not change this code path.

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

PR #1504, opened 2026-06-13
idbutil: validate maxBytes is a positive integer at construction time

natb1:
Closes #1418

## Summary

`createLruBlobCache` resolved `maxBytes` from config and used it directly as
both the eviction cap and the oversized-blob rejection threshold, without
validating it. Two misconfigurations silently corrupted the cache:

- `maxBytes: 0` — the `putEntry` guard `data.byteLength > maxBytes` rejects every
  non-empty blob, disabling the cache entirely.
- `maxBytes: -1` (any negative) — `data.byteLength > -1` is always false, so the
  oversized guard is bypassed; every blob then reaches `evictIfNeeded`, where the
  negative cap makes `totalSize + delta <= maxBytes` always false, triggering a
  full eviction loop on every put.

The `putEntry` oversized-blob guard is the core correctness fix delivered by
#1294 / PR #1414, so a misconfigured `maxBytes` that disables or inverts it is a
real regression risk.

## Change

Add a construction-time validation guard immediately after the `maxBytes`
resolution in `idbutil/src/lru-blob-cache.ts`:

```ts
if (!Number.isInteger(maxBytes) || maxBytes <= 0) {
  throw new Error(
    `LruBlobCache: maxBytes must be a positive integer, got ${maxBytes}`,
  );
}
```

Placing the guard after the `??` resolution means the default
(`DEFAULT_MAX_BYTES`, a positive integer) always passes, while an explicit `0` —
not nullish, so not replaced by the default — is correctly caught. This turns
silent misbehavior into a clear, value-bearing error at a public API boundary,
consistent with `.claude/rules/code-style.md`.

The guard is the only source change — no change to `putEntry` / `evictIfNeeded`
logic, the default value, the config type, or `connection.ts`.

## Tests

A new `describe("maxBytes validation", ...)` block in
`idbutil/test/lru-blob-cache.test.ts` covers the acceptance criteria: `maxBytes`
of `0`, `-1`, and `1.5` each throw with the offending value in the message, and a
valid `1024` constructs successfully with `cache.maxBytes === 1024` (existing
behavior preserved). The four existing `maxBytes`-related tests still pass,
confirming valid positive integers are unaffected.

Verification: `npx vitest run --root idbutil` (41 tests, all pass) and
`npm run lint --prefix idbutil` (clean).

### DIFF
--- a/idbutil/src/lru-blob-cache.ts
+++ b/idbutil/src/lru-blob-cache.ts
@@ -55,6 +55,11 @@ export interface LruBlobCache {
 
 export function createLruBlobCache(config: LruBlobCacheConfig): LruBlobCache {
   const maxBytes = config.maxBytes ?? DEFAULT_MAX_BYTES;
+  if (!Number.isInteger(maxBytes) || maxBytes <= 0) {
+    throw new Error(
+      `LruBlobCache: maxBytes must be a positive integer, got ${maxBytes}`,
+    );
+  }
 
   const { openDb, closeDb } = createDbConnection({
     name: config.name,
--- a/idbutil/test/lru-blob-cache.test.ts
+++ b/idbutil/test/lru-blob-cache.test.ts
@@ -375,7 +375,7 @@ describe("maxBytes validation", () => {
   it("throws at construction when maxBytes is -1", () => {
     expect(() =>
       createLruBlobCache({ name: uniqueDbName(), version: 1, maxBytes: -1 }),
-    ).toThrow(/got -1/);
+    ).toThrow("got -1");
   });
 
   it("throws at construction when maxBytes is a non-integer", () => {
--- a/idbutil/test/lru-blob-cache.test.ts
+++ b/idbutil/test/lru-blob-cache.test.ts
@@ -365,6 +365,32 @@ describe("onUpgrade hook", () => {
   });
 });
 
+describe("maxBytes validation", () => {
+  it("throws at construction when maxBytes is 0", () => {
+    expect(() =>
+      createLruBlobCache({ name: uniqueDbName(), version: 1, maxBytes: 0 }),
+    ).toThrow("got 0");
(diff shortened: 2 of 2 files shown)


natb1:
## QA summary — #1418 (`maxBytes` validation at construction time)

**Result: PASS** — all 6 items verified via shell-command lane (no browser component).

| # | Item | Result |
|---|------|--------|
| 1 | Full idbutil test suite (41 tests) | ✅ PASS |
| 2 | `maxBytes: 0` throws with value in message | ✅ PASS |
| 3 | `maxBytes: -1` throws with value in message | ✅ PASS |
| 4 | `maxBytes: 1.5` (non-integer) throws with value in message | ✅ PASS |
| 5 | Valid `maxBytes: 1024` constructs successfully, `cache.maxBytes === 1024` | ✅ PASS |
| 6 | Validation guard present in source | ✅ PASS |

**Counts:** 6 PASS · 0 FAIL · 0 SKIP · 0 deferred  
**Bugs:** none  

natb1:
## Review findings

### Fixed

- **`idbutil/test/lru-blob-cache.test.ts:376`** — The `maxBytes: -1` rejection test used a regex matcher (`.toThrow(/got -1/)`) inconsistent with the sibling `maxBytes: 0` and `maxBytes: 1.5` tests, which use plain string matchers. Changed to `.toThrow("got -1")` for uniformity. (fix commit: 955e3eea)

### Informational

- **`idbutil/test/lru-blob-cache.test.ts:368`** — The `maxBytes` validation test suite covers `0`, `-1`, and `1.5`, but does not exercise `NaN` or `Infinity`. `Number.isInteger(NaN)` and `Number.isInteger(Infinity)` both return `false`, so the guard already rejects them — but no test documents this. These are valid inputs to consider for future coverage.

PR #1508, opened 2026-06-13
ci: anchor prod-deploy DIFF_BASE to last successfully deployed commit

### LINKED ISSUE #1446: prod-deploy concurrency group drops intermediate merges in burst scenarios; anchor DIFF_BASE to last deployed commit
## Problem

The concurrency group serializes runs with \`cancel-in-progress: false\`, which correctly avoids cancelling an in-progress deploy. However, GitHub still cancels any *pending* run when a newer one queues. In a merge burst — merge A deploying, B and C landing while A runs — B is queued then evicted by C. C's diff base is \`github.event.before = B.sha\`, so C deploys \`B.sha...C.sha\`. The range \`A.sha...B.sha\` (B's changes) is never deployed by anyone. Apps changed only in that dropped range silently skip production until a future push happens to touch them again.

The per-push \`before...sha\` diff base is what creates this gap: each run is only responsible for its own push window, so a dropped run permanently skips its window.

**File:** \`.github/workflows/prod-deploy.yml\`, \`DIFF_BASE: \${{ github.event.before }}\`

## Recommended fix

Anchor \`DIFF_BASE\` to the last successfully deployed commit (a git tag or deploy marker written by the workflow on success, e.g. \`git tag -f last-prod-deploy && git push -f origin last-prod-deploy\`) rather than \`github.event.before\`. Each run then diffs from \`last-prod-deploy...HEAD\`, so a dropped intermediate run is automatically recovered by the next one that completes successfully. Keep \`cancel-in-progress: false\` — cancelling a mid-flight deploy risks a half-deployed state.

## Context

Out of scope for PR #1422 (which addresses the double-trigger and nondeterministic diff base for #1306).

## Acceptance criteria

- [ ] The workflow writes a \`last-prod-deploy\` git tag (or equivalent persistent marker) after a successful deploy run
- [ ] \`DIFF_BASE\` is set from the \`last-prod-deploy\` tag rather than \`github.event.before\`
- [ ] In a merge burst where intermediate runs are cancelled-pending, the surviving run diffs from the last successfully deployed commit and covers all changes since that point
- [ ] \`cancel-in-progress: false\` is preserved to avoid partial deploys
- [ ] A \`workflow_dispatch\` run still deploys all apps (the \`DEPLOY_ALL\` path is unaffected)
- [ ] CI passes after the change

natb1:
Closes #1446

## Problem

`.github/workflows/prod-deploy.yml` serializes production deploys with a
`prod-deploy` concurrency group and `cancel-in-progress: false`. That protects an
in-progress deploy, but GitHub still evicts any *pending* run when a newer one
queues. In a merge burst (A deploying; B then C landing while A runs), B is
queued then cancelled by C. Because each run's diff base was its own
`github.event.before`, the surviving run C deployed only `B.sha...C.sha` — the
range `A.sha...B.sha` (B's changes) was deployed by nobody. Apps changed only in
that dropped window silently skipped production until a later push happened to
touch them again.

The root cause is the per-push `before...sha` diff base: each run was responsible
only for its own push window, so a dropped run permanently skipped its window.

## Fix

Anchor the diff base to the **last successfully deployed commit**, recorded as a
force-updated `last-prod-deploy` git tag written by the workflow after a
successful run. Each run diffs `last-prod-deploy...HEAD`, so a dropped
intermediate run is automatically recovered by the next run that completes
successfully. `cancel-in-progress: false` is preserved — cancelling a mid-flight
deploy risks a half-deployed state.

### Changes to the `deploy-and-smoke` job (`prod-deploy.yml`)

- **Resolve deploy base** step (new, before the deploy step): for a
  `workflow_dispatch` run, deploy all apps; otherwise look up the marker on the
  remote with `git ls-remote origin refs/tags/last-prod-deploy` and set
  `DIFF_BASE` to that SHA. When the tag is absent on origin (the first run after
  this ships, and only then), bootstrap with a full deploy — a legitimate
  first-run state, not an error, and the safest bootstrap since it skips no app.
  The remote lookup is deliberate: `actions/checkout@v4` with `fetch-depth: 0`
  does not necessarily fetch the tag *ref* (`fetch-tags` defaults false), so a
  local `rev-parse` could silently miss it and degrade the workflow to
  deploy-all forever while staying green. The marker's *commit* is already in
  local history, so the diff works without the ref.
- **Removed** the inline `DEPLOY_ALL`/`DIFF_BASE` step-`env:` entries — a
  step-level `env:` key shadows anything a prior step writes to `$GITHUB_ENV`, so
  these had to go for the resolve step's output to take effect. Every secret env
  entry is untouched.
- **Record last-prod-deploy marker** step (new, after the deploy step): force-tags
  `last-prod-deploy` at `$GITHUB_SHA` and pushes it. No `if: always()`, so it runs
  only when the deploy step succeeded — a per-app deploy/smoke failure leaves the
  marker put and the next run idempotently re-diffs from the old marker.
  `if: github.ref == 'refs/heads/main'` prevents a non-main `workflow_dispatch`
  from poisoning the marker.
- **`permissions: contents: write`** at the job level so the tag push works with
  the persisted `GITHUB_TOKEN` (precedent: `budget-etl-release.yml`). The
  workflow-level `contents: read` stays, keeping `cleanup-preview` read-only.

### `run-all-prod-deploy-smoke.sh`

Comment-only: synced the explanatory comment and the `DIFF_BASE` guard's error
message to the new marker-based contract. The `DEPLOY_ALL`/`DIFF_BASE` control
flow is unchanged — only how those vars are populated (in the workflow) changed.

## Verification

No actionlint config or local workflow/script test harness exercises this path,
so end-to-end coverage cannot run locally; CI on this PR plus the first
post-merge run are authoritative. Verified locally: YAML well-formedness and
step-indentation of `prod-deploy.yml`, and the marker query
(`git ls-remote origin refs/tags/last-prod-deploy` returns empty today — the
bootstrap case).

**Note for whoever watches the first production deploy after merge:** it takes
the bootstrap deploy-all path because no `last-prod-deploy` tag exists yet — this
is expected, not a regression. That run creates the tag; every subsequent run
diffs from it.

Out of scope for #1422 (which fixed the double-trigger and nondeterministic diff
base); this is the follow-up #1422 deferred.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/run-all-prod-deploy-smoke.sh
+++ b/.claude/skills/dispatch-propagate/scripts/run-all-prod-deploy-smoke.sh
@@ -7,14 +7,16 @@ REPO_ROOT=$(git rev-parse --show-toplevel)
 # shellcheck source=lib.sh
 source "$SCRIPT_DIR/lib.sh"
 
-# The workflow checks out github.sha and passes DIFF_BASE=github.event.before, so
-# change detection diffs the merge's own before...sha range, deterministic
-# regardless of run order.
-# DEPLOY_ALL=1 deploys every app with a hosting target (for workflow_dispatch).
+# The workflow resolves DIFF_BASE from the last-prod-deploy marker (the last
+# successfully deployed commit), so change detection diffs from there. A dropped
+# intermediate run is recovered by the next successful run, which re-covers the
+# missed window.
+# DEPLOY_ALL=1 deploys every app with a hosting target; the workflow sets it for
+# workflow_dispatch and for the first-run bootstrap when no marker exists yet.
 if [ "${DEPLOY_ALL:-}" = "1" ]; then
   CHANGED_APPS=$("$SCRIPT_DIR/get-changed-apps.sh" --all)
 else
-  : "${DIFF_BASE:?DIFF_BASE required (set to github.event.before) for change detection}"
+  : "${DIFF_BASE:?DIFF_BASE required (set from the last-prod-deploy marker) for change detection}"
   CHANGED_APPS=$("$SCRIPT_DIR/get-changed-apps.sh" --base "$DIFF_BASE")
 fi
 
--- a/.github/workflows/prod-deploy.yml
+++ b/.github/workflows/prod-deploy.yml
@@ -27,6 +27,8 @@ jobs:
     concurrency:
       group: prod-deploy
       cancel-in-progress: false
+    permissions:
+      contents: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
@@ -40,6 +42,21 @@ jobs:
         env:
           FIREBASE_SERVICE_ACCOUNT_JSON: ${{ secrets.FIREBASE_SERVICE_ACCOUNT }}
         run: .github/scripts/firebase-auth.sh
+      - name: Resolve deploy base
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "workflow_dispatch: deploying all apps"
+            echo "DEPLOY_ALL=1" >> "$GITHUB_ENV"
+          else
+            SHA=$(git ls-remote origin refs/tags/last-prod-deploy | cut -f1)
+            if [ -n "$SHA" ]; then
+              echo "Diffing from last-prod-deploy ($SHA)"
+              echo "DIFF_BASE=$SHA" >> "$GITHUB_ENV"
+            else
+              echo "No last-prod-deploy tag on origin; bootstrapping with full deploy"
+              echo "DEPLOY_ALL=1" >> "$GITHUB_ENV"
+            fi
+          fi
       - name: Deploy to production and run smoke tests
         env:
           GA_MEASUREMENT_ID_AUDIO: ${{ secrets.GA_MEASUREMENT_ID_AUDIO }}
@@ -51,9 +68,12 @@ jobs:
           VITE_RECAPTCHA_SITE_KEY: ${{ secrets.RECAPTCHA_SITE_KEY }}
           CLOUDFLARE_ZONE_ID: ${{ secrets.CLOUDFLARE_ZONE_ID }}
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-          DEPLOY_ALL: ${{ github.event_name == 'workflow_dispatch' && '1' || '' }}
-          DIFF_BASE: ${{ github.event.before }}
         run: .claude/skills/dispatch-propagate/scripts/run-all-prod-deploy-smoke.sh
+      - name: Record last-prod-deploy marker
+        if: github.ref == 'refs/heads/main'
+        run: |
+          git tag -f last-prod-deploy "$GITHUB_SHA"
+          git push -f origin refs/tags/last-prod-deploy
       - name: Cleanup credentials
         if: always()
         run: .github/scripts/firebase-auth-cleanup.sh
--- a/.github/workflows/prod-deploy.yml
+++ b/.github/workflows/prod-deploy.yml
@@ -70,7 +70,7 @@ jobs:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
         run: .claude/skills/dispatch-propagate/scripts/run-all-prod-deploy-smoke.sh
       - name: Record last-prod-deploy marker
-        if: github.ref == 'refs/heads/main'
+        if: success() && github.ref == 'refs/heads/main'
         run: |
           git tag -f last-prod-deploy "$GITHUB_SHA"
           git push -f origin refs/tags/last-prod-deploy


natb1:
# QA Summary — Issue #1446 / PR #1508

**Change:** Anchor `prod-deploy` `DIFF_BASE` to `last-prod-deploy` git tag (last successfully deployed commit) instead of `github.event.before`, so burst-dropped intermediate runs are recovered by the next successful deploy.

**Files changed:** `.github/workflows/prod-deploy.yml`, `.claude/skills/dispatch-propagate/scripts/run-all-prod-deploy-smoke.sh`

**Browser component:** None (CI/workflow-only change)

---

## Results

| # | Item | Result |
|---|------|--------|
| 1 | Concurrency safety preserved (`cancel-in-progress: false`) | ✅ PASS |
| 2 | `DIFF_BASE` resolved from `last-prod-deploy` tag, not `github.event.before` | ✅ PASS |
| 3 | `workflow_dispatch` still sets `DEPLOY_ALL=1` (deploy-all path unaffected) | ✅ PASS |
| 4 | First-run bootstrap (no marker on origin) sets `DEPLOY_ALL=1` | ✅ PASS |
| 5 | `contents: write` permission + marker-record step gated on `refs/heads/main` | ✅ PASS |
| 6 | Smoke script logic intact; `DIFF_BASE` guard message updated to reference marker | ✅ PASS |
| 7 | YAML structural order valid; inline `run:` scripts pass `bash -n` syntax check | ✅ PASS |
| 8 | End-to-end burst recovery on a real GitHub Actions run | ⏳ Deferred |

**Counts:** 7 PASS · 0 FAIL · 0 SKIP · 1 deferred to office-hours

---

## Deferred to office-hours

**Item 8 — End-to-end burst recovery on a real GitHub Actions run** (`needs-human-judgment`)

Requires observing an actual merge burst on `main` after merge: intermediate runs queued behind `cancel-in-progress: false`, the surviving run resolving `DIFF_BASE` from the prior `last-prod-deploy` tag, deploying all changed apps since that point, and advancing the marker. Cannot be reproduced locally — requires actual GitHub Actions execution (tag push permission, concurrency queueing, cross-run state).

**Expected first-run behavior post-merge:** the bootstrap path fires (no `last-prod-deploy` tag on origin yet), deploying all apps. That run creates the tag; subsequent runs diff from it.

natb1:
## Review findings — #1446

### Fixed

- **`prod-deploy.yml:73` — Explicit `success()` gate on marker step** (`review`): The `Record last-prod-deploy marker` step used `if: github.ref == 'refs/heads/main'` without an explicit `success()` check. GitHub's default step condition is `success()`, but omitting it makes the load-bearing guard implicit and fragile to future edits. Changed to `if: success() && github.ref == 'refs/heads/main'`. Fix commit: ce476309.

### Informational

- **`prod-deploy.yml:74-76` — No retry on marker push** (`code-review`): The `Record last-prod-deploy marker` step runs `git push -f origin refs/tags/last-prod-deploy` with no retry or error recovery. A transient network blip here leaves the tag stale, causing the next deploy to over-deploy. Acceptable for now given the bootstrap fallback, but worth noting.

- **`prod-deploy.yml:45-59` — Inline bash for deploy-base resolution** (`code-review`): The `Resolve deploy base` logic is implemented as inline shell in the YAML step rather than a dedicated script. This limits testability and makes the step harder to version independently. A follow-up to extract it into a script would improve maintainability.

- **`prod-deploy.yml:30-31` — `contents: write` scoped to full job** (`red-team`): Finding flagged that job-level `contents: write` exposes the `GITHUB_TOKEN` during the deploy step's unversioned `npx` invocations (firebase-tools, tsx, playwright), enabling a supply-chain compromise to push arbitrary commits. Split-job architecture recommended.

  Adversarial verify: **1 upheld / 1 refuted → refuted.** The refuting skeptic noted that the deploy step already holds `FIREBASE_SERVICE_ACCOUNT`, `CLOUDFLARE_API_TOKEN`, and production Firestore write access — a supply-chain compromise in that step can serve malicious JS to all production users before `contents: write` is ever relevant. The marginal risk from `contents: write` (repo push, blocked from triggering new workflow runs by `GITHUB_TOKEN` policy; branch pushes gated by branch protection) is conditional and not an independently exploitable escalation above the already-accessible production-deploy credentials. Valid least-privilege observation — not a blocking fix for this PR.

### False-positive (security)

- **`prod-deploy.yml:47` — `github.event_name` shell interpolation** (`red-team`): `${{ github.event_name }}` is injected into a shell `run:` block via string comparison. In GitHub Actions, context values are sanitized by the runner before shell substitution; `event_name` is a controlled enum (`push`, `workflow_dispatch`, etc.) with no injection surface. Not a vulnerability.

- **`prod-deploy.yml:51` — `git ls-remote` SHA in DIFF_BASE** (`red-team`, `code-review`): The SHA from `git ls-remote origin refs/tags/last-prod-deploy | cut -f1` could be attacker-controlled if an adversary pushes the tag. The tag is protected: only the `GITHUB_TOKEN` of a successful workflow run (with `contents: write`) can push it, and that token is scoped to the repo — no external actor can push it.

- **`prod-deploy.yml:54` — SHA written to GITHUB_ENV unsanitized** (`red-team`): `echo "DIFF_BASE=$SHA"` writes the git SHA to `$GITHUB_ENV`. Git SHAs are 40-character hex strings — no shell metacharacter injection possible. Not a vulnerability.

- **`prod-deploy.yml:75-76` — Mutable force-pushed tag** (`red-team`): `git tag -f last-prod-deploy` force-pushes a shared mutable deployment anchor. The concern is that any actor triggering a successful deploy could advance the marker. This is load-bearing mutable state by design — the marker records the last successful deploy; force-updating it on each success is the intended contract, not an attack surface.

### Out-of-scope (security)

- **`prod-deploy.yml:73-76` — Non-main `workflow_dispatch` could still push tag if condition races** (`red-team`): The `if: github.ref == 'refs/heads/main'` (now `success() && github.ref == 'refs/heads/main'`) guards the tag push. The finding concerned a condition race in non-main `workflow_dispatch` runs. The guard is evaluated per-step after context resolution — no race; the `github.ref` check is deterministic. Pre-existing analysis note, not introduced by this diff.

PR #1510, opened 2026-06-13
idbutil: bound the openDb onblocked wait with a timeout

natb1:
Closes #1420

## Problem

`createDbConnection` in `idbutil/src/connection.ts` memoizes a versioned
IndexedDB connection in a closure-scoped `dbPromise`. The old `onblocked` handler
only logged a `console.warn` and never settled the `openDb()` promise. If a
same-origin tab held a connection whose `onversionchange` never closed it, the
open stayed pending forever — and because `dbPromise` cached that pending
promise, every subsequent `openDb()` returned the same stuck promise. The whole
data layer (`storeParsedData`, `getAll`, `get`, `put`, `deleteRecord`,
`clearAll`, plus the `print`, `audio`, and `local-first` callers) would freeze
with no timeout or recovery.

## Change

Adds an optional `blockedTimeoutMs` field to `DbConnectionConfig` (default
30000 ms). When `onblocked` fires and the open does not settle within the
deadline, `openDb()` rejects with a descriptive error ("…timed out waiting for
another tab to close its connection. Close other tabs using this app and
retry.") and `dbPromise` is nulled so the next call starts a fresh attempt. The
timer is cancelled when the open settles normally (`onsuccess`/`onerror`).

The default is resolved with `??` (not `||`) so a caller passing `0` is honored.
`onsuccess` clears the timer and, if the open already timed out, closes the
orphaned connection and returns *before* wiring `db.onclose`/`db.onversionchange`
— those handlers write the shared `dbPromise` closure variable, so a
late-arriving orphaned connection must not be allowed to null a freshly cached
promise belonging to a different healthy connection.

The field is optional, so every existing caller (which all pass plain object
literals) keeps the 30000 ms default with no changes.

## Tests

`idbutil/test/connection.test.ts`:

- **Timeout + cache-null**: a never-closing blocking connection makes `openDb()`
  reject with the timeout error (asserts the message carries the DB name and the
  close-tabs guidance), and a subsequent `openDb()` returns a *fresh* promise
  (proving `dbPromise` was nulled) — asserted while the first connection is still
  blocking, so the check is race-free.
- **Timer cancelled on success**: with a generous `blockedTimeoutMs`, an open
  that is briefly blocked and then unblocked resolves to a v2 database rather
  than timing out, proving the timer is cleared on `onsuccess`.

The existing "onblocked does not reject an open that later succeeds" test passes
no `blockedTimeoutMs`, covering the default path.

All 39 idbutil tests pass; `eslint src/ test/` is clean.

### DIFF
--- a/idbutil/src/connection.ts
+++ b/idbutil/src/connection.ts
@@ -2,6 +2,11 @@ interface DbConnectionConfig {
   name: string;
   version: number;
   onUpgrade: (db: IDBDatabase, oldVersion: number) => void;
+  /**
+   * Bounds how long an open waits for a blocking connection to close before
+   * rejecting. Defaults to 30000 ms.
+   */
+  blockedTimeoutMs?: number;
 }
 
 /**
@@ -19,12 +24,15 @@ export function createDbConnection(config: DbConnectionConfig): {
   if (!Number.isInteger(config.version) || config.version < 1) {
     throw new Error(`DbConnectionConfig.version must be a positive integer, got ${config.version}`);
   }
+  const blockedTimeoutMs = config.blockedTimeoutMs ?? 30000;
   let dbPromise: Promise<IDBDatabase> | null = null;
 
   function openDb(): Promise<IDBDatabase> {
     if (!dbPromise) {
       dbPromise = new Promise((resolve, reject) => {
         let upgradeError: unknown;
+        let blockedTimer: ReturnType<typeof setTimeout> | undefined;
+        let settledByTimeout = false;
         const request = indexedDB.open(config.name, config.version);
         request.onupgradeneeded = (event) => {
           try {
@@ -35,6 +43,11 @@ export function createDbConnection(config: DbConnectionConfig): {
           }
         };
         request.onsuccess = () => {
+          clearTimeout(blockedTimer);
+          if (settledByTimeout) {
+            request.result.close();
+            return;
+          }
           const db = request.result;
           db.onclose = () => { dbPromise = null; };
           db.onversionchange = () => { db.close(); dbPromise = null; };
@@ -44,8 +57,17 @@ export function createDbConnection(config: DbConnectionConfig): {
           console.warn(
             `IndexedDB upgrade of "${config.name}" is blocked by another connection; waiting for it to close.`,
           );
+          if (blockedTimer === undefined) {
+            blockedTimer = setTimeout(() => {
+              settledByTimeout = true;
+              dbPromise = null;
+              reject(new Error(
+                `IndexedDB upgrade of "${config.name}" timed out waiting for another tab to close its connection. Close other tabs using this app and retry.`,
+              ));
+            }, blockedTimeoutMs);
+          }
         };
-        request.onerror = () => { dbPromise = null; reject(upgradeError ?? request.error); };
+        request.onerror = () => { clearTimeout(blockedTimer); dbPromise = null; reject(upgradeError ?? request.error); };
       });
     }
     return dbPromise;
--- a/idbutil/src/connection.ts
+++ b/idbutil/src/connection.ts
@@ -24,6 +24,16 @@ export function createDbConnection(config: DbConnectionConfig): {
   if (!Number.isInteger(config.version) || config.version < 1) {
     throw new Error(`DbConnectionConfig.version must be a positive integer, got ${config.version}`);
   }
+  if (
+    config.blockedTimeoutMs !== undefined &&
+    (typeof config.blockedTimeoutMs !== "number" ||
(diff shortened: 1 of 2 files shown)


natb1:
## QA Summary — #1420 idbutil: openDb() blocked-timeout

**Result: ✅ All items passed — no bugs found.**

### Items executed (shell-command lane, 6 items)

| # | Title | Result |
|---|-------|--------|
| 1 | openDb() rejects with descriptive timeout error | ✅ PASS |
| 2 | dbPromise is nulled after timeout so a retry starts fresh | ✅ PASS |
| 3 | Timer is cancelled when the open succeeds before the deadline | ✅ PASS |
| 4 | blockedTimeoutMs is optional; existing callers default to 30000 ms | ✅ PASS |
| 5 | Full idbutil test suite passes | ✅ PASS |
| 6 | idbutil typechecks cleanly | ✅ PASS |

### Counts

- PASS: 6
- FAIL: 0
- SKIP: 0
- Deferred to office-hours: 0
- Bugs found: 0

### Details

**Items 1–3, 5** verified by `npx vitest run --project idbutil` → 2 test files, **39 tests passed**. The two new timeout tests cover the exact acceptance criteria: `p1` rejects with a message containing the DB name and "Close other tabs", and a subsequent `openDb()` returns a fresh promise (`p2 \!== p1`), confirming `dbPromise` was nulled. The timer-cancelled test confirms `onsuccess` clears the timeout so a briefly-blocked open resolves normally.

**Item 4** verified by grep: `blockedTimeoutMs?: number` at `idbutil/src/connection.ts:9`, default applied via `?? 30000` at line 27. All existing callers pass plain object literals with no `blockedTimeoutMs` field and keep the 30 000 ms default.

**Item 6** — idbutil has no `tsconfig.json`; CI's `run-typecheck.sh` skips it for the same reason on both `origin/main` and HEAD (baseline also fails → skip). Not a regression introduced by this PR. CI status: passing.

No browser component detected (changes limited to `idbutil/src/connection.ts` and `idbutil/test/connection.test.ts`); no server was started.

natb1:
## Review findings — #1420

### Fixed (commit 6436ba2b)

- **`idbutil/src/connection.ts:27`** — `blockedTimeoutMs` accepted non-positive and non-finite values (`??` only rejects null/undefined). Added eager validation at `createDbConnection` call time; throws a descriptive error on invalid input. Added matching input-validation tests. _(code-review, input-validation)_

- **`idbutil/test/connection.test.ts:216–247`** — Test B ("timer cancelled on success") did not verify its stated regression: it only checked that the open resolved, not that a leaked blocked timer would not later null `dbPromise`. Rewrote with vitest fake timers (`vi.useFakeTimers()`/`vi.useRealTimers()`): drives the open to completion, advances 250 ms past the 200 ms deadline, then asserts `openDb()` returns the same memoized promise — proving the timer was cleared and did not null `dbPromise`. _(review)_

### Informational

- **`idbutil/test/connection.test.ts:213`** — Cleanup line `await p2.then((db) => db.close(), () => {})` uses an empty error handler that silently swallows any rejection of `p2`. Not a product concern but the swallow could mask test failures during local iteration. _(code-review)_

- **`idbutil/src/connection.ts:76`** — If `closeDb()` is called while `openDb()` is waiting for a blocked upgrade that subsequently times out, the timeout rejection propagates to the `closeDb()` caller. This is an edge case with no caller today; noted for awareness. _(code-review)_

- **`idbutil/test/connection.test.ts:205–213`** — Test A's comment accurately documents that fake-indexeddb does not re-fire `onblocked` for a second `openDb()` while the DB is already blocked, so `p2` has no timeout timer of its own. The comment is informative, not a gap. _(review)_

- **`idbutil/src/connection.ts:57–59`** — `console.warn` fires on every `onblocked` event. The IndexedDB spec can fire `onblocked` more than once; repeated warnings are harmless here since the timer guard (`if (blockedTimer === undefined)`) prevents duplicate timers. _(review)_

- **`idbutil/src/connection.ts:61–68`** — After a timeout nulls `dbPromise` and rejects, the original `IDBOpenDBRequest` remains pending and can still fire `onsuccess`/`onerror` later. The `settledByTimeout` flag guards against the orphaned connection mutating a fresh `dbPromise`, so this is handled; noted for completeness. _(review)_

### False-positive (security)

- **`idbutil/src/connection.ts:64–65`** — `config.name` (the IndexedDB database name) is embedded in the timeout error message. The database name is a compile-time constant set by the application, not user-controlled input, so this is not an injection or data-exposure issue. _(input-validation, data-exposure)_

### Deferred

_None._

### Out-of-scope (security)

_None._ CodeQL: no open alerts. npm: no dependency changes.

PR #1457, opened 2026-06-13
dispatch: revive the input-block hook for <N>-slug workers; name office-hours sessions office-hours-<N>

natb1:
Closes #1311

## Problem

`.claude/hooks/dispatch-input-block.sh` is wired to the harness events through
which a background session blocks on user input (PreToolUse on
`ExitPlanMode`/`AskUserQuestion`, Notification `permission_prompt`, Elicitation).
On fire it parks the issue with `dispatch:office-hours` and spawns the next tick
(the baton pass), so a worker that hits a prompt does not strand its worktree
silently.

The hook was **dead code**. Its discriminator only acted when the job name
started with `dispatch-`, but since the headless-tick migration nothing spawns a
`dispatch-*` session — phase workers are named `<N>-slug` and the autonomous tick
is a headless systemd script, not a Claude session. So a phase worker that hit a
permission prompt parked silently: no label, no baton pass, and
`dispatch-select-target` skipped the worktree indefinitely.

The fix was entangled with a second drift: **office-hours sessions and phase
workers shared the `<N>-slug` name shape**, so nothing could tell them apart by
name — yet three consumers (`restore-dispatch-skill.sh`, `dispatch-stop.sh`'s
discriminator, `office-hours/SKILL.md`) already assumed office-hours sessions
were distinctly named. There is no durable non-name signal to disambiguate them,
so distinguishability required a distinct name.

## Approach

Name office-hours sessions `office-hours-<N>` (the name the codebase already
expected). A single `^[0-9]+-` discriminator then cleanly partitions the
session kinds:

- **phase workers** and **parse-job sessions** (`<N>-slug`) — autonomous
  sessions that should be parked on an unexpected block → `^[0-9]+-` matches → the
  hook fires.
- **office-hours sessions** (`office-hours-<N>`) — the interactive human session
  that legitimately blocks on input → no leading digit → excluded → never parked.

The rename makes `restore-dispatch-skill.sh:91` (`^office-hours-[0-9]+$`),
`dispatch-stop.sh`'s `^[0-9]+-` discriminator, and `office-hours/SKILL.md` all
correct at once.

## Units

1. **Name office-hours sessions `office-hours-<N>`** — `dispatch-spawn-office-hours`
   names the spawned session `office-hours-$ISSUE_NUM` instead of the worktree
   basename; stale commentary in the spawn and entry scripts updated.
2. **Extend occupancy detection** — `worktree_has_live_session` now queries both
[...]

PR #1511, opened 2026-06-13
dispatch-input-block: short read timeout so the hook never stalls 1s

natb1:
Closes #1491

## Problem

`.claude/hooks/dispatch-input-block.sh` consumed the hook payload with:

```bash
if read -t 1 -d '' PAYLOAD; then :; fi
```

`-d ''` sets the read delimiter to NUL (`\x00`). JSON hook payloads never
contain NUL, so on a stdin that stays open the `read` always runs the full
1-second timeout. The hook is wired (`.claude/settings.json`) to three
worker-blocking events — `PreToolUse` (ExitPlanMode / AskUserQuestion),
`Notification` (permission_prompt), and `Elicitation` — so every firing in every
`<N>-`slug worker session paid a guaranteed ~1-second stall. The defect
pre-existed PR #1457; #1457 only made it more visible.

## Fix

Replace the read with:

```bash
if read -rt 0.1 PAYLOAD; then :; fi
```

- **Newline delimiter** (the issue's "or equivalent") returns instantly when a
  newline-terminated payload arrives.
- **0.1 s timeout** caps the wait on an *open, idle* stdin at an imperceptible
  0.1 s. This matters because stdin is left **open** (not EOF-closed) for these
  events — the very existence of the 1-second stall proves it — and whether the
  payload is newline-terminated is undocumented. A plain `read -t 1 PAYLOAD`
  removes the stall only in the newline-terminated case; the short timeout
  sidesteps that undocumented assumption and kills the stall in **every** case.
- **`-r`** keeps backslashes in the JSON intact so the downstream `jq` parse
  cannot be corrupted.

Both knobs — newline delimiter and a shorter timeout — are options the issue
author explicitly listed, so this honors the issue's intent more reliably than
its literal primary suggestion. The `if read …; then :; fi` wrapper and the
downstream `notification_type` filter are unchanged.

## Regression test

Adds one test to the `=== dispatch-input-block ===` section of
`test-dispatch-scripts.sh` asserting the hook does **not** stall on an **open,
non-newline-terminated** stdin — the dimension every existing test missed (all
prior invocations use `< /dev/null` or `printf … | hook`, both reaching EOF
immediately, so none could distinguish the buggy `-d ''` / a `-t 1` regression
from the fix). The old code is killed by a `timeout 0.5` (rc 124 → FAIL); the
fixed code returns in ~0.1 s (rc 0 → PASS).

## Verification

- Behavioral micro-check: `read -rt 0.1 P` on an open non-newline stdin returns
  in ~0.1 s (was ~1.0 s).
- Full `test-dispatch-scripts.sh` suite passes, including the existing
  input-block Tests 1–7 (permission_prompt-acts, session_complete-passes-through)
  and the new open-stdin regression test.

## Out of scope

Two sibling hooks carry the same `read -t 1 -d ''` pattern —
`dispatch-stop.sh` and `dispatch-office-hours-strip.sh`. They fire on different
events whose stdin lifecycles were not verified here, so they are deliberately
left untouched (#1491 names only `dispatch-input-block.sh`). They are a
candidate for a separate follow-up.

### DIFF
--- a/.claude/hooks/dispatch-input-block.sh
+++ b/.claude/hooks/dispatch-input-block.sh
@@ -51,11 +51,12 @@ if ! [[ "$JOB_NAME" =~ ^[0-9]+- ]]; then
   exit 0
 fi
 
-# Consume the payload (defensive — some events pass JSON on stdin). Only
-# worker jobs reach this point; we read so the Notification filter below can
-# act. Read with a timeout so a wired event without stdin doesn't hang.
+# Consume any buffered single-line JSON payload fast. Short timeout so an
+# open, idle stdin (hook events never close stdin at EOF) returns quickly
+# rather than stalling for a full second. -r keeps backslashes in JSON intact
+# so the downstream jq parse cannot be corrupted.
 PAYLOAD=""
-if read -t 1 -d '' PAYLOAD; then :; fi
+if read -rt 0.1 PAYLOAD; then :; fi
 
 # Notification self-filter: only act on permission_prompt notifications.
 # Other notification_type values (e.g. session_complete) reach this hook and
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -14907,6 +14907,17 @@ if [[ "$rc" -ne 124 ]]; then
 else
   FAIL=$((FAIL + 1)); echo "  FAIL: open-stdin: hook was killed at 0.5s — read is still stalling"
 fi
+# Correctness: the payload must have been read and acted on — not merely
+# timed out empty (which would bypass the notification filter and park by
+# accident rather than by correct payload parsing). Mirror Test 1's check.
+apply_log=$(cat "$STUB_DIR/apply-office-hours.log" 2>/dev/null || true)
+TOTAL=$((TOTAL + 1))
+if [[ "$apply_log" == *123* ]]; then
+  PASS=$((PASS + 1)); echo "  PASS: open-stdin: dispatch-apply-office-hours invoked with issue 123 (payload was read and acted on)"
+else
+  FAIL=$((FAIL + 1)); echo "  FAIL: open-stdin: dispatch-apply-office-hours invoked with issue 123 (payload was read and acted on)"
+  echo "    apply-log: $apply_log"
+fi
 ib_teardown
 
 # ============================================================================
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -14891,6 +14891,24 @@ else
 fi
 ib_teardown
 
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — Issue #1491 / PR #1511

No browser component — all items were shell-verifiable. No QA server started.

### Items executed

| 1 | Fix landed: NUL-delimiter read replaced | script-verifiable | ✅ PASS |
| 2 | Hook returns fast on open non-newline stdin (Test 8) | script-verifiable | ✅ PASS |
| 3 | Regression: permission_prompt payloads still parsed, hook exits 0 | script-verifiable | ✅ PASS |
| 4 | Full dispatch test suite passes (CI parity) | script-verifiable | ✅ PASS |

**4 PASS / 0 FAIL / 0 SKIP**

### Details

**Item 1** — `grep` confirmed `.claude/hooks/dispatch-input-block.sh:59` contains `read -rt 0.1 PAYLOAD`; the old `read -t 1 -d '' PAYLOAD` form is absent.

**Items 2–4** — Full test suite (`test-dispatch-scripts.sh`): **2402/2402 passed, 0 failed**. New open-stdin regression Test 8 is present and green. Existing input-block Tests 1–7 (permission_prompt routing, session_complete passthrough, exit-0 paths) all still pass.

### Deferred to office-hours

None.

### Bugs found

None.

natb1:
## Review summary

### Fixed

- **Test 8 missing correctness assertion** (`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:14900`) — Test 8 asserted only the timing property (hook exits before 0.5s timeout) but not the correctness property (payload was actually read and acted on). Added an assertion that checks `apply-office-hours.log` records issue 123, mirroring what Test 1 does. Fix commit: 78d67bf7.

### Informational

- **0.1s timeout window for late-arriving stdin data** (`.claude/hooks/dispatch-input-block.sh:59`) — The 0.1s `read` timeout creates a window where stdin data that arrives after the timeout leaves `PAYLOAD` empty, bypassing the `notification_type` filter. In practice, hook stdin is populated synchronously before the hook runs, so this is a theoretical edge case rather than a real failure mode. Noted for completeness.

- **Newline delimiter assumes single-line JSON** (`.claude/hooks/dispatch-input-block.sh:59`) — Removing `-d ''` means `read` now terminates on the first newline. If the Claude harness ever sends a multi-line (pretty-printed) JSON payload, only the first line would be captured and the `jq` parse would fail. Current harness behavior sends compact single-line JSON, so this is not a regression — noted as a forward assumption.

### Dismissed

- **review-2: Test 8 correctness check** (`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:14905`) — Duplicate of the finding resolved under Fixed (code-review-1); the correctness assertion was applied once. Dismissed to avoid double-counting.

### Deferred

- **Sibling hooks carry same `read -t 1 -d ''` stall pattern** (`.claude/hooks/dispatch-input-block.sh`) — `dispatch-stop.sh` and `dispatch-office-hours-strip.sh` carry the same pattern, explicitly flagged out of scope in the PR body. Filed as #1519 (blocked_by #1491).

### Security

No open CodeQL alerts. No npm dependency changes. No security findings.

PR #1184, opened 2026-06-07
office-hours: usage-samples schema, rules, parser/serializer, demo seed

natb1:
Closes #1006

## Summary

Establishes the **data layer** for the office-hours Capacity view (epic #1005):
the `office-hours/{env}/usage-samples` Firestore collection, its security rules,
a TypeScript parser/serializer, and a bundled demo seed for the unauthenticated
tier. This is sub-issue B — no Capacity UI ships here; it mirrors how #778 built
the `items` data layer (rules + parser + seed plugin) without building a view.
The sampler that *writes* the collection is #1007; the views that *render* it are
#1008 (status band) and #1009 (history charts).

### Schema — `office-hours/{env}/usage-samples/{autoId}`

| field | Firestore | TS (`UsageSample`) |
|---|---|---|
| `sampledAt` | Timestamp | `Date` |
| `fiveHourUsedPct` | number | `number` |
| `weeklyUsedPct` | number | `number` |
| `fiveHourResetsAt` | Timestamp | `Date` |
| `weeklyResetsAt` | Timestamp | `Date` |
| `activeWorkers` | number | `number` |
| `targetWorkers` | number | `number` |
| `groupId` | string | `string` |
| `memberEmails` | string[] | `string[]` |

## What changed

- **Rules** — a new `match /office-hours/{env}/usage-samples/{sampleId}` block,
  inserted above the deny-all catch-all (which stays last). The `demo` env is
  world-readable for the public demo tier; every other env is owner-gated via the
  denormalized `memberEmails` field; client writes are denied (only the Admin-SDK
  sampler writes). Covered by `rules-test/test/firestore/office-hours.test.ts`:
  owner read, non-owner deny, unauthenticated deny, public-demo read, and
  write-deny cases.
- **Parser/serializer** — `office-hours/src/usage-samples.ts` exports
  `UsageSample`, `toUsageSample(id, data)` (duck-types Timestamps via `.toDate()`,
  validates every field, `logError` + `null` on malformed input), and
  `usageSampleToDoc(s)` (serializes back via `Timestamp.fromDate`). A vitest spec
  round-trips a sample and checks malformed-doc rejection.
- **Demo seed** — `usage-sample-seeds.ts` defines a ~14-sample synthetic series
  (offsets relative to build `now`, with 5-hour reset sawtooths, a weekly reset,
  and diverging active/target workers; no personal data). The build-time plugin
  `vite-plugin-usage-samples-seed.ts` (virtual module
[...]

PR #1107, opened 2026-06-04
agenda: Firestore data model, owner-gated rules, and demo seed data

natb1:
Closes #778

Sub-issue B of #775 — defines where the `agenda` app's reminders live and the bundled demo data for the public tier.

## Unit 1 — Firestore rules + rules-test + path doc
- `firestore.rules`: owner-gated `agenda/{env}/items/{itemId}` block above the deny-all catch-all. Read is gated to the owner via the denormalized `memberEmails` field; client writes are denied — only the Admin-SDK sync Function (`functions/src/agenda-sync.ts`, #779) writes here.
- `rules-test/test/firestore/agenda.test.ts`: covers owner read (allowed), non-owner read (denied), unauthenticated read (denied), and write (denied).
- `.claude/rules/firestore.md`: records the `agenda/{env}/items/<jitKey>` path in the path schema.

## Unit 2 — demo seed data + build-time Vite plugin
- `agenda/src/seed-reminders.ts`: synthetic, no-personal-data reminders covering the overdue / due-soon / due-later render states via signed `dueInMinutes` offsets.
- `agenda/src/vite-plugin-seed-data.ts`: the `virtual:agenda-seed-data` plugin (mirrors `budget/src/vite-plugin-seed-data.ts`). Emits module code that resolves each reminder's `dueAt` from its `dueInMinutes` offset via `Date.now()` at module-load time, so the demo stays perpetually fresh.
- `agenda/src/agenda-seed-data.d.ts`, `agenda/vite.config.ts`: virtual-module type declaration and plugin wiring.

The data-source switch, Firebase Auth, the list view, and time-until-due formatting are out of scope here — they belong to sub-issue D (#780).

## Verification
- `rules-test` vitest suite green (222 tests, incl. 4 new agenda cases).
- `npm run build --prefix agenda` succeeds with the plugin wired.
- `npm run test --prefix agenda` green (incl. the three-states coverage assertion).
- Deny-all `match /{document=**}` remains the last block in `firestore.rules`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #780, opened 2026-05-21
office-hours: the due-sorted reminder list view


natb1:
## Context

Sub-issue of #775. Builds the office-hours app's one view — the JIT reminder
list — on top of the scaffold (A, #777) and the data model (B, #778).

> **Naming — migration in progress (agenda → office-hours).** The siblings
> shipped under a mix of names and the repo is mid-rename toward the canonical
> `office-hours` (the name parent #775 uses throughout). Current merged state:
> - Sync Function: `office-hours` — writes `office-hours/{env}/items`, scans
>   `natb1/office-hours-nate` (renamed by #1014).
> - App scaffold + Firebase target + its owned Firestore collection/rules:
>   still `agenda`. Renamed to `office-hours` by **#1013 (OPEN)** — which, per
>   the app-owns-collection schema in `.claude/rules/firestore.md`, covers the
>   `match /agenda/{env}/items` rule block as well as the workspace dir and
>   hosting target.
> - Charter repo reference rename tracked by #1016 (OPEN).
>
> This view must be built against the **final `office-hours` naming**, so it is
> blocked on #1013. Until #1013 lands, the owner tier cannot read what the
> Function already writes (`office-hours/{env}/items`), because the rules still
> declare only `agenda/{env}/items`.

## Scope

- A data-source switch matching the repo's public-demo / private-auth pattern
  (the `budget` app's `SeedDataSource` / authenticated-source split):
  - **unauthenticated** — read the bundled demo reminders (sub-issue B's seed
    data), via the build-time seed-data Vite plugin;
  - **authenticated as the owner** — read the real reminders from
    `office-hours/{env}/items` in Firestore.
- Firebase Auth (Google sign-in) for the owner, via `createAppContext` with
  `enableAuth: true` (exposes `signIn` / `signOut` / `onAuthStateChanged`,
  matching landing's pattern). An unauthenticated visitor, and a signed-in
  non-owner, see the demo tier — the Firestore rules deny a non-owner the real
  data.
- The list view: one row per reminder, sorted by due time ascending —
  most-overdue first, then soonest-due. Each row shows the reminder title and a
  human-readable time-to-due ("due in 3h", "due in 2d") or overdue duration
  ("overdue 4h").
- Unit-test the sort order and the due / overdue formatting.

## Acceptance criteria

- [ ] Unauthenticated, the view renders the bundled demo reminders.
- [ ] Signed in as the owner, the view renders the real reminders from
      `office-hours/{env}/items`.
[...]

PR #1513, opened 2026-06-13
office-hours: list-query test coverage + array-contains comment for the items rule

natb1:
Closes #1117

## What

Completes the test coverage and documentation that PR #1107 (#778) deliberately
left out of scope for the `office-hours/{env}/items` collection:

1. **`firestore.rules`** — documents on the `match /office-hours/{env}/items/{itemId}`
   comment block that, because the read rule references `resource.data.memberEmails`,
   a list query (`getDocs`) must carry a matching
   `where("memberEmails", "array-contains", <viewer-email>)` filter. An unfiltered
   list query is rejected by the rules engine even when every document would satisfy
   the per-document condition. The wording mirrors the existing landing/groups
   list-query comment.

2. **`rules-test/test/firestore/office-hours.test.ts`** — adds four list-query tests
   to the `office-hours items` describe block, mirroring the existing
   `office-hours usage-samples` list-query pattern and reusing the block's `jit-demo`
   seed doc:
   - owner array-contains-filtered list query **succeeds**;
   - **unfiltered** owner list query is **denied** (the rules-engine constraint the
     usage-samples block does not cover);
   - authenticated non-member list query **for another member's items is denied**;
   - unauthenticated list query is **denied**;
   - authenticated non-member **self-targeted** filter **succeeds and returns an
     empty set** — a positive test documenting the non-obvious converse, so the
     false "self-targeted non-member filter is denied" assumption (below) can't be
     silently re-introduced.

## Deviation from the persisted plan (ratified)

The plan's third case asserted that a non-member's **self-targeted** filter
(`array-contains "stranger@test.com"` authed as `stranger@test.com`) is denied.
The emulator showed it **succeeds** — and so does the rule: a self-targeted
`array-contains` filter is tautologically rule-compliant (every doc it could
return has the requester in `memberEmails`), so the engine permits it and it
returns an empty set. The meaningful denial is a non-member filtering for
**another** member's email, which is what the corrected case 3 now asserts.

The plan inherited this false expectation by mirroring the pre-existing
`usage-samples` test added in #1184, which asserts the same false invariant.
Because `rules-test` runs in **no CI job**, that test had never run red. This PR
therefore also corrects the `usage-samples` `denies non-member list query`
assertion (one line + a clarifying comment) so the whole suite is green. This is
a scope expansion beyond the issue's stated items-only scope — flagging #1184
here so a human can ratify the change and note that a merged test asserted a
false security invariant.

The author has ratified both calls: permit-empty is the intended model (the
test was wrong, the rule stays), and the #1184 usage-samples correction stays in
this PR.

## Why

The owner's due-sorted reminder list view (#780) issues the filtered list query this
rule must permit. These tests lock in that the rule permits the filtered owner list
query and rejects every other shape, and the comment prevents future readers from
writing an unfiltered `getDocs` the rules engine would reject.

## Verification

`firestore.rules` is comment-only (no condition changes). The `rules-test` suite is
not run by any CI job, so it was run manually against the Firestore emulator:
`npm test --prefix rules-test` (firestore+storage emulator) — **all 305 tests pass**
(17 files), including the five new `office-hours items` list-query tests and the
corrected `usage-samples` non-member case.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -488,6 +488,9 @@ service cloud.firestore {
     // Office Hours reminders are the synced JIT queue, scoped to the owner via a
     // denormalized `memberEmails` field. Client writes are denied; only the
     // Admin-SDK sync Function (functions/src/office-hours-sync.ts) writes here.
+    // Because the read rule references resource.data.memberEmails, a list query
+    // (getDocs) must include where("memberEmails", "array-contains", <email>);
+    // Firestore rejects unfiltered list queries even when every document would pass.
     match /office-hours/{env}/items/{itemId} {
       allow read: if request.auth != null
         && request.auth.token.email in resource.data.memberEmails;
--- a/rules-test/test/firestore/office-hours.test.ts
+++ b/rules-test/test/firestore/office-hours.test.ts
@@ -135,14 +135,19 @@ describe("office-hours items", () => {
     );
   });
 
-  it("denies non-member list query on items collection", async () => {
+  // A non-member cannot list another member's items: filtering for an email
+  // they are not in is denied, because the matching docs carry a memberEmails
+  // the requester is absent from. (A self-targeted array-contains filter is
+  // always rule-compliant and returns an empty set, so it is not a denial
+  // case.)
+  it("denies authenticated non-member list query for another member's items", async () => {
     const ctx = authenticatedContext(env, "stranger@test.com");
     const db = ctx.firestore();
     await assertFails(
       getDocs(
         query(
           collection(db, `office-hours/${ENV}/items`),
-          where("memberEmails", "array-contains", "stranger@test.com"),
+          where("memberEmails", "array-contains", "owner@test.com"),
         ),
       ),
     );
@@ -291,14 +296,19 @@ describe("office-hours usage-samples", () => {
     );
   });
 
-  it("denies non-member list query on test-env collection", async () => {
+  // A non-member cannot list another member's samples: filtering for an email
+  // they are not in is denied, because the matching docs carry a memberEmails
+  // the requester is absent from. (A self-targeted array-contains filter is
+  // always rule-compliant and returns an empty set, so it is not a denial
+  // case.)
+  it("denies authenticated non-member list query for another member's samples", async () => {
     const ctx = authenticatedContext(env, "stranger@test.com");
     const db = ctx.firestore();
     await assertFails(
       getDocs(
         query(
           collection(db, `office-hours/${ENV}/usage-samples`),
-          where("memberEmails", "array-contains", "stranger@test.com"),
+          where("memberEmails", "array-contains", "owner@test.com"),
         ),
       ),
     );
--- a/rules-test/test/firestore/office-hours.test.ts
+++ b/rules-test/test/firestore/office-hours.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, beforeAll, beforeEach } from "vitest";
+import { describe, it, beforeAll, beforeEach, expect } from "vitest";
 import { assertSucceeds, assertFails } from "@firebase/rules-unit-testing";
 import type { RulesTestEnvironment } from "@firebase/rules-unit-testing";
 import {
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — issue #1117 / PR #1513

### Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Full rules-test suite (304 tests, 17 files) against Firestore emulator | shell-command | ✅ PASS |
| 2 | `firestore.rules` parses without syntax errors | shell-command | ✅ PASS |
| 3 | Documenting comment present and accurate in `firestore.rules` | shell-command | ✅ PASS |
| 4 | Four new list-query tests cover intended cases (allow filtered, deny unfiltered, deny non-member, deny unauthenticated) | shell-command | ✅ PASS |
| 5 | Corrected `usage-samples` non-member test filters for another member's email (`owner@test.com`), not own email | shell-command | ✅ PASS |
| 6 | Semantic correctness of non-member denial reasoning | — | ⏸ deferred to office-hours |

**Counts:** 5 PASS · 0 FAIL · 0 SKIP · 1 deferred (needs-human-judgment)

### Deferred to office-hours

**Item 6 — Semantic correctness of the non-member denial reasoning**

The suite passes operationally, but the judgment that a self-targeted `array-contains` filter is vacuously rule-compliant (and therefore not a denial case) — while a filter for another member's email is the genuine denial — is a conceptual claim. A reviewer should ratify that this reasoning is correct and that the prior `usage-samples` test (which asserted the self-targeted filter was denied) was a false security invariant.

### Bugs

None found.

natb1:
## QA update — item 6 resolved (clean pass)

The earlier summary deferred item 6 (semantic correctness of the non-member denial reasoning) to office-hours. Resolving it here with QA judgment instead: **PASS**.

**Verdict:** The reasoning is correct, both conceptually and empirically.

Firestore evaluates list queries statically against the query's filter constraints — never against stored documents. A list query is permitted only if the engine can prove every returnable document satisfies the read rule (`request.auth.token.email in resource.data.memberEmails`):

- **Self-targeted** filter (`array-contains "stranger@test.com"` authed as `stranger@test.com`) is tautologically rule-compliant → **allowed**, returning an empty set. Asserting it is *denied* is a false invariant.
- **Cross-targeted** filter (`array-contains "owner@test.com"` authed as `stranger@test.com`) → **denied**. This is the genuine non-member denial the corrected case 3 asserts.

This matches the repo's own `.claude/docs/firestore.md` "List query compatibility" guidance and is confirmed empirically: the full suite is green (304/304) against the live emulator, the authoritative implementation.

### Maintainer note — stands regardless of this PR

This PR's scope expansion is sound: the pre-existing `usage-samples` `denies non-member list query` test (added in #1184) asserted the **same false security invariant** (self-targeted filter is denied). Because `rules-test` runs in no CI job, it never ran red. This PR corrects it (one line + comment) so the suite is green.

Flagging for awareness, not as a blocker: **a merged PR (#1184) asserted a false security invariant.** No further action is required for this PR — the correction is included here — but the finding is worth a maintainer's eyes.

**Final QA disposition:** 6 PASS · 0 FAIL · 0 SKIP · 0 deferred · 0 bugs → applying `dispatch:qa-done`.

natb1:
## Review findings (dispatch:review-fix)

### Fixed

- **F1** `rules-test/test/firestore/office-hours.test.ts:117–128` — Owner list query test did not capture the snapshot, making `assertSucceeds` vacuous (the test would pass even if the seed path were wrong). Added `const snap = await assertSucceeds(...)` and `expect(snap.empty).toBe(false)` to make the assertion meaningful, consistent with the companion self-targeted test convention. Fix commit: a608cb4b.

### Informational

_None._

### Dismissed

_None._

### False-positive (security)

_None._

_None._

### Deferred

- **F3** `rules-test/test/firestore/office-hours.test.ts` (usage-samples describe block) — Usage-samples block has no unfiltered-list denial test symmetric to the items block's `denies unfiltered owner list query` case. Valid symmetry gap; out of scope for this PR's focused items additions. Filed as #1556.

### Out-of-scope (security)

- **F2** `firestore.rules:504–511` — Comment on the usage-samples rule block omits the same `where("memberEmails", "array-contains", <email>)` filter requirement documented for the items block. Pre-existing documentation gap in code this diff did not touch.

---

CodeQL: no open alerts. npm dependency audit: no `package.json`/`package-lock.json` changes.

PR #1507, opened 2026-06-13
dispatch: derive selection's claimed-set forward from live sessions + reservations

natb1:
Closes #1474

## Summary

Inverts dispatch selection's liveness data flow. Selection no longer enumerates
every registered git worktree to decide which queue items are "already in
flight." Instead it derives the claimed-issue-number set **forward** from two
small in-flight sets — live session names and reservation-ledger markers — and
consults a worktree only post-selection, for the single chosen target, via
`dispatch-resolve-worktree`.

Selection becomes O(worker cap), not O(worktrees). Its latency decouples from
worktree count (286 today, #1451), reframing that GC backlog as a pure-disk
concern. Builds on #1479, which made the per-tick scan reuse one
`claude agents --json` snapshot.

## What changed

- **`lib-claude-agents.sh`** — adds `live_session_claimed_nums`: reads the
  per-tick snapshot and emits the unique `<N>` from session names matching
  `^[0-9]+-` (phase workers `<N>-slug`) or `^office-hours-[0-9]+$` (office-hours
  sessions). Routers and job sessions match neither and are excluded. `[]` → empty
  output, return 0; UNKNOWN daemon → return 1.
- **`lib-reservation-ledger.sh`** — adds `reserved_claimed_nums` (unique `<N>`
  from marker basenames) and the convenience union `claimed_issue_nums`, which
  **fails open** on live-UNKNOWN: continues with the reserved set only and a
  one-line stderr diagnostic.
- **`dispatch-select-target`** — builds the claimed set once from
  `claimed_issue_nums` behind a `num_is_claimed` predicate; deletes the
  worktree-map build and the four worktree-walk helpers. The PR-skip and
  issue-skip now share one `<N>`-keyed predicate.
- **`dispatch-trace-leaf`** — same inversion for the queue-mode child skip; the
  descent-loop and root skip points are preserved (not folded into `--exclude`).
  The `dispatch:office-hours`-parked label check stays distinct.
- **`test-dispatch-scripts.sh`** — helper-unit tests for the new functions, plus
  integration tests for the two deliberate behavior edges and UNKNOWN fail-open
  (see below). Suite green at 2422/2422.

## Deliberate behavior changes (not byte-identical to the old walk)

1. **PR liveness now keys on the branch prefix**, not the exact branch worktree
   path. A live `10-foo` session now also marks a `10-bar` PR's `<N>=10` claimed.
   This is the safe, more-conservative-skip direction.
[...]

PR #1520, opened 2026-06-13
dispatch: extract shared num_has_live_worktree helper into lib.sh

natb1:
Closes #971

## What

Consolidate the byte-identical `child_has_live_worktree` (`dispatch-trace-leaf`,
added in PR #967) and `issue_has_live_worktree` (`dispatch-select-target`) into a
single `num_has_live_worktree` helper in `lib.sh`, where the other shared
worktree-record helpers (`list_worktree_records`, `split_worktree_record`)
already live.

## Why

The two functions were a byte-for-byte structural duplicate — same
`WORKTREE_PATHS_BY_NUM` lookup, same per-path `worktree_has_live_session` loop
with first-live early return, same no-worktree fast-path miss. Only the name and
surrounding comments differed. A code-review finding flagged this as a second
maintenance point that could silently drift; folding both into one shared helper
keeps the two scripts in lockstep with one source of truth.

## Changes

- **`lib.sh`** — add `num_has_live_worktree`, placed after `split_worktree_record`
  to cluster with the worktree-record helpers. Takes the issue-number→paths
  associative array by **nameref** (`local -n`, the established pattern in
  `dispatch-select-target`), with an underscore-prefixed nameref name to avoid a
  circular reference when callers pass `WORKTREE_PATHS_BY_NUM`.
- **`dispatch-trace-leaf`** — delete `child_has_live_worktree`; rewire
  `child_worktree_claimed` to call the shared helper; update the stale comment
  reference.
- **`dispatch-select-target`** — delete `issue_has_live_worktree`; rewire
  `issue_worktree_claimed` to call the shared helper.
- **`test-dispatch-scripts.sh`** — add a focused, daemon-free unit test (stubbed
  `worktree_has_live_session`) covering the no-worktree fast-path miss, the
  any-live early return, and the all-dead case; also exercises the nameref binding.

Behavior is unchanged — this is a pure refactor.

## Verification

- `test-dispatch-scripts.sh`: **2402/2402 passing**, including the three new
  `num_has_live_worktree` assertions.
- `bash -n` clean on `lib.sh`, `dispatch-trace-leaf`, `dispatch-select-target`.
- `grep` confirms neither old function name survives as a definition or live call
  (historical comments only).

## Out of scope

- `pr_branch_has_live_worktree` (`dispatch-select-target`) — branch-keyed, single
  path, not a duplicate of the num-array helper.
- The `reservation_exists` duplication shared by `child_worktree_claimed` /
  `issue_worktree_claimed` — a separate extraction this PR deliberately does not
  address.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -448,21 +448,6 @@ pr_branch_has_live_worktree() {
   worktree_has_live_session "$path"
 }
 
-# True when a live session owns any <num>-* worktree. No such worktree →
-# fast-path miss. Several matches → owned if any one is live.
-issue_has_live_worktree() {
-  local num="$1" p
-  local paths="${WORKTREE_PATHS_BY_NUM[$num]:-}"
-  [[ -z "$paths" ]] && return 1
-  while IFS= read -r p; do
-    [[ -z "$p" ]] && continue
-    if worktree_has_live_session "$p"; then
-      return 0
-    fi
-  done <<< "$paths"
-  return 1
-}
-
 # True when the <branch> worktree is CLAIMED — a live session owns it (#905) OR
 # it carries a reservation marker (#1046). Both close the spawn-gap re-selection
 # race: a reserved-but-not-yet-registered target must not be re-selected. An
@@ -482,7 +467,7 @@ pr_branch_worktree_claimed() {
 # worktrees (no session, no marker) stay reuse signals, not skips.
 issue_worktree_claimed() {
   local num="$1" p
-  issue_has_live_worktree "$num" && return 0
+  num_has_live_worktree WORKTREE_PATHS_BY_NUM "$num" && return 0
   local paths="${WORKTREE_PATHS_BY_NUM[$num]:-}"
   [[ -z "$paths" ]] && return 1
   while IFS= read -r p; do
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-trace-leaf
@@ -139,25 +139,6 @@ if [[ "$MODE" == "queue" ]]; then
   done < <(printf '%s' "$parked_json" | jq -r '.[].number')
 fi
 
-# True when a live Claude session owns any <num>-* worktree. No such worktree →
-# fast-path miss (return 1), no daemon round-trip. Several matches → owned if any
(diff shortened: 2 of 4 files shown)


natb1:
## QA Summary — #971 / PR #1520

**Result: 7 PASS · 0 FAIL · 0 SKIP · 1 deferred to office-hours**

No browser component (pure shell script refactor). All verification ran in the shell-command lane.

### Items executed

| # | Title | Result |
|---|-------|--------|
| 1 | Test suite passes end-to-end | ✅ PASS — 2402/2402 |
| 2 | Three new `num_has_live_worktree` assertions present | ✅ PASS — lines 6020, 6023, 6026 |
| 3 | Shared helper defined exactly once in `lib.sh` | ✅ PASS — `lib.sh:705` |
| 4 | Both old per-script helpers fully removed | ✅ PASS — no definitions found |
| 5 | No caller still invokes a removed helper name | ✅ PASS — old names absent entirely from consuming scripts |
| 6 | Both callers use identical call signature | ✅ PASS — `dispatch-select-target:470`, `dispatch-trace-leaf:150` |
| 7 | All changed scripts parse cleanly (`bash -n`) | ✅ PASS — all three ok |
| 8 | Behavior parity (is the refactor a genuine no-op?) | ⏸ Deferred to office-hours |

### Deferred to office-hours

**Item 8 — Behavior parity (needs-human-judgment)**

Confirming the extracted `num_has_live_worktree` logic is observably equivalent to the two deleted helpers (`issue_has_live_worktree`, `child_has_live_worktree`) requires a human to cross-check that no behavioral nuance was dropped (e.g. argument order, side effects). The unit tests cover the three canonical cases, but the claim of behavioral equivalence is the central assertion of this refactor.

natb1:
Closing as superseded by #1507.

This PR consolidated the byte-identical `child_has_live_worktree`
(`dispatch-trace-leaf`) and `issue_has_live_worktree`
(`dispatch-select-target`) into a shared `num_has_live_worktree` helper in
`lib.sh`. Before it could merge, #1507 ("derive selection's claimed-set forward
from live sessions + reservations") replaced the entire backward worktree-walk
mechanism in both scripts — deleting both helpers this PR deduplicates, along
with the `WORKTREE_PATHS_BY_NUM` / `WORKTREE_PATH_BY_BRANCH` arrays they read —
and unified the logic forward behind `num_is_claimed` / `claimed_issue_nums`.

The dedup target no longer exists: #1507 already achieved a stronger version of
this refactor's "one source of truth" goal. There is no merge that preserves
this PR's intent against main, so the work is moot. The `num_has_live_worktree`
helper added here would land orphaned with no caller.

PR #1523, opened 2026-06-13
dispatch: consolidate parse_duration into lib.sh

natb1:
Closes #1065

## Summary

Consolidates the duplicated `parse_duration` helper (a `<int><unit>` → seconds
parser, units `s|m|h|d|w`) into the shared `lib.sh`, and removes the four inline
copies. All four scripts already `source lib.sh`, so no new sourcing was needed —
only the inline definitions were removed.

## The copies had already diverged

The issue framed the copies as "duplicated" (implying verbatim). They were not —
they had already drifted into **three variants**, which is exactly the divergence
the consolidation prevents going forward:

1. `dispatch-select-target` (most correct): `10#` base-10 fix so a leading-zero
   spec like `08h`/`012h` parses as decimal, not octal; on bad input printed a
   descriptive `error: invalid duration …` to stderr and `exit 1`.
2. `dispatch-jit-engine` / `dispatch-jit-calendar-import`: identical to each
   other; **no** `10#` fix; silent `return 1` on bad input.
3. `dispatch-statements-scan`: silent `return 1`, no `10#` fix; minor cosmetic
   differences.

## Canonical contract chosen for `lib.sh`

- **Keeps the `10#` base-10 fix** (required by test JS9). Single-source-of-truth
  means the other three scripts now inherit it.
- **Silent `return 1` on bad input** (pure-library contract) — the caller owns
  the error message. This already matched 3 of the 4 call sites, which do
  `if ! X=$(parse_duration …); then <own error>; fi`.

## Forced behavior change in the three silent scripts

Inheriting the `10#` fix is a **forced, strict-improvement behavior change** in
`dispatch-statements-scan`, `dispatch-jit-engine`, and
`dispatch-jit-calendar-import`: a leading-zero duration like `08h`/`012h` now
parses as decimal where it previously hit a bash octal-literal error. There is
no test coverage for these three scripts' parser (they are not copied into the
test harness), so this change is unguarded by a test — acceptable for this
consolidation.

## dispatch-select-target call sites

`dispatch-select-target` was the only caller relying on the function's *internal*
`exit 1` + message. Its three `jit_scan` call sites are now wrapped to check the
return and emit the descriptive error themselves — preserving its clear-error
behavior (per `.claude/rules/code-style.md`) and its non-zero exit (test JS10).
The assignment inside each `if !` stays a plain assignment (not `local …`), so
`local` does not mask the command-substitution exit status.

## Verification

- Full dispatch script suite: **2401/2401 pass**, including JS9 (leading-zero
  parses base-10) and JS10 (invalid duration → non-zero exit).
- `grep 'parse_duration *() *{'` over the scripts dir → exactly one hit:
  `lib.sh`.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-jit-calendar-import
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-jit-calendar-import
@@ -113,27 +113,6 @@ set -uo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 source "$SCRIPT_DIR/lib.sh"
 
-# ---- parse_duration <str> — duration string to seconds ----------------------
-# Copied inline from dispatch-jit-engine. Accepts <int><unit> where unit is one
-# of s|m|h|d|w. Prints the duration in seconds on stdout and returns 0. On
-# unparseable input prints nothing and returns 1.
-parse_duration() {
-  local spec="$1" num unit
-  if [[ ! "$spec" =~ ^([0-9]+)([smhdw])$ ]]; then
-    return 1
-  fi
-  num="${BASH_REMATCH[1]}"
-  unit="${BASH_REMATCH[2]}"
-  case "$unit" in
-    s) printf '%s\n' "$(( num * 1 ))" ;;
-    m) printf '%s\n' "$(( num * 60 ))" ;;
-    h) printf '%s\n' "$(( num * 3600 ))" ;;
-    d) printf '%s\n' "$(( num * 86400 ))" ;;
-    w) printf '%s\n' "$(( num * 604800 ))" ;;
-  esac
-  return 0
-}
-
 # ---- Step 0: no-config gate -------------------------------------------------
 # Any of the three OAuth env vars unset → silent no-op.
 if [[ -z "${GOOGLE_CALENDAR_CLIENT_ID:-}" \
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-jit-engine
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-jit-engine
@@ -63,27 +63,6 @@ set -uo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 source "$SCRIPT_DIR/lib.sh"
 
-# ---- parse_duration <str> — duration string to seconds ----------------------
-# Accepts <int><unit> where unit is one of s|m|h|d|w. Prints the duration in
-# seconds on stdout and returns 0. On unparseable input prints nothing and
-# returns 1 — the caller treats that as a hard error for the jit.
-parse_duration() {
-  local spec="$1" num unit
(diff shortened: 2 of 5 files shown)


natb1:
## QA Summary — #1065 (`parse_duration` consolidation)

**Result: ✅ All checks passed**

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Full dispatch script suite | shell-command | PASS |
| 2 | JS9 — leading-zero parses base-10 | shell-command | PASS |
| 3 | JS10 — invalid duration → non-zero exit | shell-command | PASS |
| 4 | Single source of truth — no remaining inline definitions | shell-command | PASS |
| 5 | lib.sh `parse_duration` callable and correct | shell-command | PASS |

### Details

- **Full suite:** 2401/2401 passed, 0 failed.
- **JS9:** `PASS: leading-zero duration is base-10 → earlier-due jit-y wins` — confirms `10#` base-10 fix survived the move into `lib.sh`.
- **JS10:** `PASS: invalid duration → exits non-zero` — confirms `dispatch-select-target` wrapped call sites correctly surface parse failures.
- **Single definition:** `grep 'parse_duration *() *{'` returns exactly one hit: `lib.sh:37`.
- **Correctness spot-check:** `30s`→30, `2m`→120, `1h`→3600, `1d`→86400, `1w`→604800, `012h`→43200 (base-10), `24` (no unit) → rejected with exit 1.

### Deferred to office-hours

None.

### Bugs found

None.

No browser walkthrough was needed (pure shell script refactor, no frontend component).

natb1:
## Review — PR #1523

_No code changes applied (all findings dismissed or out-of-scope). No fix commit._

---

### False-positive (security)

- **`lib.sh:42` — `10#` base-10 prefix accepted as correct** (`red-team`): The `10#` prefix was flagged as a silent behavior change for callers that previously passed leading-zero specs like `08h`. This is intentional and documented: the PR body explicitly describes it as a "forced, strict-improvement behavior change" — leading-zero specs now parse as decimal instead of triggering a bash octal error. The three affected scripts (`dispatch-statements-scan`, `dispatch-jit-engine`, `dispatch-jit-calendar-import`) had no test coverage for this case and no callers rely on the old broken behavior. Not a vulnerability.

---

### Out-of-scope (security)

- **`lib.sh:39-51` — no upper bound on digit-string length** (`red-team`): `parse_duration` accepts an arbitrarily long digit string (`[0-9]+` has no length cap), so a value like `99999999999999999999w` passes the regex and produces bash arithmetic overflow — wrapping to a negative or zero duration. This is a pre-existing class of input; the diff did not introduce the parsing pattern. All call sites in this repo read durations from config files the operator controls, not from untrusted user input, so exploitability is low. Classified out-of-scope for this PR; not filed as a follow-up (below meaningfulness threshold for a config-only attack surface).

---

### CodeQL

No open alerts on this PR ref.

### Dependency audit

Not applicable (`deps=false` — no `package.json`/`package-lock.json` changes).

PR #1385, opened 2026-06-11
office-hours: source capacity sampler member-emails from Secret Manager (PII-safe)

natb1:
Closes #1377

## What

Brings the local capacity sampler onto the same canonical PII source the
Cloud-Function side already uses, so the office-hours Capacity HISTORY chart can
populate with the owner's real `office-hours/prod/usage-samples` series.

The sampler was a silent opt-in keyed off the **PII** env var
`DISPATCH_USAGE_SAMPLES_MEMBER_EMAILS`. That var is unset on the dispatch
machine, so the deployed sampler wrote nothing every tick. The member-email list
is PII and must not enter the public repo, a shell env var, a local file, or a
local secret store.

This change:

- **Sources member emails from Secret Manager at runtime.** The writer
  (`usage-sample-writer.mjs`) resolves the list from the `OFFICE_HOURS_MEMBER_EMAILS`
  Firebase Secret Manager secret — the same secret #1257 created for the
  office-hours Cloud Function — via the `@google-cloud/secret-manager` SDK,
  authenticating with the same Application Default Credentials the writer already
  needs for its firebase-admin Firestore write. No second secret; one canonical
  source for the function side and the local sampler.
- **Decouples the opt-in gate from the PII.** `dispatch-sample-usage` now gates
  per-tick sampling on a non-PII `DISPATCH_USAGE_SAMPLES_ENABLED=1` switch
  (anything other than exactly `1` → no-op). Enabling sampling no longer requires
  putting emails in the environment.
- **Keeps dry-run dependency-free.** The Secret Manager fetch lives only on the
  real-mode path (mirroring how firebase-admin is dynamically imported only in
  real mode). `--dry-run` reads a test-only `DISPATCH_USAGE_SAMPLES_SECRET_OVERRIDE`
  seam and never imports the SDK, so the bash unit suite runs without
  `@google-cloud/secret-manager` or `firebase-admin` installed.
- **Documents the setup** in `dispatch-propagate/reference.md`: the opt-in switch,
  the secret name, ADC authentication, the required
  `roles/secretmanager.secretAccessor` IAM role, the non-PII config defaults, and
  the fail-safe contract — with no PII.

## Fail-safe contract (preserved)

When enabled, a missing/failed secret fetch or missing credentials produces one
stderr diagnostic prefixed `usage-sample-writer:`, a non-zero exit, and no
document written — never a doc with empty/wrong `memberEmails`. The
`dispatch-tick` wrapper logs the sampler failure and proceeds, so the tick is
[...]

PR #1530, opened 2026-06-13
dispatch-propagate: test DISPATCH_USAGE_SAMPLES_SECRET_NAME empty/slash guards

natb1:
Closes #1386

## What

Adds two writer unit-test cases to
`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` covering the
two `DISPATCH_USAGE_SAMPLES_SECRET_NAME` guards in `usage-sample-writer.mjs`:

- **W8** — empty `DISPATCH_USAGE_SAMPLES_SECRET_NAME` → asserts non-zero exit and
  a diagnostic containing `must be non-empty`.
- **W9** — slash in `DISPATCH_USAGE_SAMPLES_SECRET_NAME` (`a/b`) → asserts non-zero
  exit and a diagnostic containing `must not contain a slash`.

## Why

`DISPATCH_USAGE_SAMPLES_SECRET_NAME` is interpolated directly into a GCP Secret
Manager resource path. The empty-string and slash guards that defend that
interpolation (`usage-sample-writer.mjs:126-131`) were previously exercised by no
test — the variable appeared only in the teardown `unset` list, so a future
refactor could silently break either guard with no test failure. These two cases
pin both error paths, including the guard-specific diagnostic so a test cannot
pass for the wrong reason.

The guards were added by the now-merged #1385, and the missing coverage was
surfaced by its review pass (backlink #1385).

## Notes

- Each case uses a valid `DISPATCH_USAGE_SAMPLES_SECRET_OVERRIDE` and `GROUP_ID`
  so `SECRET_NAME` is the only invalid input. `SECRET_NAME` is validated first in
  `loadConfig` and `?? DEFAULT_SECRET_NAME` falls back only on null/undefined, so
  an exported empty string stays empty and the length-0 guard fires.
- Named W8/W9 rather than the issue's proposed W7/W8 — a W7 case already exists.
- No change to `usage-sample-writer.mjs` (guards already work) and no teardown
  edit (`SECRET_NAME` is already in the shared `su_teardown` `unset` block).

## Verification

**2415/2415 passed, 0 failed** (up from 2411; the four new assertions are green
and no existing case regressed).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -22756,6 +22756,30 @@ assert_eq "writer W7 → diagnostic message contains DISPATCH_USAGE_SAMPLES_SECR
   "$([[ "$out" == *"DISPATCH_USAGE_SAMPLES_SECRET_OVERRIDE"* ]] && echo 1 || echo 0)"
 su_teardown
 
+# W8 — empty DISPATCH_USAGE_SAMPLES_SECRET_NAME → non-zero exit; writer prints
+# "DISPATCH_USAGE_SAMPLES_SECRET_NAME must be non-empty".
+su_setup
+export DISPATCH_USAGE_SAMPLES_SECRET_OVERRIDE="a@b.com"
+export DISPATCH_USAGE_SAMPLES_GROUP_ID="grp-1"
+export DISPATCH_USAGE_SAMPLES_NOW="$SU_NOW"
+export DISPATCH_USAGE_SAMPLES_SECRET_NAME=''
+if out=$(WRITER "$W1_PAYLOAD" 2>&1); then rc=0; else rc=$?; fi
+assert_eq "writer W8 empty SECRET_NAME → non-zero exit" "1" "$([[ "$rc" -ne 0 ]] && echo 1 || echo 0)"
+assert_eq "writer W8 → diagnostic mentions must be non-empty" "1" "$([[ "$out" == *"must be non-empty"* ]] && echo 1 || echo 0)"
+su_teardown
+
+# W9 — slash in DISPATCH_USAGE_SAMPLES_SECRET_NAME → non-zero exit; writer prints
+# "DISPATCH_USAGE_SAMPLES_SECRET_NAME must not contain a slash".
+su_setup
+export DISPATCH_USAGE_SAMPLES_SECRET_OVERRIDE="a@b.com"
+export DISPATCH_USAGE_SAMPLES_GROUP_ID="grp-1"
+export DISPATCH_USAGE_SAMPLES_NOW="$SU_NOW"
+export DISPATCH_USAGE_SAMPLES_SECRET_NAME='a/b'
+if out=$(WRITER "$W1_PAYLOAD" 2>&1); then rc=0; else rc=$?; fi
+assert_eq "writer W9 slash SECRET_NAME → non-zero exit" "1" "$([[ "$rc" -ne 0 ]] && echo 1 || echo 0)"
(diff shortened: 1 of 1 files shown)


natb1:
## QA Summary — #1386 (PR #1530)

**Result: PASS** — All 5 items passed, 0 failed, 0 skipped.

### Items executed

| # | Title | Result |
|---|-------|--------|
| 1 | Golden path: full bash unit suite passes with new cases | PASS |
| 2 | W8 exists: empty SECRET_NAME → non-zero exit + "must be non-empty" | PASS |
| 3 | W9 exists: slash SECRET_NAME → non-zero exit + "must not contain a slash" | PASS |
| 4 | Teardown already covers SECRET_NAME via shared `su_teardown` | PASS |
| 5 | No existing W*/S* cases regressed | PASS |

**5 PASS / 0 FAIL / 0 SKIP**

### Detail

- Full suite: **2437/2437 passed, 0 failed** (exit 0). The PR adds 4 new assertions (2 per case × 2 cases) accounting for the increase from 2411 to 2415 noted in the PR, plus additional tests that landed on `main` since the branch was cut.
- W8 assertion strings verified present in test file (grep-c = 1 each).
- W9 assertion strings verified present in test file (grep-c = 1 each).
- `DISPATCH_USAGE_SAMPLES_SECRET_NAME` appears 7× in the test file — shared `su_teardown` handles cleanup for both new cases.

### Deferred to office-hours

None.

### Bugs found

None.

### Browser walkthrough

Not applicable — the only changed file is `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` (no frontend/browser component).

natb1:
## Review summary

### Fixed

- **test-dispatch-scripts.sh:23049** — W8's message assertion was scoped to the generic substring `must be non-empty` rather than `DISPATCH_USAGE_SAMPLES_SECRET_NAME must be non-empty`. Changed to the full variable-prefixed substring, consistent with W7. (commit 3a00ba34)
- **test-dispatch-scripts.sh:23061** — W9's message assertion was scoped to the generic substring `must not contain a slash` rather than `DISPATCH_USAGE_SAMPLES_SECRET_NAME must not contain a slash`. Changed to the full variable-prefixed substring, consistent with W7. (commit 3a00ba34)

- **test-dispatch-scripts.sh:23048–23049** — Red-team finding claimed W8's `must be non-empty` substring could be satisfied by a groupId or projectId check instead of the secretName check. Refuted: W8 sets `GROUP_ID=grp-1` (non-empty) and leaves `PROJECT_ID` unset (defaults to `commons-systems`, also non-empty). Neither alternate `must be non-empty` path can fire under these inputs. The nullish-coalescing operator (`??`) does not substitute the default for an empty string, so `SECRET_NAME=''` correctly trips the secretName guard before groupId is even reached. Two skeptics voted refuted.
- **test-dispatch-scripts.sh:23060–23061** — Red-team finding claimed W9's `must not contain a slash` substring could be satisfied by a groupId or projectId slash check instead of the secretName check. Refuted: W9 supplies `GROUP_ID=grp-1` (no slash) and leaves `PROJECT_ID` unset (defaults to `commons-systems`, no slash). Any regression that deleted the secretName slash guard would cause the writer to exit 0, immediately breaking W9's exit-code assertion before the message assertion is even reached. Two skeptics voted refuted.

### Out-of-scope (security)

- **test-dispatch-scripts.sh:23042–23062** — Neither W8 nor W9 covers `SECRET_NAME` values containing other characters (beyond slash) that could affect a GCP Secret Manager resource path. Out of scope for this PR, which targets the two existing guards specifically. No follow-up filed (below meaningfulness threshold for a test-only change).

### Informational

_None._

### Dismissed

_None._

### False-positive (security)

_None._

### Deferred

_None._

PR #1533, opened 2026-06-13
dispatch-tick-recover: ensure_daemon_service before arming the recovery reseed

### LINKED ISSUE #1370: dispatch-tick-recover schedules dispatch-tick directly via systemd-run, bypassing dispatch-schedule-reseed and its ensure_daemon_service call
## Context

`dispatch-tick-recover` arms a recovery reseed by calling `systemd-run` directly (mirroring the final step of `dispatch-schedule-reseed`), but it does not call `ensure_daemon_service` before doing so. `dispatch-spawn-tick` and `dispatch-schedule-reseed` both call `ensure_daemon_service || true` before arming a tick or reseed — this call idempotently installs, resets, and activates the durable `claude-daemon.service` unit.

If the daemon service enters systemd's `failed` state (e.g. `StartLimit` exhausted) and the chain is being revived via the recover handler, `ensure_daemon_service` is never called to reset or revive the service. The next tick spawned by the recovery reseed runs without the durable daemon service active, so the daemon spawns inside the transient reseed cgroup — reverting to the #1196 behavior that #1217 is designed to supersede.

See `.claude/skills/dispatch-propagate/scripts/dispatch-tick-recover` (Step 5 / `systemd-run` call) vs. `dispatch-spawn-tick` (line 163) and `dispatch-schedule-reseed` (line 385).

Surfaced by the #1217 review pass; out of scope for that PR.

## Recommended fix

Add `ensure_daemon_service || true` to `dispatch-tick-recover` before the `systemd-run` call that arms the recovery reseed, mirroring the calls already present in `dispatch-spawn-tick` and `dispatch-schedule-reseed`:

```bash
# Before arming the reseed (Step 5 of dispatch-tick-recover):
ensure_daemon_service || true
```

## Acceptance criteria

- [ ] `dispatch-tick-recover` calls `ensure_daemon_service || true` before the `systemd-run` call that arms the recovery reseed.
- [ ] `test-dispatch-scripts.sh` covers the path where `ensure_daemon_service` is called from `dispatch-tick-recover` (e.g. verifying it is invoked when a reseed is armed, matching the existing coverage pattern for `dispatch-spawn-tick` and `dispatch-schedule-reseed`).
- [ ] After the fix, when the daemon service is in `failed` state and `dispatch-tick-recover` fires, the service is reset and activated before the reseed runs — `/proc/<daemon-pid>/cgroup` reflects the durable service unit, not a transient reseed cgroup.

natb1:
Closes #1370

## What

`dispatch-tick-recover` is the `OnFailure=` systemd handler that guarantees a
dispatch-chain continuation after an abnormal tick/reseed exit. When no
continuation exists and the failure count is under the cap, its Step 5 arms a
bounded-backoff recovery reseed by calling `systemd-run --user` directly.

This PR adds `ensure_daemon_service || true` to that arming path — after
`UNIT_NAME` is computed and before the `systemd-run` call — matching the three
sibling launchers (`dispatch-spawn-tick`, `dispatch-schedule-reseed`,
`dispatch-schedule-convergence-reseed`), which all install/activate the durable
`dispatch-claude-daemon.service` (#1197) before arming.

## Why

Without this call, if the daemon service is in systemd's `failed` state (e.g.
`StartLimit` exhausted) and the chain is revived via the recover handler,
`ensure_daemon_service` never runs to reset/revive the service, so the recovery
reseed's tick reverts to the transient-cgroup behavior that #1217 superseded
(the on-demand daemon spawning inside the transient unit's cgroup, #1196).
Surfaced by the #1217 review pass; out of scope for that PR.

## Placement — arming path only

The call sits on the arming path **only**, not early/unconditional like the
siblings. The continuation no-op path (Step 2) and the cap-escalation path
(Step 4) do not spawn a tick, so they need no durable daemon. This matches the
issue's wording "before the `systemd-run` call that arms the recovery reseed".
`ensure_recover_unit` is deliberately **not** added — the recover unit must not
chain to itself.

## Tests

`test-dispatch-scripts.sh`, `dispatch-tick-recover` section:

- `tr_setup` now installs a dedicated `daemon-systemctl` logging stub and exports
  the three `DISPATCH_DAEMON_*` overrides, so the real `ensure_daemon_service`
  never writes to `~/.config/systemd/user/` or runs real `systemctl --user`
  during the suite. `tr_teardown` unsets them.
- New positive test: after a first-failure arm, the daemon log contains
  `enable --now dispatch-claude-daemon.service` — proof the call fires on the
  arming path.
- Negative assertions on the busy-worker and pending-timer continuation no-op
  paths: the daemon log stays empty, locking in arming-path-only placement.

The `/proc/<daemon-pid>/cgroup` runtime property (AC#3) is satisfied by
construction and is not unit-testable under the stub harness — it is covered by
the call now existing on the arming path plus `ensure_daemon_service`'s own
existing install/activate tests.

Full suite: 2415/2415 passing.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-tick-recover
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-tick-recover
@@ -344,6 +344,13 @@ fi
 RESEED_AT=$(( NOW + BACKOFF ))
 UNIT_NAME="dispatch-reseed-${RESEED_AT}"
 
+# Install + activate the durable dispatch-claude-daemon.service (#1197) before
+# arming the recovery reseed, so the reseed's tick attaches to a daemon already
+# running in a stable cgroup rather than spawning one in the transient reseed
+# unit's cgroup. Best-effort: a failure here is non-fatal — the on-demand daemon
+# then falls back to surviving via KillMode=process (#1196).
+ensure_daemon_service || true
+
 RUN_STDERR=$(mktemp)
 trap 'rm -f "$RUN_STDERR"' EXIT
 # Capture the exit code directly — after `if cmd; then ... fi`, `$?` is the exit
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -12207,6 +12207,23 @@ exit 0
 STUB
   chmod +x "$TMPDIR_TEST/bin/gh"
 
+  # ensure_daemon_service (lib.sh) would otherwise write to the real
+  # ~/.config/systemd/user/ and run a real `systemctl --user`. Wire a separate
+  # logging stub (distinct from the recover systemctl) so ensure_daemon_service
+  # calls land in daemon-systemctl-log and are separately assertable.
+  cat > "$TMPDIR_TEST/bin/daemon-systemctl" <<STUB
+#!/usr/bin/env bash
+echo "\$*" >> "$TMPDIR_TEST/daemon-systemctl-log"
+exit 0
+STUB
+  chmod +x "$TMPDIR_TEST/bin/daemon-systemctl"
+  export DISPATCH_DAEMON_UNIT_DIR="$TMPDIR_TEST/daemon-systemd-user"
+  export DISPATCH_DAEMON_SYSTEMCTL_CMD="$TMPDIR_TEST/bin/daemon-systemctl"
+  # ensure_daemon_service only needs a non-empty, newline/quote-free path; it
+  # does not execute this binary in tests — the claude stub already exists for
+  # the CLAUDE_AGENTS_CMD path, so reuse it.
+  export DISPATCH_DAEMON_CLAUDE_CMD="$TMPDIR_TEST/bin/claude"
+
   export DISPATCH_TICK_RECOVER_SYSTEMD_RUN_CMD="$TMPDIR_TEST/bin/systemd-run"
   export DISPATCH_TICK_RECOVER_SYSTEMCTL_CMD="$TMPDIR_TEST/bin/systemctl"
   export DISPATCH_TICK_RECOVER_GH_CMD="$TMPDIR_TEST/bin/gh"
@@ -12235,6 +12252,7 @@ tr_teardown() {
   unset DISPATCH_TICK_RECOVER_BASE_BACKOFF
   unset DISPATCH_TICK_RECOVER_MAX_BACKOFF
   unset DISPATCH_TICK_RECOVER_RESET_WINDOW
+  unset DISPATCH_DAEMON_UNIT_DIR DISPATCH_DAEMON_SYSTEMCTL_CMD DISPATCH_DAEMON_CLAUDE_CMD
 }
 
 # tr_seed_state <count> <last_failure> — write the consecutive-failure state.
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — #1370 / PR #1533

### Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | `dispatch-tick-recover` calls `ensure_daemon_service` before arming the reseed | shell-command | ✅ PASS |
| 2 | Test suite covers the `ensure_daemon_service` call from `dispatch-tick-recover` | shell-command | ✅ PASS |
| 3 | Failed daemon service is reset and activated before the reseed runs (durable unit, not transient cgroup) | — | ⏸ Deferred to office-hours |

**Counts:** 2 PASS, 0 FAIL, 0 SKIP, 1 deferred to office-hours.

### Deferred to Office-Hours

**Item 3 — Runtime cgroup property (AC#3):** Verifying that `/proc/<daemon-pid>/cgroup` reflects the durable `dispatch-claude-daemon.service` unit (not a transient reseed cgroup) when the daemon is in `failed` state requires a live systemd environment with real cgroup namespacing. This cannot be reproduced under the stub harness and is noted in the PR body as "satisfied by construction." A human reviewer should confirm by inspection that `ensure_daemon_service || true` is correctly ordered before the `systemd-run` arming call — or accept the PR body's construction argument.

### Bugs Found

None.

### Notes

- No browser component detected; no QA server was started.
- Test suite: 2437/2437 passing after merging `origin/main` (PR reports 2415 pre-merge; the delta is from new tests on `main`).

natb1:
## Review summary — #1370

### Fixed

- **test-dispatch-scripts.sh:12804** — Test 4 (cap-escalation path) was missing a negative assertion confirming `ensure_daemon_service` did not run, unlike the matching assertions added to the continuation no-op paths (Tests 2 and 3). Added `assert_eq 'cap: daemon service not touched (cap-escalation path)'` after the cap/cap-latched teardown. (ca4dc273)

### Informational

- **test-dispatch-scripts.sh:12757** — Test 1b runs the same scenario as Test 1 (first failure, no continuation) through a full `tr_setup/invoke/tr_teardown` cycle; its sole purpose is to assert the daemon log. This is minor duplication but the separation is intentional for readability — no action needed.

### False-positive

- **test-dispatch-scripts.sh:12664** — The `daemon-systemctl` stub is written via a `<<STUB` heredoc with `$TMPDIR_TEST` interpolated. `$TMPDIR_TEST` is set unconditionally earlier in the suite to a `mktemp -d` path; it is safe and intentional shell interpolation, not an injection risk.
- **dispatch-tick-recover:355** — `ensure_daemon_service || true` silencing all failure codes is intentional best-effort semantics: the PR body and the inline comment both document that this call is non-fatal and the fall-through behavior (on-demand daemon in KillMode=process) is the documented fallback (#1196). The silencing is correct by design.

### Out-of-scope

The following findings are in `lib.sh` — pre-existing code the diff does not touch — and are not introduced by this PR:

- **lib.sh:1090** — `safe_path` in `ensure_daemon_service` strips newline and double-quote but not backslash or other metacharacters.
- **lib.sh:1092** — `DISPATCH_DAEMON_UNIT_DIR` env-var override accepted without path sanitization.
- **lib.sh:1079** — `CLAUDE_CMD` validation rejects newline/double-quote but not backslash.
- **dispatch-tick-recover:355** — `DISPATCH_DAEMON_CLAUDE_CMD` interpolated into a heredoc; the sanitization lives in `ensure_daemon_service` itself (lib.sh), not in this diff.

_All out-of-scope findings are in pre-existing code; no follow-up issues filed (below the meaningfulness threshold for this surface)._

PR #1541, opened 2026-06-13
budget-etl: bound the CSV duplicate-ID suffix scan with a defensive cap

### LINKED ISSUE #1373: budget-etl: unbounded suffix loop in collision-free duplicate-ID dedup (DoS vector)
The duplicate-ID suffix search in \`budget-etl/internal/parse/csv.go\` uses an unbounded \`for n := 2; ; n++\` loop (lines 104–114). For a pathological input where the original file contains IDs \`X\`, \`X-2\`, \`X-3\`, ..., \`X-1000000\`, and two rows with duplicate \`X\`, the loop iterates 999999 times before finding a free suffix. While this terminates (the original set is finite and the suffix space is infinite), it represents a denial-of-service vector for adversarially crafted input files processed by an ETL pipeline.

**Root cause:** The \`original\` set populated in the first pass (lines 62–71) is finite, so the loop always terminates, but the cost is O(|original ∩ {X-2, X-3, …}|) per duplicate row — unbounded by any explicit cap.

**Recommended fix:** Cap the loop at \`len(records)\` (the number of rows in the file) and return an error if the cap is exceeded. This bound is always sufficient for non-adversarial input (you cannot have more duplicate suffixes than rows) and surfaces corruption for adversarial cases.

```go
for n := 2; n <= len(records); n++ {
    candidate := fmt.Sprintf("%s-%d", txnID, n)
    _, inUsed := used[candidate]
    _, inOriginal := original[candidate]
    if !inUsed && !inOriginal {
        txnID = candidate
        break
    }
    if n == len(records) {
        return ParseResult{}, fmt.Errorf("%s: line %d: cannot find unique suffix for duplicate ID %q", path, i+2, txnID)
    }
}
```

**Files:** \`budget-etl/internal/parse/csv.go\`

Surfaced by the review pass on #1275; out of scope for that PR.

## Acceptance criteria

- [ ] The suffix search loop is bounded by \`len(records)\` (or equivalent).
- [ ] When the cap is exceeded, \`parseCSV\` returns a descriptive error identifying the row and the original ID.
- [ ] Existing \`TestParseCSV_DuplicateIDs\` and \`TestParseCSV_DuplicateIDCollision\` tests continue to pass.
- [ ] A new test \`TestParseCSV_DuplicateIDCapExceeded\` exercises the error path: a CSV where the original set contains \`X-2\` through \`X-{N}\` for all \`N <= len(records)\`, leaving no free suffix.

natb1:
Closes #1373

## What

`budget-etl/internal/parse/csv.go` dedups duplicate transaction IDs by appending
a numeric suffix (`X` → `X-2`, `X-3`, …), scanning from `n=2` until it finds a
suffix free of both already-assigned IDs and real later IDs. The scan loop was
written unbounded (`for n := 2; ; n++`), which #1373 flagged as a potential DoS
vector.

This PR bounds the scan at `len(records)` and documents the bound as a defensive
termination guard.

## Why this shape

Two facts established during planning drive the chosen fix:

1. **The loop is already implicitly bounded.** Every blocking `X-n` candidate
   must be an ID literally present in the finite file, so a free suffix always
   exists at or below `n = len(records)-1`. The loop can never iterate more than
   ~`len(records)` times even today.

2. **The cap-exceeded error branch is provably unreachable.** To exhaust
   `X-2 … X-{len(records)}` you would need more distinct `X-n` blocker rows than
   the file can hold once the base row and the triggering duplicate are counted —
   you always fall short. There is no input that reaches the error.

`budget-etl` only processes trusted input (the user's own bank exports, run by a
local CLI / the `budget-parse-job` path over the user's own files) — there is no
untrusted submitter, so the "DoS" is theoretical. The right-sized goal is
**defensive hardening**: make the loop's termination *explicit in source* (so a
future refactor that breaks the implicit-finiteness invariant is caught) without
pretending to fix an attack that cannot occur. The explicit cap is therefore a
documented defensive guard, and its error path is unreachable on any input.

## Test

Because the cap-exceeded branch is unreachable, the issue's
originally-requested `TestParseCSV_DuplicateIDCapExceeded` cannot exist. Per the
user-approved resolution it is replaced by `TestParseCSV_DuplicateIDDenseCollision`,
which exercises the scan skipping a dense run of taken suffixes (file order
`X, X-2, X-3, X` → the duplicate `X` lands on `X-4`) and asserts a clean,
no-error result.

The pre-existing `TestParseCSV_DuplicateIDs` and `TestParseCSV_DuplicateIDCollision`
(which encode the collision-safe scan from #1275's review) stay green.

## Verification

```
cd budget-etl && go test ./internal/parse/   # all pass, incl. the new dense-collision test
cd budget-etl && go vet ./internal/parse/    # clean
```

### DIFF
--- a/budget-etl/internal/parse/csv.go
+++ b/budget-etl/internal/parse/csv.go
@@ -102,14 +102,25 @@ func parseCSV(path string) (ParseResult, error) {
 		}
 
 		if _, taken := used[txnID]; taken {
-			for n := 2; ; n++ {
+			// Defensive termination guard. The scan is implicitly bounded: every
+			// blocking X-n candidate must be an ID present in the finite file, so a
+			// free suffix always exists at or below n = len(records)-1. This explicit
+			// cap makes termination guaranteed in source against a future change that
+			// breaks that invariant; the error is not reachable on any input (you
+			// cannot pack more distinct X-n blockers into the file than it has rows).
+			found := false
+			for n := 2; n <= len(records); n++ {
 				candidate := fmt.Sprintf("%s-%d", txnID, n)
 				_, inUsed := used[candidate]
 				_, inOriginal := original[candidate]
 				if !inUsed && !inOriginal {
 					txnID = candidate
+					found = true
 					break
 				}
+			}
+			if !found {
+				return ParseResult{}, fmt.Errorf("%s: line %d: cannot find unique suffix for duplicate ID %q within %d candidates", path, i+2, txnID, len(records))
 			}
 		}
 		used[txnID] = struct{}{}
--- a/budget-etl/internal/parse/csv.go
+++ b/budget-etl/internal/parse/csv.go
@@ -109,7 +109,7 @@ func parseCSV(path string) (ParseResult, error) {
 			// breaks that invariant; the error is not reachable on any input (you
 			// cannot pack more distinct X-n blockers into the file than it has rows).
 			found := false
-			for n := 2; n <= len(records); n++ {
+			for n := 2; n <= len(records)-1; n++ {
 				candidate := fmt.Sprintf("%s-%d", txnID, n)
 				_, inUsed := used[candidate]
 				_, inOriginal := original[candidate]
@@ -120,7 +120,7 @@ func parseCSV(path string) (ParseResult, error) {
 				}
 			}
 			if !found {
-				return ParseResult{}, fmt.Errorf("%s: line %d: cannot find unique suffix for duplicate ID %q within %d candidates", path, i+2, txnID, len(records))
+				return ParseResult{}, fmt.Errorf("%s: line %d: cannot find unique suffix for duplicate ID %q within %d candidates", path, i+2, txnID, len(records)-1)
 			}
 		}
 		used[txnID] = struct{}{}
--- a/budget-etl/internal/parse/csv_test.go
+++ b/budget-etl/internal/parse/csv_test.go
@@ -148,3 +148,44 @@ func TestParseCSV_DuplicateIDCollision(t *testing.T) {
 		t.Errorf("txn 2: ID = %q, want %q", txns[2].TransactionID, "X-2")
 	}
 }
+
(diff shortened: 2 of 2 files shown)


natb1:
## QA summary — issue #1373 / PR #1541

All items script-verifiable (pure Go backend change, no browser component).

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Loop bound exists in source (`n <= len(records)` at csv.go:112) | PASS |
| 2 | Defensive cap-exceeded error guard in source (`found` sentinel at csv.go:111,118,122) | PASS |
| 3 | Existing regression tests: `TestParseCSV_DuplicateIDs`, `TestParseCSV_DuplicateIDCollision` | PASS |
| 4 | New dense-collision test: `TestParseCSV_DuplicateIDDenseCollision` | PASS |
| 5 | `go vet ./internal/parse/` clean | PASS |

**5 PASS / 0 FAIL / 0 SKIP**

No `needs-human-judgment` items. No bugs found.

### Notes

The cap-exceeded error path (`!found` guard) is documented as a provably unreachable defensive guard — per the user-approved resolution, criterion #4 was replaced by the dense-collision test rather than an error-path test. CI is passing.

natb1:
## Review findings (commit 1130969f)

### Fixed

- **`budget-etl/internal/parse/csv.go:112`** — Loop bound off-by-one: condition was `n <= len(records)` but the comment stated "at or below `n = len(records)-1`". Tightened to `n <= len(records)-1` to align code with the documented bound (which is sufficient — a free suffix is always guaranteed at or below that index). (code-review)

- **`budget-etl/internal/parse/csv.go:123`** — Error message reported "`within %d candidates`" using `len(records)`, but the loop attempts candidates `n=2…len(records)-1`, which is `len(records)-1` candidates. Changed format argument to `len(records)-1`. (code-review)

### Security

No attack surface found beyond `app_or_rules=true` scope; no open CodeQL alerts on this PR ref; no dependency changes. Security review: clean.

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

### Out-of-scope

_None._

PR #1543, opened 2026-06-13
dispatch-stop & office-hours-strip: short read timeout so the hooks never stall 1s

natb1:
Closes #1519

## Summary

Two sibling hooks — `dispatch-stop.sh` (Stop) and
`dispatch-office-hours-strip.sh` (UserPromptSubmit) — still carried the
`read -t 1 -d '' PAYLOAD` drain pattern that #1491 (merged as PR #1511) fixed in
`dispatch-input-block.sh`. `-d ''` sets the read delimiter to NUL (`\x00`); JSON
hook payloads never contain NUL, so on an open-but-idle stdin (hook events do not
close stdin at EOF) the `read` always burned its full 1-second timeout. That
stalled the hook by ~1s on **every** Stop event and **every** UserPromptSubmit.

In both hooks the payload is **drain-only** — read into a variable
(`PAYLOAD` / `_PAYLOAD`) that is never referenced again. Neither hook parses the
payload; both read solely to drain buffered stdin and avoid an upstream pipe
deadlock. So the `read -rt 0.1` change from PR #1511 applies cleanly and is
strictly safer here than in input-block (which *does* parse its payload
downstream) — the unused `-r` is kept only as the conventional safe form, not for
any backslash-preservation reason.

Outcome: both hooks drain stdin in ≤0.1s instead of stalling 1s.

## Changes

- `dispatch-stop.sh`: `read -t 1 -d '' PAYLOAD` → `read -rt 0.1 PAYLOAD`, with an
  updated comment explaining the short timeout.
- `dispatch-office-hours-strip.sh`: `read -t 1 -d '' _PAYLOAD` →
  `read -rt 0.1 _PAYLOAD`, same comment update.
- `test-dispatch-scripts.sh`: one regression test per hook, each feeding an open,
  non-newline-held stdin under `timeout 0.5` and asserting the hook returns
  before the timeout (`rc != 124`). Each test also asserts the hook's normal
  behavior is unaffected — the stop test confirms the advance disposition still
  spawns exactly once; the strip test confirms the label strip (`gh pr edit …
  --remove-label`) still fires. Both mirror PR #1511's input-block stall test
  shape.

## Verification

Full dispatch test suite passes (2437/2437), including the four new
`open-stdin` PASS lines.

### DIFF
--- a/.claude/hooks/dispatch-office-hours-strip.sh
+++ b/.claude/hooks/dispatch-office-hours-strip.sh
@@ -16,9 +16,11 @@
 set -uo pipefail
 trap 'echo "[dispatch-office-hours-strip] WARNING: unexpected error on line $LINENO" >&2; exit 0' ERR
 
-# Consume the payload (UserPromptSubmit delivers JSON on stdin; we do not read
-# any field, but consuming the stdin avoids upstream pipe deadlocks).
-if read -t 1 -d '' _PAYLOAD; then :; fi
+# Drain the payload fast (UserPromptSubmit delivers JSON on stdin; we read no
+# field from it, only drain to avoid an upstream pipe deadlock). Short timeout
+# so an open, idle stdin returns in ~0.1s rather than stalling a full second
+# on the NUL delimiter.
+if read -rt 0.1 _PAYLOAD; then :; fi
 
 BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null) || exit 0
 ISSUE_NUM=$(printf '%s\n' "$BRANCH" | grep -oE '^[0-9]+') || exit 0
--- a/.claude/hooks/dispatch-stop.sh
+++ b/.claude/hooks/dispatch-stop.sh
@@ -96,9 +96,12 @@ if ! [[ "$JOB_NAME" =~ ^[0-9]+- ]]; then
   exit 0
 fi
 
-# Consume the payload (defensive — Stop may pass JSON on stdin). Unused.
+# Drain any buffered single-line JSON payload fast (defensive — Stop may
+# pass JSON on stdin, but this hook never reads a field from it). Short
+# timeout so an open, idle stdin (hook events never close stdin at EOF)
+# returns in ~0.1s rather than stalling a full second on the NUL delimiter.
 PAYLOAD=""
-if read -t 1 -d '' PAYLOAD; then :; fi
+if read -rt 0.1 PAYLOAD; then :; fi
 
 # Resolve issue number from the validated JOB_NAME (<N>-<slug>). Discriminator 2
 # guarantees JOB_NAME matches ^[0-9]+-, so the numeric prefix is non-empty.
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -16267,6 +16267,32 @@ else
 fi
 stop_teardown
 
+# --- Test: open non-newline stdin → hook returns fast (no 1s read stall) (#1519) ---
+echo "Test: stop hook open non-newline stdin → returns fast, normal disposition still fires"
+stop_setup
+echo "123-foo-bar" > "$STUB_DIR/current-branch.txt"
+echo "456" > "$STUB_DIR/find-pr-output"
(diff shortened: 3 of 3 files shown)


natb1:
## QA Summary — Issue #1519

**PR #1543** · No browser component (pure shell hook change)

### Items Executed

| 1 | `dispatch-stop.sh` uses `read -rt 0.1` | script-verifiable | PASS |
| 2 | `dispatch-office-hours-strip.sh` uses `read -rt 0.1` | script-verifiable | PASS |
| 3 | Both hooks syntactically valid bash | script-verifiable | PASS |
| 4 | Regression tests present in `test-dispatch-scripts.sh` | script-verifiable | PASS |
| 5 | Full dispatch test suite passes | script-verifiable | PASS |

**Counts:** 5 PASS · 0 FAIL · 0 SKIP · 0 deferred to office-hours

### Details

- `dispatch-stop.sh`: confirmed `read -rt 0.1 PAYLOAD`; old `read -t 1 -d ''` pattern gone
- `dispatch-office-hours-strip.sh`: confirmed `read -rt 0.1 _PAYLOAD`; old pattern gone
- Both hooks pass `bash -n` syntax check
- 3 `timeout 0.5` regression tests present (input-block, strip, stop) with `rc != 124` assertions
- Full suite: **2437/2437 passed, 0 failed**
- New open-stdin tests all PASS:
  - `stop open-stdin: hook completed before the 0.5s timeout` ✓
  - `stop open-stdin: spawn invoked exactly once` ✓
  - `strip open-stdin: hook completed before the 0.5s timeout` ✓
  - `strip open-stdin: 'gh pr edit 456 --remove-label' still invoked` ✓

### Bugs Found

None.

natb1:
## Review findings

### Fixed

- **test-dispatch-scripts.sh:15252** — Misleading FAIL message in `strip open-stdin` correctness test: the `FAIL` branch said "still invoked" (identical to the PASS message). Fixed to "NOT invoked (drain broke strip)" so failures are distinguishable. Also added a symmetric `gh-issue-edit.log` assertion mirroring Test 1's structure — the prior test only checked `gh-pr-edit.log`, not that the issue label was also stripped. *(sources: code-review, review)* — fix commit 87fb35fa

### Informational

- **test-dispatch-scripts.sh:16318** — The new stop-hook open-stdin test fires Branch B (phase advanced: `MARKER_PHASE=implement`, `CURRENT_PHASE=fix-checks`). Branch B calls `dispatch-spawn` once, so `spawn_calls=1` is the expected assertion. This is correct; no change needed.
- **dispatch-stop.sh:99** — Minor comment phrasing inconsistency: `dispatch-stop.sh` says "single-line JSON payload" while `dispatch-office-hours-strip.sh` says "JSON on stdin". Both are accurate descriptions of the same behavior; no change needed.

### Dismissed

- **dispatch-stop.sh:99–104** — Comment says "single-line JSON payload" but `read -rt 0.1` only drains the first newline-terminated line, so a multi-line payload would leave remainder in the pipe buffer. The hooks are drain-only — neither parses nor uses the payload downstream — so partial drain is correct behavior; full NUL-delimited drain is not needed. Not a bug. *(source: review)*
- **dispatch-stop.sh:103** — `PAYLOAD=""` initializer immediately before `read -rt 0.1 PAYLOAD` is redundant (`read` sets it to empty on timeout). No behavioral impact; retained for clarity in the original PR. *(source: review)*

### False-positive (security)

- **dispatch-stop.sh:104** — Concern that switching from `read -t 1 -d '' PAYLOAD` (NUL delimiter, full drain) to `read -rt 0.1 PAYLOAD` (newline delimiter, partial drain) could leave data in the pipe buffer causing upstream deadlock. Both hooks are drain-only; the upstream hook writer (`claude`/`anthropic`) does not block on the drain completing. The original 1-second NUL-wait was itself the problem (#1491). Not a real vulnerability. *(source: red-team)*
- **dispatch-office-hours-strip.sh:23** — Same partial-drain concern as above for the UserPromptSubmit hook. Same reasoning applies. Not a real vulnerability. *(source: red-team)*
- **test-dispatch-scripts.sh:16318** — Concern that `wc -l < "$STUB_DIR/spawn-calls.log" 2>/dev/null || echo 0` could miscount if spawn-calls.log has no trailing newline. The stub writes one line per call via `echo`, which always adds a newline; `wc -l` is correct. Not a bug. *(source: red-team)*

PR #1548, opened 2026-06-13
dispatch-select-target: count only landed appends so a dropped leaf cannot short-circuit mixed-phase root descent

### LINKED ISSUE #1403: root_recorded increments on bucket_append no-op, can short-circuit mixed-phase root descent in dispatch-select-target --top
## Context

Found during review of PR #1402 (issue #1317: dispatch fan-out worker concurrency fix).

## Finding

\`bucket_append\` silently drops an entry when the phase bucket is already at TOP capacity (\`dispatch-select-target:667\`). \`root_recorded\` is incremented unconditionally at line 887 regardless of whether \`bucket_append\` stored the leaf or dropped it.

For TOP=1 this is harmless: a root only ever contributes one leaf, and the bucket can only hold one entry, so \`bucket_append\` never drops in practice.

For TOP>1: when a root has leaves in two phases (implement and plan) and the first phase bucket was already filled by a prior root, \`bucket_append\` silently drops the leaf. \`root_recorded\` still increments, and if it reaches TOP on the dropped leaf, the loop breaks before descending to the plan-phase leaves of this root. The plan slot for this root remains unfilled for the current fan-out run.

This requires a multi-leaf, multi-phase root with one phase bucket pre-filled by a prior root — an edge case not covered by current tests.

**Location:** \`.claude/skills/dispatch-propagate/scripts/dispatch-select-target:882-894\`

Relevant code:
\`\`\`bash
bucket_append FIRST_ISSUE "$leaf_bucket" "$leaf"   # line 885 — silently drops if full
root_excl["$leaf"]=1
root_recorded=$(( root_recorded + 1 ))             # line 887 — increments unconditionally
\`\`\`

## Recommended fix

Only increment \`root_recorded\` when \`bucket_append\` actually inserts. One approach: make \`bucket_append\` return 1 on a no-op drop and 0 on an actual insert, then gate the increment:

\`\`\`bash
bucket_append() {
  local -n _ba_ref="$1"
  local _ba_key="$2" _ba_entry="$3"
  (( $(bucket_count "$1" "$_ba_key") >= TOP )) && return 1   # 1 = dropped
  local _ba_cur="${_ba_ref[$_ba_key]:-}"
  if [[ -z "$_ba_cur" ]]; then
    _ba_ref[$_ba_key]="$_ba_entry"
  else
    _ba_ref[$_ba_key]="$_ba_cur"$'\n'"$_ba_entry"
  fi
}

# in the issue scan loop:
if bucket_append FIRST_ISSUE "$leaf_bucket" "$leaf"; then
  root_recorded=$(( root_recorded + 1 ))
fi
\`\`\`

Alternatively, track `root_contributed` (actual inserts) separately from `root_attempted` (all trace returns). Does not affect TOP=1 behavior.

## Acceptance criteria

- [ ] \`bucket_append\` signals whether it inserted or dropped (exit status or return value).
- [ ] \`root_recorded\` (or an equivalent counter) increments only when a leaf was actually inserted into a bucket, not when it was dropped due to the cap.
- [ ] For TOP=1, behavior is byte-identical to the current implementation.
- [ ] A test covers the mixed-phase root edge case: a root with implement and plan leaves where one phase bucket is pre-filled by a prior root — the second-phase leaf is still reached and recorded.

natb1:
Closes #1403

## Problem

`dispatch-select-target`'s single-pass `--top N` fan-out scan descends each
help-wanted root to up to `N` startable leaves, filing each leaf into a
per-`(category, priority, phase)` bucket via `bucket_append` (capped at `TOP`).
The per-root descent is bounded by a `root_recorded` counter and the guard
`(( root_recorded >= TOP )) && break`.

`root_recorded` was incremented **unconditionally** after every `bucket_append`,
including when the append was a silent no-op drop (the bucket was already at
`TOP`). So a root whose leaves traced into an already-full bucket inflated
`root_recorded` on the drops, and once it hit `TOP` the loop broke **before
descending to that root's leaves in the other phase** — e.g. a root's
higher-rank `implement` leaf was skipped after its `plan` leaves dropped into a
full `plan` bucket.

## Fix

`root_recorded` now counts only leaves actually **inserted**:

- `bucket_append` returns `0` on insert and `1` on a full-bucket drop (was
  always `0`). The doc comment now states this contract.
- Both call sites are guarded against `set -euo pipefail`:
  - PR queue site: `bucket_append … || true` (it tracks no counter; a routine
    full-bucket PR drop must not abort the script — PR buckets drop routinely,
    including at `TOP=1`).
  - Issue-leaf site: `if bucket_append …; then root_recorded=$(( … + 1 ))`. The
    leaf is added to `root_excl` unconditionally first (a traced leaf is excluded
    from re-tracing whether it landed or dropped, unchanged from before).
- A deliberate `(( TOP == 1 )) && break` in the drop branch keeps `TOP=1` output
  **byte-identical** to the pre-fix loop: a root that hits a full bucket on a
  landed-append attempt still stops descending at `TOP=1`. `TOP>1` now descends
  past the drop to the root's other-phase leaves.

## Tests

Two new `--top N` tests in `test-dispatch-scripts.sh`, each isolating one half of
the fix (verified red-green in both directions):

- **T4e** (`--top 2`): prior roots fill the `plan` bucket to `TOP=2`; the test
  root's two plan leaves drop, and descent still reaches its `implement` leaf,
  which emission ranks first. RED against the pre-fix script (which broke before
  the implement leaf), GREEN after.
- **T4f** (`--top 1`): the `(( TOP == 1 )) && break` freeze keeps the output
  byte-identical to the plain no-flag call. RED against a no-freeze "naive" fix
  (which would descend to the implement leaf and change the single `TOP=1`
  winner), GREEN with the freeze.

Full suite: 2438/2438 pass.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -651,17 +651,20 @@ bucket_count() {
 # holds fewer than TOP entries (the per-bucket cap). An empty bucket is set to
 # the bare entry (no leading newline), so a TOP=1 bucket expands to exactly the
 # entry — preserving the former single-entry map's byte layout for the emission
-# path.
+# path. Returns 0 when the entry is inserted, 1 when the bucket is already at TOP
+# and the entry is dropped (the caller observes this to decide whether the entry
+# actually landed).
 bucket_append() {
   local -n _ba_ref="$1"
   local _ba_key="$2" _ba_entry="$3"
-  (( $(bucket_count "$1" "$_ba_key") >= TOP )) && return 0
+  (( $(bucket_count "$1" "$_ba_key") >= TOP )) && return 1
   local _ba_cur="${_ba_ref[$_ba_key]:-}"
   if [[ -z "$_ba_cur" ]]; then
     _ba_ref[$_ba_key]="$_ba_entry"
   else
     _ba_ref[$_ba_key]="$_ba_cur"$'\n'"$_ba_entry"
   fi
+  return 0
 }
 
 # group_emittable_count <category> <priority-bit> — total entries the emission
@@ -814,7 +817,7 @@ while IFS=$'\t' read -r pr_num _created pr_branch pr_closing; do
   pr_cat=$(category_of_labels "$pr_combined")
   pr_pri=$(has_priority_in_labels "$pr_combined")
   pr_key="$pr_cat|$pr_pri|$phase"
-  bucket_append FIRST_PR "$pr_key" "$pr_num $pr_branch"
+  bucket_append FIRST_PR "$pr_key" "$pr_num $pr_branch" || true
 done <<< "$PR_SORTED"
 
 if [[ "$QA_MODE" == "true" ]]; then
@@ -1017,9 +1020,16 @@ while IFS=$'\t' read -r issue_num issue_pri issue_catrank issue_cat; do
       leaf_phase=plan
       [[ -n "${ISSUE_PLANNED[$leaf]:-}" ]] && leaf_phase=implement
       leaf_bucket="$issue_cat|$issue_pri|$leaf_phase"
-      bucket_append FIRST_ISSUE "$leaf_bucket" "$leaf"
       root_excl["$leaf"]=1
-      root_recorded=$(( root_recorded + 1 ))
+      if bucket_append FIRST_ISSUE "$leaf_bucket" "$leaf"; then
+        root_recorded=$(( root_recorded + 1 ))
+      else
+        # Bucket already at TOP — this leaf was dropped. Criterion #3 freezes
+        # TOP=1: a root that hits a full bucket on a landed-append attempt stops
+        # descending, byte-identical to the pre-fix loop. TOP>1 keeps descending
+        # past the drop to this root's other-phase leaves (#1403).
+        (( TOP == 1 )) && break
+      fi
     else
       trace_status=$?
       [[ "$trace_status" -eq 2 ]] && break
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -985,8 +985,9 @@ while IFS=$'\t' read -r issue_num issue_pri issue_catrank issue_cat; do
       # Phase-key the leaf (NOT the help-wanted root — the traced leaf may differ
       # from issue_num). dispatch:planned → implement, else plan, mirroring
       # dispatch-phase. bucket_append caps each phase bucket at TOP and keeps the
-      # oldest entries; a leaf traced into an already-full bucket is silently
-      # dropped (the cap subsumes the former first-seen guard).
+      # oldest entries; a leaf traced into an already-full bucket is dropped — the
+      # return code gates root_recorded and, at TOP=1, short-circuits descent
+      # (the cap subsumes the former first-seen guard).
       leaf_phase=plan
       [[ -n "${ISSUE_PLANNED[$leaf]:-}" ]] && leaf_phase=implement
       leaf_bucket="$issue_cat|$issue_pri|$leaf_phase"
@@ -995,7 +996,7 @@ while IFS=$'\t' read -r issue_num issue_pri issue_catrank issue_cat; do
         root_recorded=$(( root_recorded + 1 ))
       else
         # Bucket already at TOP — this leaf was dropped. Criterion #3 freezes
-        # TOP=1: a root that hits a full bucket on a landed-append attempt stops
+        # TOP=1: a root that hits a full bucket on an append attempt stops
         # descending, byte-identical to the pre-fix loop. TOP>1 keeps descending
         # past the drop to this root's other-phase leaves (#1403).
         (( TOP == 1 )) && break
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -3320,6 +3320,62 @@ assert_eq "--top 3 descent skips parked leaf 5501 → 5500,5502" "$(printf 'issu
 assert_eq "--top 3 descent: parked leaf 5501 absent" "0" "$(printf '%s\n' "$result" | grep -c '^issue 5501$')"
 teardown
 
+# T4e. Mixed-phase root descent is not short-circuited by full-bucket drops
+#      (#1403). `root_recorded` counts only leaves actually INSERTED, so a root
+#      whose first leaves drop into an already-full bucket still descends to its
+#      other-phase leaf. Prior roots 6010, 6011 (no sub-issues → each its own
+#      plan leaf) fill the `plan` bucket to TOP=2. Root 6020 then descends three
+#      sub-issues in order: 6021, 6022 (both plan → both DROP into the full plan
+#      bucket) then 6023 (dispatch:planned → implement). With the fix the two
+#      plan drops do not advance `root_recorded`, so descent reaches 6023, which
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — Issue #1403 / PR #1548

**Branch:** `1403-root-recorded-increments-on`
**Scope:** Pure-script fix — `dispatch-select-target` + `test-dispatch-scripts.sh`. No browser component.

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | `bucket_append` signals insert vs drop (interface contract) | shell-command | PASS |
| 2 | `root_recorded` increments only on a real insert | shell-command | PASS |
| 3 | PR call-site drop guard survives `set -e` | shell-command | PASS |
| 4 | TOP=2 mixed-phase descent reaches implement leaf past full-plan drops | shell-command | PASS |
| 5 | TOP=1 byte-parity freeze | shell-command | PASS |
| 6 | Full regression suite is green | shell-command | PASS |
| 7 | Byte-identical for TOP=1 holds universally (not just on the two fixtures) | — | deferred |

**Counts:** 6 PASS · 0 FAIL · 0 SKIP · 1 deferred to office-hours

### Test suite

2510/2510 passed, 0 failed. T4e (mixed-phase TOP=2) and T4f (TOP=1 byte-parity freeze) both green.

### Deferred to office-hours

**Item 7 — "Byte-identical for TOP=1" holds universally (needs-human-judgment):**
Two test fixtures (T4f, T5) are necessary but not sufficient to prove the universal claim. A reviewer needs to confirm by inspection that the sole TOP=1 behavioral change introduced by this fix is the `(( TOP == 1 )) && break` in the drop branch — which mirrors the pre-fix first-seen stop — and that no other code path diverges from pre-fix behavior at TOP=1.

### Bugs found

None.

natb1:
## QA follow-up — item 7 resolved (TOP=1 byte-parity holds universally)

The earlier QA comment deferred item 7 (acceptance criterion #3: *"For TOP=1, behavior is byte-identical to the current implementation"*) as `needs-human-judgment`, because two passing fixtures (T4f, T5) are necessary but not sufficient to prove a *universal* claim. Resolving it by code inspection:

**Claim:** at `TOP=1`, the patched `dispatch-select-target` produces byte-identical output to the pre-fix script for **all** inputs, not just the two test fixtures.

**`root_recorded` has exactly three uses:** reset to 0 per root (line 940), the descent break-guard `(( root_recorded >= TOP )) && break` (line 944), and the gated increment (line 995). Nothing else reads it.

**Per-root descent at `TOP=1`, by outcome:**

- **Readiness-skip branches** (lines 977–983: `root_excl[leaf]=1; continue`) — untouched by the fix; no `root_recorded` change. Identical.
- **Trace failure** (lines 1003–1008) — untouched by the fix. Identical.
- **Leaf reaches `bucket_append`** (lines 990–1002):
  - **Insert** (bucket count `0 < 1`): post-fix `bucket_append` returns 0 → `root_recorded`→1 → next iteration line 944 breaks. Pre-fix: insert, `root_recorded`→1, next iteration breaks. **Identical** — same bucket contents, same counter.
  - **Drop** (bucket count `1 ≥ 1`): post-fix `bucket_append` drops without mutating, returns 1 → `(( TOP==1 )) && break` fires immediately. Pre-fix: drop without mutating, `root_recorded`→1, next iteration line 944 breaks. **Same observable result** — both leave the bucket unchanged, both set `root_excl[leaf]`, both terminate this root after one bucket-section leaf. The sole difference is that post-fix leaves `root_recorded` at 0 instead of 1, but that value is reset to 0 for the next root (line 940) and never read again — a **dead store**.

**Other surfaces:**
- **PR call site** (line 789, `|| true`): return value discarded at any `TOP`; bucket-mutation logic unchanged → PR buckets byte-identical.
- **`bucket_append` internal:** the only change is the cap-path return value (`0`→`1`); the data mutation is byte-identical. The return value is consumed at exactly two call sites (confirmed by grep), both handled.

Global emission depends only on bucket contents, which are identical at `TOP=1` → **output is byte-identical for all inputs**. The deliberate `(( TOP == 1 )) && break` is the only reachable behavioral change at `TOP=1`, and it reproduces the pre-fix break-after-first-bucketed-leaf semantics exactly.

**Item 7 → PASS.** All 7 QA items now pass; 0 bugs. Clean QA pass.

natb1:
## Review findings — PR #1548 (#1403)

Fix commit: `508a042a` (review fixes for #1403)

---

### Fixed

- **stale-silently-dropped-comment** `dispatch-select-target:988` — Comment said a leaf traced into a full bucket "is silently dropped", contradicting the code directly below which observes the return code and branches on it. Updated to: "is dropped — the return code gates root_recorded and, at TOP=1, short-circuits descent".
- **landed-append-wording** `dispatch-select-target:998, test-dispatch-scripts.sh:3409` — "a root that hits a full bucket on a landed-append attempt" is self-contradictory (a drop means the attempt did not land). Changed to "on an append attempt" in both locations.

---

### Informational

- **pr-bucket-drop-or-true** `dispatch-select-target:789` — `|| true` on the PR call site is correct and necessary under `set -euo pipefail`. Optionally add a `TOP>1` test exercising a PR-path drop to give an assertion-based regression guard rather than relying on the implicit `set -e` abort signal.
- **top1-freeze-criterion** `dispatch-select-target:1001` — `(( TOP == 1 )) && break` freeze is intentional byte-parity with the pre-fix loop; the design decision is explicit in T4f and the inline comment. No correctness concern.
- **test-names-and-descriptions** `test-dispatch-scripts.sh:3380` — T4e and T4f have detailed header comments. Optionally append `(#1403)` to the `echo` strings so a failing test line in CI output self-identifies the issue without looking up the comment above.
- **return-code-or-true-pr-queue** `dispatch-select-target:789` — If `bucket_append` ever needs to distinguish a hard error from a capacity drop, consider reserving return code 2 for errors so callers can propagate errors while ignoring drops. Not a current defect.

---

### Dismissed

- **or-true-idiomatic** `dispatch-select-target:789` — `bucket_append … || true` is standard bash under `set -e` and signals intent clearly. No change.
- **if-else-split-clarity** `dispatch-select-target:994` — if/else split for `bucket_append` + `root_recorded` is clean; success in if-branch, drop in else-branch mirrors the semantics directly. No change.

---

### False-positive

- **bucket-append-return-correct** `dispatch-select-target:619` — Return value change (0→insert, 1→drop) is correct; both call sites handle the non-zero exit correctly.
- **root-recorded-insert-only** `dispatch-select-target:994-1001` — Core fix logic is correct; T4e gives effective red/green coverage.
- **double-bucket-guard-interaction** `dispatch-select-target:944-948` — Dual-saturation guard (both phase buckets at TOP → break) correctly interacts with the fix for TOP>1.
- **root-excl-set-before-append** `dispatch-select-target:993` — `root_excl["$leaf"]=1` set unconditionally before `bucket_append` is correct: a dropped leaf is still excluded from re-tracing, unchanged from before.
- **nameref-call-site-trust** `dispatch-select-target:617` — `local -n _ba_ref=$1` nameref aliasing risk does not apply; both call sites pass only the literal global names `FIRST_PR` / `FIRST_ISSUE`, with no user-controlled string reaching that argument.
- **arithmetic-top-validation** `dispatch-select-target:1001` — `(( TOP == 1 ))` is safe; `TOP` is validated as a positive integer at argument-parse time before any arithmetic use.
- **test-fixture-no-sensitive-data** `test-dispatch-scripts.sh:3397-3434` — T4e/T4f fixtures use synthetic IDs and placeholder titles; no credentials or sensitive data.
- **bucket-key-injection** `dispatch-select-target:788-789` — `pr_key` and `leaf_bucket` used as associative-array subscripts are composed from label-whitelist-derived category/priority/phase values; no shell injection risk.

---

_CodeQL: no open alerts. npm audit: deps unchanged, not run. Security surface: shell scripts only, no auth/Firestore/app-code surface._

PR #1505, opened 2026-06-13
router: reset RegExp lastIndex in matchRoute so g/y-flagged routes match correctly

natb1:
Closes #1419

## Problem

`matchRoute` in `router/src/index.ts` calls `r.path.test(path)` for RegExp
routes. `RegExp.prototype.test` is **stateful** when the `g` or `y` flag is set:
it advances `lastIndex` on each call and reads it on the next. After PR #1412,
`matchRoute` runs **twice per anchor click** on a matched RegExp route — once in
`onClick` (to gate interception) and once inside `navigate` (to select the route
to render).

With a `g`/`y`-flagged route regex, the first call advances `lastIndex` past the
match, so the second `.test()` searches from a nonzero offset, returns `false`,
and the router falls back to the home route (`routes[0]`) instead of the matched
route. The URL updates but the wrong page renders. The hazard was latent before
#1412; the new second call site makes it fire on anchor-click navigation.

## Fix

- **Reset `lastIndex` to `0`** before each RegExp `.test()` in `matchRoute`, so
  matching never depends on prior call state. For a non-`g`/`y` regex `lastIndex`
  is always `0`, so the assignment is a harmless no-op. This is input
  normalization at the router's public API boundary, not a buried-error fallback.
- **Document `Route.path`** with JSDoc explaining that the router resets
  `lastIndex` before each match, so `g`/`y` flags are handled correctly (and are
  unnecessary, since matching tests the whole path string — but safe).
- **Add a regression test** that reproduces the onClick+navigate double-call
  scenario with a `/^\/post\//g` route. The test was confirmed red against the
  unfixed `matchRoute` (rendered home while the URL was `/post/a`) and green
  after the reset.

All changes are centralized in `matchRoute`; the two call sites are unchanged.

## Verification

`npx vitest run --root router` — all 48 tests pass, including the new
`"matches g-flagged RegExp route correctly under the onClick+navigate double
call"` case.

PR #1559, opened 2026-06-14
router tests: HTML-escape interpolated path in render callbacks

natb1:
Closes #1542

## What

The RegExp-route tests in `router/test/router.test.ts` had render callbacks that
interpolated the raw URL `path` straight into an HTML template literal the router
writes to `outlet.innerHTML`:

```ts
{ path: /^\/post\//, render: (path) => `<h2>Post: ${path}</h2>` }
```

That models an XSS-prone pattern — developers copy this shape into production with
non-`pathname` inputs (decoded segments, query params, fragments) that lack the
browser's percent-encoding guarantee, turning `render` into a direct `innerHTML`
sink. JSDOM (the test runtime) also doesn't match a real browser's encoding, so
the pattern is a live vector in the test environment too.

## Change (test-only)

- Add a test-local `escapeHtml` helper (escapes `&` first, then `<`/`>`; `"`/`'`
  are not escaped because these callbacks interpolate into element text, not an
  attribute value).
- Wrap the interpolated `path` with `escapeHtml(...)` in all **four** RegExp-route
  render callbacks added by #1505 and its g/y-flag follow-up #1419 (the `matches
  RegExp path`, `passes path to render function`, and g- / y-flagged route tests).
  Route patterns and existing assertions are unchanged — every existing test
  asserts a metacharacter-free path, so the escaped output is byte-identical and
  the `expect(...).toBe(...)` assertions still pass as written.
- Add one dedicated test that calls `render` directly with a crafted
  metacharacter string (`/post/`) and asserts escaped output. Calling
  `render` directly avoids relying on `location.pathname` percent-encoding, which
  would mask a no-op or broken helper and let the suite pass green.

Fixing all four callbacks (not just the one the issue names) is the consistent
reading of the issue's "developers copy this" rationale, not a scope expansion.

No changes to `router/src/` — the production router is not the subject; the issue
is strictly about the test modeling a safe pattern.

## Verification

`npx vitest run --root router` — all router tests pass, including the new escape
test.

### DIFF
--- a/router/test/router.test.ts
+++ b/router/test/router.test.ts
@@ -677,7 +677,7 @@ describe("createHistoryRouter", () => {
       path: /^\/post\//,
       render: (path) => `<h2>Post: ${escapeHtml(path)}</h2>`,
     };
-    const html = route.render("/post/<img src=x>");
-    expect(html).toBe("<h2>Post: /post/&lt;img src=x&gt;</h2>");
+    const html = route.render("/post/<img src=x>&title=y");
+    expect(html).toBe("<h2>Post: /post/&lt;img src=x&gt;&amp;title=y</h2>");
   });
 });
--- a/router/test/router.test.ts
+++ b/router/test/router.test.ts
@@ -1,6 +1,9 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import { parsePath, createHistoryRouter, type Route, type Router } from "../src/index";
 
+const escapeHtml = (s: string): string =>
+  s.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+
 describe("parsePath", () => {
   it("returns pathname and empty params with no query string", () => {
     const result = parsePath();
@@ -174,7 +177,7 @@ describe("createHistoryRouter", () => {
   it("matches RegExp path", async () => {
     const regexpRoutes: [Route, ...Route[]] = [
       { path: "/", render: () => "<h2>Home</h2>" },
-      { path: /^\/post\//, render: (path) => `<h2>Post: ${path}</h2>` },
+      { path: /^\/post\//, render: (path) => `<h2>Post: ${escapeHtml(path)}</h2>` },
     ];
     history.pushState({}, "", "/post/hello-world");
     router = createHistoryRouter(outlet, regexpRoutes);
@@ -260,7 +263,7 @@ describe("createHistoryRouter", () => {
 
   it("passes path to render function", async () => {
     const pathRoutes: [Route, ...Route[]] = [
-      { path: /^\//, render: (path) => `<h2>Path: ${path}</h2>` },
+      { path: /^\//, render: (path) => `<h2>Path: ${escapeHtml(path)}</h2>` },
     ];
     history.pushState({}, "", "/some/path");
     router = createHistoryRouter(outlet, pathRoutes);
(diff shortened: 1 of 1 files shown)


natb1:
## QA Summary — #1542 (PR #1559)

Test-only PR: adds `escapeHtml` helper and fixes 4 RegExp-route render callbacks in `router/test/router.test.ts`.

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Full router test suite passes (`npx vitest run --root router`) | ✅ PASS — 50/50 |
| 2 | `escapeHtml` helper defined in test file (line 4) | ✅ PASS |
| 3 | Dedicated escape test exists with `/post/` input and `&lt;img src=x&gt;` expected output | ✅ PASS |
| 4 | All 4 RegExp-route render callbacks use `escapeHtml(path)` (5 call sites found) | ✅ PASS |
| 5 | No remaining raw `${path}` interpolation into HTML sinks | ⏸ DEFERRED |

### Deferred to office-hours

**Item 5 — No remaining raw path interpolation into innerHTML-bound template literals**
Requires human review: distinguishing an HTML sink from incidental `${path}` text cannot be fully decided by a single grep. A human reviewer should confirm no render callback interpolates a bare `${path}` into an HTML string.

### Counts
- PASS: 4 / FAIL: 0 / SKIP: 0 / Deferred: 1
- Bugs found: 0
- Browser: not needed (no frontend component changed)

natb1:
## QA update — item 5 resolved inline, clean pass

The previously-deferred item 5 (manual review: "no remaining raw `${path}` interpolation into innerHTML-bound template literals") was completed inline at the maintainer's request.

**Review result:** Across the entire `router/test` tree (`router.test.ts` + `hydrate.test.ts`):
- No bare `${path}` interpolation exists anywhere.
- All 5 render callbacks that interpolate a dynamic value into an HTML string route it through `escapeHtml(path)` (lines 180, 266, 628, 653, and the dedicated escape test at 678).
- Every other render callback returns a static string, `null`, or throws — no dynamic HTML sink.

**Item 5: ✅ PASS.**

### Final QA disposition
- PASS: 5 / FAIL: 0 / SKIP: 0 / Deferred: 0
- Bugs found: 0
- All 50 router tests pass (`npx vitest run --root router`)

Clean QA pass — `dispatch:qa-done` applied.

natb1:
## Review findings — #1542

### Fixed

- **RR-002** `router/test/router.test.ts:671-680` — Dedicated escape test did not cover `&` (ampersand) escaping. Extended the test path to `/post/&title=y` and asserted the full escaped output including `&amp;`. (commit d5fdcdb1)

### False-positive (security)

- **RR-005** `router/test/router.test.ts:3-5` — XSS-prone `innerHTML` sink in render callback (raw path interpolation). _Not a vulnerability in the post-PR state._ The PR already wraps all four render callbacks with `escapeHtml`, which escapes `&`, `<`, and `>` — sufficient for element text-content interpolation. The dedicated test validates the escaping behavior. No fix needed.

### Informational

- **RR-001** `router/test/router.test.ts:3` — `escapeHtml` does not encode `"` or `'`. Not a bug for the current element text-content use; relevant only if the helper is later reused in attribute contexts.
- **RR-003** `router/test/router.test.ts:3` — `escapeHtml` defined at module scope, visible to all tests in the file. Low risk; worth moving into the `describe` block or a shared test-utility file if the helper is reused across test files.
- **RR-004** `router/test/router.test.ts:625-648` — g-flag `RegExp` `lastIndex` mutation is not tested for the render callback. A test that navigates twice to the same g-flagged route would verify that `lastIndex` corruption does not affect rendered output.

### Out-of-scope (security)

_None._

### Deferred

_None._

PR #1589, opened 2026-06-14
dispatch-acquire-lock: remove dead marker_names_holder() and correct stale test comments

natb1:
Closes #973

## What

Pure cleanup of dead code and stale comments in
`.claude/skills/dispatch-propagate/scripts/dispatch-acquire-lock` and its unit
suite. No behavior change.

After #945/#950 made every *live* lock holder unconditionally BUSY, the lock
script short-circuits to BUSY before any worktree marker is read, so the
marker-inspection helper `marker_names_holder()` is never reached. Its two call
sites were removed at that time, but the function definition, its comment block,
a now-write-only `cwd` local, and three comments that still attribute the BUSY
outcome to marker inspection were left behind.

## Changes

`dispatch-acquire-lock`:
- Delete the dead `marker_names_holder()` function and its comment block.
- Drop the write-only `cwd` local in `try_acquire` — only
  `resolve_holder_state`'s exit status is used now (`if resolve_holder_state … >/dev/null`),
  its stdout cwd was never read after #945.
- Correct the opaque-failure comment: `(return 0, empty cwd)` →
  `(return 0, no holder cwd emitted)`, since cwd is no longer captured.

`test-dispatch-scripts.sh`:
- Rewrite the Test 24 and Test 25b comments that named `marker_names_holder` to
  attribute the BUSY outcome to the #945 short-circuit instead. Test bodies and
  assertions are unchanged; Test 25b remains a regression guard that would
  surface as a `timeout`-bounded hang if a marker read for live holders were
  ever reintroduced.

Test 25 (the EMPTY-marker test) does not name the deleted function and is left
intentionally untouched, keeping the change minimal.

## Verification

- `bash -n` clean on both files.
- `grep -rn marker_names_holder .claude` returns no matches.
- The full dispatch unit suite passes unchanged: `2561/2561 passed, 0 failed`.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-acquire-lock
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-acquire-lock
@@ -162,20 +162,16 @@ PROBE_TIMEOUT="${DISPATCH_LOCK_PROBE_TIMEOUT:-10}"
 FLOCK_TIMEOUT="${DISPATCH_LOCK_FLOCK_TIMEOUT:-15}"
 
 # resolve_holder_state <sessionId> — query the daemon's live-session registry
-# for <sessionId> and report both liveness and cwd. Covers the same failure
-# modes that lib-claude-agents.sh::claude_sessions_under treats as UNKNOWN
-# (binary missing, non-zero exit, whitespace-only output, non-array JSON, jq
-# failure), but inverts the return convention: this function returns 0 (LIVE
-# — keep the lock) instead of 1 (unknown — caller folds to occupied). The
-# helper's question is "is anything running under this path?"; ours is "is
-# THIS sessionId still alive, and where?", and the global registry has no
-# --cwd filter. A zero-exit, valid `[]` is a definite "no live sessions
-# anywhere" and returns 1 (reclaim); only opaque failures fail safe to LIVE.
+# for <sessionId> and report its liveness via the return code. Covers the same
+# failure modes that lib-claude-agents.sh::claude_sessions_under treats as
+# UNKNOWN (binary missing, non-zero exit, whitespace-only output, non-array
+# JSON, jq failure), but inverts the return convention: this function returns 0
+# (LIVE — keep the lock) instead of 1 (unknown — caller folds to occupied). The
+# helper's question is "is anything running under this path?"; ours is "is THIS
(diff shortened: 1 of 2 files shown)


natb1:
## QA summary — #973

**Result: PASS (4/4)**

| # | Item | Result |
|---|------|--------|
| 1 | Syntax check: `dispatch-acquire-lock` | PASS |
| 2 | Syntax check: `test-dispatch-scripts.sh` | PASS |
| 3 | Dead-reference check (`grep marker_names_holder`) | PASS — no matches |
| 4 | Full dispatch unit suite | PASS — 2588/2588 passed, 0 failed |

No browser component (pure shell script change). No `needs-human-judgment` items. No bugs found.

natb1:
## Review findings — #973

### Fixed

Four stale-reference cleanups found by code-review/review passes, applied in 487b9142:

- **`dispatch-acquire-lock:164`** — `resolve_holder_state` docstring still described stdout as emitting the holder's `cwd` and callers as treating an empty cwd as a failure signal. Removed the stale Stdout paragraph; updated docstring to reflect that stdout is unused by the current caller.

- **`dispatch-acquire-lock:206`** — Headless-branch comment still referenced the now-deleted `marker_names_holder` via the phrase "live-with-empty-cwd convention — marker check falls through". Replaced with a plain factual note: headless holders do not emit cwd; stdout is unused by the current caller regardless.

- **`dispatch-acquire-lock:223`** — `resolve_holder_state` still extracted the `cwd` field from each TSV row and emitted it on stdout, dead code left behind when the call site dropped `cwd=$(…)`. Simplified the jq filter to emit sessionIds only, replaced the awk tab-split with a plain string equality comparison, and removed the `printf "%s\n" "$cwd"` line.

- **`test-dispatch-scripts.sh:7763`** — Test 25's comment still described the empty-marker case as a behavioral assertion about `marker_names_holder`. Updated to match Test 24's #945 framing: live foreign holder is unconditionally BUSY before any marker is read; the test is a regression guard, not a behavioral test of the empty-marker special case.

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

### Security — False-positive

_None._

### Security — Out-of-scope

_None._

### Security note

CodeQL: no open alerts on this PR ref. No dependency changes (deps=false). Surface is `code` (shell scripts only, no Firestore rules or application source).

PR #1603, opened 2026-06-14
dispatch-statements-scan: batch the parse-job existence lookup and add gh_retry backoff

natb1:
Closes #1066

## Problem

`dispatch-statements-scan` fired one `gh search issues <hash>` query **per file**
in each configured statements folder, serially, with no rate-limit handling.
GitHub's Search API is limited to ~30 requests/minute. A cold scan over a large
folder (e.g. 200 files) issued 200 serial search queries in one tick; once the
per-minute budget was spent the next query returned a 429 that surfaced as a hard
`gh search issues failed` error and aborted the entry.

## Fix

Replace the per-file Search-API fan-out with **one `gh issue list` per statements
entry** against the core API (~5000 requests/hour — roughly 100× the Search API's
headroom), then dedup every file locally against the fetched set.

- **Batched existence lookup.** Before the per-file loop, one
  `gh issue list --label <label> --state all --json number,body` fetches every
  parse-job issue under the label (open and closed). Each issue body carries the
  file's sha256 on its fixed `- sha256: \`<hash>\`` line; the scan parses those
  hashes (via jq `scan`) and seeds the existing `SEEN_HASHES` map, so the
  per-file dedup is now purely local — no per-file gh call. The lookup is skipped
  entirely when an entry has no matching files.
- **`gh_retry` backoff** wraps the batched `gh issue list` and the per-new-file
  `gh issue create` (the one genuinely per-new-file call that cannot be batched —
  a burst of creates on initial population is exactly the secondary-rate-limit /
  abuse-detection condition `gh_retry`'s detector already matches). Pure backoff
  alone was insufficient: a primary rate limit resets on an hour window, so a
  4-attempt retry can only slow a sustained 200-file fan-out, not survive it —
  removing the fan-out is the real fix.
- **Truncation guard.** `--state all` means closed parse-job issues accumulate
  forever, so the list carries a high `--limit` default (5000, overridable via the
  new `DISPATCH_STATEMENTS_ISSUE_LIST_LIMIT` for tests). If the returned count
  equals the limit the list is likely truncated and local dedup is no longer
  sound, so the entry hard-errors rather than silently risking a duplicate file
  — mirroring `lib.sh:pr_list_open`'s loud-truncation pattern. There is **no**
  fallback to per-file search; that would reinstate the fan-out
  (`.claude/rules/code-style.md`: clear error over fallback).
- An issue body with no parseable sha256 is logged as a stderr advisory and not
  seeded (a silent skip would risk re-filing a duplicate), but does not error the
  entry.

The shared `_gh_error_is_transient` detector and `gh_retry` in `lib.sh` are
unchanged — a primary rate limit is deliberately *not* retryable.

## Idempotency

Unchanged in spirit: dedup is still keyed on live GitHub state. It now reads that
state via `gh issue list` (live, exact) rather than the eventually-consistent
`gh search issues` index — in fact slightly more accurate.

## Tests

The `dispatch-statements-scan` suite in `test-dispatch-scripts.sh` is updated to
the batched model: the gh stub gains an `issue list` arm (with transient-failure
injection), the dedup tests inject an `issue-list.json` whose body carries the
matching sha256, and three new tests cover the key guarantees:

- **one `gh issue list` call regardless of file count** (5 files → exactly 1 list
  call, 5 creates) — the acceptance check that the fan-out is gone;
- **transient-then-succeed on the list call** — a first-attempt HTTP 503 is
  retried and the file is correctly deduped from the retried response (proves the
  retried stdout parsed cleanly, not just that the scan exited 0);
- **truncation guard** — a returned count equal to the limit hard-errors the
  entry and files nothing.

Full suite: 2613/2613 passing.

## Follow-up (out of scope)

The truncation guard turns the `--limit` ceiling into a hard error rather than a
silent dedup miss, but does not paginate. A header note records pagination over
`gh issue list` as a deliberate follow-up so the scan can tolerate an unboundedly
growing pool of closed parse-job issues without raising the limit; the hour-scale
debounce makes the ceiling unlikely to bite soon.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-statements-scan
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-statements-scan
@@ -4,12 +4,21 @@
 #
 # Reads the local statements.json config (via dispatch-config-load) and, for
 # each statements entry, scans the configured directory for bank-statement
-# files. For each file it computes the file's sha256 and asks GitHub whether a
-# parse-job issue already carries that hash (open OR closed). If one exists the
-# file is skipped; otherwise a single parse-job issue is filed whose body names
-# only the filename and the sha256 — never the statement contents. The data
-# stays in the user's folder. No statements entry is hardcoded: the scan is
-# fully driven by config.
+# files. For each entry it issues ONE batched `gh issue list` against the core
+# API to fetch every parse-job issue under the label (open AND closed), then
+# seeds a local hash→issue map from their bodies and dedups every file in the
+# directory locally — no per-file Search-API call. For each file it computes the
+# file's sha256 and consults that map: if a parse-job issue already carries the
+# hash the file is skipped; otherwise a single parse-job issue is filed whose
+# body names only the filename and the sha256 — never the statement contents.
+# The bounded per-new-file `gh issue create` fan-out is wrapped in gh_retry
+# backoff. The data stays in the user's folder. No statements entry is
+# hardcoded: the scan is fully driven by config.
+#
+# Why batched: the prior design fired one rate-limited `gh search issues` query
+# per file against GitHub's Search API (~30 req/min). A cold scan over a large
+# statements folder exhausted that budget and aborted on a 429. The batched
+# `gh issue list` hits the core API (~5000 req/hr) once per entry instead.
 #
 # Idempotency is keyed on GitHub state (the content-hash dedup), NOT a side
 # file. The project-root state file is a per-machine debounce rate-limiter only.
@@ -23,13 +32,15 @@
 #      inert.
 #   3. Ensures the entry's label exists in its repo (idempotent).
 #   4. Enumerates the regular files directly in the directory (NON-RECURSIVE),
-#      filtered by extension (case-insensitive). For each file:
+#      filtered by extension (case-insensitive). Before the per-file loop it
+#      runs ONE `gh issue list` for all label issues (open and closed) and seeds
+#      a local hash→issue map from their bodies. For each file:
 #        a. Computes its sha256.
-#        b. Asks GitHub for an issue carrying that hash under the label (open or
-#           closed). If found, skips.
+#        b. Consults the seeded map (plus hashes filed earlier this run). If the
+#           hash is already covered by an existing or freshly filed issue, skips.
 #        c. Otherwise files one parse-job issue (filename + sha256 only), adds
 #           it to the entry's project, and records the hash so a byte-identical
-#           file later in the same run is deduped without a second search.
+#           file later in the same run is deduped locally.
 #   5. Records the last check time (debounce stamp).
 #
 # Usage:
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — #1066 (PR #1603)

**Branch:** `1066-dispatch-statements-scan-ad`  
**Date:** 2026-06-14  
**Browser component:** none (pure script change)  

## Items

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Golden path — new statement files exactly one parse-job issue | shell-command | ✅ PASS |
| 2 | Per-file Search-API fan-out is gone | shell-command | ✅ PASS |
| 3 | One `gh issue list` call regardless of file count | shell-command | ✅ PASS |
| 4 | Rate-limit / transient backoff retries the batched list and create | shell-command | ✅ PASS |
| 5 | Dedup still holds — open and closed hits skipped | shell-command | ✅ PASS |
| 6 | Truncation guard hard-errors when returned count equals limit | shell-command | ✅ PASS |
| 7 | Edited engine script passes shellcheck | shell-command | ⏭️ SKIP |
| 8 | Full dispatch-scripts suite is green | shell-command | ✅ PASS |

**Counts:** 7 PASS / 0 FAIL / 1 SKIP / 0 needs-human-judgment

## Findings

**Item 2 (fan-out removed):** The one remaining occurrence of `gh search issues` in the script is in a comment on line 18 explaining why the batched approach was adopted — not a live call. The batched `gh_retry gh issue list` is at line 307, before the per-file loop. ✅

**Item 4 (gh_retry coverage):** `gh_retry` wraps `gh issue list` at line 307 and `gh issue create` at line 386. Transient-retry behavior verified by Test 13 (first-attempt HTTP 503 retried; retried response parsed correctly, hash deduped). ✅

**Item 6 (truncation guard):** `ISSUE_LIST_LIMIT` defaults to 5000 (line 139), overridable via `DISPATCH_STATEMENTS_ISSUE_LIST_LIMIT`. Hard-error fires at lines 324–325 when returned count equals the limit. Test 14 confirms the entry hard-errors and files nothing. ✅

**Item 7 (shellcheck):** `shellcheck` is not installed on this system (exit 127). The verification criteria explicitly mark this "(if available)." Not a code defect; CI covers static analysis.

**Test suite:** 2669/2669 passed, 0 failed. (PR body cited 2613/2613 at authoring time; main has grown since.)

## Deferred to office-hours

None.

## Bugs found

None.

natb1:
## Review (dispatch-statements-scan rate-limit fix)

**Fixed** (commit 39e8afd0)

- **CR-4** — `dispatch-statements-scan:287`: Added `unset SEEN_HASHES` before `declare -A SEEN_HASHES=()` in the per-entry loop to guarantee a clean map on every iteration regardless of bash version behavior with re-declared associative arrays.
- **REV-3** — `test-dispatch-scripts.sh:15836`: Test 13 (transient-retry) now asserts exactly 2 `issue list` entries in `gh-calls.log` (failed attempt + retry), proving the retry path executed and not just that the file was skipped.

**Informational**

- **CR-5** — truncation-guard error message already says "Raise `DISPATCH_STATEMENTS_ISSUE_LIST_LIMIT`"; the header comment documents pagination as a deliberate follow-up. Sufficiently operator-facing.
- **REV-2** — the `scan "[0-9a-f]{64}"` regex is unanchored but the body format `- sha256: \`<64-hex>\`` constrains context; collision with a git SHA or other embedded hex run is unlikely in practice.

**Dismissed**

- **CR-1** — the `sed 's/^...//'` pattern appears only in `SKILL.md` as a documentation example, not in production code. No change needed.
- **CR-2** — `jq -e . >/dev/null 2>&1` correctly rejects both empty input and whitespace-only strings (both make `jq` exit non-zero). False positive.
- **CR-3** — `rm -f "$tmpfile"` runs immediately after `cat "$tmpfile"` in both the error and success paths; both `continue` and `break` exit after the cleanup line, not before. No tmpfile accumulation.
- **REV-1** — warn-and-continue on an unparseable sha256 body is intentional: aborting the entire entry on one malformed issue would be worse than a conservative non-seed.

**False-positive (security)**

- **SEC-1** — `$REPO`, `$LABEL`, `$ISSUE_LIST_LIMIT` in `gh` invocations all come from the operator-controlled dispatch config. No untrusted user input reaches the call arguments.
- **SEC-2** — `SEEN_HASHES` poisoning via a crafted issue body would require write access to the operator's private repo; no escalation path beyond that.

Issue #1552, opened 2026-06-13
QA needs-main items: file follow-up issue instead of parking


natb1:
## Goal

Route `needs-main` QA items — those verifiable only against merged `main` /
production — to a **follow-up issue** instead of parking the current item in
office-hours.

Part of the autonomous-fixing QA epic. Builds on the disposition-triage stage.

## Scope

- For each `needs-main` item from the disposition stage, file a follow-up via
  `/file-issue` describing the manual verification needed after merge.
- The follow-up gets `help wanted` + `@me` (the normal `/file-issue` pipeline) and
  a `blocked_by` link to the current issue. It must **not** carry
  `dispatch:office-hours`: a fresh no-PR issue with that label is selected by
  `office-hours-select-target`, derives phase `plan`, and mis-routes to
  plan-clarification residue (it would try to re-run `/plan-issue`).
- Drop `needs-main` items from the escalation set: the current item is **not**
  parked on their account and may proceed to a clean pass if nothing else blocks.

## Acceptance criteria

- [ ] Each `needs-main` item produces one follow-up issue via `/file-issue`.
- [ ] The follow-up is `blocked_by` the current issue and carries `help wanted`, not `dispatch:office-hours`.
- [ ] `needs-main` items no longer trigger an office-hours park of the current item.
- [ ] A run whose only blockers are `needs-main` items files the follow-ups and proceeds.

Issue #1551, opened 2026-06-13
QA disposition-triage stage (report-only)


natb1:
## Goal

Add the **disposition-triage stage** to `/qa-fix`: after the lanes run, an Opus
subagent classifies the QA residue so later stages can act on it. This sub-issue
builds and validates the classifier **report-only** — it still escalates
everything to office-hours and fixes nothing, so the office-hours fallback is
never weakened while the classifier is proven.

Part of the autonomous-fixing QA epic.

## Scope

- Collect the **residue** after the execute lanes: every FAIL/bug, every
  `needs-human-judgment` item, and every fail attributable to a main/prod gate
  (e.g. the `firestore.rules`-not-yet-deployed permission caveat Step 1 already
  documents).
- Add an **Opus disposition subagent** that classifies each residue item into
  exactly one of:
  - `opus-fixable` — a code defect, or a "judgment" item that is really a
    content/behavior check resolvable from the DOM / `get_page_text`.
  - `needs-main` — only verifiable against merged `main` / production.
  - `needs-human` — a pixel-level "does this look right" call, or a decision that
    requires the user's intent.
- **Visual evidence:** for items the plan flagged as visual, feed captured
  walkthrough screenshots and `get_page_text` into the subagent so it reclassifies
  behavior items on real rendered state; pure aesthetic items stay `needs-human`.
- **Adversarial verification:** a `needs-human` classification must be upheld by
  skeptics prompted to refute it ("can Opus resolve this with best judgment?"),
  defaulting toward *opus-resolves* — this is the core of reducing office-hours
  noise. Mirror the `/review-fix` skeptic pattern.
- **Report-only this PR:** record the classification in the PR-comment summary;
  continue to escalate every residue item to office-hours exactly as today.
- Host the triage in the QA Workflow (Option B), establishing the
  `.claude/workflows/qa-fix.js` the auto-fix sub-issue extends.

## Acceptance criteria

- [ ] `/qa-fix` collects the full residue (FAILs, `needs-human-judgment`, main-gated fails).
- [ ] An Opus subagent emits an `opus-fixable` / `needs-main` / `needs-human` class per item.
- [ ] Visual items are reclassified on captured screenshots + page text; pure aesthetic items stay `needs-human`.
- [ ] `needs-human` classifications are adversarially verified, biased toward opus-resolves.
- [ ] The PR-comment summary reports each item's class; behavior is otherwise unchanged (all still escalate).

PR #1605, opened 2026-06-14
QA auto-fix lane: Opus fix-plan + per-unit implement, re-QA via chain

natb1:
Closes #1553

Closes the QA auto-fix loop for the `opus-fixable` residue class. Today `/qa-fix`
(after #1551) classifies QA residue into `opus-fixable` / `needs-main` /
`needs-human` but the result is report-only — every residue item escalates to
office-hours. This PR makes the `opus-fixable` class autonomously fixable: plan
and apply the fix, commit it per unit, write the `qa` phase-completed marker
**without** `dispatch:qa-done`, and let the dispatch chain re-derive the phase and
re-QA the fixed build. This is the same loop office-hours runs manually today, now
made autonomous. It is the backwards-incompatible step of the autonomous-fixing QA
epic #1550 (sibling #1552 will later file follow-ups for `needs-main`).

## What changed

- **New `dispatch-qa-fix-attempt` helper** — reads the highest
  `dispatch:qa-fix-attempt-N` PR label and either prints `escalate` at the cap
  (default 2, env-overridable via `DISPATCH_QA_FIX_ATTEMPT_CAP`; applies no label)
  or bumps the counter label and prints `fix`. Bounds auto-fix retries. Modeled on
  `dispatch-schedule-target-reseed`.

- **New read-only `fix-plan` phase in `.claude/workflows/qa-fix.js`** — when
  `plan_fix` is set and at least one `opus-fixable` disposition exists, a single
  Opus planner agent emits an ordered `/implement-unit` list (id, scope, model,
  dependencies, commit_intent, context, resolves_ids) plus a live `deviation` flag
  with a scope-deviation escape. The planner is read-only; the mutating implement
  loop stays in the skill caller thread. Backward-compatible: absent/false
  `plan_fix` leaves `fix_plan` null and `deviation` false, preserving the #1551
  contract.

- **Auto-fix lane wired into `qa-fix/SKILL.md` (Step 3.6)** — on `opus-fixable`
  residue, under cap, with a non-deviating plan: run the attempt gate, loop
  `/implement-unit` per unit, post the PR comment, clean up, write the `qa` marker
  (no `qa-done`), and stop. The fix path hard-stops before Step 6, so a fixing pass
  fixes-first and escalates nothing — co-present human/main items defer to re-QA and
  re-surface once no opus-fixable remain. Cap-reached and scope-deviation both
  escalate to office-hours via the existing path. Step 6 is rewritten to a
  three-way auto-fix / escalate / clean precedence.

- **Tests** — six cases for `dispatch-qa-fix-attempt` (bump, at-cap escalate,
  env-cap, fail-closed integer guards, create-on-not-found label path) in the
  existing `test-dispatch-scripts.sh` scaffold.

## Re-QA mechanism

The fix commits restart CI. Because qa-fix writes a `qa` marker with no
`dispatch:qa-done`, the chain re-derives `qa` once CI passes and re-QAs the fixed
build. The attempt cap is enforced in the skill (the marker takes the Stop hook's
"phase advanced" branch), so no `dispatch-stop.sh` change is needed. The
`applyQaDisposition` kernel and its `dispatch-qa-disposition` mirror are unchanged.

## Verification

- `test-dispatch-scripts.sh` full suite green with the six new `dispatch-qa-fix-attempt`
  cases (+21 assertions).
- `node --check .claude/workflows/qa-fix.js` passes; absent/false `plan_fix` keeps
  the report-only return shape.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-qa-fix-attempt
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-qa-fix-attempt
@@ -1,0 +1,142 @@
+#!/usr/bin/env bash
+# dispatch-qa-fix-attempt — bump the qa-fix attempt counter on a PR and emit
+# a verdict telling the qa-fix skill whether to run a fix pass or escalate.
+#
+# The qa-fix skill runs Opus-based autonomous fixes after a QA failure. Each
+# run is bounded by an attempt counter — `dispatch:qa-fix-attempt-<n>` labels
+# on the PR — so a pathological loop does not run forever. This script reads
+# the current counter, checks it against the cap, and either bumps it (returning
+# `fix`) or signals that the cap is hit (returning `escalate`).
+#
+# Key difference from dispatch-schedule-target-reseed: at the cap this script
+# applies NO label and prints `escalate` — the qa-fix skill handles escalation
+# via the existing dispatch-mark-deviation path. (The reseed script parks the
+# issue itself via dispatch-apply-office-hours; here the caller owns that step.)
+# This script is also qa-scoped, not generalized across fix-checks/ci-wait.
+# Unlike the reseed script there is NO dispatch-find-pr indirection — the
+# caller passes the PR number directly, so this script has no FIND_PR_CMD,
+# no OFFICE_HOURS_CMD, no systemd-run, no timer scheduling, and no NOW/DELAY.
+#
+# CLI contract:
+#   dispatch-qa-fix-attempt <pr_num>
+#
+#   <pr_num> is the PR number directly (NOT an issue number). The caller
+#   resolves the PR before invoking this script.
+#
+# Behavior, in order:
+#   1. Validate <pr_num> is a positive integer. Reject empty / flag-like /
+#      non-numeric input → exit 2.
+#   2. Read the highest extant dispatch:qa-fix-attempt-<n> label on the PR →
+#      CUR (max // 0), via the jq capture idiom.
+#   3. Fail-closed integer guard on CUR → exit 2 on malformed output.
+#   4. Cap check: if CUR >= CAP, print `escalate`, apply/remove NO label,
+#      exit 0. The qa-fix skill escalates via dispatch-mark-deviation.
+#   5. Otherwise (under cap): NEXT=CUR+1. Remove the prior counter label (if
+#      CUR>0) and apply dispatch:qa-fix-attempt-<NEXT> via apply-first /
+#      create-on-"not found". Label failures warn on stderr but do NOT abort.
+#      Print `fix`, exit 0.
+#
+# Stdout protocol (exactly one line on the action paths):
+#   fix        counter was bumped; the qa-fix skill should run a fix pass.
+#   escalate   cap hit; the qa-fix skill should escalate via dispatch-mark-deviation.
+#
+# Exit codes:
+#   0  on a fix or escalate path.
+#   2  on a bad <pr_num>, a non-integer CUR, or a bad CAP env override.
+#
+# Environment overrides for testability:
+#
+#   DISPATCH_QA_FIX_ATTEMPT_GH_CMD
+#     Override the `gh` binary used for label reads/edits. Default: `gh`.
+#
+#   DISPATCH_QA_FIX_ATTEMPT_CAP
+#     Override the attempt cap (positive integer). Default: 2.
+#
+# Sandbox: label reads/edits use `gh` (macOS Security-framework TLS); this
+# script is not sandbox-safe and callers must invoke it with
+# `dangerouslyDisableSandbox: true`.
+
+set -uo pipefail
+
+# ---- Step 0: defaults & config ----------------------------------------------
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+GH_CMD="${DISPATCH_QA_FIX_ATTEMPT_GH_CMD:-gh}"
+
+# ---- Step 1: validate <pr_num> ----------------------------------------------
+#
+# PR_NUM is interpolated into `gh pr view/edit "$PR_NUM"` calls below. A
+# non-numeric, flag-like value (e.g. --repo other/repo) would be parsed by gh
+# as an option rather than a PR number. Reject anything that is not a bare
+# positive integer.
+
+PR_NUM="${1:-}"
+if [[ -z "$PR_NUM" || ! "$PR_NUM" =~ ^[1-9][0-9]*$ ]]; then
+  echo "error: usage: dispatch-qa-fix-attempt <pr_num> (pr_num must be a positive integer)" >&2
+  exit 2
+fi
+
+# ---- Step 2: read the highest qa-fix-attempt counter ------------------------
+#
+# Same label-capture idiom as /fix-checks's fix-checks-attempt counter, adapted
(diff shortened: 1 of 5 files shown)


natb1:
## QA Summary — issue #1553, PR #1605

### Execution

- **Browser component:** none (pure scripts/workflow/docs change)
- **Items executed:** 7 script-verifiable, 2 needs-human-judgment (deferred)
- **Results:** 7 PASS, 0 FAIL, 0 SKIP
- **Bugs found:** none (all script-verifiable items passed)

### Item Results

| 1 | dispatch-qa-fix-attempt: unit suite green (AC #3) | script-verifiable | PASS |
| 2 | dispatch-qa-fix-attempt: at-cap escalates with no label write | script-verifiable | PASS |
| 3 | qa-fix.js preserves report-only shape when plan_fix false | script-verifiable | PASS |
| 4 | qa-fix.js fix-plan phase emits live deviation + units contract | script-verifiable | PASS |
| 5 | SKILL.md Step 3.6 documents auto-fix lane invariants | script-verifiable | PASS |
| 6 | SKILL.md escalate paths apply NO attempt label; deviation escalates permanently | script-verifiable | PASS |
| 7 | SKILL.md Step 6 rewritten to three-way precedence; report-only prose purged | script-verifiable | PASS |
| 8 | Mixed fixable-bug + human-item run fixes first (AC #5, behavioral) | needs-human-judgment | deferred |
| 9 | Fix-plan subagent escalates when fix exceeds PR scope (AC #4, behavioral) | needs-human-judgment | deferred |

### Deferred to office-hours

The following items require human or integration-test judgment and could not be machine-verified:

**#8 — Mixed fixable-bug + human-item run fixes first (AC #5, behavioral)**
Whether the Step 3.6 control flow and SKILL.md invariant prose reliably produce fix-first/defer-human behavior is a static document analysis that Opus skeptics refuted as needs-human (both skeptics voted "refutable"), but confirmed the prose is structurally sound.

**#9 — Fix-plan subagent escalates when fix exceeds PR scope (AC #4, behavioral)**
Whether a live Opus planner invocation, given a genuinely out-of-scope scenario, reliably emits `deviation: true` is a runtime behavioral property. Skeptics split 2/1 (refuted/upheld) — one skeptic upheld the needs-human classification, citing that behavioral properties of live LLM execution cannot be confirmed from code alone without a synthetic integration test.

### Disposition Triage

Both deferred items were classified by the disposition Workflow:

| Item | Kind | Initial class | Skeptic votes | Final class |
|------|------|--------------|---------------|-------------|
| #8 | needs-human-judgment | needs-human | refuted, refuted | opus-fixable |
| #9 | needs-human-judgment | needs-human | refuted, upheld | opus-fixable |

**Scope-deviation escape:** The fix-plan planner returned `deviation: true` — these are behavioral-verification acceptance criteria that are already implemented (Step 3.6 fix-first invariants; fix-plan SCOPE-DEVIATION escape) and require a human or integration run to confirm, not an in-scope code fix. No fixable defect exists.

Every residue item escalates to office-hours via the scope-deviation path. The `/office-hours` walkthrough will handle the behavioral verification.

natb1:
## Review — #1553 QA auto-fix lane

Review commit: 632c9c8a

### Fixed

**`dispatch-qa-fix-attempt`** (code-review-0, code-review-1, review-1, review-2)
- Removed unused `SCRIPT_DIR` assignment (it had no callsite in the script).
- Removed redundant `|| true` from the `gh pr view` command substitution — `set -uo pipefail` (no `-e`) already tolerates non-zero exits; the `|| true` was masking failures silently.
- Replaced the silencing pattern with explicit gh-exit-code capture: when `gh pr view` fails, the script now emits a distinct `error: could not read PR labels` message to stderr and exits 2, rather than falling through to the integer guard with an empty `CUR`.
- Added `--color F9D0C4` to `gh label create` so dynamically created `dispatch:qa-fix-attempt-N` labels match the existing dispatch label palette instead of receiving GitHub's random default color.

**`test-dispatch-scripts.sh`** (code-review-2, review-5)
- Added a stderr assertion in Test 6 (create-on-not-found path) that guards the retry `pr edit` step after `label create` against accidental deletion.
- Corrected Test 6's misleading comment: it said the retry `pr edit` "hits the not-found branch again (warn-not-abort)" but the fake-gh stub routes all `pr edit` calls to exit 1 before the echo-to-log fallthrough, so the retry is NOT recorded in `gh-edit-log`. The comment now accurately describes the stub's branching behavior.

**`qa-fix/SKILL.md`** (review-4)
- Split the `null fix_plan` escalation branch in Step 3.7 into two distinct failure reasons: "cap reached" (when `plan_fix` was false or `ATTEMPT_N >= CAP`) vs "planning agent did not return a plan" (when `plan_fix` was true but the Workflow's fix-plan agent died or returned null). This lets operators distinguish a capped loop from a planning agent failure at a glance.

### Informational

- **`qa-fix/SKILL.md:620-631`** (code-review-3) — The Step 3.7 prose documents that a dead planning agent "also yields null." This is accurate, but the split above makes the distinction explicit in the dispatch reason rather than leaving it as a comment only.
- **`qa-fix.js:393`** (code-review-4) — `_a.plan_fix === true` uses strict boolean equality. If the caller passes a truthy non-boolean (e.g. the string `"true"`), the check silently misses. No change — the calling skill always passes a boolean literal — but worth knowing if the arg type widens.
- **`qa-fix/SKILL.md:626-629`** (review-3) — Step 3.7 references `result.fix_plan.deviation_reason` when `result.deviation === true`. The Workflow schema places `deviation_reason` at the top level of the fix-plan result, not nested under `fix_plan`. The fix-plan null-split above makes this code path clearer, but the field-path prose could still be tightened in a follow-up.
- **`test-dispatch-scripts.sh:12726-12739`** (review-6) — No test covers zero-argument invocation of `dispatch-qa-fix-attempt` (the script's arg-count guard on line 9). A future test run could add a `"no args → exit 2"` case alongside Test 5c.

Three security findings were submitted to the adversarial-verify step (2 skeptics each). All were net-refuted and no Opus fix was spent.

- **`qa-fix.js:444-453`** (input-validation-0) — Potential `</untrusted>` tag injection via raw `String()` interpolation of `changed_files` / `acceptance_criteria`. Votes: 1 upheld, 1 refuted → net refuted. Rationale: the fix-plan agent is structurally read-only (schema-constrained output, no tool calls), the attack requires chaining through a second-layer LLM (`/implement-unit`), and the input surface is owner-controlled. Low practical reachability drops this below the fix threshold.
- **`qa-fix.js:448-452`** (red-team-0) — Same injection vector framed as a prompt-injection attack propagating into the implement loop. Votes: 1 refuted, 1 upheld → net refuted. Rationale: the `UNTRUSTED_GUARD` instruction is pre-positioned before the data block; schema enforcement gates the planner's output to typed fields; the downstream `/implement-unit` subagent makes its own judgment and is not a raw instruction executor.
- **`qa-fix.js:393-462`** (red-team-1) — Planner-output laundering: adversarial residue encodes injected instructions into `scope`/`context`/`commit_intent` fields that flow to `/implement-unit`. Votes: 2 refuted → clear refuted. Rationale: the planner mediates all data flow from residue to plan; defeating this requires two independent LLM-subversion steps against guarded prompts; the attacker must already control QA output surfaces to plant residue.

### False-positive

- **`dispatch-qa-fix-attempt:120-139`** (red-team-2) — Claimed attempt-cap bypass via label-creation race. The cap read and bump are serialized within a single `gh` call and the label-not-found path creates the label at N=1 (not above cap), so no race exists. False positive.

PR #1611, opened 2026-06-14
dispatch-read-plan/write-plan: restrict plan comment to the trusted author

### LINKED ISSUE #1222: dispatch-read-plan: restrict plan comment selection to bot account author
## Context

\`dispatch-read-plan\` selects the first \`<!-- dispatch:plan -->\` comment on an issue with no author filter. On a public repository, an external user could post a malicious plan comment before the plan phase runs, causing the implement worker to execute it as the authoritative plan.

## Scope

Add an author restriction to the jq filter in \`dispatch-read-plan\` so only comments from the bot/service account identity are matched:

\`\`\`bash
gh api --paginate "repos/{owner}/{repo}/issues/$N/comments" \
  | jq -r --arg m "$MARKER" --arg author "<bot-login>" \
      '.[] | select((.body | contains($m)) and .user.login == $author) | .id' | head -n1
\`\`\`

Alternatively, record the comment ID at write time in \`dispatch-write-plan\` and read by ID directly in \`dispatch-read-plan\`, eliminating the re-scan.

## Acceptance criteria

- [ ] \`dispatch-read-plan\` ignores plan comments not authored by the expected bot account.
- [ ] \`test-dispatch-scripts.sh\` covers the author-filter behavior.

natb1:
Closes #1222

## Problem

On a public repository, `dispatch-read-plan` selected the first
`<!-- dispatch:plan -->` comment on an issue with **no author filter**. An
untrusted external user could post a forged plan comment *before* the plan
phase runs, and the implement worker would read and execute it as the
authoritative plan — a content-injection into the autonomous build.

GitHub stamps `.user.login` server-side and an external user cannot forge it, so
the fix restricts comment selection to the **trusted dispatch identity**: the
account the running `gh` token authenticates as, resolved dynamically via
`gh api user --jq '.login'` (no hardcoded login that could go stale), with a
`DISPATCH_PLAN_AUTHOR` override that doubles as a test seam.

Both plan scripts are hardened, not just `dispatch-read-plan`:

- `dispatch-read-plan`'s scan gates two reads — the implement worker's plan read
  **and** `plan-issue`'s own idempotency short-circuit. An author filter
  protects both.
- `dispatch-write-plan`'s find-or-update scan also matched by marker only.
  Fixing read-plan alone would convert the injection into a **denial of
  service**: an attacker pre-posts a marker comment; the filtered read-plan
  ignores it so planning proceeds, but the unfiltered write-plan finds the
  attacker's comment first and PATCHes it with the real plan; then the
  implement-phase read-plan ignores that attacker-authored comment → no plan
  found → exit 1. So write-plan applies the same filter: it only ever updates a
  trusted-authored comment and otherwise POSTs a fresh trusted-authored one.

## Changes

**Unit 1 — author-filter both plan scripts.** In `dispatch-read-plan` and
`dispatch-write-plan`, resolve `PLAN_AUTHOR` after the marker line and fail
clearly if it cannot be resolved (no silent fallback to an unfiltered scan).
Switch each marker scan from gh's built-in `--jq` to a real `jq` pipe so
`--arg` carries both the marker and the author:
`select((.body | contains($m)) and (.user.login == $author))`. The direct
`gh … | jq` pipe is safe under `.claude/rules/shell-json.md` (no `echo`
in between). Header comments note the author restriction.

**Unit 2 — test coverage.** Extend the `dispatch-write-plan / dispatch-read-plan`
block in `test-dispatch-scripts.sh`: the fake `gh` POST handler now stamps
`.user.login`, the block pins `DISPATCH_PLAN_AUTHOR=plan-bot` (the fake has no
`user` endpoint), and a new foreign-author injection case asserts read-plan
ignores an attacker-authored marker comment (exit 1) and write-plan POSTs a
fresh trusted comment rather than PATCHing the attacker's — leaving the forged
comment untouched. Existing positive round-trip tests still pass because they
POST as the trusted author.

## Verification

`bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` —
2666/2666 passed, 0 failed, including the new foreign-author cases.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-read-plan
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-read-plan
@@ -8,6 +8,11 @@
 # implement phase reads the plan from here in a fresh context. Exits non-zero
 # with a clear diagnostic when no plan comment exists.
 #
+# The comment must be authored by the trusted dispatch identity — the account
+# the running gh token authenticates as, overridable via DISPATCH_PLAN_AUTHOR.
+# A marker comment authored by any other account is ignored, closing the
+# forged-plan injection vector on public repos.
+#
 # The repo is resolved explicitly (GH_REPO from the common dir's remote.origin.url)
 # so gh addresses the right repository from any caller cwd. A missing arg or a
 # non-numeric issue-num is a clear error, not a fallback (exit 2).
@@ -38,11 +43,20 @@ repo="${url%.git}"; repo="${repo#*github.com[:/]}"
 export GH_REPO="$repo"
 
 MARKER='<!-- dispatch:plan -->'
+PLAN_AUTHOR="${DISPATCH_PLAN_AUTHOR:-$(gh api user --jq '.login')}"
+if [[ -z "$PLAN_AUTHOR" ]]; then
+  echo "dispatch-read-plan: could not resolve expected plan-comment author (gh api user)" >&2
+  exit 1
+fi
 
 # Fetch the comment id first, then fetch that single comment's body. Do NOT pipe
 # --jq .body through head — a multi-line body would truncate to its first line.
+# Filter to comments authored by the trusted dispatch identity to prevent a
+# forged-plan injection from an untrusted external user on a public repo.
 CID=$(gh api --paginate "repos/{owner}/{repo}/issues/$N/comments" \
-  --jq '.[] | select(.body | contains("'"$MARKER"'")) | .id' | head -n1)
+  | jq -r --arg m "$MARKER" --arg author "$PLAN_AUTHOR" \
+      '.[] | select((.body | contains($m)) and (.user.login == $author)) | .id' \
+  | head -n1)
 if [[ -z "$CID" ]]; then
   echo "dispatch-read-plan: no <!-- dispatch:plan --> comment found on issue #$N" >&2
   exit 1
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-write-plan
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-write-plan
@@ -5,8 +5,11 @@
 #
 # Reads plan markdown from STDIN and writes it to a sentinel
 # <!-- dispatch:plan --> comment on issue <issue-num>. Find-or-update: if a
-# comment containing the marker already exists, it is edited in place (PATCH);
-# otherwise a new comment is created (POST). The operation is idempotent —
+# comment containing the marker already exists AND is authored by the trusted
+# dispatch identity (the gh-token account, overridable via DISPATCH_PLAN_AUTHOR),
+# it is edited in place (PATCH); otherwise a new comment is created (POST). A
+# marker comment authored by any other account is ignored and a fresh
+# trusted-authored comment is POSTed instead. The operation is idempotent —
 # re-running replaces the plan, it never stacks a second comment. The full plan
 # lives as a comment, not in the issue body, so the machine-authored plan never
 # clobbers the human-authored acceptance criteria.
@@ -41,6 +44,11 @@ repo="${url%.git}"; repo="${repo#*github.com[:/]}"
 export GH_REPO="$repo"
 
 MARKER='<!-- dispatch:plan -->'
+PLAN_AUTHOR="${DISPATCH_PLAN_AUTHOR:-$(gh api user --jq '.login')}"
+if [[ -z "$PLAN_AUTHOR" ]]; then
+  echo "dispatch-write-plan: could not resolve expected plan-comment author (gh api user)" >&2
+  exit 1
+fi
 
 CONTENT="$(cat)"
 if [[ -z "${CONTENT//[[:space:]]/}" ]]; then
@@ -53,7 +61,9 @@ trap 'rm -f "$tmp"' EXIT
 printf '%s\n%s\n' "$MARKER" "$CONTENT" > "$tmp"
 
 CID=$(gh api --paginate "repos/{owner}/{repo}/issues/$N/comments" \
-  --jq '.[] | select(.body | contains("'"$MARKER"'")) | .id' | head -n1)
+  | jq -r --arg m "$MARKER" --arg author "$PLAN_AUTHOR" \
+      '.[] | select((.body | contains($m)) and (.user.login == $author)) | .id' \
+  | head -n1)
 
 if [[ -n "$CID" ]]; then
   gh api --method PATCH "repos/{owner}/{repo}/issues/comments/$CID" \
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -25500,7 +25500,7 @@ if [[ "$method" == "POST" && "$endpoint" == *"/comments" ]]; then
   next=$(( $(cat "$COUNTER") + 1 ))
   echo "$next" > "$COUNTER"
   newbody=$(cat "$bodyfile")
-  updated=$(jq -c --argjson id "$next" --arg body "$newbody" '. + [{id:$id, body:$body}]' "$STORE")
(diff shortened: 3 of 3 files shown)


natb1:
# QA Summary — #1222 (PR #1611)

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Full dispatch-scripts test suite passes | shell-command | PASS |
| 2 | dispatch-read-plan rejects a foreign-authored plan comment | shell-command | PASS |
| 3 | dispatch-write-plan does not adopt a foreign-authored comment | shell-command | PASS |
| 4 | PLAN_AUTHOR resolution: DISPATCH_PLAN_AUTHOR override short-circuits `gh api user` | shell-command | PASS |
| 5 | jq filter shape is correct | shell-command | PASS |
| 6 | Test harness hygiene: author stamping and DISPATCH_PLAN_AUTHOR scoping | shell-command | PASS |
| 7 | Maintainer judgment: dynamic-author resolution is the right trust anchor and fails safe | — | Deferred to office-hours |

**Counts:** 6 PASS, 0 FAIL, 0 SKIP, 1 deferred to office-hours.

## Test suite

```
Results: 2666/2666 passed, 0 failed
```

New foreign-author cases (6a/6b/6c) all passed:
- PASS: read-plan ignores a forged foreign-authored plan comment
- PASS: write-plan leaves the attacker comment body unchanged
- PASS: write-plan POSTs a fresh trusted-authored plan comment
- PASS: read-plan returns the trusted plan, not the forged one

## Deferred to office-hours

**Item 7 — Maintainer judgment: dynamic-author resolution is the right trust anchor and fails safe**

Requires confirmation that plan-write and plan-read always run under the same `gh` identity in all production paths (dispatch tick, office-hours re-plan, etc.), so the trusted-author filter never rejects a legitimate plan. Also: confirm fail-closed (exit 1 on unresolved author) is the correct failure mode vs. the prior fail-open behavior.

| Item | Class | Verify verdict |
|------|-------|---------------|
| 7 — Maintainer judgment | `opus-fixable` | Refuted (1 refuted / 1 upheld) |

**Rationale (skeptic):** The three sub-questions (same gh identity across phases, staleness of dynamic resolution, fail-closed acceptability) are all answerable from the code and project rules. Both scripts derive `PLAN_AUTHOR` identically via `${DISPATCH_PLAN_AUTHOR:-$(gh api user --jq '.login')}`, and both plan and implement phases run under the same dispatch-tick ambient gh token. Dynamic resolution is always fresh. `code-style.md` endorses fail-closed. Opus can confirm this without a maintainer call.

**Counter-argument (skeptic upholding needs-human):** The discriminating sub-claim — that plan-write and plan-read always run under the same gh identity in *all* production paths — is not pinned in any in-repo artifact (`DISPATCH_PLAN_AUTHOR` absent from `.env.example`, `settings.json`, dispatch scripts). The office-hours re-plan flow (`/office-hours` step 2b re-runs `/plan-issue`) uses human-driven sessions with potentially different gh credentials than the headless tick that runs implement. If identities diverge between plan and implement, `dispatch-read-plan` exits 1 and the implement phase parks — operational risk the maintainer should confirm.

natb1:
## QA item 7 resolved in-session — trust model confirmed sound

The earlier QA comment deferred item 7 (maintainer judgment on the dynamic-author trust anchor) to office-hours. The maintainer has delegated that judgment back, so it was resolved live with direct evidence rather than parked. Recording the resolution here so the PR record reflects the actual outcome.

**Question:** Could the trusted-author filter be *too strict* and reject a legitimate plan — i.e. do plan-write and plan-read always run under the same `gh` identity in all production paths?

**Finding — the trust model holds:**

1. **Single ambient gh identity across all phases.** The dispatch deployment runs every phase (headless tick *and* office-hours sessions) on one machine under one `gh auth`. `gh api user --jq .login` resolves to `natb1`. There is no separate bot account, no per-session credential, and no token rotation between the plan and implement phases.
2. **Empirical confirmation on this very issue.** The existing `<!-- dispatch:plan -->` comment on #1222 is authored by `natb1` — exactly the identity `gh api user` resolves to. The author-filtered read selects it correctly.
3. **`DISPATCH_PLAN_AUTHOR` is not pinned anywhere** (absent from config, settings, and every dispatch script), so both scripts fall through to the same dynamic `$(gh api user)` resolution — they cannot diverge.
4. **The "office-hours uses different credentials" concern does not apply.** Office-hours and the headless tick share the same machine-level `gh auth`. The only way the identities could diverge is a reconfigured `gh auth` between phases — which would break *every* dispatch `gh` call, not just plan reading, so it is not a risk this fix introduces.

**Secondary sub-questions (both confirmed):**
- *Staleness:* `gh api user` is resolved fresh on every invocation in both scripts; no stored state, so it cannot go stale.
- *Fail-closed (exit 1) on unresolved author:* correct per `.claude/rules/code-style.md` (prefer clear errors over defensive fallbacks). It is strictly safer than the prior fail-open behavior — a misconfigured environment errors loudly instead of silently accepting a forged plan.

**Conclusion:** All items pass. The fix is a narrowly-scoped, fail-safe security hardening. QA is complete; applying `dispatch:qa-done`.

natb1:
## Review: PR #1611

**Surface:** `code` | **CodeQL:** no open alerts | **Dependency audit:** skipped (no dependency changes)

---

### Fixed

_None._

### Required (security)

_None._

### Informational

- **cr-1** — `dispatch-read-plan:46-50`, `dispatch-write-plan:47-51`: PLAN_AUTHOR guard diagnostic is unreachable on `gh` failure; literal `'null'` bypasses empty check. (`gh api user` failing silently sets PLAN_AUTHOR to the empty string in some shells, but if it prints the literal string `null`, the `-z` guard passes and the script proceeds with an unintended author.) Recommended: also reject the literal string `'null'`.
- **cr-2** — `test-dispatch-scripts.sh:26371-26481`: Test blocks 5 and 6 appear out of numerical order in the file (block 6 is inserted before block 5's bare+worktree test).
- **cr-3** — `dispatch-read-plan:46`, `dispatch-write-plan:47`: Write-read identity coupling is implicit and undocumented — both scripts must run under the same gh-token identity; a mismatch surfaces as "no plan comment found" rather than an auth error.
- **rv-2-iv-4-sec-7** — `test-dispatch-scripts.sh:26308`: The `gh api user` default-resolution path is entirely untested; the fake `gh` has no `/user` endpoint. A test without `DISPATCH_PLAN_AUTHOR` set would cover this path and the non-zero exit case.
- **rv-3** — `dispatch-write-plan:63-66`: The attacker's forged comment persists on the issue after write-plan POSTs the trusted replacement. No security impact (read-plan ignores it), but operators may not expect the lingering comment.
- **rv-5** — `dispatch-read-plan:46`, `dispatch-write-plan:47`: Each script invocation adds one extra `gh api user` call when `DISPATCH_PLAN_AUTHOR` is unset. No impact at current call frequency.
- **rv-7** — `dispatch-read-plan:46-50`, `dispatch-write-plan:47-51`: PLAN_AUTHOR resolution block is copy-pasted verbatim into both scripts. Extraction to `lib.sh` is optional; only worth doing if a third script needs the same logic.
- **sec-1** — `dispatch-read-plan:56-59`, `dispatch-write-plan:63-66`: Core injection fix is complete. `jq --arg` argument passing is correct; direct `gh … | jq` pipe is safe under `.claude/rules/shell-json.md`.
- **sec-3** — `dispatch-read-plan:46`, `dispatch-write-plan:47`: `DISPATCH_PLAN_AUTHOR` env var override is a test seam not reachable from attacker-controlled data. No action required.

### Dismissed

_None._

### False-positive (security)

- **rv-4-iv-2** — Pagination with `gh --paginate` piped to `jq` is correct — not a regression. Direct pipe is safe.
- **sec-4** — TOCTOU between comment-id scan and body fetch is not exploitable.

### Deferred

- **sec-5-iv-3** — `.user.login` is mutable; account rename + login reclaim could permit a forged comment to pass the author filter → filed as #1644 (blocked by #1222).
- **sec-6** — SIGPIPE from `head -n1` on multi-comment issues may cause spurious exit under `pipefail` → filed as #1643 (blocked by #1222).

### Out-of-scope (security)

_None._

PR #1616, opened 2026-06-14
dispatch-select-target: extract shared category/priority jq defs from ISSUE_SORTED + helpers

### LINKED ISSUE #1490: dispatch-select-target: ISSUE_SORTED jq duplicates category/priority logic from `category_of_labels` / `has_priority_in_labels`
## Context

The \`ISSUE_SORTED\` jq filter (lines 895-906 of \`dispatch-select-target\`) inlines duplicates of the category-resolution and priority-detection logic from the \`category_of_labels\` and \`has_priority_in_labels\` shell helpers (lines 610-628). The comment acknowledges this ('the exact ones used by category_of_labels / has_priority_in_labels').

This creates a maintenance hazard: if either shell helper changes its logic (e.g., a new label convention, a change in how category names are matched), the inlined jq must be updated separately, with no static check ensuring the two stay in sync. A divergence would silently mis-sort roots, potentially breaking the early-exit soundness argument that relies on consistent rank ordering between the walk and the emission loops.

Surfaced by the review pass on #1479; out of scope for that PR.

## Recommended fix

Extract the category/priority jq fragments into a shared jq library or jq `def` block referenced by both the ISSUE_SORTED filter and the shell helpers. Alternatively, add a comment block at both sites that cross-references the other by line number, and add a test that validates ISSUE_SORTED's rank order against the output of `category_of_labels`/`has_priority_in_labels` for a representative label set.

## Files

- `.claude/skills/dispatch-propagate/scripts/dispatch-select-target`

## Acceptance criteria

- [ ] The category-resolution logic (first-matching-category from TOPIC_CATEGORIES) appears in exactly one canonical location, or both sites have a cross-reference comment pointing to the other with a line-number anchor.
- [ ] The priority-detection logic (`index("priority")`) appears in exactly one canonical location, or both sites have a cross-reference comment.
- [ ] A test validates that ISSUE_SORTED's rank ordering (pri desc, catrank asc, createdAt asc) produces the same per-issue category/priority assignments as `category_of_labels`/`has_priority_in_labels` for a representative label set (e.g., each TOPIC_CATEGORIES entry plus "other").
- [ ] No regression in the existing dispatch-select-target test suite.

natb1:
Closes #1490

## Why

`dispatch-select-target` classifies the dispatch queue's open issues by topic
category and priority. Two shell helpers owned that logic —
`category_of_labels()` (first matching `TOPIC_CATEGORIES` entry, else `"other"`)
and `has_priority_in_labels()` (`1` if an exact `priority` label is present).
The `ISSUE_SORTED` jq filter re-implemented **both** expressions inline to
precompute its `pri`/`catname` sort columns, and the two copies were kept in
sync only by a comment — no static check. A helper change not mirrored into the
inline jq would silently mis-sort the queue's roots, breaking the early-exit
soundness argument that relies on the walk and the emission loops sharing one
rank order (#1452/#1548).

## What changed

**Unit 1 — single-source extraction.** The two jq expressions now live in one
`JQ_LABEL_DEFS` shell variable (defs `has_priority` and `category_of($cats)`)
local to the script. `category_of_labels()`, `has_priority_in_labels()`, and the
`ISSUE_SORTED` filter all call those defs — exactly one copy of each expression.
The `pri`/`catname`/`catrank` columns and sort order of `ISSUE_SORTED` are
preserved byte-for-byte; this is a pure dedup. `catrank` (the `CATEGORIES`
iteration-order rank) has no shell-helper analog and stays inline, out of scope
for the dedup.

**Unit 2 — machine-checkable agreement proof.** A hidden `--debug-classify`
subcommand emits, for each surviving issue, the classification from BOTH the
real inline `ISSUE_SORTED` jq and the real `category_of_labels` /
`has_priority_in_labels` shell functions. A new test in
`test-dispatch-scripts.sh` asserts the two agree across every `TOPIC_CATEGORIES`
entry, the `"other"` fallback, and a `priority` case. Because def-level text
equality is true by construction, the real subject under test is the adapter
wiring (`(.labels // []) | the_def` and threading `$cats`).

## Design note

The canonical defs stay in `dispatch-select-target`, not `lib.sh`:
`category_of` depends on `TOPIC_CATEGORIES` and this script's ranking policy with
zero cross-script reuse, so moving it to the generic `lib.sh` would disperse the
policy and re-introduce a list the test would have to mirror. `--debug-classify`
gives the test reachability without that dispersal.

## Verification

The full dispatch script suite passes (2687/2687, exit 0), including the new
criterion-#3 block and every existing select-target subprocess test — confirming
the `ISSUE_SORTED` output is unchanged.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-select-target
@@ -562,14 +562,25 @@ CATEGORIES=("${TOPIC_CATEGORIES[@]}" other)
 # Same priority list as a JSON array, for category_of_labels' jq filter.
 TOPIC_CATEGORIES_JSON=$(printf '%s\n' "${TOPIC_CATEGORIES[@]}" | jq -Rcn '[inputs]')
 
+# Single source of the category/priority label expressions. Each def takes a
+# JSON array of {name:...} label objects as input and derives [.[].name]
+# itself, so callers pass the raw label array. category_of_labels /
+# has_priority_in_labels wrap these, and the ISSUE_SORTED filter below calls
+# the same defs — there is now exactly one copy of each expression.
+JQ_LABEL_DEFS='
+  def has_priority: [.[].name] | if index("priority") then 1 else 0 end;
+  def category_of($cats):
+    [.[].name] as $names
+    | first($cats[] | select(. as $c | $names | index($c))) // "other";
+'
+
 # Highest-priority topic category among a JSON array of {name:...} label
 # objects — the first TOPIC_CATEGORIES entry that appears as an exact label
 # name, else "other". Exact match, so the phase label "dispatch:reviewed" never
 # reads as topic "dispatch".
 category_of_labels() {
-  printf '%s' "$1" | jq -r --argjson cats "$TOPIC_CATEGORIES_JSON" '
-    [.[].name] as $names
-    | first($cats[] | select(. as $c | $names | index($c))) // "other"'
+  printf '%s' "$1" | jq -r --argjson cats "$TOPIC_CATEGORIES_JSON" \
+    "$JQ_LABEL_DEFS"' category_of($cats)'
 }
 
 # Union of a PR's closing issues' labels into a JSON array of {name:...} label
@@ -583,7 +594,7 @@ pr_combined_labels() {
 
 # 1 if the label set contains an exact "priority" label, else 0.
 has_priority_in_labels() {
-  printf '%s' "$1" | jq -r '[.[].name] | if index("priority") then 1 else 0 end'
+  printf '%s' "$1" | jq -r "$JQ_LABEL_DEFS"' has_priority'
 }
 
 # Top-N ranked PRs per (category, priority-bit, phase): key =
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — Issue #1490

**PR #1616** | All acceptance criteria verified via script checks. No browser component.

## Items

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Full dispatch-script test suite passes offline | shell-command | PASS |
| 2 | Category/priority logic in exactly one canonical location | shell-command | PASS |
| 3 | `--debug-classify` subcommand wiring | shell-command | PASS |

**Totals:** 3 passed, 0 failed, 0 skipped

## Detail

**Item 1 — Full test suite (2687/2687, exit 0):**  
All existing select-target subprocess tests pass unchanged, and the new `#1490 --debug-classify` block passes — 22 assertions covering every `TOPIC_CATEGORIES` entry (`security`, `bug`, `testing infrastructure`, `dispatch`, `landing`, `fellspiral`, `budget`, `print`, `audio`), the `"other"` fallback, and the `priority` case, plus row-count and spot-value checks.

**Item 2 — Single canonical source (criteria #1 and #2):**
- `has_priority` body (`index("priority")`) appears **1** time (expect 1) ✓
- `category_of` body (`first($cats[]...)`) appears **1** time (expect 1) ✓
- `$JQ_LABEL_DEFS` referenced at **3** sites (category_of_labels, has_priority_in_labels, ISSUE_SORTED) ✓

**Item 3 — `--debug-classify` wiring:**
- Arg branch (`DEBUG_CLASSIFY=true`) at line 179 ✓
- Emit block (`if [[ "$DEBUG_CLASSIFY" == "true" ]]`) at line 893 ✓
- 5-column printf at line 904 ✓

## Deferred to office-hours

None.

PR #1622, opened 2026-06-14
test-aggregate-usage.sh: cover cmd_prefix env-var prefix stripping

natb1:
Closes #1588

## What

Add a fixture line and two assertions to `test-aggregate-usage.sh` that cover
`cmd_prefix`'s env-var-assignment stripping in the dispatch token-audit's
`tool_sequences` lens.

## Why

`cmd_prefix` (`aggregate-usage.sh`) normalizes a Bash command into an n-gram
token, and its first step strips leading env-var assignments
(`sub("^([A-Za-z_][A-Za-z0-9_]*=[^ ]* +)+"; "")`), so
`VITE_GITHUB_BRANCH=foo npm run build` keys as `Bash:npm run`, not
`Bash:VITE_GITHUB_BRANCH=foo npm`. The suite previously exercised `cmd_prefix`
only via plain, non-prefixed commands — the env-var-stripping branch had no
coverage. A silent regression of that `sub(...)` regex would corrupt the
`tool_sequences` n-gram keys with no failing test.

## Change

`test-aggregate-usage.sh` only — no change to `aggregate-usage.sh` (the strip
already works correctly):

- A new **zero-usage** worker assistant fixture line with one Bash tool call
  whose command begins with an env-var assignment. All-zero `usage` keeps it
  total-neutral: every totals / price / `baseline_context` / session-count
  assertion is unchanged; only the worker `tool_calls` order gains a trailing
  `Bash:npm run` token (`A,B,A,B,NPM`).
- Two new assertions: a positive one confirming the stripped key
  `Bash:npm run` appears in a top sequence, and a negative regression guard
  confirming no n-gram token retains an env-var assignment (`=`).
- Comment upkeep: renumbered the shifted fixture-line headers and updated the
  `tool_sequences` order comment to `A,B,A,B,NPM`.

## Verification

`bash .claude/skills/dispatch-token-audit/scripts/test-aggregate-usage.sh` →
`Results: 28/28 passed, 0 failed` (26 existing + 2 new). Deleting the
`sub(...)` strip from a scratch copy fails exactly the two new assertions, the
other 26 still passing — proving the fixture is total-neutral and the
assertions genuinely guard the strip.

### DIFF
--- a/.claude/skills/dispatch-token-audit/scripts/test-aggregate-usage.sh
+++ b/.claude/skills/dispatch-token-audit/scripts/test-aggregate-usage.sh
@@ -66,14 +66,23 @@ setup() {
   printf '%s\n' '{"type":"assistant","attributionSkill":"plan-implement","isSidechain":false,"gitBranch":"999-fixture","message":{"model":"claude-opus-4-8","content":[{"type":"tool_use","id":"toolu_003","name":"Bash","input":{"command":".claude/skills/dispatch-propagate/scripts/dispatch-context-pack 999 --pr"}},{"type":"tool_use","id":"toolu_004","name":"Bash","input":{"command":"gh issue view 999 --json labels"}}],"usage":{"input_tokens":100,"cache_creation_input_tokens":200,"cache_read_input_tokens":400,"output_tokens":50}}}' \
     >> "$worker_jsonl"
 
-  # line 4: tool-error user line — normalizes to "Exit code N"
+  # line 4: assistant — zero-usage fixture line exercising cmd_prefix's env-var
+  # stripping (#1588). One Bash tool call whose command begins with an env-var
+  # assignment; cmd_prefix must strip it, keying the n-gram as "Bash:npm run".
+  # All-zero usage so every totals / price / baseline_context / session-count
+  # assertion is untouched; only the worker tool_calls order gains a trailing
+  # NPM token (A,B,A,B,NPM).
+  printf '%s\n' '{"type":"assistant","attributionSkill":"plan-implement","isSidechain":false,"gitBranch":"999-fixture","message":{"model":"claude-opus-4-8","content":[{"type":"tool_use","id":"toolu_005","name":"Bash","input":{"command":"VITE_GITHUB_BRANCH=foo npm run build"}}],"usage":{"input_tokens":0,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":0}}}' \
+    >> "$worker_jsonl"
+
+  # line 5: tool-error user line — normalizes to "Exit code N"
   # The \n inside the content string is the two-character JSON escape sequence
   # (backslash + n), not a real newline. printf '%s\n' with single-quoted string
   # keeps the backslash literal in the file bytes.
   printf '%s\n' '{"type":"user","message":{"content":[{"type":"tool_result","is_error":true,"content":"Exit code 1\nsome detail"}]}}' \
     >> "$worker_jsonl"
 
-  # line 5: tool-result user line for the toolu_001 Bash call — known-length
+  # line 6: tool-result user line for the toolu_001 Bash call — known-length
   # ASCII payload "PAYLOAD_0123456789" (18 bytes) attributed to Bash. The error
   # result above has NO tool_use_id → attributed to "unknown" (23 bytes).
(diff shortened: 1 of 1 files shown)


natb1:
## QA Summary — #1588 / PR #1622

**Branch:** `1588-test-aggregate-usage-sh-add` | **No browser component** (pure script change)

### Results: 5 PASS · 0 FAIL · 0 SKIP · 1 deferred

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Run aggregate-usage unit suite end-to-end | shell | ✅ PASS — 28/28 passed, 0 failed |
| 2 | Positive assertion: stripped key `Bash:npm run` present | shell | ✅ PASS |
| 3 | Negative guard: no n-gram token retains `=` | shell | ✅ PASS |
| 4 | Fixture line `VITE_GITHUB_BRANCH=foo npm run build` present | shell | ✅ PASS — 2 matches (JSON line + comment) |
| 5 | Zero-usage fixture leaves totals/price assertions unchanged | shell | ✅ PASS — 0 FAIL lines |
| 6 | Judge assertions genuinely coupled to strip regex | — | ⏸ Deferred to office-hours |

### Deferred to office-hours

**Item 6 — Judge that the new assertions genuinely test the strip regex, not a tautology**

Requires confirming that removing the `sub(...)` strip from `aggregate-usage.sh` would flip exactly the two new assertions to FAIL while leaving the other 26 passing. Deferred as `needs-human-judgment`.

### Bugs found

None.

---

### Disposition triage (report-only)

This section is informational only. Every residue item still escalates to office-hours below, exactly as before. The disposition triage changes no terminal behavior.

| # | Item | Class | Rationale |
|---|------|-------|-----------|
| 6 | Judge assertions genuinely coupled to strip regex | `opus-fixable` | Resolving this is a deterministic technical question fully answerable from the script source — not a pixel-level call or user-intent question. Coupling trace: the strip regex removes `VITE_GITHUB_BRANCH=foo ` from the fixture command, yielding token key `Bash:npm run`; under a strip regression the key becomes `Bash:VITE_GITHUB_BRANCH=foo npm`. The positive assertion requires `Bash:npm run` (would yield 0 vs expected 1 → fails) and the negative guard counts any token matching `=` (would be ≥1 vs expected 0 → trips). Both assertions break precisely under the regression. |

natb1:
## QA follow-up — item 6 resolved → clean pass

The one deferred `needs-human-judgment` item (#6 — "Judge that the new assertions genuinely test the strip regex, not a tautology") has been resolved empirically.

**Test:** copied `aggregate-usage.sh` + `test-aggregate-usage.sh` to a scratch dir, deleted the env-var strip line (`sub("^([A-Za-z_][A-Za-z0-9_]*=[^ ]* +)+"; "")`) from the scratch `aggregate-usage.sh`, and ran the suite.

**Result:**
```
  FAIL: tool_sequences env-var prefix stripped to "Bash:npm run"
  FAIL: tool_sequences: no n-gram token keeps an env-var assignment
Results: 26/28 passed, 2 failed
```

Deleting the strip fails **exactly** the two new assertions and leaves the other 26 passing. The assertions are genuinely coupled to the strip regex — not tautological — so they provide the regression signal the issue asks for.

All QA items now pass with no bug and no outstanding human-judgment residue. Marking `dispatch:qa-done`.

natb1:
## Review: #1622

No code changes required. 3 findings across 2 buckets.

---

### Informational

- **test-aggregate-usage.sh:212** — The positive env-var-stripping assertion converts `length > 0` (a boolean) to 1/0 via `if . then 1 else 0 end` and then asserts `"1"`. This is indirect: `length > 0` is a boolean but `assert_eq` compares strings, so the `if`-branch is needed. Not wrong, but a direct `any(.sequence[]; . == "Bash:npm run") | if . then 1 else 0 end` would be marginally clearer. No change required — the current form is correct and already exercised passing tests. _(sources: code-review, review)_

---

### False-positive

- **test-aggregate-usage.sh:214** — Reviewers flagged that `test("=")` matches any token containing `=`, not only env-var-prefixed ones (e.g. `--flag=value` style args). Classified false-positive: no token in the fixture's tool sequences contains a bare `=` outside the env-var assignment under test, and the assertion's purpose is to confirm the stripping regex removes the env-var fragment — a broader `=` check is correct-enough for the fixture and doesn't introduce false failures. The specific concern (a future fixture adding `--flag=value` commands could cause spurious failures) is a hypothetical, not a current defect. _(sources: code-review, review)_

---

_No security findings (no dependency changes; no open CodeQL alerts). No fixes applied. No follow-up issues filed._

PR #1629, opened 2026-06-14
blog: render link labels through parseInline to sanitize SSR HTML

natb1:
Closes #1191

## Problem

The blog's `marked` link renderer (`blog/src/marked-config.ts`) inserted the
link label directly into the anchor output. In `marked@15`, the object link
renderer's `text` argument is the token's **raw source text**, not Marked's
pre-rendered/escaped inner HTML. Two consequences:

- **XSS in the SSR/prerender path.** `[](https://e.com)`
  emitted `<a ...></a>` — raw HTML passed straight
  through. The client GitHub-fetch hydration path re-sanitizes with DOMPurify,
  but the SSR/prerender path and the `outlet.innerHTML` path insert
  `content.html` **raw**, so bots, crawlers, and no-JS visitors received
  unsanitized HTML.
- **Lost inline formatting.** `[*emphasis*](https://e.com)` rendered the literal
  `*emphasis*` instead of `<em>emphasis</em>`, because the raw source text was
  never inline-parsed.

The renderer's own comment claimed a server-side sanitizer was unnecessary
because `html: () => ""` strips all raw HTML — but that invariant was **false**
for link labels, which bypassed the `html` renderer entirely.

## Fix

Render the link label through the renderer's own parser:
`const label = this.parser.parseInline(tokens)`. This walks the label's inline
tokens through the same renderer, so a raw-HTML token hits `html: () => ""`
(stripped — XSS closed) and an `em`/`strong`/`code` token renders via Marked's
escaping defaults (formatting restored). One fix at the renderer covers every
consumer — SSR, prerender, hydration, both `landing` and `fellspiral` apps —
with no per-sink DOMPurify pass. The now-true comment is rewritten and states
the boundary precisely: unsafe URL protocols (e.g. `javascript:` hrefs) are
**not** neutralized here and remain out of scope, tracked in #1192.

## Tests

Two tests added to the `link renderer` block in
`blog/test/marked-config.test.ts`, asserting on `createMarked().parse()` output
**directly** (not through the DOMPurify-bearing hydration path, which would
give a false green):

- raw HTML in a link label is stripped (no `` wrapper intact;
- `*emphasis*` in a link label renders as `<em>emphasis</em>`.

## Verification

- `npx vitest run --root blog` — 364 tests pass, including the 2 new ones; no
  existing `marked-config` / `home` / `link-renderer` test regresses.
- Landing build compiles cleanly (`marked-config.ts` adds zero type errors).
- `grep -rn '\[!\[' --include='*.md' landing fellspiral` — no output (no live
  post nests an image in a link label, which would now route through the image
  renderer).


natb1:
# QA Summary — PR #1629 (Issue #1191)

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | XSS fix: new test passes (raw HTML stripped from link labels) | shell-command | PASS |
| 2 | Inline formatting fix: emphasis inside link labels renders correctly | shell-command | PASS |
| 3 | Regression check: all pre-existing link-renderer tests still pass | shell-command | PASS |
| 4 | No live post nests an image inside a link label | shell-command | PASS |
| 5 | Build succeeds: TypeScript types are satisfied after signature change | shell-command | PASS |
| 6 | Verify the comment at marked-config.ts:39-46 accurately describes the current mechanism | — | deferred |

## Counts

- PASS: 5
- FAIL: 0
- SKIP: 0
- Deferred to office-hours: 1

## Bugs found

None.

## Deferred to office-hours

- **Item 6 — Verify the comment at marked-config.ts:39-46 accurately describes the current mechanism:**
  Triage classified as `needs-human-judgment` (subjective documentation quality). The disposition triage reclassified it `opus-fixable` (comment accuracy is verifiable from the code alone), but `opus-fixable` items remain report-only in this phase and still escalate to office-hours.

## Disposition triage

One residue item entered disposition triage (Step 3.5). The Workflow classified it and returned:

| # | Title | Kind | Class | Verify | Rationale |
|---|-------|------|-------|--------|-----------|
| 6 | Verify the comment at marked-config.ts:39-46 accurately describes the current mechanism | needs-human-judgment | opus-fixable | n/a | Comment accuracy is verifiable from the source code alone — no human, main-gated, or aesthetic judgment required. Reading `marked-config.ts` confirms the comment accurately describes `this.parser.parseInline` as the label mechanism, raw HTML routing through `html: () => ""`, and the `javascript:` scope-out (tracked in #1192). |

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR. `opus-fixable` and `needs-human` items remain report-only here and still escalate to office-hours below, exactly as before.

No `needs-main` items — Step 3.6 was skipped. No filed follow-ups.

natb1:
## Review pass — PR #1629 / issue #1191

### Fixed

- **`blog/src/marked-config.ts:65`** — Removed redundant `tokens as Token[]` cast: `Tokens.Link.tokens` is already typed `Token[]` in the marked type definitions, so the assertion was a no-op cast. Dropped the now-unused `Token` import. (commit `92ee49ea`)

### Dismissed

- **`blog/test/marked-config.test.ts:143`** — No test for other inline-format interactions beyond `*emphasis*` (e.g. `**bold**`, `` `code` ``, `~~strike~~`). The existing test is representative and the `parseInline` path is general-purpose; covering every token type would be exhaustive without adding signal. _No change._

### False-positive (security)

- **`blog/src/marked-config.ts:65`** — Claim that a nested image token inside a link label (e.g. `[x](https://a.com)`) would crash the image renderer via `parseInline`. `parseInline` routes inline tokens; a nested `` in a link label is parsed as a `link` or `image` token and handled by the image renderer, which validates against `IMAGE_DIMENSIONS` (unknown paths throw). The renderer's existing throw-on-unknown guard is the correct behavior, not a vulnerability. _No change._

### Deferred

- **`blog/src/marked-config.ts:58`** — `escapeHtml(href)` does not neutralize `javascript:` (or `vbscript:`, `data:`) URL schemes. In the SSR path, `DOMPurify` is never applied to pre-rendered content, so a blog post with `[click](javascript:alert(document.cookie))` produces a live XSS vector. → filed as follow-up #1192 (blocked by #1191)
- **`blog/test/marked-config.test.ts:135`** — XSS test does not cover the `javascript:` href vector, leaving the acknowledged gap untested and unguarded against future regression. → filed as new follow-up #1670 (blocked by #1191)
- **`blog/src/marked-config.ts:50`** — Comment explicitly acknowledges `javascript:` hrefs are not neutralized; same underlying risk as above. → tracked in #1192 (blocked by #1191)

### Out-of-scope (security)

- **`blog/src/marked-config.ts:49–50`** (input-validation, security-review, auth, firebase) — Four finders independently flagged that `escapeHtml(href)` does not neutralize unsafe URL protocols. All are pre-existing in the unchanged portion of the renderer and are tracked in #1192. _No action in this PR._

---

_No deviation. No required security findings left unresolved. CI passing._

PR #1630, opened 2026-06-14
landing: reject unsafe href protocols in marked link renderer

natb1:
Closes #1192

## What

Add an allowlist guard to the `link()` renderer in `blog/src/marked-config.ts`
so it emits an anchor only for safe href schemes. Any other scheme renders the
link's label as plain (escaped) text with no anchor.

```typescript
const SAFE_HREF = /^(https?:|mailto:|tel:|\/|#)/i;
// …
link({ href, text, title }) {
  if (!SAFE_HREF.test(href)) {
    return escapeHtml(text);
  }
  // … unchanged safe-link rendering
}
```

## Why

The renderer escaped hrefs with `escapeHtml` before placing them in an `href=`
attribute but never rejected dangerous URI schemes (`javascript:`, `data:`,
`vbscript:`). `escapeHtml` only escapes `& < > " '` — it does not neutralize the
`javascript:` scheme, so `[click](javascript:alert(1))` rendered a live
`<a href="javascript:alert(1)" …>`.

The client hydration path runs DOMPurify, which strips these — but the
build-time prerender / virtual-module paths have no such guard, so the dangerous
anchor shipped verbatim in the prerendered HTML that bots, crawlers, and
JS-disabled browsers receive. This closes that SSR/prerender XSS exposure.

## Design notes

- **Allowlist, not denylist** — anything that does not begin cleanly with a safe
  scheme prefix is rejected, which is robust against whitespace/case/control-char
  obfuscation of `javascript:`.
- The rejected branch returns `escapeHtml(text)`. In marked v15's
  renderer-object API the `text` parameter is the raw source label (not
  pre-parsed inline HTML), so this does not double-escape and it defensively
  neutralizes raw HTML in a rejected link's label.
- **No legitimate link is broken.** The allowlist accepts `http(s):`,
  `mailto:`, `tel:`, leading-slash, and `#anchor`. A grep of every inline link
  target in current post content (`landing/post/`, `fellspiral/post/`) shows all
  targets are `http(s):` or `/`-prefixed. Bare relative paths (`foo/bar`) are
  rejected as an intentional side effect — the blog's internal-link convention
  is leading-slash — and a test documents this.

## Out of scope

- The raw-label-HTML sanitization of the *safe* branch is tracked separately in
  #1191; the safe branch keeps `${text}` unchanged.
- The `image()` renderer, `isExternalHref()`, and the `pages/home.ts` DOMPurify
  config are unchanged.

## Tests

Added cases to the `link renderer` describe block in
`blog/test/marked-config.test.ts` covering `javascript:`/`data:`/`vbscript:`
rejection, case/leading-space obfuscation variants, safe-scheme regression
guards, escaping of HTML metacharacters in a rejected label, and the
intentional bare-relative-path rejection. Full blog suite: 374/374 pass.


natb1:
# QA Summary — PR #1630 / Issue #1192

## Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Unit suite confirms unsafe schemes are neutralized and safe schemes still work | shell-command | PASS |
| 2 | Full blog suite passes (no regression from the renderer change) | shell-command | PASS |
| 3 | Source guard matches the acceptance criteria (allowlist present in the renderer) | shell-command | PASS |
| 4 | Reviewer judgment: allowlist scope vs. the issue's named protocols and relative-path policy | needs-human-judgment | deferred |

**Counts: 3 PASS · 0 FAIL · 0 SKIP · 1 deferred**

No browser walkthrough (changed files are in `blog/` library package — no known frontend app package modified).

## Disposition triage

**Item 4** was triage-classified `needs-human-judgment`. The skeptic fan-out (2/2 votes) **refuted** that classification — the bare-relative-path question is machine-verifiable from the codebase: the plan comment on #1192 already records a grep of `landing/post/` and `fellspiral/post/` showing zero `./`-relative or bare-relative links, and the committed test at `blog/test/marked-config.test.ts:218` documents the rejection as intentional. Both skeptics concluded no human judgment is required to answer the factual content question.

**Fix-plan outcome — scope deviation.** The Opus fix planner classified this as `opus-fixable` (matching the skeptic verdict) but refused to author a fix. The planner's reason:

> Finding 4 is a human-confirmation item the QA reviewer deliberately punted ("human confirms acceptable OR flags as gap"); the acceptance criteria list `./` under "continue to work as before," but the impl's `SAFE_HREF=/^(https?:|mailto:|tel:|\/|#)/i` rejects `./`/`../` (only leading-slash `/` and `#` pass) and the PR codified that rejection as an intentional, tested choice (test at `marked-config.test.ts:218`) — so the human must decide whether the author's deliberate bare-relative rejection overrides the `./` criterion or is a content-breaking gap, a behavior decision the issue does not authorize an autonomous fix to make.

**`needs-main` items:** none — none were filed as follow-ups, and none are dropped from the escalation set.

**Terminal behavior (scope-deviation escalation):** The fix planner returned `deviation=true` with no units. This is a permanent escalation, not a retry — no attempt label is applied. Escalating to office-hours for human resolution of whether the `./` relative-path rejection is acceptable (no existing blog content uses `./`-relative links, per the plan grep) or whether the acceptance criteria should be updated to reflect the intentional choice.

natb1:
# QA item #4 resolved in-session (supersedes the prior scope-deviation escalation)

Per direct user instruction to use best judgment on the opus-fixable item, the `./` relative-path gap is now **fixed** rather than escalated.

**The gap:** Issue #1192's acceptance criteria list `./` among the relative paths that must "continue to work as before," but `SAFE_HREF` only accepted leading-slash (`/`) and `#`, so `[x](./sibling)` rendered as inert text — a latent footgun for future authors.

**The fix** (commit `e9caac78`): broadened the allowlist regex from `\/` to `\.{0,2}\/`, so root-relative (`/`), same-dir (`./`), and parent-dir (`../`) paths all produce anchors.

**Security preserved:** a reference beginning with `.` or `/` cannot encode a URI scheme (RFC 3986 schemes must start with a letter), so no `javascript:`/`data:`/`vbscript:` payload slips through. `.javascript:` (no slash) stays rejected; `./javascript:alert(1)` is parsed by browsers as a relative path segment, never a scheme. Bare relative paths (`foo/bar`) stay rejected per the leading-slash convention.

**Tests:** added `./` and `../` anchor-regression cases. Full blog suite: 376/376 pass.

All four QA acceptance criteria are now satisfied.

natb1:
## Review findings — #1192

### Fixed

- **`blog/src/marked-config.ts:68` — Silent link downgrade on non-allowlisted hrefs** (code-review): Bare-relative paths (e.g. `foo/bar`), query-only hrefs (`?x=1`), and unrecognised schemes silently degraded to plain label text with no build-time signal. Added `assertSafeHrefs()` in `renderPostContents()` that lexes every published post at build time and throws on any link href that fails `SAFE_HREF`, surfacing the silent downgrade as a loud build error. _(fix commit: dae21480)_

### Informational

- **`blog/src/marked-config.ts:47` — `SAFE_HREF` also matches protocol-relative URLs** (code-review, review): The sub-pattern `\.{0,2}\/` allows zero dots, so it also matches `//evil.com`. In this renderer the value still passes through `escapeHtml(href)` and is never executed as a script, so there is no XSS vector — a protocol-relative URL rendered into `href="//evil.com"` just navigates to that host, identical to `https://evil.com`. No change required; noted for awareness.

### False-positive (security)

- **`blog/src/marked-config.ts:47` — `SAFE_HREF` and protocol-relative URLs** (red-team): The regex correctly rejects `javascript:`, `data:`, and `vbscript:` and the protocol-relative case (`//evil.com`) is navigational, not script-execution. Not a vulnerability.

- **`blog/src/marked-config.ts:68-69` — `escapeHtml(text)` on rejected label** (red-team): `text` in marked v15's renderer API is the raw source label, not pre-parsed inline HTML. `escapeHtml` correctly neutralises HTML metacharacters; no double-escaping or injection path exists. Not a vulnerability.

PR #1660, opened 2026-06-14
issue-sub-issues: guard against empty-string FIELDS arg

### LINKED ISSUE #1631: issue-sub-issues: guard against empty-string FIELDS arg
## Problem

\`issue-sub-issues\` uses \`FIELDS="${2:-number,state,stateReason}"\` which falls back to the default only when \`$2\` is unset or null. An explicit empty string \`""\` passes through and produces \`gh issue view $sub --json ""\`, which the \`gh\` CLI rejects with a confusing error.

No current caller triggers this path (both callers either omit \`$2\` or pass a non-empty string), but it is a latent trap for any future caller.

## Proposed fix

Add a guard after the FIELDS assignment:

\`\`\`bash
[[ -n "$FIELDS" ]] || { echo "issue-sub-issues: FIELDS arg must not be empty" >&2; exit 2; }
\`\`\`

Add a test case that passes an empty string as \`$2\` and asserts a non-zero exit with a descriptive error message.

## Scope

Single-file change to \`issue-sub-issues\` plus one test case in \`test-dispatch-scripts.sh\`.

## Acceptance criteria

- [ ] \`issue-sub-issues <N> ""\` exits non-zero (exit code 2) and prints a descriptive error to stderr
- [ ] The error message includes "FIELDS arg must not be empty" or equivalent
- [ ] \`issue-sub-issues <N>\` (no second arg) still works correctly with the default \`number,state,stateReason\` fields
- [ ] \`issue-sub-issues <N> "number,state"\` (explicit non-empty arg) still works correctly
- [ ] A test case in \`test-dispatch-scripts.sh\` covers the empty-string path and asserts the non-zero exit + error message

natb1:
Closes #1631

## What

`issue-sub-issues` assigned `FIELDS="${2:-number,state,stateReason}"` and forwarded
`$FIELDS` to `gh issue view --json` in the per-child loop. The `:-` default
substitution fires only when `$2` is unset or null, so an explicit empty string
(`issue-sub-issues <N> ""`) passed straight through to `gh --json ""`, which `gh`
rejects with a confusing error deep inside the loop. No current caller triggers
this, but it was a latent trap for any future caller.

This adds a guard immediately after the `FIELDS=` assignment, before any API call,
that exits 2 with `issue-sub-issues: FIELDS arg must not be empty` when `$2` is
supplied as an explicit empty string.

## Why the guard checks `$2`, not `$FIELDS`

Because `${2:-default}` treats an explicit empty `$2` identically to an unset `$2`,
`$FIELDS` is never empty by the time it is assigned — a `[[ -n "$FIELDS" ]]` check
could never fire. The guard instead inspects the raw positional argument:

```bash
[[ $# -lt 2 || -n "$2" ]] || { echo "issue-sub-issues: FIELDS arg must not be empty" >&2; exit 2; }
```

This fires only when `$2` is present and empty; the no-`$2` (default) and explicit
non-empty paths are untouched.

## Scope

- The `:-` default substitution stays — it still handles the unset/omitted case.
- No caller changes (no caller passes an empty string).
- No change to the default field set or the field-forwarding behavior.

## Test

Adds case E to the `issue-sub-issues field forwarding` block in
`test-dispatch-scripts.sh`, asserting `issue-sub-issues 84 ""` exits 2 with the
descriptive stderr. The guard runs before `gh_api_array`, so no stub is needed.
Full dispatch-scripts suite stays green.


natb1:
# QA Summary — Issue #1631

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Empty-string FIELDS arg is rejected with exit 2 and a descriptive error | shell-command | PASS |
| 2 | Full dispatch-scripts test suite passes (authoritative regression check) | shell-command | PASS |
| 3 | No second arg still uses the default fields (golden path, unchanged behavior) | shell-command | PASS |
| 4 | Explicit non-empty FIELDS arg still works (golden path, unchanged behavior) | shell-command | PASS |
| 5 | Guard wording stays actionable and matches the documented message | deferred (needs-human-judgment) | — |

**Counts:** 4 PASS, 0 FAIL, 0 SKIP, 1 deferred to disposition triage.

Test suite result: **2888/2888 passed, 0 failed.**

## Bugs found

None from the script-verifiable items. All four acceptance criteria checks passed.

## Disposition triage

Item 5 was a `needs-human-judgment` deferral from the triage plan (subjective judgment on error message wording and exit-code convention). The disposition Workflow reclassified it as **`opus-fixable`** based on code evidence — exit 2 is the established usage-error convention in this script family, and `test-dispatch-scripts.sh` already asserts the exact message with a passing test.

However, the fix-planning phase returned a **scope deviation** (`deviation: true`, empty `units`):

> *"The sole opus-fixable finding (id 5) is a human-judgment deferral — not a code defect; the guard, message, and test already satisfy every acceptance criterion, so there is no autonomous code change to make."*

`needs-main` handling: no `needs-main` items — the `needs-main` note is not applicable here.

**Terminal behavior (escalation pass):** The scope-deviation escape path fires when `result.deviation === true` and no fix units were authored. No attempt label is applied. This pass escalates to office-hours so a human can confirm in seconds that the wording and exit-code convention are acceptable — the deviation_reason already documents the confirmation work (green test, matching convention).

natb1:
## QA update — marked complete

Superseding the escalation in the prior comment. Per maintainer direction, the
single residue item (#5) was the only `opus-fixable` finding, and the fix planner
confirmed there is **no code change to make** — the guard, error message
(`issue-sub-issues: FIELDS arg must not be empty`), exit-2 convention, and the
green test already satisfy every acceptance criterion. There were no other residue
items to address.

With nothing to fix and nothing else outstanding, QA is marked complete:
`dispatch:qa-done` applied. The chain advances past the QA phase.

natb1:
## Review: #1631 guard for empty FIELDS arg

No code changes applied — all findings are informational, dismissed, or false-positive.

### Informational

- **CR-1** (`issue-sub-issues:7`): Guard placement after assignment — the placement after `FIELDS=` is intentional and explained in the PR body (the guard must check `$2`, not `$FIELDS`, because `:-` substitution already masks an empty `$2` by the time `FIELDS` is assigned).
- **REV-2** (`test-dispatch-scripts.sh:2250`): Test copies `issue-sub-issues` to `TMPDIR_TEST` before invoking it — consistent with the existing pattern used throughout the test block.

### Dismissed

- **CR-2** (`issue-sub-issues:7`): Concern about `$# -lt 2 || -n "$2"` idiom — condition is correct; the negated short-circuit `|| { … exit 2; }` is idiomatic bash. No change needed.
- **REV-1** (`issue-sub-issues:7`): Missing guard for whitespace-only FIELDS arg — whitespace-only is a separate latent trap outside the stated scope of this PR (the PR body explicitly limits scope to the empty-string case).
- **REV-3** (`test-dispatch-scripts.sh:2251`): `2>&1 1>/dev/null` ordering in err capture — redirect ordering is correct standard bash idiom. No change needed.

### False-positive (security)

- **SEC-1** (`issue-sub-issues:7`): No injection vector from empty FIELDS — guard exits before any `gh` call; non-empty path passes `$FIELDS` as a direct argument to `gh --json`, not through shell expansion. Not a vulnerability.
- **SEC-2** (`test-dispatch-scripts.sh:2248`): Test copies executable to TMPDIR — `cp`+`chmod` pattern is used consistently across the test suite in a controlled environment. Not a security risk.

_CodeQL scan: no open alerts for this PR._

PR #1663, opened 2026-06-14
dispatch-tick-recover: arm a heartbeat reseed instead of trusting a point-in-time busy worker

natb1:
Closes #1647

## Problem

`dispatch-tick-recover` is the dispatch chain's continuation backstop — the
systemd `OnFailure=` handler that guarantees the chain continues after an
abnormal tick exit. Its `continuation_present()` check trusted a point-in-time
"busy worker present" reading (`claude_agents_count_busy_workers > 0`) as a
guarantee that the worker's future Stop hook would carry the chain. On that
signal it reset the consecutive-failure count to 0 and armed **no** reseed.

But `claude agents` liveness is ~50% stale, and a busy worker can be killed or
hang. If that phantom carrier never fires its Stop hook, no tick is ever spawned
and no reseed was armed → **permanent stall**. Worse, the reset-to-0 meant the
consecutive-failure cap never advanced, so escalation to a
`dispatch:chain-stalled` issue never fired either.

The 2026-06-14 incident: an `empty` tick called recover, logged `continuation
present; no-op`, armed no reseed; the bg daemon then churned 217 mostly-`(killed)`
claim/settle events over 6 hours with no `dispatch-tick` until an external kick —
`recover-state.json` showed `count: 0`, the cap never advanced.

## Fix (heartbeat reseed)

After any no-spawn disposition the chain is now *guaranteed* a continuation: a
busy worker may carry it, but a bounded **deferred heartbeat reseed** is armed as
a backstop so a phantom carrier cannot strand the chain — and the
consecutive-failure cap still advances to escalation when no real continuation
materializes across repeated failures.

This adopts the issue's acceptance-criterion 1a (heartbeat reseed) over 1b (a
busy-worker liveness gate): a liveness gate would be built on the same ~50%-stale
`claude agents` signal that *caused* the bug. A heartbeat reseed depends only on
systemd's own timer.

### What changed in `dispatch-tick-recover`

- **Continuation detection narrows to the pending-timer branch only.** A pending
  `dispatch-reseed*` timer is the only systemd-*guaranteed* continuation; a busy
  worker is no longer trusted.
- **The pending-timer no-op no longer resets the count.** The count heals only
  via the existing `RESET_WINDOW` aging (a failure older than the window starts a
  fresh episode) — the true "healthy for a while" signal. The state file is left
  untouched on the no-op, preserving the real `last_failure` that drives aging.
- **Busy-aware window selection in Step 5.** A busy worker now selects a flat,
  deferred `HEARTBEAT` window (so the real Stop-hook path wins the happy-path
  race); no busy worker (or daemon UNKNOWN) keeps the existing exponential
  backoff. Either way the count was already advanced, so escalation can be
  reached under a persistent phantom-busy stall.
- **New tunable `DISPATCH_TICK_RECOVER_HEARTBEAT` (default 1800)**, validated as a
  positive integer, with a load-bearing guard that it must be `< RESET_WINDOW`
  (else the count would age back to a fresh episode every generation and the cap
  would never be reached — re-introducing the bug one level up).

The accepted tradeoff: a redundant heartbeat tick on the happy path, harmlessly
absorbed by dispatch's per-worktree dedup + concurrency gate.

`dispatch-self-close` carries a *related* busy-worker check (#1010) but it parks
a router visibly for a human rather than trusting silently — a different
mechanism with its own backstop, deliberately left untouched.

## Tests

`test-dispatch-scripts.sh` `=== dispatch-tick-recover ===` section:

- **Rewrote Test 2** — a busy worker now arms a deferred HEARTBEAT reseed at
  `NOW + 1800` and advances the count, instead of no-op-resetting it.
- **Added a count-preservation assertion** to the pending-timer test (the timer
  continuation branch no longer resets the count).
- **Added the load-bearing phantom-busy regression** — ≥ CAP+1 consecutive
  phantom-busy generations must advance the count through escalation (rounds 1-3
  arm flat heartbeats, round 4 escalates), proving the chain cannot stall
  forever. A 2-invocation test would pass while masking the bug.
- **Added two rejection tests** — `HEARTBEAT=0` and `HEARTBEAT >= RESET_WINDOW`
  both exit 2 with no reseed.

The full bash suite passes (2879/2879). It is not in CI, so it was run manually.


natb1:
## QA Summary — Issue #1647 / PR #1663

### Items executed

All items ran in the **shell-command lane** (no browser component detected — pure backend/script change to `dispatch-tick-recover` and `test-dispatch-scripts.sh`).

| # | Title | Result |
|---|-------|--------|
| 1 | Phantom-busy worker self-heals the chain instead of stalling permanently | ✅ PASS |
| 2 | Repeated phantom-busy generations advance to escalation within bound | ✅ PASS |
| 3 | Pending reseed timer is the only trusted continuation; no-op does not mask failure episode | ✅ PASS |
| 4 | Cap-defeating HEARTBEAT config is rejected loudly with no reseed | ✅ PASS |
| 5 | Full recover suite passes — no regressions in surrounding recover behavior | ✅ PASS |

### Full test suite

`bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — **2879/2879 passed, 0 failed**

### Detailed assertions

**Item 1 (busy-worker acceptance criterion 1):**
- `PASS: busy-worker: dispatch-tick-recover exits 0`
- `PASS: busy-worker: heartbeat reseed armed at NOW + 1800`
- `PASS: busy-worker: state count advances to 3`
- `PASS: busy-worker: ensure_daemon_service ran (arming path)`

**Item 2 (phantom-busy regression — acceptance criteria 2 + 3):**
- `PASS: phantom-busy: exactly 3 heartbeat reseeds (rounds 1-3)`
- `PASS: phantom-busy: round 4 escalates (chain-stalled issue)`
- `PASS: phantom-busy: state count advances to 4`

**Item 3 (pending-timer, count-preservation — acceptance criterion 2):**
- `PASS: pending-timer: dispatch-tick-recover exits 0`
- `PASS: pending-timer: no reseed armed (systemd-run log empty)`
- `PASS: pending-timer: daemon service not touched (continuation no-op)`
- `PASS: pending-timer: state count preserved (not reset)`

**Item 4 (HEARTBEAT validation — guard against cap-defeating config):**
- `PASS: heartbeat-zero: dispatch-tick-recover exits 2`
- `PASS: heartbeat-zero: no reseed armed (systemd-run log empty)`
- `PASS: heartbeat-ge-reset: dispatch-tick-recover exits 2`
- `PASS: heartbeat-ge-reset: no reseed armed (systemd-run log empty)`

### Conclusion

All three acceptance criteria are verified by the test suite:
1. A point-in-time busy-worker reading no longer permanently suppresses the reseed — a HEARTBEAT reseed is armed as a deferred backstop.
2. The consecutive-failure cap advances to escalation under a persistent phantom-busy stall (count no longer reset-to-0 on every generation).
3. The load-bearing phantom-busy regression (Test 10: ≥ CAP+1 consecutive phantom-busy rounds → escalation) passes.

Clean pass — no residue, no bugs found.

natb1:
## Review: #1647 (PR #1663)

No code changes applied — all findings are informational or false-positive.

---

### Informational

**CR-1** — `.claude/skills/dispatch-propagate/scripts/dispatch-tick-recover:362`
The clamp `if (( BACKOFF <= 0 || BACKOFF > MAX ))` on the HEARTBEAT path is a no-op (HEARTBEAT is validated positive and less than RESET_WINDOW, so it can never be <= 0 or > MAX=3600 given defaults). The inline comment notes this but calls it a "no-op" — minor stale wording since the clamp applies uniformly to both windows. No change required; the guard is harmless and the note is accurate.

**CR-2** — `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:14498`
Test 10 relies on the default RESET_WINDOW (3600 from `tr_setup`) without exporting `DISPATCH_TICK_RECOVER_RESET_WINDOW` explicitly. A future change to `tr_setup`'s default could silently break the phantom-busy stall assertion. Informational only; the test is correct as-is.

**CR-3** — `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:14492`
Test 10 calls `tr_busy_worker` once, then re-runs `dispatch-tick-recover` four times in a loop. Between iterations, the stub's busy-worker file is not reset, so all four rounds read `busy > 0`. This matches intent (persistent phantom-busy stall) but is implicit — a future test author may not realize the stub persists across loop iterations. Informational; the behavior is correct.

**CR-4** — `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:14349`
Test 3's new count-preservation assertion (`assert_eq "pending-timer: state count preserved (not reset)" "2" "$(tr_state_count)"`) is placed after the daemon-service check rather than directly after the reseed check. Ordering is a style note only; the assertion is valid and tests the correct invariant.

**CR-5** — `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:14487`
The phantom-busy regression test (Test 10) is hardcoded to CAP=3 (4 rounds). If `DISPATCH_TICK_RECOVER_CAP` is changed from its default, the loop count and expected systemd-log line count would need updating. Informational; the test is correct for current defaults.

---

### False-positive (security)

**SEC-1** — `.claude/skills/dispatch-propagate/scripts/dispatch-tick-recover:358-370`
_(OWASP A01:2021 Broken Access Control / STRIDE: Denial of Service)_
The security finder flagged `ensure_daemon_service` running before the reseed is armed as a potential unintended side-effect path. This is a false positive: `ensure_daemon_service` is an idempotent availability safeguard (it starts the Claude daemon so the scheduled reseed timer has a target). It does not introduce an auth bypass or injection path. The function's scope is already limited and the behaviour is intentional.

---

_Fixed: None. Deferred: None. Out-of-scope: None._

PR #1667, opened 2026-06-14
fellspiral: test that the live-DOM skip persists across Home -> Post -> Home

natb1:
Closes #1409

## What

Adds one test case to `fellspiral/test/main-routing.test.ts` covering the
extended scope of the live-DOM skip that PR #1396 (for #1285) introduced in
`fellspiral/src/main.ts`.

That change replaced the one-shot `isFirstHomeRender` flag with a live-DOM
check: the home route's `render` returns `null` (preserving the existing DOM,
skipping `loadPosts`) whenever `app.querySelector("#posts")` is present, and
runs `loadPosts` otherwise. The skip therefore fires whenever `#posts` is
genuinely live — not only on the initial prerender, but also after a prior
`loadPosts()` render produced a fresh `#posts`. That extended scope was
intentional but untested.

## The new case

`"keeps skipping while #posts is live from loadPosts across Post and repeat-Home navigations (#1409)"`
walks a faithful Home → Post → Home flow:

1. Enter via `/admin` so the first Home nav finds no live `#posts` and runs
   `loadPosts` → `renderHomeHtml`.
2. Navigate Home — a fresh `#posts` renders; `renderHomeHtml` called once.
3. Navigate to `/post/example` — `/post/...` and `/` share the same route
   regex (`/^\/(?:post\/.*)?$/`); `render` returns `null` because `#posts` is
   live, so `#posts` stays put and `renderHomeHtml` stays at one call.
4. Navigate Home again — also skips for the same reason; `renderHomeHtml`
   remains at exactly one call.

The `toHaveBeenCalledTimes(1)` assertion is the faithful invariant: across the
whole flow the `#posts` from the first `loadPosts()` stays live, so the home
render is reachable exactly once. It fails if `main.ts`'s live-DOM skip were
reverted to a one-shot flag or removed.

## Scope

Test-only — no changes to `fellspiral/src/main.ts` or any other file. Reuses
the file's existing scaffolding (`scaffoldDom`, `importMainTrackingListeners`,
the hoisted spies and `vi.mock` block); no new mocks or helpers.

All three cases in the suite pass.

### DIFF
--- a/fellspiral/test/main-routing.test.ts
+++ b/fellspiral/test/main-routing.test.ts
@@ -208,4 +208,55 @@ describe("fellspiral main routing (#1285)", () => {
     expect(renderHomeHtml).not.toHaveBeenCalled();
     expect(app.querySelector("[data-prerendered]")).not.toBeNull();
   });
+
+  // Issue #1409 — extended scope of the live-DOM skip.
+  //
+  // The live-DOM check (app.querySelector("#posts")) fires whenever #posts is
+  // present, not only on the initial prerender. After a prior loadPosts() run
+  // renders a fresh #posts, repeat navigations to "/" and "/post/..." both
+  // return null from render and leave #posts untouched. This case exercises
+  // the full Home → Post → Home flow to confirm renderHomeHtml is called
+  // exactly once (the initial Home) and #posts remains live throughout.
+  it("keeps skipping while #posts is live from loadPosts across Post and repeat-Home navigations (#1409)", async () => {
+    // 1. Enter via /admin so the first Home nav finds no live #posts and
+    //    therefore runs loadPosts → renderHomeHtml. Mirror the /admin setup.
+    history.pushState({}, "", "/admin");
+    scaffoldDom();
+    await importMainTrackingListeners();
+    const app = document.getElementById("app")!;
+    // Wait for the admin route to clear the prerendered #posts.
+    await vi.waitFor(() => {
+      expect(app.querySelector("#posts")).toBeNull();
+    });
+
+    // 2. Navigate Home via the real router's popstate listener.
+    history.pushState({}, "", "/");
+    window.dispatchEvent(new PopStateEvent("popstate"));
+    // Wait until a fresh #posts with data-test="home" is present.
+    await vi.waitFor(() => {
+      expect(app.querySelector("#posts")).not.toBeNull();
+      expect(app.innerHTML).toContain('data-test="home"');
+    });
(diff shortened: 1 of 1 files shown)


natb1:
# QA Summary — Issue #1409

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Correct test case exists in main-routing.test.ts | shell-command | PASS |
| 2 | All three cases in the main-routing suite pass | shell-command | PASS |
| 3 | PR is test-only (no changes to src/main.ts or app source) | shell-command | PASS |

## Details

- **Item 1** — `grep` confirmed the test case `"keeps skipping while #posts is live from loadPosts across Post and repeat-Home navigations (#1409)"` is present at `fellspiral/test/main-routing.test.ts:220`.
- **Item 2** — `npx vitest run --root fellspiral main-routing` reported **3 passed (3)** in 805ms. All three cases in the suite pass, including the new Home → Post → Home case.
- **Item 3** — `git diff --name-only origin/main...HEAD` shows only `fellspiral/test/main-routing.test.ts` changed; no app source files (`fellspiral/src/main.ts` or others) were modified.

## Disposition triage

No residue — all items passed. No disposition triage needed.

## Result

Clean autonomous pass. No bugs found, no human judgment needed.

natb1:
## Review — #1409 (PR #1667)

**CodeQL:** No open alerts. ✓
**npm audit:** No dependency changes — skipped. ✓

### Informational

- `fellspiral/test/main-routing.test.ts:247` — Step 3 (navigate to `/post/example`) uses `await new Promise((r) => setTimeout(r, 0))` as the sole synchronization before asserting `#posts` remains live. This is a microtask-flush idiom; it works here because the post route's `render` returns `null` synchronously (no async side-effects), so the zero-delay settle is sufficient. Noted for awareness; no change required.
- `fellspiral/test/main-routing.test.ts:257` — Step 4 (second Home navigation) similarly uses `await new Promise((r) => setTimeout(r, 0))` before the final `toHaveBeenCalledTimes(1)` assertion. Same rationale applies.

### Dismissed

- `fellspiral/test/main-routing.test.ts:249` — The test does not assert the specific post content after navigating to `/post/example`. The PR's intent is to verify that `#posts` stays live and `renderHomeHtml` is not re-called, not to test post-route rendering. An assertion on post content would be out-of-scope for this case; dismissed as not applicable.

### Fixed
_None._

### Deferred
_None._

### Security
_No attack surface — diff is test-only (no executable, config, dependency, or Firestore-rules changes). Security finders skipped._

PR #1681, opened 2026-06-15
budget-etl: validate category-filter fields on transaction-specific categorization rules

natb1:
Closes #1371

## What

`budget-etl` rejects category-filter fields (`MatchCategory`, `ExcludeCategory`,
`Category`) on general categorization rules — #1366 added that check in
`convertExportRules`. But transaction-specific rules (those with a
`TransactionID`) are split off by `splitRules` *before* `convertExportRules`
runs, and are applied by `applyTransactionRules`, which writes `r.Target`
directly without ever consulting the category-filter fields. As a result, a
`MatchCategory`/`ExcludeCategory`/`Category` set on a transaction-specific
categorization rule was silently inert — the same class of author confusion
#1276 / #1366 fixed for general rules.

## How

- Extract a single shared helper `validateCategoryFilterFields(export.Rule) error`
  in `budget-etl/main.go`, reproducing the existing condition and error message
  byte-for-byte.
- Wire it into `convertExportRules` (replacing the previous inline check — no
  behavior change, the error string is identical).
- Wire it into `applyTransactionRules` at the top of the rule loop, *before* the
  transaction-index lookup, so validation is unconditional and a bad txn rule
  errors even when its target transaction is not in the current batch — matching
  `convertExportRules`, which validates every rule.

The validation logic and error message now live in one place and stay identical
across both paths.

## Test

- New table-driven test `TestApplyTransactionRulesRejectsCategoryFilterOnCategorization`
  mirroring the existing `convertExportRules` test: the three reject cases
  (categorization rule with `MatchCategory`, `ExcludeCategory`, `Category` set)
  plus the pass case (`budget_assignment` rule with the same fields set).
- `go test ./...` passes from `budget-etl/`.

### DIFF
--- a/budget-etl/cmd/patch/integration_test.go
+++ b/budget-etl/cmd/patch/integration_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"math"
 	"os"
 	"testing"
@@ -102,9 +103,19 @@ func ruleIDPresent(rs []export.Rule, id string) bool {
 // convertExportRules. It is duplicated here because that helper is
 // package-private to the parent budget-etl/main.go and cmd/patch can't
 // import it.
+//
+// DIVERGENCE RISK: this is a hand-maintained copy. When main.go's
+// convertExportRules changes (e.g. the validateCategoryFilterFields check
+// added below), this helper must be updated in lockstep or the integration
+// test will silently accept inputs the real binary rejects.
 func convertExportRulesForTest(exportRules []export.Rule) ([]rules.Rule, error) {
 	out := make([]rules.Rule, len(exportRules))
 	for i, r := range exportRules {
+		// Mirror convertExportRules' validateCategoryFilterFields call:
+		// category-filter fields are only valid on budget_assignment rules.
+		if r.Type == "categorization" && (r.MatchCategory != "" || r.ExcludeCategory != "" || r.Category != "") {
+			return nil, fmt.Errorf("rule %s: category-filter fields (matchCategory/excludeCategory/category) are only valid on budget_assignment rules, not categorization rules", r.ID)
+		}
 		out[i] = rules.Rule{
 			ID:              r.ID,
 			Type:            r.Type,
--- a/budget-etl/main.go
+++ b/budget-etl/main.go
@@ -517,12 +517,20 @@ func buildRule(r rules.Rule, minAmountDollars, maxAmountDollars *float64) (rules
 	return r, nil
 }
 
+// validateCategoryFilterFields rejects category-filter fields on categorization rules.
+func validateCategoryFilterFields(r export.Rule) error {
+	if r.Type == "categorization" && (r.MatchCategory != "" || r.ExcludeCategory != "" || r.Category != "") {
+		return fmt.Errorf("rule %s: category-filter fields (matchCategory/excludeCategory/category) are only valid on budget_assignment rules, not categorization rules", r.ID)
+	}
+	return nil
+}
+
 // convertExportRules converts export.Rule to rules.Rule for the rules engine.
 func convertExportRules(exportRules []export.Rule) ([]rules.Rule, error) {
 	ruleSet := make([]rules.Rule, len(exportRules))
 	for i, r := range exportRules {
-		if r.Type == "categorization" && (r.MatchCategory != "" || r.ExcludeCategory != "" || r.Category != "") {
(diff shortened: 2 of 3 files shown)


natb1:
# QA Summary — PR #1681 / Issue #1371

## Results

**Items executed:** 6 (5 script-verifiable, 1 needs-human-judgment)

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Transaction-specific categorization rule with MatchCategory is rejected | shell-command | PASS |
| 2 | Transaction-specific categorization rule with ExcludeCategory is rejected | shell-command | PASS |
| 3 | Transaction-specific categorization rule with Category is rejected | shell-command | PASS |
| 4 | Transaction-specific budget_assignment rule with category fields is NOT rejected | shell-command | PASS |
| 5 | Validation fires even when the target transaction is absent from the batch | (deferred — see below) | — |
| 6 | Full budget-etl Go test suite passes | shell-command | PASS |

No bugs found. No browser walkthrough (pure backend change).

## Disposition triage

Item 5 ("Validation fires even when the target transaction is absent from the batch") was deferred from the initial triage as `needs-human-judgment` — verifying the validation-before-idIndex-lookup ordering required code inspection rather than a shell command. The disposition Workflow classified it **`opus-fixable`**: code inspection confirmed the production ordering is already correct (`validateCategoryFilterFields` runs at line 568 in `applyTransactionRules`, before the `idIndex` lookup at line 571 and the `!ok { continue }` skip at lines 572–574); the gap was a missing test case covering the absent-target scenario.

`needs-main` items: none. No `blocked_by` follow-ups were filed.

## Fix applied this pass

**Fixed:** Item 5 — added a regression test case to `TestApplyTransactionRulesRejectsCategoryFilterOnCategorization` in `budget-etl/main_test.go`: a `categorization` rule with `MatchCategory: "Food"` and `TransactionID: "doc-absent"` (absent from the `docIDs` slice), asserting `wantErr: true`. This locks in the validate-before-idIndex-lookup invariant against future refactors.

Commit: `ae767e83` — CI restarted; the chain re-QAs once CI passes.

natb1:
# QA Pass — PR #1681 / Issue #1371

**Attempt:** re-QA after fix attempt #1  
**Browser component:** None (pure backend Go change)

## Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Full Go test suite passes (regression gate) | shell-command | PASS |
| 2 | Each category-filter field rejected on a transaction-specific categorization rule | shell-command | PASS |
| 3 | Category-filter fields on a transaction-specific `budget_assignment` rule do not error | shell-command | PASS |
| 4 | Validation runs before the transaction-index lookup (bad txn rule errors even when target absent) | shell-command | PASS |
| 5 | Shared helper is the single source of the error string (byte-for-byte parity) | shell-command | PASS |

No bugs found. No browser walkthrough (pure backend change). No residue.

## Detail

- `go test ./...` from `budget-etl/` passes all packages.
- `TestApplyTransactionRulesRejectsCategoryFilterOnCategorization` — all 5 subtests pass:
  - categorization + MatchCategory → error ✓
  - categorization + ExcludeCategory → error ✓
  - categorization + Category → error ✓
  - budget_assignment + all three fields → no error ✓
  - categorization + MatchCategory, target transaction absent → error ✓
- `validateCategoryFilterFields` is called at line 568 in `applyTransactionRules`, before the `idIndex` lookup, confirming unconditional validation.
- The error string `category-filter fields (matchCategory/excludeCategory/category) are only valid on budget_assignment rules, not categorization rules` appears exactly once in `main.go`.

Clean pass — no items deferred to office-hours, no follow-ups filed.

natb1:
## Review — #1371 (`budget-etl`: validate category-filter fields on transaction-specific categorization rules)

### Fixed

- **`convertExportRulesForTest` divergence from `convertExportRules`** (`budget-etl/cmd/patch/integration_test.go:105`) — the integration test helper `convertExportRulesForTest` was not updated when `validateCategoryFilterFields` replaced the inline check in `convertExportRules`, so the helper would silently accept category-filter fields on categorization rules that `convertExportRules` now rejects. Inlined an equivalent `validateCategoryFilterFields` call into the helper and added a divergence-risk comment. (fix commit: 61c30c63)

### Informational

- **`TestApplyTransactionRulesRejectsCategoryFilterOnCategorization` success case** (`budget-etl/main_test.go:282`) — the `budget_assignment` pass case sets all three category-filter fields simultaneously, whereas the matching `TestConvertExportRulesRejectsCategoryFilterOnCategorization` tests them individually (one per case). Either style is correct; noting the asymmetry for awareness. No change needed.

### Security

No attack-surface findings. The diff adds a validation helper and wires it into two call sites — no external inputs, no new permissions, no data exposure.

PR #1684, opened 2026-06-15
budget: round-trip the snapshot schema through a single collection registry

natb1:
Closes #1268

## What

The budget web app serializes its IndexedDB stores to a snapshot JSON on export
and reconstructs them on upload. Three code paths (export, load, IDB setup) each
hand-maintained their own parallel list of collections, so when one path omitted
a collection the others kept, data was silently lost on a round-trip. The live
instance of that trap: `statementItems` and `reconciliationNotes` were
**exported** but **dropped on upload** (`toParsedData` hardcoded them to `[]`,
and neither entity had a parse path), so a user who exported then re-uploaded
lost both collections with no error.

This PR introduces a single **collection registry** (`budget/src/collection-registry.ts`)
that pairs the three round-trip adaptors (`parseRaw` / `toIdbRecord` /
`toRawJson`) per data collection, and rewires export, load, and IDB setup to
derive their collection lists and types from it. Omitting a collection on one
side is now a **compile error** (or a test failure) rather than silent data
loss, and every exported collection — including the two previously
export-only ones — round-trips losslessly.

## How

- **`collection-registry.ts`** (added earlier in the branch) — a non-widening
  `defineCollection` identity helper forces each entry's three adaptors to chain
  (a mispaired adaptor is a compile error at that entry), and registry-derived
  mapped types (`CollectionIdbData`, `CollectionDomainData`, `CollectionRawData`,
  `DataStoreName`) become the single source of truth for the set of collections.
- **`statement-item.ts` / `reconciliation-note.ts`** (added earlier) — the two
  missing upload/parse paths (`RawStatementItem` / `RawReconciliationNote`,
  `parseRaw*`, `*ToIdbRecord`), with their `*ToRawJson` return types tightened
  from `object` to the specific `Raw*` type.
- **`idb.ts`** — `STORE_NAMES`, `StoreName`, `ParsedData`, and the
  `storeParsedData` write loop now derive from the registry; `meta` stays
  special-cased and last.
- **`upload.ts`** — `ParsedUpload` becomes `CollectionDomainData & { … }` (every
  collection a required key), and `parseUploadedJson` / `toParsedData` now wire
  `statementItems` and `reconciliationNotes` through instead of dropping them.
- **`export.ts`** — the `output` object is constrained by `CollectionRawData & { … }`,
  so omitting a collection from export is a compile error.
- **`export.test.ts`** — the hardcoded `etlStores` round-trip list is replaced by
  a registry-driven exhaustiveness test that asserts its fixture-map keys
  deep-equal `Object.keys(collectionRegistry)`, drives a non-empty fixture for
  every collection, and asserts no store is emptied by a round-trip — now
  covering `statementItems`, `reconciliationNotes`, and `reconciliationEvents`.
  A compile-time `AssertEqual<keyof ParsedData, DataStoreName | "meta">` guard
  surfaces registry drift as a tsc error.

## Implementation note

`RawOutput` (the upload *input* type) is deliberately kept hand-written with the
specific `Raw*` field types rather than derived from `RawOf<K>`. Eight
`*ToRawJson` adaptors still return the wide `object` type mid-migration, so a
derived input type would narrow `object` → a specific `Raw*` in the existing
`.map` callbacks and fail `tsc`. This matches the plan's intentionally-loose
input-type decision — real uploaded files legitimately omit collections, and the
round-trip guarantee comes entirely from the **required** `ParsedUpload` /
`ParsedData` return types, not the input type.

## Verification

- `npm run build --prefix budget` — green (`tsc` + vite). A collection omitted
  from `toParsedData`, `parseUploadedJson`, or the export `output`, or a
  mispaired registry adaptor, now fails here.
- `npx vitest run --root budget` — full suite green (50 files, 992 tests),
  including the new registry-driven round-trip test proving the two
  previously-dropped collections survive a full export → parse → store cycle.

### DIFF
--- a/budget/src/collection-registry.ts
+++ b/budget/src/collection-registry.ts
@@ -1,0 +1,77 @@
+/**
+ * Single source of truth for the budget snapshot collections.
+ *
+ * Each entry pairs the three adaptors for one IndexedDB store — parse a raw
+ * JSON record into the domain shape, convert the domain shape into the IDB
+ * record, and serialize the IDB record back to raw JSON. `defineCollection`
+ * forces those three to chain, so a mispaired adaptor is a compile error at
+ * the offending entry rather than a silent data loss on round-trip.
+ *
+ * The export/upload/idb paths derive their collection lists from this registry,
+ * so adding a store here is the single edit that wires it through every path.
+ *
+ * This module must NOT import from idb.ts — idb.ts imports the registry, and a
+ * back-import would create a cycle. The Idb/Domain/Raw types are derived from
+ * the adaptor signatures via ReturnType, not imported by name.
+ */
+import { parseRawTransaction, transactionToIdbRecord, transactionToRawJson } from "./entities/transaction.js";
+import { parseRawBudget, budgetToIdbRecord, budgetToRawJson } from "./entities/budget.js";
+import { parseRawBudgetPeriod, budgetPeriodToIdbRecord, budgetPeriodToRawJson } from "./entities/budget-period.js";
+import { parseRawRule, ruleToIdbRecord, ruleToRawJson } from "./entities/rule.js";
+import { parseRawNormalizationRule, normalizationRuleToIdbRecord, normalizationRuleToRawJson } from "./entities/normalization-rule.js";
+import { parseRawStatement, statementToIdbRecord, statementToRawJson } from "./entities/statement.js";
+import { parseRawStatementItem, statementItemToIdbRecord, statementItemToRawJson } from "./entities/statement-item.js";
+import { parseRawReconciliationNote, reconciliationNoteToIdbRecord, reconciliationNoteToRawJson } from "./entities/reconciliation-note.js";
+import { parseRawAccount, accountToIdbRecord, accountToRawJson } from "./entities/account.js";
+import { parseRawJournalEntry, journalEntryToIdbRecord, journalEntryToRawJson } from "./entities/journal-entry.js";
+import { parseRawJournalLeg, journalLegToIdbRecord, journalLegToRawJson } from "./entities/journal-leg.js";
+import { parseRawReconciliationEvent, reconciliationEventToIdbRecord, reconciliationEventToRawJson } from "./entities/reconciliation-event.js";
+import { parseRawWeeklyAggregate, weeklyAggregateToIdbRecord, weeklyAggregateToRawJson } from "./entities/weekly-aggregate.js";
+
+/**
+ * Identity helper that infers the domain (D), IDB (I), raw-in (RIn), and
+ * raw-out (ROut) types from the three adaptor functions and forces them to
+ * chain. The two links the registry guarantees:
+ *   - parseRaw output type === toIdbRecord input type (both D)
+ *   - toRawJson input type === toIdbRecord output type (both I)
+ *
+ * RIn and ROut are kept as separate type parameters on purpose. Several
+ * `*ToRawJson` adaptors still return the wide `object` type mid-migration,
+ * which cannot satisfy a single shared R that is both a supertype of `object`
+ * (toRawJson's return) and a subtype of the specific `Raw*` parse input. Two
+ * params let each side infer independently without widening any entry to `any`.
+ */
+function defineCollection<D, I, RIn, ROut>(a: {
+  parseRaw: (raw: RIn, i: number) => D;
+  toIdbRecord: (d: D) => I;
+  toRawJson: (idb: I) => ROut;
+}): typeof a {
+  return a;
+}
+
+const collectionRegistry = {
+  transactions: defineCollection({ parseRaw: parseRawTransaction, toIdbRecord: transactionToIdbRecord, toRawJson: transactionToRawJson }),
+  budgets: defineCollection({ parseRaw: parseRawBudget, toIdbRecord: budgetToIdbRecord, toRawJson: budgetToRawJson }),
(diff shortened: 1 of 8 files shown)


natb1:
# QA Summary — #1268 (PR #1684)

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Compile-time enforcement: omitting a collection from export fails tsc | shell-command | PASS |
| 2 | Registry-driven round-trip test passes (no store emptied on re-upload) | shell-command | PASS |
| 3 | IDB schema unchanged: store list still matches the on-disk DB (no version bump needed) | shell-command | PASS |

**Counts:** 3 PASS, 0 FAIL, 0 SKIP

## Details

**Item 1 — `npm run build --prefix budget`:** tsc + vite build completed cleanly. The `CollectionRawData & {...}` annotation on `export.ts`, `CollectionDomainData & {...}` on `upload.ts`, and `CollectionIdbData & { meta }` on `idb.ts` all satisfy tsc — and the `AssertEqual<keyof ParsedData, DataStoreName | "meta">` compile-time guard in the test file holds.

**Item 2 — `npx vitest run --root budget`:** 50 test files, 992 tests, all passing. Includes `round-trip empties no registry store` which builds a non-empty IDB fixture for all 13 registry collections, asserts fixture keys equal `Object.keys(collectionRegistry)`, drives `exportToJson → parseUploadedJson → toParsedData`, and asserts every registry store survives non-empty. `statementItems`, `reconciliationNotes`, and `reconciliationEvents` (the #1268 fix) all round-trip correctly.

**Item 3 — version/collection count check:** `idb.ts` declares `version: 5` (line 37); `collection-registry.ts` has 13 `defineCollection(` calls. The on-disk DB schema is unchanged; an existing user's database opens without migration.

## No browser walkthrough required

All changes are to TypeScript data-model files (`collection-registry.ts`, `upload.ts`, `idb.ts`, `export.ts`, entity files). No React components, HTML templates, or CSS changed. All acceptance criteria are verifiable by compile and test suite.

natb1:
## Review — PR #1684

### Fixed

- **reconciliation-note.ts:123** — `parseRawReconciliationNote` rejected empty strings for the `note` field via `requireUploadString`, but Firestore and IDB tolerate empty notes. Switched to permissive validation so empty-string notes survive a round-trip. (sources: code-review, firebase, red-team, review) — fix commit 52fbc132
- **reconciliation-note.ts:121** — `entityId` was validated with `requireUploadId` but the error label used the entity name alone instead of the `entity[i].field` pattern used elsewhere. Switched to `requireUploadString` with consistent index-then-field labeling. (source: code-review) — fix commit 52fbc132
- **statement-item.ts:139** — `requireUploadId` for `statementItemId` emitted `statementItem.statementItemId[i]` instead of `statementItem[i].statementItemId`. Added an optional `field` param to `requireUploadId` and updated the call site for accurate field-path error messages. (source: code-review) — fix commit 52fbc132

### Informational

- **idb.ts:31** — `STORE_NAMES` is built from `Object.keys(collectionRegistry)` spread plus `'meta'`. `Object.keys()` returns `string[]` and the `as DataStoreName[]` cast is necessary but silences type errors if the registry gains a key that is not a `DataStoreName`. Acceptable under current constraints; worth revisiting if the registry diverges from the IDB store set.
- **upload.ts:55-56** — `statementItems` and `reconciliationNotes` are marked optional (`?`) in `RawOutput` while other collections are required. This reflects that real uploaded snapshots may legitimately omit these newer collections. Documented as intentional in the PR body.
- **reconciliation-note.ts:148** — `parseRawReconciliationNote` and `reconciliationNoteToIdbRecord` both set `groupId` to `null`. Since the snapshot format omits `groupId`, a re-uploaded snapshot will always produce null — consistent with the current app behavior but worth noting if groupId semantics change.
- **reconciliation-note.ts:65 / :154** — `updatedBy` holds a user email address and is included verbatim in the exported snapshot JSON. This is intentional (the field is user-authored), but means a snapshot file carries PII. No code change needed; user awareness is the mitigation.

### False-positive

- **auth-1 (reconciliation-note.ts:125–126)** — `updatedBy` is accepted verbatim from the uploaded file with no check against the authenticated user. This is a single-user personal finance app; there is no multi-user trust boundary where impersonation matters.
- **data-exposure-0 (_helpers.ts:24)** — `parseISOTimestamp` echoes the raw timestamp string in the validation error. Error messages are only shown to the uploading user themselves; no cross-user exposure path.
- **data-exposure-1 (_helpers.ts:161)** — `requireUploadEnum` echoes `JSON.stringify(value)` in the validation error. Same reasoning — error is surfaced only to the uploader.

### Out-of-scope

- **auth-0 (reconciliation-note.ts:123)** — `note` field has no upper length bound; a pathological upload could store a very large string. Pre-existing constraint, not introduced by this diff.
- **auth-2 (statement-item.ts:136–149)** — `amount` accepts any finite number including `Number.MAX_VALUE`. Pre-existing; not introduced by this diff.

### Deferred

- **export.ts:32** — `export.ts` still manually enumerates all 13 stores in the `Promise.all` and `output` object despite the collection registry existing. The compile-time `CollectionRawData` annotation catches missing `output` keys but not missing `getAll` calls. Tracked as #1697 (blocked_by #1268).
- **export.ts:32–47** — Same root issue from the review pass: `getAll` calls are a hand-written flat list not enforced against the registry; a new registry collection would need two manual edits. Tracked as #1696 (blocked_by #1268).

### Security

No open CodeQL alerts. No dependency changes (`deps=false`). No security findings require action.

Issue #473, opened 2026-04-07
PDF full-document search


natb1:
## Context

Sub-issue of #470. Implements the document search methods on the PDF renderer using
the pdf.js text content API.

## Requirements

- Implement search, goToResult, and clearSearch on the PDF renderer
- Use page.getTextContent() to extract and search text across all pages
- Return results with page number and surrounding context snippet
- goToResult navigates to the page and highlights the match in the text layer
- Active result highlight is visually distinct

## Acceptance criteria

- [ ] Searching returns matches from all pages of the PDF
- [ ] Each result shows page number and a context snippet
- [ ] goToResult navigates to the correct page
- [ ] The matched text is highlighted in the text layer on the target page
- [ ] clearSearch removes all highlights
- [ ] Works on multi-page PDFs with varying text layouts

Issue #472, opened 2026-04-07
Document search UI and ContentRenderer interface


natb1:
## Context

Sub-issue of #470. Adds full-document search to the info panel and extends the
ContentRenderer interface so renderers can opt in to search.

## Requirements

- Extend ContentRenderer with optional search methods:
  - search?(query: string): Promise<SearchResult[]> where SearchResult includes location, snippet, and page/chapter label
  - goToResult?(result: SearchResult): Promise<void>
  - clearSearch?(): void
- Add search input to the info panel (below navigation, above metadata)
- Display result count and scrollable list of results with context snippets
- Clicking a result calls goToResult and navigates the viewer
- Show search section only when the active renderer exposes search methods
- Clear results when search input is emptied or cleared

## Acceptance criteria

- [ ] ContentRenderer interface includes optional search/goToResult/clearSearch methods
- [ ] Info panel shows search input for renderers that implement search
- [ ] Searching displays a count and scrollable result list with context snippets
- [ ] Clicking a result navigates to that location in the document
- [ ] Clearing the input clears results and any highlights
- [ ] Search section is hidden for image archive viewer

PR #1688, opened 2026-06-15
print: EPUB full-document search

### LINKED ISSUE #474: EPUB full-document search
## Context

Sub-issue of #470. Implements the document search methods on the EPUB renderer using
the epubjs search and annotation APIs.

## Requirements

- Implement search, goToResult, and clearSearch on the EPUB renderer
- Use epubjs book search or spine-level search for text matching across chapters
- Return results with chapter label and surrounding context snippet
- goToResult navigates to the CFI location and highlights the match
- Active result highlight is visually distinct

## Acceptance criteria

- [ ] Searching returns matches from all chapters of the EPUB
- [ ] Each result shows chapter label and a context snippet
- [ ] goToResult navigates to the correct chapter and location
- [ ] The matched text is highlighted in the rendered content
- [ ] clearSearch removes all highlights
- [ ] Search works across chapter boundaries

natb1:
Closes #474

## What

Implements full-document search for EPUBs in the print viewer. The shared
document-search UI and the `ContentRenderer` `search` / `goToResult` /
`clearSearch` interface already landed in #472, but the EPUB renderer did not
implement those three methods, so the search box stayed hidden for EPUBs. This
PR adds them, so searching works across all chapters of an EPUB.

The PDF counterpart (#473) is a separate open sibling and is out of scope here.

## How

- **`search(query)`** iterates `book.spine.spineItems`, loading each section
  (`section.load` / `section.find` / `section.unload` in a `finally`) so matches
  are found across every chapter. Results map epubjs's `{ cfi, excerpt }` to the
  `SearchResult` shape: the range CFI becomes `location`, the
  whitespace-collapsed excerpt becomes `snippet`, and `matchStart` / `matchLength`
  are computed (case-insensitive) so the `SearchResult` invariant always holds.
  Chapter labels come from `book.navigation.get(section.href)?.label`, falling
  back to `Ch. <n>`.
- After iterating, a **monotonic epoch guard** discards a slow earlier search
  superseded by a faster later one before it touches highlight state — full-document
  search loads every spine section and routinely outlasts the UI's 300ms debounce
  (and the Enter-key handler bypasses the debounce entirely). The current search
  then clears prior highlights and runs a single synchronous base-highlight burst.
- **`goToResult(result)`** navigates via `rendition.display(cfi)` (awaiting the
  relocated event through the existing `waitForRelocated()` helper) and promotes
  the target match to a visually distinct active highlight, demoting the previous
  active match back to base.
- **`clearSearch()`** removes all tracked annotation highlights and resets state.
- Highlight colors come from the epubjs annotation `styles` param (SVG
  `fill` / `fill-opacity`), so the base-vs-active distinction needs no CSS change.
- Narrow local type casts cover the epubjs members missing from the package's
  `.d.ts` (`spineItems`, `Section.find/load/unload`, `rendition.annotations`,
  `book.load`, `book.navigation.get`), following the existing `spine.length`
  precedent — no `any`.

## Tests

`print/test/viewer/epub.test.ts` extends the epubjs mock (fake sections with
`find`/`load`/`unload`, `navigation.get`, and `rendition.annotations`) and covers:
matches from all spine sections, chapter labels including the `Ch. N` fallback,
snippet/`matchStart`/`matchLength` invariants (including a `"..."`-padded excerpt
and a case-insensitive match), the base-highlight burst, the empty-query
short-circuit, `goToResult` active-highlight promotion/demotion, `clearSearch`
teardown and no-op-on-second-call, and the overlapping-search epoch race (only
the latest query's highlights survive).

## Out of scope / follow-up

Capping the result/annotation count for degenerate common-term queries (e.g.
"the" → thousands of annotations in one burst — jank, not a crash, and only
in-view highlights actually render). That is a shared-UI product decision
spanning all renderers, not this issue's scope.

## Manual / QA verification

Two behaviors only real epubjs exercises (the unit tests cannot reach them):

1. **Deferred annotation injection** — `annotations.highlight(cfi)` on a
   not-yet-rendered section is stored by epubjs and injected when that section is
   later displayed. Search a term matching in a later chapter, navigate there,
   confirm the highlight appears.
2. **No stale highlights under the race** — search a common term in a
   multi-chapter EPUB, type fast / press Enter repeatedly, then clear the query,
   and confirm zero stale in-content highlights remain.


natb1:
## QA pass — issue #474 (EPUB full-document search)

**Attempt:** 0 of 2 · **Auto-fix applied:** yes · **Needs human:** yes (1 item)

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Unit tests pass (search/goToResult/clearSearch) | PASS |
| 2 | Search returns matches across all chapters | PASS |
| 3 | Each result shows chapter label and context snippet | PASS |
| 4 | Active result highlight visually distinct from base | needs-human |
| 5 | goToResult navigates to correct location | PASS |
| 6 | clearSearch removes all highlights | PASS |
| 7 | Epoch guard: superseded search leaves no stale highlights | PASS |

### Auto-fix applied

**`waitForRelocated` timeout too short for large EPUB sections.**
During the walkthrough of Confessions of St. Augustine (`gutenberg-3296`), `waitForRelocated` timed out 11 times in ~8 minutes, calling `reportError` on every navigation (next/prev/TOC/goToResult). The 5s timeout is too short for rendering large EPUB sections. Increased to 30s — still guards against genuine hangs.

Commit: a5425063

### Needs human judgment

**Item 4 — Active vs base highlight color distinction.**
DOM confirms both fills are present: active `#f59e0b` (fill-opacity:0.5) and base `#fde68a`. Whether the amber shade difference is perceptible enough for typical users is a subjective UX call. Escalating to office-hours for human review.

---
*QA pass by dispatch:qa-fix — escalating to office-hours for item 4.*

natb1:
## Review: #474 EPUB full-document search

### Fixed

- **CR-3** `print/src/viewer/epub.ts:362` — `destroy()` did not reset `_searchEpoch`. If a renderer is destroyed mid-search, a subsequent `search()` call on the same instance could start at a non-zero epoch, allowing stale-epoch comparisons to pass. Added `_searchEpoch = 0` to `destroy()` alongside the existing `_searchCfis` / `_activeCfi` resets. (commit 33b35742)
- **CR-4** `print/src/viewer/epub.ts:312` — Loader binding used an awkward double-cast through `unknown` to reach `.bind`. Simplified to cast `book.load` directly as `(this: Book, req: unknown) => Promise<unknown>` before calling `.bind(book)`, which is clearer and still type-safe. (commit 33b35742)

### Informational

- **CR-1** `print/src/viewer/epub.ts:102` — The 5s→30s `waitForRelocated` timeout change is not verified at the exact interval in tests. The test that exercises the timeout path was updated from `advanceTimersByTime(5000)` to `30000`, confirming the test passes, but it validates non-hang behavior rather than the exact duration. Low-value to add a dedicated interval assertion.
- **CR-2** `print/src/viewer/epub.ts:80-84` — The `clearSearchHighlights` guard `if (_activeCfi && !_searchCfis.includes(_activeCfi))` is dead code in practice: `goToResult` does not remove the CFI from `_searchCfis`, so `_activeCfi` will always be in `_searchCfis` when set. The guard protects a state combination the current API cannot produce. No harm, no action needed.
- **RV-1** `print/src/viewer/epub.ts:330-345` — The epoch guard looks like it could allow `clearSearchHighlights` from a stale search to wipe a newer search's highlights. In practice, both the highlight burst and `clearSearchHighlights` are synchronous JS after the async iteration, so they cannot interleave. The race exists only in async-mental-model reasoning, not in the execution model.

### False-positive (security)

- **SEC-1** `print/src/viewer/epub.ts:323` — Query string passed to `epubjs section.find()` without sanitization. `section.find` is a pure in-memory text comparison with no DOM write or eval path. No injection risk.
- **SEC-2** `print/src/viewer/epub.ts:339` — CFI values from `section.find()` passed to `annotations.highlight()` without validation. In a client-side EPUB reader loading local files, an attacker controlling the EPUB already has full file-system read access and can inject via the EPUB's own HTML/CSS — the CFI path does not escalate capabilities. False positive; noted as a defense-in-depth observation.
- **SEC-3** `print/src/viewer/epub.ts:44` — Monotonic epoch counter flagged as predictable to an in-page attacker. The epoch is a UI race guard, not a security primitive; it provides no guarantees an adversary needs to subvert. False positive.

### Deferred

_None._

### Out-of-scope (security)

_None._ (CodeQL: no open alerts on this PR ref. npm: no dependency changes.)

PR #1690, opened 2026-06-15
budget-etl: data-driven virtual transactions + budget upsert

### LINKED ISSUE #321: budget-etl: configurable virtual transaction generation
## Context

`generateVirtualSynchrony` in `main.go` hardcodes institution (`pnc`), account (`5111`), description match (`SYNCHRONY`), category (`Transfer:CardPayment`), output category (`Pet:Veterinarian`), and budget (`pet`). The companion `computePetBudget` and `appendPetBudgetIfNeeded` are similarly coupled to this one account.

Adding a second virtual account source would require duplicating the function. These values are configuration masquerading as code.

Identified during code quality review of #310 (Finding 9, iteration 3).

## Requirements

Define virtual transaction generation as data-driven configuration rather than compiled Go code. The configuration should specify:

- Source filter criteria (institution, account, category, description pattern)
- Output fields (category, budget, institution, account)
- Budget computation parameters (allowance derivation from matched transactions)

## Acceptance criteria

- [ ] Virtual transaction generation is driven by configuration (JSON export format or similar), not hardcoded Go functions
- [ ] `generateVirtualSynchrony`, `computePetBudget`, and `appendPetBudgetIfNeeded` are replaced by a generic implementation
- [ ] Existing Synchrony/pet behavior is preserved (same output for same input)
- [ ] Adding a new virtual account source requires only configuration changes, not code changes
- [ ] ETL tests cover the generic implementation with at least two different virtual account configurations

natb1:
Closes #1277
Closes #321

## What this does

`budget-etl` baked one user's PNC/5111 → Synchrony pet-vet flow into the binary:
`generateVirtualSynchrony`, `computePetBudget`, and `appendPetBudgetIfNeeded`
hardcoded the source filter, the virtual-transaction output, and the pet
allowance, plus a `"synchrony-virtual-"` statement-prefix filter. Adding a new
virtual source required editing Go code, and `appendPetBudgetIfNeeded`
early-returned when a `pet` budget already existed — so a seeded allowance froze
forever and never tracked changing inputs (the bug).

This replaces all of that with a generic, data-driven implementation. Virtual
transactions are now generated from `virtualTransactionRules` carried in the
snapshot, and budgets are upserted (not frozen). Adding a new virtual source now
needs only configuration — no code.

This satisfies the union of two issues — #1277 (frozen-allowance fix) and #321
(genericity bar: a new virtual source needs only config; tests cover ≥2
configs). The union design meets both, so the PR closes both.

## Design

- **`export.VirtualTransactionRule`** (new) drives generation: a source filter
  (`institution`/`account`/`category`/`descriptionContains`, empty field =
  match-any), an output shape (`target*`), and budget-seeding fields
  (`budgetId`/`budgetName`/`allowancePeriod`/`rollover`). Added to
  `Output.virtualTransactionRules`.
- **`generateVirtualTransactions`** generalizes `generateVirtualSynchrony`: loops
  over rules × transactions, skips already-virtual source rows, dedups by
  `(date, amount)` per rule, and emits virtual transactions + one zero-balance
  virtual statement per (rule, period).
- **`upsertVirtualBudgets`** replaces `computePetBudget` + `appendPetBudgetIfNeeded`:
  same monthly-average math (allowance values are identical), but matches by
  `budgetId` and **overwrites** an existing budget's allowance instead of
  early-returning — this is the frozen-allowance fix. Rules with no matched
  transactions are skipped.
- **`virtualStmtPrefix`** is the single source of truth for the statement-ID
  prefix (`targetInstitution + "-" + targetAccount + "-"`), used by both the
  generator and the `runInputJSON` drop-filter so they cannot drift. With the
  legacy values it reproduces `"synchrony-virtual-"` exactly.

## Migration (brownfield)

No snapshot in the wild carries `virtualTransactionRules` yet, so behavior must
be preserved. Both ETL paths seed-on-nil: when `inp.VirtualTransactionRules ==
nil` (absent in JSON), they seed the legacy Synchrony→pet rule, log the seed
once, and write the seeded rules back into the output. The seed is the only place
the legacy `pnc`/`5111`/`SYNCHRONY`/`synchrony`/`pet` constants survive, and it
fires at most once per snapshot (the write-back makes the field non-nil).

The `virtualTransactionRules` field deliberately has **no** `omitempty`: the
design needs three distinguishable states — absent (`nil`) seeds; explicit `[]`
disables generation permanently; populated is used as-is. `omitempty` would omit
a `len==0` slice, collapsing `[]` back to `nil` on round-trip and re-firing the
seed. Dropping it preserves the sentinel. (This reconciles Unit 1's initial
`omitempty` with the design's stated three-state sentinel; resolved in favor of
the design.)

The statement-filter generalization is asymmetric: only `runInputJSON` filters
virtual input statements (dropping just those matching a rule's
`virtualStmtPrefix`, retaining non-re-derivable monthly anchors). `runMerge`'s
`mergeStatements` already drops all virtual input statements unconditionally and
re-appends this run's freshly generated ones, so it gets **no** prefix logic.

## Downstream

`budget/src/upload.ts` parses the snapshot with `JSON.parse` + destructuring and
ignores unknown top-level fields; `export.ReadFile` is likewise permissive.
Adding `virtualTransactionRules` is backward compatible — no version bump, no app
change.

## Tests

`TestGenerateVirtualTransactions` covers ≥2 configs: the legacy Synchrony→pet
rule (asserting parity — exact statement IDs, transaction IDs, periods, target
fields, budget index) and a second distinct rule proving genericity, plus
per-rule dedup, skip-virtual-source, and match-any-empty-field. `TestUpsertBudget…`
covers the frozen-allowance overwrite, insert-when-absent, and skip-when-no-txns.
The legacy tests they replace were removed in the final unit.

## Verification

`go build ./... && go vet ./... && go test ./...` green after every commit
(coexist-then-delete ordering). The manual end-to-end check over a real
pre-change snapshot (confirm the seed fires once, virtual transactions/statements
are identical, the allowance updates, and the seed does not re-fire) is QA-phase
work — it needs the encrypted-snapshot password, out of scope for implement.


natb1:
# QA Summary — PR #1690 (issue #1277)

## Results

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Full Go test suite passes | shell-command | PASS |
| 2 | Build and vet are clean | shell-command | PASS |
| 3 | TestGenerateVirtualTransactions (AC: rule-driven generation) | shell-command | PASS |
| 4 | TestUpsertBudgetUpdatesExisting (AC: allowance updates) | shell-command | PASS |
| 5 | Hardcoded functions removed; generic replacements present | shell-command | PASS |
| 6 | VirtualTransactionRule type and Output field exist | shell-command | PASS |
| 7 | Seed-on-nil preserves legacy behavior | deferred | needs-human-judgment |
| 8 | End-to-end parity on real pre-change snapshot | deferred | needs-human-judgment |

**Counts:** PASS: 6 · FAIL: 0 · SKIP: 0 · Deferred: 2

No browser walkthrough (pure Go backend change — no frontend component).

## Deferred to office-hours

The following items require human judgment or credentials unavailable to automation:

- **Item 7: Seed-on-nil behavior on a real legacy snapshot** — verifying that `virtualTransactionRules == nil` seeds the legacy Synchrony→pet rule exactly once and writes it back requires running `budget-etl` over a real pre-change encrypted snapshot (at `/mnt/g/My Drive/budget`) using the GPG-protected `BUDGET_ETL_PASSWORD`.

- **Item 8: End-to-end parity and allowance-update on a real pre-change snapshot** — confirming that virtual transactions/statements are identical to pre-change behavior and that the frozen-allowance bug is fixed in production requires the real encrypted snapshot and the GPG credential.

## Disposition triage

| # | Title | Class | Verify verdict |
|---|-------|-------|----------------|
| 7 | Seed-on-nil preserves legacy behavior | needs-human | **refuted** (3/3 skeptics) |
| 8 | End-to-end parity on real pre-change snapshot | needs-human | **upheld** (0/3 skeptics) |

**`needs-main` items:** none. No follow-ups were filed.

**Item 7 — refuted by skeptics:** All 3 skeptics independently concluded that seed-on-nil behavior is verifiable with a synthetic plaintext JSON fixture (no GPG credentials needed). The nil-vs-non-nil distinction is a JSON deserialization property: omitting `virtualTransactionRules` from the JSON → nil Go slice → seed fires; including it → non-nil slice → seed skipped. The existing test suite (`TestRunInputJSONPreservesDerivedAnchors`, `TestRunMerge`) already uses synthetic plaintext fixtures with the same infrastructure — a dedicated `TestSeedOnNil` test could cover this path without any production data. **Suggested follow-up:** add a `TestSeedOnNil` test covering the three-state sentinel (nil seeds, `[]` disables, populated is used as-is).

**Item 8 — upheld:** The `/mnt/g/My Drive/budget` path is not mounted in the automated environment, and the GPG credential requires interactive pinentry. Running `budget-etl` over the actual production `.benc` snapshot cannot be done without human involvement.

**Terminal behavior:** Items 7 and 8 escalate to office-hours. The manual end-to-end check (item 8) requires the user to warm the GPG agent and run the ETL over their real snapshot. Item 7 (seed-on-nil) could be added as an automated test in a follow-up.

natb1:
## QA update — items 7 & 8 resolved (supersedes the prior escalation)

My earlier comment escalated items 7 and 8 to office-hours. That is no longer the disposition — both are resolved per user direction.

**Item 7 — Seed-on-nil migration (resolved with a new test).** The disposition skeptics were unanimous (3/3 refuted the needs-human call): the seed-on-nil behavior is verifiable with a synthetic plaintext fixture — no encrypted snapshot or GPG credential needed. Added `TestRunInputJSONSeedsVirtualRulesOnNil` (commit `2105ad20`), which drives the real `runInputJSON` migration sentinel end-to-end and asserts the three distinguishable states:
- `nil` (field absent) → seeds the legacy Synchrony→pet rule **once**, applies it (generates the virtual transactions + the `pet` budget), and writes the seeded rule back into the output;
- seeded output fed back in → **no re-seed**, rule count and virtual generation stable;
- explicit `[]` → generation stays disabled, nothing seeded.

CI is green on the new commit (`go-tests` pass).

**Item 8 — End-to-end parity on a real pre-change snapshot (skipped per user direction).** This check requires running `budget-etl` over the user's real encrypted `.benc` snapshot, which needs the unmounted `/mnt/g/My Drive/budget` share and the interactive GPG-protected `BUDGET_ETL_PASSWORD`. It cannot be automated and was skipped at the user's request.

**Result:** QA concludes with `dispatch:qa-done`. The escalation set is empty; the issue is not parked.

natb1:
## Review — PR #1690

### Fixed

- **F1** `budget-etl/main.go:1150-1180` — Duplicate migration seed block in `runMerge`: extracted both inline seed blocks into a shared `seedLegacyVirtualTransactionRules` helper, called from both `runInputJSON` and `runMerge`. Fix commit: 8f2b5a37.

### Informational

- **I1** `budget-etl/main.go:418` — Approximate month calculation uses `hours / (24 * 30.44)`. This is the same approximation as the legacy `computePetBudget` and intentional for a monthly-average heuristic; no change needed.
- **I2** `budget-etl/main.go:450-484` — The `idx` map in `upsertVirtualBudgets` is built once from the input `budgets` slice and mutated in-place as rules append new entries. This is correct and intentional; no reset between rules is needed.

### Dismissed

- **S1** `budget-etl/main.go:295-307` — Unconstrained match-any empty-field filter: a misconfigured rule with all-empty filters would match every transaction. This is a runtime misconfiguration risk, not a code vulnerability — virtual source rows are already skipped to prevent feedback loops, and rule validation is the caller's responsibility.
- **S2** `budget-etl/main.go:425` — Snapshot deserialization without strict field validation: the snapshot is a trusted, user-controlled local file, not an external or network input. Field validation at this boundary is not a security concern.

### False-positive (security)

- **S3** `budget-etl/main.go:498-511` — Migration seed constants (`pnc`/`5111`/`SYNCHRONY`) flagged as hardcoded sensitive values: these are bank account routing identifiers, not credentials. The design explicitly states the seed block is the only surviving location for the legacy values.

### Deferred

- **D1** `budget-etl/main.go:491-513` — MIGRATION comment lacks a concrete removal condition: the `MIGRATION: deletable once all snapshots carry virtualTransactionRules` comment has no date or explicit trigger, so the migration code could persist indefinitely. Filed as #1693.

PR #484, opened 2026-04-07
Document search UI and ContentRenderer interface

natb1:
## Summary

- Extend ContentRenderer with optional `search`, `goToResult`, and `clearSearch` methods plus a `SearchResult` type
- Add search input to the viewer info panel (between navigation and metadata) with debounced input, result count, scrollable result list with highlighted snippets, and click-to-navigate
- Search section hidden by default, shown only when the active renderer exposes search methods
- Arrow key page navigation suppressed while search input is focused
- Unit tests (20 new), shell integration tests, E2E acceptance test, and smoke test

Closes #472

PR #483, opened 2026-04-07
PDF text layer overlay for selectable text and browser find

natb1:
## Summary
- Render a pdfjs TextLayer over the canvas so browser Ctrl+F finds text and users can select/copy passages
- Wrap canvas + text layer div in a positioned wrapper; text layer uses CSS-scale viewport for alignment
- Narrow fullscreen tap zones from 50% to 20% so center area allows text selection
- Add acceptance tests for text layer rendering, page navigation updates, tap zone width, and image archive exclusion
- Add smoke test verifying text layer presence on PDF viewer pages

Closes #471

PR #1691, opened 2026-06-15
print: PDF full-document search

### LINKED ISSUE #473: PDF full-document search

natb1:
Closes #473

Implements full-document search for PDFs in the print viewer by adding the
three renderer methods the existing search UI gates on — `search`,
`goToResult`, and `clearSearch` — to `print/src/viewer/pdf.ts`. The document
search UI (PR #484) and the PDF text layer (PR #483) already landed; until now
the search panel stayed hidden for PDFs because `print/src/viewer/pdf.ts` did
not implement these methods. This is the last piece needed for PDF
full-document search.

## How it works

- **`search(query)`** walks every page via pdf.js `getPage(i).getTextContent()`,
  reconstructs each page's text with a single load-bearing rule —
  `items.map(i => "str" in i ? i.str : "").join("")`, an empty join, no injected
  separators — and returns `SearchResult`s with a `Page N` label, a 35-char
  context snippet, and a `page:offset:length` location token. Page text is cached
  in a module-scoped `Map` so debounced keystrokes don't re-fetch. Results are
  capped at 200 (documented in code) to avoid an unbounded list.
- **`goToResult(result)`** decodes the location, navigates to the page via the
  internal `renderPage` (not public `goToPage`), and arms a `pendingHighlight`.
  The highlight is applied as the final step of `renderPage`, under the same
  `renderGen` generation-token guard that already serializes concurrent renders
  (#1162/#1383) — so a fast result-to-result navigation never annotates a stale
  page. The highlight offset is mapped back onto the text-layer divs using the
  **same** empty-join rule as `search()`, walking pdf.js's index-aligned
  `textContentItemsStr` / `textDivs` getters, then wrapping the matched
  substring in a `<span class="search-highlight active">` via DOM nodes (no
  `innerHTML`, so the text layer's transparent-text CSS still applies). The
  pending highlight intentionally survives a resize re-render so it re-applies.
- **`clearSearch()`** actively unwraps the highlight spans (the user can clear
  the query without navigating, so no re-render fires to clean them up) and
  disarms `pendingHighlight`. Manual navigation (`goToPage`/`next`/`prev`) and
  `destroy()` also unwrap; `destroy()` clears the page-text cache.

## Highlight mechanism

The highlight span keeps `color: transparent` (inherited from the text layer)
so it shows a visible background over the canvas glyphs without double-rendering
text. The active match uses a distinct stronger color.

## Known limitation (intentional, not a bug)

The empty-join rule means a query spanning two text items or lines loses the
separator — e.g. `"the"` + `"quick"` is searchable only as `"thequick"`. This
mirrors how pdf.js's own `textContentItemsStr` concatenates, and keeping the
search side and the highlight-mapping side byte-identical is what guarantees
highlights land on the right characters. Cross-item separator handling is out of
scope for this issue.

## Tests

- **Unit** (`print/test/viewer/pdf.test.ts`, new): the pure helpers
  (`buildPageText`, `findMatches`, `buildSnippet`, `encode`/`decodeLocation`,
  `offsetToDivRanges` incl. spanning-div mapping); `search()` against a mocked
  `pdfjs-dist` returning multi-page matches with correct labels and location
  round-trips; `clearSearch()` removing the highlight DOM via a full
  `goToResult` simulation. 30 tests, all green.
- **e2e** (`print/e2e/viewer.spec.ts`): on the seeded 3-page `plato-republic`
  PDF — search UI un-hides, a query returns matches from more than one page,
  clicking a result navigates to the right page and shows an active
  `.search-highlight.active` in the text layer, and clearing the query removes
  the highlights.


natb1:
## QA — issue #473 (PDF full-document search)

**Acceptance tests:** CI PASS (pre-QA gate) ✅

### Items

| # | Item | Result |
|---|------|--------|
| 1 | Search panel accessible on PDF | ✅ PASS |
| 2 | Multi-page results with labels | ✅ PASS |
| 3 | Click result navigates to page | ✅ PASS |
| 4 | Active highlight visible after navigation | ✅ PASS |
| 5 | Active highlight visually distinct | ⏸ DEFERRED (needs-human-judgment) |
| 6 | Clear search removes highlights | ✅ PASS |
| 7 | EPUB search panel hidden | ✅ PASS |
| 8 | CSS selectors present | ✅ PASS |

### Bug fixed during QA (commit c2b5dce3)

**Spread mode: navigation + highlight** — two bugs found and fixed:

1. `shell.ts` `onNavigate` callback called `controller.enter()` on every search navigation, creating duplicate spread-page containers. Fixed to call `controller.goToPage(renderer.currentPage)`.

2. `pdf.ts` `renderPageInto()` rendered spread text layers without applying `pendingHighlight`. Fixed to apply `pendingHighlight` after the text layer renders, mirroring the existing logic in `renderPage()`.

Verified: left-page highlight (page 1 solo spread), right-page highlight (page 3 in pages 2–3 spread), clear search removes spread highlights, spread toggle off→on works cleanly.

### Deferred to office-hours

- Item 5: Whether the active highlight is visually distinct enough — subjective UX call.

natb1:
## Review — PR #1691

**Surface:** `code` (5 files: `pdf.ts`, `shell.ts`, `viewer.css`, `viewer.spec.ts`, `pdf.test.ts`)
**Security:** No open CodeQL alerts. No dependency changes (deps audit skipped).

> **Partial coverage:** The red-team security finder failed (rate limited); that domain was not reviewed. The classify agent timed out; all findings defaulted to Deferred.

---

### Fixed

_None._ No code changes applied — all in-scope findings are out-of-scope for this PR (see Deferred).

### Required (security)

_None._

_None._

### Informational

_None._

### Dismissed

_None._

### Deferred

All findings are real issues in code this PR adds, but each is a follow-up improvement rather than a pre-ship blocker. Filed as `blocked_by #473` tracking issues:

| Finding | Location | Follow-up |
|---|---|---|
| `goToResult()` re-arms `pendingHighlight` without calling `unwrapHighlights()` first — stale detached DOM nodes accumulate in `highlightRestores` on repeated result clicks | `pdf.ts:577` | #1717 |
| `findMatches` stores `query.length` instead of `needle.length` — highlight span misplaced for Unicode queries where lowercasing changes code-unit length | `pdf.ts:42` | #1718 |
| Spread mode: clicking a result fires two full page renders (one via `goToResult`'s internal `renderPage`, one via the `onNavigate` callback's `controller.goToPage`) — wasted render hidden by CSS | `shell.ts:444` | #1719 |
| E2E clear-query assertion fires without `poll` — may race the async `clearSearch` path under slow conditions | `viewer.spec.ts:47` | #1720 |
| `applyHighlight` reads `div.textContent` before calling `unwrapHighlights` — records already-span-wrapped content as `originalText`, corrupting restore on double-highlight | `pdf.ts:319` | #1721 |
| `pdfDoc!` non-null assertion in `search()` loop — a concurrent `destroy()` between `await`s nulls `pdfDoc` and throws `TypeError` | `pdf.ts:362` | #1722 |
| `.textLayer .search-highlight` CSS rule (no `.active`) is dead — `applyHighlight` always sets `className = 'search-highlight active'`; the non-active rule is never matched | `viewer.css:132` | #1723 |
| E2E test hardcodes `plato-republic` — if it is a `testOnly` seed item, the test fails in environments that don't seed `testOnly` items | `viewer.spec.ts:10` | #1724 |
| `search()` missing early null guard for `pdfDoc` — calling before `init()` resolves or after `destroy()` hits the `pdfDoc!` assertion | `pdf.ts:518` | #1725 |
| `applyHighlight()` missing `unwrapHighlights()` call before mutating divs — second injection records already-wrapped content as original | `pdf.ts:311` | #1726 |
| `renderPageInto()` (spread mode) applies `pendingHighlight` without a gen-based race guard — two concurrent calls can both inject spans without an `unwrapHighlights()` between them | `pdf.ts:413` | #1727 |
| `onNavigate` callback calls `controller.goToPage()`, which unconditionally sets `pendingHighlight = null` — search-result highlights never visible in spread mode | `shell.ts:444` | #1729 |

### Out-of-scope (security)

_None._ No CodeQL alerts; no high/critical pre-existing npm advisories.

---

No code changes were applied in this review pass. All findings are deferred follow-ups filed as issues blocked by #473.

PR #1189, opened 2026-06-07
print: LocalFolderMediaSource — open a folder of PDFs/EPUBs as the library

natb1:
Closes #1035

## Summary

Adds an alternate, unauthenticated media source to `print`: a folder of the user's
own PDFs/EPUBs, read in place with no upload, unioned with the existing Firebase
Storage cloud library. Implements P1 (#1035) of the print local-first reading-library
epic (P).

The change is purely additive. The Firebase Storage + authenticated path stays the
default and behaviorally unchanged — its viewer resolution and download paths are
untouched. Only the library *listing* is routed through the new
`createFirebaseMediaSource`, which is behaviorally identical (`list()` calls the same
`getPublicMedia` / `getAllAccessibleMedia`).

## What changed

**`@commons-systems/mediautil`** — new `createLocalFolderMediaSource<T>`
(`mediautil/src/local-folder.ts`): a generic `MediaSource<T>` over a directory
handle. Lists and resolves top-level files in place, skips a file that disappears
mid-scan (never throws the whole scan), sorts newest-first, and throws a clear
`Local file no longer present` error when a resolved item is gone. Uses structural
handle types so it is unit-testable with a fake directory, decoupled from FSA lib
types.

**`print`** — the local-first library:
- `print/src/library.ts` — aggregation seam: builds the cloud `FirebaseMediaSource`,
  maps local files to `MediaItem`s tagged `origin: "local"` with id `local:<name>`,
  and binds the active `LocalFolderMediaSource`. `MediaItem` gains an optional
  `origin` field (`"cloud" | "local"`).
- `print/src/local-folder-ui.ts` + `home.ts` — a directory picker
  (`showDirectoryPicker`) and a permission flow persisted via the F1
  `FsaHandleStore`. A returning session needs no picker when permission is granted,
  or one click when in `prompt` state. Local items are unioned into the existing
  `#media-list` (user's own folder first), and a window-focus re-scan picks up
  files added/removed in the folder. On a non-Chromium (non-FSA) browser the local
  controls are absent and the cloud path is unchanged.
- `print/src/pages/view.ts` + `main.ts` — the viewer resolves local PDFs/EPUBs via
  `resolveLocalBlob`, forces localStorage reading position (no `local:` ids written
  to Firestore — the folder-synced sidecar is P2 #1036), and surfaces the existing
  `#view-error` message gracefully when a file has moved/disappeared. The cloud
  viewer path is byte-for-byte unchanged.

## Scope notes

[...]

PR #1711, opened 2026-06-15
print: make library.test.ts before-binding block order-independent

natb1:
Closes #1239

## What

Makes the `"local source — before binding"` describe block in
`print/test/library.test.ts` order-independent. Its four assertions
(`hasLocalSource()` is `false`, `listLocal()` returns `[]`, `getLocalItem()`
returns `null`, `resolveLocalBlob()` throws) all depend on the module-level
`localSource` singleton in `print/src/library.ts` being `null`. In declaration
order they run before the `"after binding"` block calls `createLocalSource(dir)`,
so they pass — but under Vitest `--shuffle` the after-binding block can run
first, leaving `localSource` bound and breaking every before-binding assertion.

## How

- `print/src/library.ts` — add a test-only `resetLocalSource()` that sets the
  module-private `localSource` singleton back to `null`, alongside the existing
  `localSource` accessors.
- `print/test/library.test.ts` — add `beforeEach(() => resetLocalSource())` as
  the first statement in the before-binding block so each assertion runs against
  a guaranteed-null source, regardless of suite execution order. Imports and the
  block's leading comment are updated to match.

`--shuffle` is not enabled in `print/vite.config.ts` today, so this is a
latent-correctness fix with no behavior change to shipped code — deferred from
PR #1189 review. The after-binding block is untouched (it binds its own source).

## Verification

`library.test.ts` passes both in declaration order and under
`--sequence.shuffle` (13/13 across multiple shuffle seeds):

```bash
npx vitest run --root print test/library.test.ts
npx vitest run --root print --sequence.shuffle test/library.test.ts
```


natb1:
# QA Summary — Issue #1239

**PR:** #1711  
**Result:** ✅ Clean pass — all items passed

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | library.test.ts passes in declaration order | shell-command | ✅ PASS |
| 2 | library.test.ts passes under shuffled execution order | shell-command | ✅ PASS |
| 3 | resetLocalSource is exported and resets the singleton to null | shell-command | ✅ PASS |

## Details

**Item 1 — Declaration order:** `npx vitest run --root print test/library.test.ts` → 13/13 passed

**Item 2 — Shuffle order:** `npx vitest run --root print --sequence.shuffle test/library.test.ts` → 13/13 passed (seed 1781568425647). The `"local source — before binding"` assertions (`hasLocalSource()` false, `listLocal()` returns `[]`, `getLocalItem()` returns null, `resolveLocalBlob()` throws) all pass regardless of execution order — confirming the fix's core correctness property.

**Item 3 — Mechanism check:** `resetLocalSource` exported at `print/src/library.ts:131`; `beforeEach(() => resetLocalSource())` wired at `print/test/library.test.ts:122`.

No bugs found. No residue. No deferred items.

natb1:
## Review (automated)

No code changes applied — all findings are informational or false-positive.

### Informational

- **`print/test/library.test.ts:144`** — The `"local source — after binding"` describe block binds a local source inside its `it()` but has no `afterEach` to clean up. With `--shuffle`, a later describe block could see bound state from a prior run. Low urgency since `--shuffle` is not enabled today, but worth a follow-up alongside this fix.
  _Sources: code-review, review_

- **`print/src/library.ts:131`** — `resetLocalSource()` nulls `localSource` only. The module also holds `localFolderReadyPromise` / `resolveLocalFolder` mutable state; if the after-binding block's promise setup ever races, those fields are not reset. Not triggered by current tests but a latent concern if the before-binding block is extended.
  _Sources: code-review, review_

### False-positive

- **`print/src/library.ts:131`** — `resetLocalSource()` is exported from a production module with no `import.meta.env.TEST` guard. Flagged by red-team / auth / data-exposure reviewers as a risk of callers nulling the live singleton. Ruled false-positive: the function is deliberately test-only (JSDoc says so), it ships in a library package that is not deployed as a server and has no privileged callers, and the existing `createLocalSource` / `hasLocalSource` exports follow the same pattern. Tree-shaking removes it from production bundles.
  _Sources: auth, code-review, data-exposure, red-team, review_

### Fixed

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1713, opened 2026-06-15
print: debounce local-folder focus rescan and serialize LocalFolderMediaSource scans

### LINKED ISSUE #1237: print: debounce focus handler and guard concurrent directory scans in LocalFolderMediaSource
`print/src/local-folder-ui.ts:63` and `mediautil/src/local-folder.ts:38`:

The `window` focus handler calls `scan()` with no debounce and no concurrent-scan guard. Rapid focus events on a large directory stack concurrent scans; each scan starts with `index.clear()`, causing a torn-index race where `resolveToBlob` incorrectly reports present files as missing mid-flight.

Two fixes needed:
1. Debounce the focus handler (300–500 ms) in `local-folder-ui.ts` so consecutive focus events are coalesced.
2. Serialize scans in `LocalFolderMediaSource.scan()` with a shared pending-scan promise: `if (!pendingScan) { pendingScan = doScan().finally(() => { pendingScan = null; }); } return pendingScan;`

Surfaced during security review of PR #1189; deferred because the fix requires concurrent design decisions.

natb1:
Closes #1237

## Summary

Fixes the torn-index race in the print library's "open a folder of PDFs/EPUBs"
feature, where rapid `window` focus events (e.g. alt-tabbing back to a large
directory) stacked concurrent directory rescans and could make `resolveToBlob`
spuriously throw `"Local file no longer present"` for a file that is actually
there. Two coordinated fixes, one per package.

## Unit 1 — Serialize scans behind a shared pending-scan promise (`mediautil`)

`LocalFolderMediaSource.scan()` rebuilds the `index` (id → handle) and `items`
(id → record) maps, beginning each pass with `index.clear(); items.clear()`.
Concurrent `scan()` calls meant one scan could wipe the index while another was
mid-population — the torn-index window `resolveToBlob` fell into.

The scan body is now an inner `doScan()`, and `scan()` guards it behind a shared
`pendingScan` promise so concurrent callers join the in-flight scan instead of
starting a competing one. The guard uses `.finally` (not `.then`) so
`pendingScan` resets on **both** resolve and reject — a failed scan cannot
permanently wedge the guard.

**Accepted tradeoff:** scan-on-miss in `metadata`/`resolveToBlob` now *joins* an
in-flight scan rather than forcing a fresh one. A brand-new file added *during* a
scan that predates it won't appear until the next rescan. This is by design: the
next focus rescan catches it, and it is strictly better than the torn-index
false-missing it replaces. Joined callers share the in-flight scan's outcome
(including rejection); the `.finally` reset guarantees no permanent wedge.

## Unit 2 — Debounce the focus-rescan handler (`print`)

`bindAndRender` in `print/src/local-folder-ui.ts` rebinds a `window` `"focus"`
listener on every folder bind, with no debounce. The handler now coalesces rapid
focus events behind a 400 ms trailing-edge debounce (`FOCUS_RESCAN_DEBOUNCE_MS`,
within the issue's 300–500 ms range), matching the existing inline
`setTimeout`/`clearTimeout` idiom in `print/src/viewer/shell.ts` (`scheduleSave`).
`focusCleanup` also clears any pending timer, so a rebind or teardown cannot leave
a queued timer that fires a stale rescan against a detached container.

## Verification

- `npx vitest run --root mediautil` — coalescing test (a focus burst yields a
  single `directory.values()` traversal) and torn-index test (`resolveToBlob`
  returns bytes for a present file while a `list()` scan is mid-flight) pass
  alongside the existing suite.
- `npx vitest run --root print` — debounce tests (a focus burst collapses to one
  rescan; no rescan before the debounce elapses; cleanup cancels a pending timer)
  pass alongside the existing suite (390 passing).
- `npm run build --prefix print` — type-check + bundle clean (`mediautil` is a
  source-only workspace package with no build step; `print` is its consuming
  type-checker).


natb1:
# QA — #1237 print: debounce focus handler and guard concurrent directory scans

## Items

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | mediautil scan serialization — coalescing and torn-index race | shell-command | ✅ PASS |
| 2 | print focus-handler debounce | shell-command | ✅ PASS |
| 3 | print package type-checks and bundles clean | shell-command | ✅ PASS |

**Counts:** 3 PASS · 0 FAIL · 0 SKIP

## Detail

**Item 1 — `npx vitest run --root mediautil`**
18 tests passed across 2 files. Includes the new coalescing test (concurrent `scan()` calls share a single pending-scan promise) and torn-index test (`resolveToBlob` returns bytes for a present file while a `list()` scan is mid-flight).

**Item 2 — `npx vitest run --root print`**
457 tests passed across 28 files. Includes all 3 new debounce tests: focus burst → single rescan, no rescan before 400 ms elapses, cleanup clears pending timer.

**Item 3 — `npm run build --prefix print`**
Build clean: 191 modules transformed. Warnings (dynamic import, chunk size) are pre-existing and not introduced by this PR.

## No residue — clean pass

natb1:
## Review — PR #1713 (#1237)

### Fixed

- **[print/test/local-folder-ui.test.ts:131]** `initLocalFolder — focus-rescan debounce` tests 1 and 2 did not call the cleanup returned by `initLocalFolder`, leaving a window focus listener registered after each test. Added `afterEach` in that describe block that calls `cleanup?.()` to remove the listener. (code-review)
- **[print/test/local-folder-ui.test.ts:147]** Debounce timeout was hardcoded as the literal `400` (and `399` for the boundary test) in three test assertions. Exported `FOCUS_RESCAN_DEBOUNCE_MS` from `local-folder-ui.ts` and imported it in the test; replaced magic numbers with `FOCUS_RESCAN_DEBOUNCE_MS` and `FOCUS_RESCAN_DEBOUNCE_MS - 1`. (code-review)
- **[print/test/local-folder-ui.test.ts:78]** The `initLocalFolder` describe block registered a window focus listener via `initLocalFolder` in each test with no cleanup, allowing listeners to accumulate and leak across tests. Added `afterEach` that invokes the cleanup returned by `initLocalFolder`. (review)
- **[print/src/local-folder-ui.ts:33]** `FOCUS_RESCAN_DEBOUNCE_MS` was not exported, forcing the test file to hardcode `400`/`399`. Exported the constant so tests stay correct regardless of future changes. (review)

Fix commit: 13f70f8e

### Informational

- **[mediautil/test/local-folder.test.ts:159]** The two new coalescing/torn-index tests do not have an `afterEach` teardown, but since `createLocalFolderMediaSource` is a pure in-memory object with no global side effects, no teardown is needed — this is acceptable. (code-review)
- **[mediautil/test/local-folder.test.ts:175]** The torn-index test covers only the happy path (file present, `resolveToBlob` succeeds). No test covers `resolveToBlob` when the scan itself rejects mid-flight. This is a valid coverage gap but out of scope for this PR's fix. (review)

### Security

No attack surface change to authentication, authorization, data exposure, or injection paths. CodeQL: no open alerts. Dependency audit: not applicable (no `package.json`/`package-lock.json` changes).

PR #1714, opened 2026-06-15
print: skip pure-extension filenames (e.g. .pdf) in fileToLocalItem

natb1:
Closes #1238

## What

`fileToLocalItem` in `print/src/library.ts` derived a local item's `title` from
the filename basename: `title = name.slice(0, idx)` where
`idx = name.lastIndexOf(".")`. For a file named *exactly* `.pdf` — a Unix dotfile
whose whole name is the extension — `idx === 0`, so the extension check passed
(`.pdf` is supported) but the title became the empty string and the library card
rendered with no visible title.

## Fix

Widen the basename guard from `if (idx < 0)` to `if (idx <= 0)`, folding the
pure-extension case (`idx === 0`) into the existing no-dot case (`idx === -1`).
A pure-extension filename is now skipped as unsupported (`null`), matching how
the function already handles a name with no dot (`README` → `null`) and an
unsupported extension (`notes.txt` → `null`). A comment marks the `<= 0` so a
future reader does not "simplify" it back to `< 0` and reintroduce the bug.

The doc comment is updated to say the function returns `null` for an
"unsupported or title-less filename." No change to `title` derivation — once the
guard skips `idx === 0`, the slice is only ever reached with `idx >= 1`, so it
always yields a non-empty title.

A pure-extension filename is not a real-world library entry; it was surfaced and
deferred during review of PR #1189.

## Test

Adds one case to the existing `fileToLocalItem` suite asserting `.pdf` and
`.epub` pure-extension names return `null`. All `fileToLocalItem` cases pass
(`npx vitest run --root print test/library.test.ts`): `book.pdf` → `book`,
`book.PDF` → pdf (case-insensitive), `notes.txt` → `null`, `README` → `null`.


natb1:
# QA Summary — Issue #1238

**Branch:** `1238-print-handle-pure-extension`  
**PR:** #1714  

## Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Pure-extension filename (.pdf) is skipped, no title-less card rendered | shell-command | PASS |
| 2 | Pure-extension .epub filename is also skipped | shell-command | PASS |
| 3 | Normal supported filename still yields a non-empty title | shell-command | PASS |
| 4 | Existing no-dot and unsupported-extension cases still return null | shell-command | PASS |
| 5 | Guard comment is present (prevents reverting `<= 0` to `< 0`) | shell-command | PASS |

## Test Results

- `npx vitest run --root print test/library.test.ts` — **14/14 passed** (new pure-extension case + all existing `fileToLocalItem` cases)
- `npm run test --prefix print` — **537 passed, 1 skipped** (full print test suite)
- `grep -n "idx <= 0" print/src/library.ts` — guard comment confirmed at line 62

## Verdict

Clean pass. All QA items verified by the test suite. No residue, no browser walkthrough needed (change is a pure TypeScript library function). No bugs found.

natb1:
## Review findings — PR #1714

### Informational

- **CR-001** `print/src/library.ts:62` — Comment style: inline comment on the guard line is appropriate for documenting this subtle invariant; no change required.
- **RV-001** `print/test/library.test.ts:96-102` — Test covers both `.pdf` and `.epub` in one `it` block; functionally identical cases. Could be split into two blocks for clarity, but current form is acceptable.

### Dismissed

- **RV-002** `print/src/library.ts:52` — Doc comment semicolon after "filename." is grammatically valid (joins two independent clauses). Not an actual issue.

### False-positive (security)

- **SEC-001** `print/src/library.ts, print/test/library.test.ts` — No security-relevant attack surface in this diff. The change is a pure guard-extension on a local client-side filename parser with no network I/O, no auth, no storage writes, and no injection vector.

### Fixed

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1736, opened 2026-06-15
print: drain stale highlights in goToResult() before re-arming

### LINKED ISSUE #1717: print: goToResult() does not call unwrapHighlights() before re-arming pendingHighlight
`goToResult()` sets `pendingHighlight` and calls `renderPage()` without first calling `unwrapHighlights()`. Every successive result click pushes new entries into `highlightRestores` via `applyHighlight()` without draining the entries from the previous call. `renderPage()` calls `textLayerDiv.replaceChildren()` which detaches the previously highlighted divs from the DOM, but `highlightRestores` still holds references to those detached nodes. Over repeated result clicks the array grows unboundedly, leaking detached DOM node references.

Contrast with `goToPage()`, `next()`, and `prev()`, which all call `unwrapHighlights()` before proceeding (lines 471-472, 479-480, 488-489 in `print/src/viewer/pdf.ts`).

**Reproduction steps:**
1. Open any multi-page PDF in the print viewer.
2. Enter a search query that produces multiple results.
3. Click the first result to navigate to it.
4. Click a second result — `highlightRestores` now holds the stale entries from step 3.
5. Repeat: each click accumulates additional detached node references.

**Expected behavior:** Each call to `goToResult()` drains stale `highlightRestores` entries before arming the new highlight, so the array never holds references to detached DOM nodes.

**Actual behavior:** `highlightRestores` grows without bound across result clicks, retaining references to detached nodes from every previous call.

**Recommended fix:** Add `unwrapHighlights()` at the top of `goToResult()` (before setting `pendingHighlight`), mirroring the pattern in `goToPage()`, `next()`, and `prev()`. This drains stale entries before the new highlight is applied.

Related: #1691

**Files:** `print/src/viewer/pdf.ts`

**Acceptance criteria:**
- [ ] `unwrapHighlights()` is called at the start of `goToResult()`, before `pendingHighlight` is set.
- [ ] After clicking two different search results in sequence, `highlightRestores` contains only the entries from the most recent result (not an accumulation from both).
- [ ] No detached DOM node references remain in `highlightRestores` between result navigations.
- [ ] Existing behavior of `goToPage()`, `next()`, `prev()`, and `clearSearch()` is unchanged.

natb1:
Closes #1717

## Problem

`goToResult()` in `print/src/viewer/pdf.ts` armed `pendingHighlight` and called
`renderPage()` **without first calling `unwrapHighlights()`**. Each result click
runs `applyHighlight()`, which pushes a `{ div, originalText }` entry into the
closure-private `highlightRestores` array. `renderPage()` rebuilds the text layer
via `replaceChildren()`, detaching the previously-highlighted divs from the DOM —
but `highlightRestores` kept referencing those detached nodes. Across repeated
result clicks the array grew without bound, leaking detached DOM node references.

The sibling navigation methods `goToPage()`, `next()`, and `prev()` all call
`unwrapHighlights()` before proceeding, draining stale entries. `goToResult()` was
the lone arming path that did not.

## Fix

Drain `highlightRestores` at the start of `goToResult()`, before the new
`pendingHighlight` is armed — mirroring the sibling methods.

The `unwrapHighlights()` call is placed **after** the `if (!decoded) return;`
guard (not before it), so a malformed/undecodable result — which performs no
navigation and no re-render — does not clear the currently-visible highlight. No
`pendingHighlight = null` is added: `goToResult` is *arming* a highlight (unlike
`goToPage`/`next`/`prev`, which *discard* highlights on manual navigation), so
only the drain is wanted. The existing `unwrapHighlights()` helper is reused
verbatim — no new helper.

## Test

Adds one regression test to the section-C `describe("clearSearch()")` block in
`print/test/viewer/pdf.test.ts`. The leak is of *detached* node references, not
directly observable via `container` queries — so the test holds a JS reference to
the first result's tracked text-layer div and asserts it is restored (its
`.search-highlight` span removed) when the second result is armed, and that
exactly one highlight remains live in the DOM. The test fails on the pre-fix
`goToResult` and passes with the drain.

The full `print` suite passes (454 passed, 1 pre-existing skip).

### DIFF
--- a/print/src/viewer/pdf.ts
+++ b/print/src/viewer/pdf.ts
@@ -597,6 +597,11 @@ export function createPdfRenderer(onError?: (err: unknown) => void): ContentRend
     async goToResult(result: SearchResult): Promise<void> {
       const decoded = decodeLocation(result.location);
       if (!decoded) return;
+      // Drain stale highlightRestores entries — they reference now-detached
+      // nodes from the previous result's text layer — before arming the new
+      // highlight, mirroring goToPage/next/prev. Without this the array grows
+      // unboundedly across result clicks, leaking detached DOM node references.
+      unwrapHighlights();
       _currentPage = decoded.page;
       // Arm the highlight before rendering; renderPage applies it as its final
       // gen-guarded step. Call the internal renderPage directly (like
--- a/print/test/viewer/pdf.test.ts
+++ b/print/test/viewer/pdf.test.ts
@@ -479,4 +479,26 @@ describe("clearSearch()", () => {
 
     expect(container.querySelector(".search-highlight.active")).toBeNull();
   });
+
+  it("goToResult() drains the previous result's highlight before arming the next", async () => {
+    const renderer = createPdfRenderer();
+    await renderer.init(container, "fake://source.pdf");
+
+    const results = await renderer.search!("the");
+    const r1 = results.find((r) => r.label === "Page 1")!;
+    const r2 = results.find((r) => r.label === "Page 2")!;
+
+    await renderer.goToResult!(r1);
+    // Capture the text-layer div that holds the first result's highlight span.
+    const div1 = container.querySelector(".search-highlight")!.parentElement!;
+    expect(div1.querySelector(".search-highlight")).not.toBeNull();
+
+    // Arming the second result must drain the first entry (unwrapHighlights),
+    // restoring div1 instead of leaking it in highlightRestores.
+    await renderer.goToResult!(r2);
+    expect(div1.querySelector(".search-highlight")).toBeNull();
+
+    // Exactly one highlight is live in the DOM (the current result).
+    expect(container.querySelectorAll(".search-highlight.active").length).toBe(1);
+  });
 });


natb1:
# QA Summary — Issue #1717

**PR:** #1736  
**Branch:** `1717-print-gotoresult-does-not-c`  
**App:** `print`

## Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Regression test passes — goToResult drains previous highlight before arming next | shell-command (vitest) | ✅ PASS |
| 2 | goToResult calls unwrapHighlights before arming pendingHighlight (ordering) | shell-command (file check) | ✅ PASS |
| 3 | Golden path — search navigation between results still works | browser walkthrough | ✅ PASS |

**Counts:** 3 PASS, 0 FAIL, 0 SKIP

## Details

### Item 1 — Regression test (vitest)

```
npx vitest run --root print print/test/viewer/pdf.test.ts
Test Files  1 passed (1)
Tests       31 passed (31)
```

All 31 tests pass including the new regression test:
`goToResult() drains the previous result's highlight before arming the next`

### Item 2 — Code ordering check

```
awk '/async goToResult/,/^    },/' print/src/viewer/pdf.ts | grep -n 'unwrapHighlights\|pendingHighlight ='
8:      unwrapHighlights();
13:      pendingHighlight = { ... };
```

`unwrapHighlights()` at line 8 — correctly precedes `pendingHighlight=` at line 13.

### Item 3 — Browser walkthrough (Phaedrus PDF, search "Socrates")

Searched "Socrates" in the Phaedrus viewer — 10 results returned. Clicked results 1 → 2 → 3 in sequence:

- After clicking result 1: "SOCRATES" in persons-section highlighted. DOM: 2 active highlights (word split across spans: "S" + "OCRATES").
- After clicking result 2: persons-section highlight cleared; 227a "SOCRATES" highlighted. DOM: still exactly 2 active highlights.
- After clicking result 3: 227a highlight cleared; 227b "SOCRATES" highlighted. DOM: still exactly 2 active highlights.

No stale accumulation across result clicks. No console errors. No 4xx/5xx network requests.

**GIF:** `tmp/qa-fix-walkthrough-1717.gif`

## Acceptance criteria

- [x] `unwrapHighlights()` is called at the start of `goToResult()`, before `pendingHighlight` is set. *(Item 2)*
- [x] After clicking two different search results in sequence, `highlightRestores` contains only the entries from the most recent result. *(Items 1, 3)*
- [x] No detached DOM node references remain in `highlightRestores` between result navigations. *(Item 1 — regression test)*
- [x] Existing behavior of `goToPage()`, `next()`, `prev()`, and `clearSearch()` is unchanged. *(Item 1 — full suite passes)*

## Notes

Pre-QA acceptance check: Playwright headless Chromium cannot run on NixOS (dynamically linked executable fails). This is a known environment limitation — CI runs the acceptance suite and is passing. The acceptance test failure is environmental, not a product bug.

natb1:
## Review — PR #1736

### Fixed
_None._

### Required (security)
_None._ — CodeQL: no open alerts. Dependency audit: no `package.json`/`package-lock.json` changes (skipped).

### Informational

- **print/test/viewer/pdf.test.ts:493** — The non-null assertions (`container.querySelector('.search-highlight')!.parentElement!`) are evaluated before the guard that would fire if no highlight exists; a failure here produces an opaque `Cannot read properties of null` rather than a clear test-failure message. Low-risk in practice since `goToResult(r1)` precedes the query, but fragile for future test authors. (`code-review`)
- **print/test/viewer/pdf.test.ts:483** — The new test covers two different results on two different pages, but does not test clicking the same result twice (same page, same div). That path exercises the `if (pendingHighlight?.match(result.location))` early-return guard added elsewhere. Not a gap in coverage for this PR's fix, but worth noting for completeness. (`review`)

### Dismissed
_None._

### False-positive (security)
_None._

### Deferred

- **print/src/viewer/pdf.ts:590** — `goToOutlineEntry` does not call `unwrapHighlights()` or clear `pendingHighlight` before navigating. Filed as follow-up → #1745 (blocked by #1717). (`code-review`)
- **print/src/viewer/pdf.ts:500** — `goToPosition` does not call `unwrapHighlights()` or clear `pendingHighlight`. Filed as follow-up → #1744 (blocked by #1717). (`code-review`)

### Out-of-scope (security)
_None._

PR #1738, opened 2026-06-15
blog: unsubscribe onAuthStateChanged auth listener in destroy()

natb1:
Closes #1715

## Summary

`createBlogApp` wired the auth-state listener via
`config.firebase.onAuthStateChanged(cb)` but consumed the return value only as a
`PromiseLike` for error handling (`.then(undefined, errHandler)`), never
capturing the unsubscribe. `destroy()` therefore never unsubscribed — after
teardown an auth-state event could still fire `refreshAfterAuthChange` on the
destroyed instance, writing to detached DOM and calling the already-destroyed
router.

## The production return shape

The value passed in production is firebaseutil's **async** wrapper from
`createAppContext` (`landing/src/main.ts`, `fellspiral/src/main.ts`), which
returns `Promise<() => void>` — the unsubscribe resolves *inside* the promise. A
literal `typeof === "function"` check would be false in production and drop the
unsubscribe, so the fix captures it in **both** shapes.

## Changes (entirely within `blog/src/create-blog-app.ts`)

- **Config type** — widened `onAuthStateChanged`'s return to
  `(() => void) | PromiseLike<(() => void) | void>`. This admits the production
  `Promise<() => void>`, a synchronous `() => void`, and the existing
  `Promise<void>` test mocks without any caller edits.
- **Capture + race guard** — register a teardown (before the listener call) that
  flips an `authDestroyed` flag and calls the captured unsubscribe; capture from
  the synchronous-function arm and the promise arm. If `destroy()` ran before an
  async unsubscribe resolved, the resolved unsubscribe is invoked immediately.
  The existing auth callback body and the `auth-init` error handler
  (`deferProgrammerError` / `logError`) are preserved unchanged; only the
  previously-`undefined` success arm now captures the resolved unsubscribe.

## Tests (`blog/test/create-blog-app.test.ts`)

Three new cases on the existing harness:

1. Production-representative `Promise<() => void>` — unsubscribe called once on
   `destroy()`.
2. Destroy-before-resolve race — `destroy()` runs first, then the promise
   resolves; unsubscribe still called once.
3. Synchronous-function return — unsubscribe called once on `destroy()`.

All pre-existing cases still pass (the default mock returning `Promise.resolve()`
is unaffected).

## Verification

- `npx vitest run --root blog test/create-blog-app.test.ts` — 11 passed.
- `npm run build --prefix landing` and `npm run build --prefix fellspiral` — both
  type-check clean; the widened union accepts the production `Promise<() => void>`
  with no caller changes.

`blog` is a library workspace with no standalone DOM-resolving typecheck, so the
consuming-app builds and vitest are the authoritative checks.

### DIFF
--- a/blog/src/create-blog-app.ts
+++ b/blog/src/create-blog-app.ts
@@ -57,7 +57,9 @@ export interface CreateBlogAppConfig {
     initAppCheck: (() => Promise<void>) | undefined;
     signIn: () => Promise<unknown> | void;
     signOut: () => Promise<unknown> | void;
-    onAuthStateChanged: (cb: (user: User | null) => void) => PromiseLike<unknown>;
+    onAuthStateChanged: (
+      cb: (user: User | null) => void,
+    ) => (() => void) | PromiseLike<(() => void) | void>;
   };
   adminGroupId: GroupId;
   // optional hooks (consumed in units 2/3)
@@ -267,7 +269,28 @@ export function createBlogApp(config: CreateBlogAppConfig): BlogAppHandle {
     updateInfoPanel();
   }
 
-  config.firebase.onAuthStateChanged((user) => {
+  // Capture the auth-state unsubscribe so destroy() can tear it down; without
+  // this an auth event can still fire refreshAfterAuthChange on a destroyed
+  // instance. Production passes firebaseutil's async wrapper resolving to the
+  // unsubscribe (Promise<() => void>), so we handle both a sync function return
+  // and a promise resolving to one, with a guard for the destroy-before-resolve
+  // race. The teardown is registered before the call so a synchronous capture
+  // and the race guard share the same closure state.
+  let authUnsub: (() => void) | undefined;
+  let authDestroyed = false;
+  teardowns.push(() => {
+    authDestroyed = true;
+    authUnsub?.();
+  });
+
+  const captureUnsub = (unsub: (() => void) | void): void => {
+    if (typeof unsub !== "function") return;
+    // destroy() may have run before an async unsubscribe resolved.
+    if (authDestroyed) unsub();
+    else authUnsub = unsub;
+  };
+
+  const authResult = config.firebase.onAuthStateChanged((user) => {
     if (user?.uid === currentUser?.uid) return;
     currentUser = user;
     // Intentional silent degradation — user sees stale content rather than an error.
@@ -275,10 +298,16 @@ export function createBlogApp(config: CreateBlogAppConfig): BlogAppHandle {
       if (deferProgrammerError(err)) return;
       logError(err, { operation: "auth-change-refresh" });
     });
-  }).then(undefined, (err) => {
-    if (deferProgrammerError(err)) return;
-    logError(err, { operation: "auth-init" });
   });
+
+  if (typeof authResult === "function") {
+    captureUnsub(authResult);
+  } else {
+    authResult.then(captureUnsub, (err) => {
+      if (deferProgrammerError(err)) return;
+      logError(err, { operation: "auth-init" });
+    });
+  }
 
   deferAppCheckInit(
     config.firebase.initAppCheck,
--- a/blog/test/create-blog-app.test.ts
+++ b/blog/test/create-blog-app.test.ts
@@ -420,6 +420,98 @@ describe("createBlogApp routing and panel behavior", () => {
     });
   });
 
+  // Case 8 — #1715: onAuthStateChanged unsubscribe is called on destroy().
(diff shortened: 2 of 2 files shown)


natb1:
## QA Pass — Issue #1715 / PR #1738

**Result: Clean pass.** All acceptance criteria verified.

No browser component detected (changes are in `blog/` library only). No QA server started.

### Items executed (6 total — shell-command lane only)

| # | Item | Result |
|---|------|--------|
| 1 | Blog vitest suite (11 tests incl. 3 new #1715 cases) | ✅ PASS |
| 2 | `blog` standalone tsc (`npx tsc --noEmit -p blog/tsconfig.json`) | ⚠️ FAIL — pre-existing |
| 3 | `landing` build/typecheck (`npm run build --prefix landing`) | ✅ PASS |
| 4 | `fellspiral` build/typecheck (`npm run build --prefix fellspiral`) | ✅ PASS |
| 5 | Config type shape (`onAuthStateChanged` return union) | ✅ PASS |
| 6 | Capture/teardown wiring (`captureUnsub`, `authUnsub`, `authDestroyed`) | ✅ PASS |

**Counts:** 5 PASS / 1 FAIL (pre-existing) / 0 SKIP

### Item 1 detail — vitest

`npx vitest run --root blog test/create-blog-app.test.ts` → **11 passed** in 1.00s.

The three new cases from this PR all passed:
- Case 8: `Promise<() => void>` unsubscribe called once on `destroy()` (production-representative)
- Case 9: destroy-before-resolve race — `destroy()` before promise resolves; unsubscribe still called once
- Case 10: synchronous-function return — unsubscribe called once on `destroy()`

All pre-existing cases unaffected.

### Item 5 detail — config type

`onAuthStateChanged` declared at `blog/src/create-blog-app.ts:60–62`:
```ts
onAuthStateChanged: (
  cb: (user: User | null) => void,
) => (() => void) | PromiseLike<(() => void) | void>;
```
Union admits `() => void` (AC #1 ✅).

### Item 6 detail — capture/teardown wiring

All wiring criteria confirmed in source:
- `authUnsub` / `authDestroyed` variables declared (AC #2 ✅)
- `teardowns.push(() => { authDestroyed = true; authUnsub?.(); })` registered before the listener call (AC #3 ✅)
- `authResult.then(captureUnsub, errHandler)` — error path retained as second arg (AC #4 ✅)
- Three tests confirm `destroy()` calls unsubscribe (AC #5 ✅)

### Disposition triage

The standalone `blog` tsc check (item 2) produced errors about missing DOM types (`document`, `window`, `HTMLElement`, `DOMParser`, etc.) throughout the package. None of the errors mention the PR's new types or helpers. The PR itself documents this: *"blog is a library workspace with no standalone DOM-resolving typecheck, so the consuming-app builds and vitest are the authoritative checks."*

This FAIL was passed to the Opus disposition classifier, which returned zero dispositions — confirming it is a pre-existing known limitation, not a regression introduced by this PR.

`needs-main` items: none. No follow-up issues filed.

Escalation set: **empty** → clean pass.

natb1:
## Review — PR #1738

### Fixed

- **CR-4** `blog/test/create-blog-app.test.ts:446` — Comment said "Flush the microtask queue" but `setTimeout(r, 0)` is a macrotask; corrected to clarify that pending Promise `.then()` callbacks run before the macrotask fires. (commit a08e1cc4)

### Informational

- **CR-2** `blog/src/create-blog-app.ts:281-285` — `captureUnsub` is declared as a named `const` but called only twice in a local scope. Could be inlined at the two call sites (the `if/else` block at lines 288-294). Acceptable as-is; readability trade-off only.
- **CR-3** `blog/test/create-blog-app.test.ts:420-517` — The three new test cases each repeat the full `firebase` mock config inline rather than using a shared helper. Could extract a `makeFirebaseConfig()` helper or add a firebase override to `makeConfig()`. Low priority; no correctness impact.
- **CR-5** `blog/src/create-blog-app.ts:104-106` — `authDestroyed` is never reset to `false` after being set. This is intentional — it is a one-way flag. Informational only.

### Dismissed

- **CR-1** `blog/src/create-blog-app.ts:272-278` — Comment block before the unsub setup explains what the code does rather than why. Project code-style rule says to omit comments where well-named identifiers suffice. Dismissed: the comment does carry a non-obvious constraint (production async shape + race guard), so keeping it is defensible; no change required.

### False-positive (security)

- **SEC-1** — No security vulnerabilities found in the diff. The change is a resource-management fix (listener cleanup) with no attack surface: no input validation boundary, no data exposure, no auth bypass, no Firebase security rule changes. CodeQL scan: no open alerts on this PR ref.

PR #1742, opened 2026-06-15
print: findMatches uses needle.length, not query.length

natb1:
Closes #1718

## Problem

`findMatches` in `print/src/viewer/pdf.ts` runs a case-insensitive search by
lowercasing both the haystack (`pageText.toLowerCase()`) and the query
(`query.toLowerCase()` → `needle`), then finds `needle` in the haystack. But it
stored `query.length` — not `needle.length` — in both the result's `length`
field and the `from` advance.

For Unicode queries whose lowercased form has a different code-unit count,
`needle.length !== query.length`. For example `"İ"` (U+0130, LATIN CAPITAL
LETTER I WITH DOT ABOVE) has `length` 1, but `"İ".toLowerCase()` is `"i" + U+0307`
(combining dot above) with `length` 2. The match in the lowercased haystack spans
`needle.length` (2) code units at the found `offset`, so storing `query.length`
(1) meant:

- The `length` baked into the `SearchResult.location` token was wrong, so
  `offsetToDivRanges` walked the wrong number of characters and misplaced or
  truncated the highlight span.
- The `from = offset + query.length` advance under-stepped the real match end, so
  a subsequent match could be missed or misfound.

## Fix

Use `needle.length` in both places. The match in `haystack` spans exactly
`needle.length` characters starting at `offset`, and `buildPageText` joins items
with an empty separator, so the raw `pageText` at that offset spans the same
`needle.length` characters used for the highlight.

## Tests

- `findMatches`: an expanding-case query (`"İ"`) now yields `{ offset: 1, length:
  2 }`, not `{ length: 1 }`. The combining string is built via `"İ".toLowerCase()`
  rather than a fragile raw literal.
- `offsetToDivRanges`: the corrected `offset`/`length` maps cleanly onto its item,
  covering the highlight-placement path end to end.

The existing ASCII `"length equals query length"` test (where `needle.length ===
query.length`) is unchanged and still passes. All 32 tests in
`print/test/viewer/pdf.test.ts` pass.

### DIFF
--- a/print/src/viewer/pdf.ts
+++ b/print/src/viewer/pdf.ts
@@ -39,8 +39,8 @@ export function findMatches(
   while (from < haystack.length) {
     const offset = haystack.indexOf(needle, from);
     if (offset === -1) break;
-    results.push({ offset, length: query.length });
-    from = offset + query.length;
+    results.push({ offset, length: needle.length });
+    from = offset + needle.length;
   }
   return results;
 }
--- a/print/test/viewer/pdf.test.ts
+++ b/print/test/viewer/pdf.test.ts
@@ -136,11 +136,12 @@ describe("findMatches", () => {
     expect(findMatches("hello world", "xyz")).toEqual([]);
   });
 
-  it("length equals query length", () => {
+  it("length equals needle (lowercased query) length", () => {
     const query = "needle";
+    const needle = query.toLowerCase();
     const matches = findMatches("needle in a needlestack", query);
     for (const m of matches) {
-      expect(m.length).toBe(query.length);
+      expect(m.length).toBe(needle.length);
     }
   });
 
@@ -158,6 +159,19 @@ describe("findMatches", () => {
     const matches = findMatches(pageText, query);
     // Match spans needle.length (2) at offset 1, not query.length (1).
     expect(matches).toEqual([{ offset: 1, length: 2 }]);
+  });
+
+  it("advances by needle (lowercased) length between adjacent expanding-case matches", () => {
+    // "İ" (U+0130) lowercases to "i" + U+0307 (combining dot above): length 1 → 2.
+    const query = "İ";
+    const needle = query.toLowerCase(); // "i̇", length 2
+    const matches = findMatches(needle + needle, query);
+    // Two adjacent matches: the next search must resume at needle.length (2),
+    // not query.length (1), or it would start inside the consumed needle.
+    expect(matches).toEqual([
+      { offset: 0, length: needle.length },
+      { offset: needle.length, length: needle.length },
+    ]);
   });
 });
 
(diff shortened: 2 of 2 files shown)


natb1:
# QA Summary — #1718 (PR #1742)

**Result: Clean pass — all items passed.**

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | findMatches and offsetToDivRanges unit tests pass | shell-command | PASS |
| 2 | Source no longer references query.length in findMatches | shell-command | PASS |
| 3 | PDF search highlight renders on the correct span (no-regression) | needs-browser | PASS |

## Item details

**Item 1** — `npx vitest run --root print test/viewer/pdf.test.ts`: 32/32 tests passed, including the new expanding-case Unicode test (`"İ"` → `{ offset: 1, length: 2 }`) and the new `offsetToDivRanges` mapping test.

**Item 2** — `grep -n "query.length" print/src/viewer/pdf.ts` returned no matches: both occurrences replaced with `needle.length`.

**Item 3** — Opened Phaedrus PDF in the viewer, searched for "Socrates" (10 results), navigated between matches. Orange highlight covered each "SOCRATES:" occurrence exactly — no shift, no truncation. Zoomed verification confirmed pixel-accurate alignment. No console errors, no 4xx/5xx network errors.

## Walkthrough recording

`qa-fix-walkthrough-1718.gif` (downloaded to browser)

## Pre-QA acceptance gate note

The local Playwright acceptance gate could not run — `apt-get` is unavailable in this Nix/WSL environment (Playwright's browser install fails). This is a known infrastructure limitation, not a product bug; the fix touches only `print/src/viewer/pdf.ts` and `print/test/viewer/pdf.test.ts` and cannot have caused it. CI acceptance job is authoritative and **passing** (confirmed via `gh pr checks 1742`).

natb1:
## Review — PR #1742

**Code surface:** `print/src/viewer/pdf.ts`, `print/test/viewer/pdf.test.ts`
**Security surface:** `code` (app source changed) · CodeQL: no open alerts · deps: unchanged

---

### Fixed (2)

- **`print/test/viewer/pdf.test.ts:139`** — Stale test name and assertion: the existing `'length equals query length'` test asserted `expect(m.length).toBe(query.length)`. After the fix the invariant is `needle.length` (the lowercased form), not `query.length`; the test only happened to pass because the ASCII query has equal lengths. Renamed to `'length equals needle (lowercased query) length'` and updated the assertion to `toBe(needle.length)`. (commit 375f9f3e)

- **`print/test/viewer/pdf.test.ts:153`** — The new expanding-case test covered the `length` fix but not the `from`-advancement fix (`from = offset + needle.length`). Added an adjacent-expanding-case-match test that places two consecutive expanding-case matches in the haystack and asserts both are found at the correct offsets — directly verifying the `from`-advancement path. (commit 375f9f3e)

---

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1743, opened 2026-06-15
print: guard search() against null pdfDoc before the page-fetch loop

natb1:
Closes #1725

## Summary

`search()` in `print/src/viewer/pdf.ts` dereferenced `pdfDoc!` inside its
page-fetch loop with no prior null check. `pdfDoc` is `null` both before
`init()` resolves and after `destroy()` runs (which sets `pdfDoc = null`). The
in-loop `destroyed` guards only fire *after* each `await`, so they catch
destruction that races an in-progress `getPage` call but not a fresh `search()`
invoked on an already-null `pdfDoc` — in that case `pdfDoc!.getPage` threw
`TypeError: Cannot read properties of null`.

This adds an early `if (!pdfDoc) return [];` guard before the page-fetch loop,
mirroring the existing guard in `getOutline()` verbatim. `search()` now returns
`[]` without throwing when called before `init()` resolves or after `destroy()`.

Follow-up hardening to #1691 (PDF full-document search), which introduced
`search()`.

## Changes

- `print/src/viewer/pdf.ts`: add `if (!pdfDoc) return [];` in `search()` after
  the trimmed-query guard, before the page-fetch loop. No other behavior change
  — the existing `pdfDoc!.getPage(i)` assertion and in-loop `destroyed` checks
  are unchanged, so the non-null search path is unaffected.
- `print/test/viewer/pdf.test.ts`: two regression tests in the `search()` block
  — `search()` before `init()` and after `destroy()` each resolve to `[]`
  rather than throwing.

## Verification

```
npx vitest run --root print test/viewer/pdf.test.ts   # 32 passed
npx vitest run --root print test/viewer                # 286 passed
```

### DIFF
--- a/print/src/viewer/pdf.ts
+++ b/print/src/viewer/pdf.ts
@@ -546,6 +546,7 @@ export function createPdfRenderer(onError?: (err: unknown) => void): ContentRend
     async search(query: string): Promise<SearchResult[]> {
       const trimmed = query.trim();
       if (!trimmed) return [];
+      if (!pdfDoc) return [];
 
       const results: SearchResult[] = [];
 
--- a/print/test/viewer/pdf.test.ts
+++ b/print/test/viewer/pdf.test.ts
@@ -367,6 +367,20 @@ describe("search()", () => {
     const decoded = decodeLocation(page2!.location);
     expect(decoded).toEqual({ page: 2, offset: 0, length: 3 });
   });
+
+  it("returns [] without throwing when called before init() resolves", async () => {
+    const renderer = createPdfRenderer();
+
+    await expect(renderer.search!("the")).resolves.toEqual([]);
+  });
+
+  it("returns [] without throwing after destroy()", async () => {
+    const renderer = createPdfRenderer();
+    await renderer.init(container, "fake://source.pdf");
+    renderer.destroy();
+
+    await expect(renderer.search!("the")).resolves.toEqual([]);
+  });
 });
 
 // ---------------------------------------------------------------------------


natb1:
# QA Pass — Issue #1725

## Items executed

| 1 | Null guard present in `search()` source | script-verifiable | ✅ PASS |
| 2 | `search()` before `init()` and after `destroy()` return `[]` without throwing | script-verifiable | ✅ PASS |
| 3 | Existing search functionality and full viewer suite remain unaffected | script-verifiable | ✅ PASS |

## Counts

- PASS: 3
- FAIL: 0
- SKIP: 0

## Details

**Item 1** — `grep` confirmed `if (!pdfDoc) return [];` is present in `search()` after the trimmed-query guard, before the page-fetch loop.

**Item 2** — `npx vitest run --root print test/viewer/pdf.test.ts`: 32 tests passed, including the two new regression tests:
- `"returns [] without throwing when called before init() resolves"` ✅
- `"returns [] without throwing after destroy()"` ✅

**Item 3** — `npx vitest run --root print test/viewer`: 286 tests passed across all viewer test files. The `DOMException` log lines are from `epub.test.ts` testing error-handling paths with `blob:` and `https://example.com` URLs in a jsdom/happy-dom environment — they are expected stderr from the epub tests, not failures.

## Disposition

No residue — all acceptance criteria verified by machine checks. Clean autonomous pass.

natb1:
## Review — PR #1743 (issue #1725)

**Summary:** All findings are informational. No bugs or security vulnerabilities found. No code changes applied.

---

### Informational

- **`print/src/viewer/pdf.ts:549` — Guard placement correct** (`F1+GUARD-PLACEMENT+SEC-001`)
  Null guard on `pdfDoc` before the page-fetch loop is correctly placed and mirrors the existing `getOutline()` guard verbatim. No change needed.

- **`print/test/viewer/pdf.test.ts:371-375` — Pre-init test wording** (`F2+TEST1+SEC4-pre`)
  The "before init() resolves" regression test passes because `pageCount` is `0` before `init()`, so the loop never runs — the early `[]` return comes from the loop condition, not the null guard. The guard itself is genuinely exercised by the post-destroy test. Consider renaming the test description to `'returns [] when called before init() (pageCount is 0)'` to accurately reflect which guard fires. No code change is required.

- **`print/test/viewer/pdf.test.ts:377-383` — Post-destroy test provides genuine regression coverage** (`F3+TEST2+SEC4-post`)
  After `destroy()`, `pdfDoc` is null but `pageCount` retains its pre-destroy value, so the loop would execute and `pdfDoc!.getPage(i)` would throw without the guard. This test correctly exercises the actual bug scenario. No change needed.

- **`print/src/viewer/pdf.ts:557-568` — In-loop checks and non-null assertion are correct** (`F4+NONNULL+RACE`)
  The `!` assertion is required because TypeScript does not narrow across `await` boundaries. The post-`await` `destroyed` checks prevent `getPage` from being called after `destroy()` transitions `pdfDoc` to null. No change needed.

- **`print/src/viewer/pdf.ts:546-549` — No DoS vector** (`SEC-003`)
  Repeated `search()` calls on a null `pdfDoc` return `[]` in O(1) after `query.trim()`. Upstream debounce in `search.ts` already rate-limits calls. No action needed.

---

### Out-of-scope (security)

- **`print/src/viewer/search.ts:17-21,68` — Pre-existing innerHTML pattern** (`SEC-002`)
  `innerHTML` is used in `search.ts` but all PDF-content-derived fields are passed through `escapeHtml()` before interpolation. Pattern is pre-existing, unaffected by this diff. No action needed.

---

### Security scan coverage

- **CodeQL:** No open alerts on `refs/pull/1743/head`.
- **Dependency audit:** Not run — diff does not touch `package.json` / `package-lock.json`.
- **No required findings. Deviation: false.**

PR #1746, opened 2026-06-15
print: capture pdfDoc into a local in search() to drop the non-null assertion

natb1:
Closes #1722

## Problem

`search()` in `print/src/viewer/pdf.ts` read the module-level `pdfDoc`
through a non-null assertion inside its async page loop
(`await pdfDoc\!.getPage(i)`). `destroy()` sets `destroyed = true` and, later
in the same call, `pdfDoc = null`. If `pdfDoc` was already `null` when
`search()` was entered — e.g. the viewer was torn down and a debounced search
keystroke still fired — the first `pdfDoc\!.getPage(1)` dereferenced `null` and
threw an unhandled `TypeError` instead of returning gracefully.

Every other method in the file already guards `pdfDoc` before use
(`getOutline`, `renderPage`, `renderPageInto`). `search()` was the lone
holdout.

## Fix

Capture `pdfDoc` into a local at the top of `search()`, guard on the local
(and `destroyed`), and use the local in the loop — removing the non-null
assertion:

```ts
const doc = pdfDoc;
if (\!doc || destroyed) return results;
```

The `destroyed` flag is set *before* `pdfDoc` is nulled, and each `await` in
the loop is already followed by `if (destroyed) return results`, so a
separate-task `destroy()` cannot interleave mid-loop — the genuinely reachable
crash was **pdfDoc-null-at-entry** only. Capturing the local makes the
graceful path explicit and future-proofs the loop against an added `await`.

## Tests

Two tests added to the existing `describe("search()", …)` block in
`print/test/viewer/pdf.test.ts`:

1. **pdfDoc-null-at-entry → `[]` without throwing** — the load-bearing
   regression guard; fails on the old code, passes on the fixed code.
2. **destroy() mid-loop → resolves, never rejects** — a graceful-degradation
   check driven by a racing document whose `getPage` calls `destroy()` while
   the loop is in flight.

## Verification

- `npx vitest run --root print test/viewer/pdf.test.ts` — 32 tests pass,
  including the 2 new ones.
- `npm run build --prefix print` — `tsc` types the `doc` local cleanly.

### DIFF
--- a/print/src/viewer/pdf.ts
+++ b/print/src/viewer/pdf.ts
@@ -565,9 +565,21 @@ export function createPdfRenderer(onError?: (err: unknown) => void): ContentRend
         // debounced keystroke would be wasteful; cache across search calls.
         let pageText = pageTextCache.get(i);
         if (pageText === undefined) {
-          const page = await doc.getPage(i);
-          if (destroyed) return results;
-          const tc = await page.getTextContent();
+          let tc;
+          try {
+            const page = await doc.getPage(i);
+            if (destroyed) return results;
+            tc = await page.getTextContent();
+          } catch (err) {
+            // A concurrent destroy() tears down the pdfjs worker, which rejects
+            // pending getPage()/getTextContent() calls with an
+            // UnknownErrorException. Degrade silently — same intent as the
+            // `destroyed` short-circuits — instead of surfacing "Search failed".
+            if (destroyed || (err as { name?: string })?.name === "UnknownErrorException") {
+              return results;
+            }
+            throw err;
+          }
           if (destroyed) return results;
           pageText = buildPageText(tc.items as { str?: string }[]);
           pageTextCache.set(i, pageText);
--- a/print/src/viewer/pdf.ts
+++ b/print/src/viewer/pdf.ts
@@ -553,12 +553,19 @@ export function createPdfRenderer(onError?: (err: unknown) => void): ContentRend
       // very short and matches thousands of positions across a long document.
       const MAX_RESULTS = 200;
 
+      // Capture pdfDoc into a local so a concurrent destroy() — which nulls
+      // pdfDoc — cannot turn a non-null assertion into a TypeError. Matches the
+      // guard pattern in getOutline()/renderPage(). The destroyed flag still
+      // short-circuits after each await below.
+      const doc = pdfDoc;
+      if (!doc || destroyed) return results;
+
       for (let i = 1; i <= _pageCount; i++) {
         // Lazily populate the page-text cache. Re-fetching text for every
         // debounced keystroke would be wasteful; cache across search calls.
         let pageText = pageTextCache.get(i);
         if (pageText === undefined) {
-          const page = await pdfDoc!.getPage(i);
+          const page = await doc.getPage(i);
           if (destroyed) return results;
           const tc = await page.getTextContent();
           if (destroyed) return results;
--- a/print/test/viewer/pdf.test.ts
+++ b/print/test/viewer/pdf.test.ts
@@ -367,6 +367,42 @@ describe("search()", () => {
     const decoded = decodeLocation(page2!.location);
     expect(decoded).toEqual({ page: 2, offset: 0, length: 3 });
   });
+
+  it("returns [] without throwing when destroy() ran before search()", async () => {
+    const renderer = createPdfRenderer();
+    await renderer.init(container, "fake://source.pdf");
+    renderer.destroy(); // sets destroyed = true and pdfDoc = null
+    // Old code: `await pdfDoc!.getPage(1)` dereferences null → TypeError.
+    await expect(renderer.search!("the")).resolves.toEqual([]);
+  });
+
+  it("resolves (does not reject) when destroy() fires during the page loop", async () => {
+    const pdfjs = await import("pdfjs-dist");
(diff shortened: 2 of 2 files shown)


natb1:
# QA Summary — Issue #1722

## Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | search() no longer uses pdfDoc! | shell-command | PASS |
| 2 | search() returns [] when pdfDoc is null at entry | shell-command | PASS |
| 3 | Concurrent destroy() mid-loop resolves instead of rejecting | shell-command | PASS |
| 4 | Existing search() unit tests continue to pass | shell-command | PASS |
| 5 | Full print suite and TypeScript build remain green | shell-command | PASS |

## Verification

- `grep` confirms no `pdfDoc!` token inside `search()` — the local capture `const doc = pdfDoc` is in place.
- `npx vitest run --root print test/viewer/pdf.test.ts` — **32/32 tests pass**, including both new regression tests.
- `npx vitest run --root print` — **456/456 tests pass** across the full suite.
- `npm run build --prefix print` — TypeScript build succeeds; the `doc` local types cleanly.

## Clean Pass

No residue items. No browser walkthrough needed (all acceptance criteria are unit-test verifiable). No `needs-human-judgment` items deferred to office-hours.

natb1:
## Review: PR #1746

### Fixed

- **`search()` loop unguarded against destroy-during-search errors** (`print/src/viewer/pdf.ts:568`): The `doc.getPage()`/`page.getTextContent()` calls inside the loop were not wrapped in error handling. If pdf.js internally tears down the document mid-loop (e.g. throwing `UnknownErrorException`), the error would propagate as an unhandled rejection. Added a try/catch that catches destroy-triggered errors and returns `results` early, matching the silent-degradation intent of the existing `if (destroyed) return results` guards. (fix commit: bfd9579b)
- **`spy.mockRestore()` not in try/finally in race test** (`print/test/viewer/pdf.test.ts:397-404`): The `mockRestore()` call at the end of the "destroy() mid-loop" test was not guarded by try/finally. If an assertion or await rejected before reaching the restore, the `getDocument` mock would leak into subsequent tests. Wrapped the test body in try/finally so `spy.mockRestore()` always runs. (fix commit: bfd9579b)

### Informational

- **Race test `destroy()` call is synchronous inside `getPage` callback** (`print/test/viewer/pdf.test.ts:385`): The race test calls `renderer.destroy()` synchronously inside the `getPage` stub before returning `Promise.resolve(...)`. Because `destroy()` sets `destroyed = true` synchronously before nulling `pdfDoc`, and the stub returns a resolved promise, the test correctly exercises the `if (destroyed) return results` guard after the first `await`. This is a valid approach for the graceful-degradation intent being tested; no change needed.

### Dismissed

- **`resolves during page loop` test + ResizeObserver** (`print/test/viewer/pdf.test.ts:379`): Concern that `renderer.init()` registers a ResizeObserver that could cause issues in the test environment. The test environment already mocks/stubs the necessary APIs and all 32 tests pass including the new ones; no observable issue. Dismissed.
- **`getOutline()` reads `pdfDoc` again after the null guard** (`print/src/viewer/pdf.ts:590-594`): Reviewer flagged that `getOutline()` checks `if (!pdfDoc) return []` and then calls `await pdfDoc.getOutline()` without re-capturing. The local-capture pattern from `search()` would be cleaner, but `getOutline()` does not use `pdfDoc` across multiple awaits — it's a single call — so the race window is negligible. The existing pattern is consistent with `renderPage`/`renderPageInto`. Dismissed as out-of-scope for this PR.
- **Non-null assertion `renderer.search!` in regression test** (`print/test/viewer/pdf.test.ts:376`): The `search` property is optional on `ContentRenderer` and the test uses `renderer.search!`. Minor style point; does not affect correctness or test validity. Dismissed.

### False-positive

- **Local `doc` reference keeps `PDFDocumentProxy` alive after `destroy()`** (`print/src/viewer/pdf.ts:560-569`): Concern that capturing `doc = pdfDoc` prevents garbage collection after `destroy()` calls `pdfDoc.destroy()`. The local `doc` is scoped to the `search()` call frame and released when the loop exits — no persistent retention. The pdf.js `PDFDocumentProxy.destroy()` call handles internal cleanup synchronously; the local reference does not interfere. False-positive.
- **`pageTextCache` written unconditionally after `destroyed` check** (`print/src/viewer/pdf.ts:572-573`): Concern that the cache write races with teardown. The `if (destroyed) return results` check immediately precedes the cache write; no `await` separates the check from the write, so no concurrent `destroy()` can interpose. False-positive.
- **`_pageCount` not snapshotted alongside `doc`** (`print/src/viewer/pdf.ts:563`): Concern that a second `init()` call could change `_pageCount` mid-loop. A second `init()` is not a supported concurrent usage pattern in this viewer; the `destroyed` flag and the `doc` local guard the documented race. False-positive.

### Deferred

- **`resolveOutlineItems` uses `pdfDoc!` across awaits** (`print/src/viewer/pdf.ts:184`): `resolveOutlineItems` calls `pdfDoc!.getDestination()` (line 184) and `pdfDoc!.getPageIndex()` (line 189) across awaits without capturing `pdfDoc` into a local first. A concurrent `destroy()` can null `pdfDoc` between these awaits, producing a spurious TypeError caught by the enclosing try/catch. Pre-existing pattern, not introduced by this PR; filed as #1752 (blocked by #1722).

PR #1747, opened 2026-06-15
print: make applyHighlight idempotent by unwrapping before re-highlighting

natb1:
Closes #1721

## Problem

`applyHighlight` in `print/src/viewer/pdf.ts` wraps search-match text in
`<span class="search-highlight active">` elements and records each mutated div's
pre-mutation `textContent` into the module-level `highlightRestores` array, so
`unwrapHighlights` can later restore the text layer to pristine plain text. But
`applyHighlight` did **not** unwrap first.

It is called on every re-render of a page whose `pendingHighlight` is still
armed — the single-page resize path (`renderPage`, via the debounced
ResizeObserver) and the spread path (`renderPageInto`). Each re-render pushed a
**new** `highlightRestores` entry without clearing the prior ones. Because every
render builds fresh `textDiv` objects (the previous divs are detached by
`replaceChildren`), the accumulated entries point at **detached** divs:
`highlightRestores` grows unboundedly while a search highlight is active and the
user resizes, and `applyHighlight` is non-idempotent.

## Fix

Call `unwrapHighlights()` as the first statement of `applyHighlight()` — the
issue's recommended one-liner. `unwrapHighlights` restores every recorded div
(including the now-detached prior ones) and resets `highlightRestores.length = 0`
before the new highlight is recorded, making `applyHighlight` idempotent and
self-cleaning regardless of caller. The doc comments on `applyHighlight` and
`highlightRestores` are updated to state the one-entry-per-highlighted-div
invariant.

## Test

Adds one regression test in the `clearSearch()` block of
`print/test/viewer/pdf.test.ts` that drives the debounced resize re-render path
with the highlight still armed and asserts the detached first-render div is
restored (its `.search-highlight` span removed and `textContent` back to plain
text). Validated as a red→green discriminator: the test fails with the
`unwrapHighlights();` line removed and passes with it in place.

## Verification

- `npx vitest run --root print test/viewer/pdf.test.ts` — all pass; red→green
  discriminator confirmed.
- `npx vitest run --root print` — full print suite passes (454 passed, 1
  pre-existing unrelated skip).

### DIFF
--- a/print/src/viewer/pdf.ts
+++ b/print/src/viewer/pdf.ts
@@ -160,8 +160,12 @@ export function createPdfRenderer(onError?: (err: unknown) => void): ContentRend
   // re-render (renderPage(_currentPage)) re-applies it after a resize. Cleared
   // by the public manual-nav methods so a stale highlight never reappears.
   let pendingHighlight: { page: number; offset: number; length: number } | null = null;
-  // Records the original text content of every div mutated by applyHighlight so
-  // unwrapHighlights can restore the text layer to its pristine state.
+  // Records the original text content of every div mutated by the most recent
+  // applyHighlight call so unwrapHighlights can restore the text layer to its
+  // pristine state. Invariant: after applyHighlight returns, this array holds
+  // exactly one entry per currently-highlighted div (the divs from the most
+  // recent call only — earlier entries are cleared by the unwrapHighlights()
+  // call at the start of applyHighlight).
   const highlightRestores: { div: HTMLElement; originalText: string }[] = [];
   interface SpreadPage { renderTask: RenderTask; textLayer: TextLayer | null; }
   const spreadPages: SpreadPage[] = [];
@@ -308,8 +312,15 @@ export function createPdfRenderer(onError?: (err: unknown) => void): ContentRend
    *
    * Children are rebuilt with createTextNode/createElement (never innerHTML) so
    * the text layer's transparent-text CSS keeps applying.
+   *
+   * Calls unwrapHighlights() first so it is idempotent: any prior highlight
+   * (including detached divs from a superseded render) is restored and
+   * highlightRestores is cleared before the new entries are recorded. After
+   * this function returns, highlightRestores holds exactly one entry per
+   * currently-highlighted div.
    */
   function applyHighlight(tl: TextLayer, h: { offset: number; length: number }): void {
+    unwrapHighlights();
     const itemsStr = tl.textContentItemsStr;
     const divs = tl.textDivs;
     const segments = offsetToDivRanges(itemsStr, h.offset, h.length);
--- a/print/test/viewer/pdf.test.ts
+++ b/print/test/viewer/pdf.test.ts
@@ -464,6 +464,33 @@ describe("clearSearch()", () => {
     expect(restored).toBeDefined();
   });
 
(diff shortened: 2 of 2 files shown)


natb1:
## QA Pass — Issue #1721

**PR:** #1747
**Branch:** `1721-print-applyhighlight-reads`

### Items Executed

| 1 | Search highlight appears on first navigation to a result | needs-browser | ✅ PASS |
| 2 | Resize-while-highlighted then clear-search leaves the visible text layer clean | needs-browser | ✅ PASS |
| 3 | Idempotent-applyHighlight regression mechanism is covered and green | script-verifiable | ✅ PASS |

**Counts:** 3 passed, 0 failed, 0 skipped

### Results

**Item 1 — Search highlight appears on first navigation**
Searched for "Lysias" in Phaedrus, navigated to the first result. Confirmed `.search-highlight.active` span present in DOM wrapping the matched text. The new `unwrapHighlights()` call at the start of `applyHighlight` does not suppress the initial highlight. ✅

**Item 2 — Resize-while-highlighted then clear-search leaves visible text layer clean**
With the search highlight active, resized the browser window (triggering the debounced ResizeObserver re-render with `pendingHighlight` still armed). Highlight re-appeared correctly on the re-rendered page. Cleared the search — confirmed 0 `.search-highlight` spans remaining in the DOM. No console errors. ✅

**Item 3 — Regression test suite (vitest)**
`npx vitest run --root print test/viewer/pdf.test.ts` — 31/31 passed, including the new "re-rendering a page with the highlight still armed restores the detached div (idempotent applyHighlight)" regression test that is a red→green discriminator for this fix. ✅

### Walkthrough Recording
GIF: `qa-fix-walkthrough-1721.gif`

### Notes
- Pre-QA acceptance tests could not run locally (known NixOS/Playwright environment mismatch — chromium version incompatibility). CI is passing and is authoritative for acceptance coverage.
- The core correctness guarantee (detached-div cleanup, `highlightRestores` invariant) is verified by the regression test (Item 3). The browser walkthrough confirmed the user-visible highlight lifecycle is intact.

natb1:
## Review: PR #1747 (for #1721)

### Fixed

- **F001** (`print/src/viewer/pdf.ts:319`) — `unwrapHighlights()` is correctly called as the first statement in `applyHighlight()`. This is the right fix: it restores previously-highlighted divs (including detached ones from superseded renders) before recording new entries, making `applyHighlight` idempotent and self-cleaning. No additional code changes required.

### Informational

- **F003** (`print/src/viewer/pdf.ts:163`) — Doc comment already states the invariant clearly (`"earlier entries are cleared by the unwrapHighlights() call at the start of applyHighlight()"`). No change needed.

### Dismissed

- **F002** (`print/test/viewer/pdf.test.ts:488`) — The hardcoded `'the cat sat'` string is deliberate; it matches the mock fixture's text content for the page-1 text div, which the test controls. More readable than abstracting to a shared constant.

### False-positive (security)

- **F004** (`print/src/viewer/pdf.ts:308-330`) — No XSS or injection risk. The code consistently uses `createTextNode`/`createElement` (never `innerHTML`). The fix does not introduce any new DOM manipulation path. (OWASP A03:2021 Injection / STRIDE Tampering — confirmed not applicable.)

### Security

No open CodeQL alerts on this PR ref. No dependency changes (`package.json`/`package-lock.json` untouched).

---

_No code changes applied — the implementation is correct as submitted._

PR #1748, opened 2026-06-15
print: in spread mode, a search-result click renders once (shell routes the render)

natb1:
Closes #1719

## Problem

In the print viewer's **spread mode**, clicking a search result fired **two**
full page renders instead of one:

1. `search.ts` calls `renderer.goToResult(result)`, which armed the highlight,
   set `_currentPage`, and rendered the **single-page** canvas. In spread mode
   that canvas is `display:none`, so the entire render (`getPage` +
   `getTextContent` + canvas `render` + `applyHighlight`) was immediately hidden
   waste — and it still appended a stale set of `highlightRestores` entries.
2. The shell's `onNavigate` callback then rendered the visible spread, appending
   a **second** set of `highlightRestores` entries for the same logical
   highlight.

Every spread-mode result click burned one wasted full single-page render, and
`clearSearch()` had to restore two sets of entries (the stale hidden set plus
the spread set).

## Root cause

`goToResult` unconditionally rendered the renderer's own single-page view, but
in spread mode the **shell** (which owns the `SpreadController`) is what renders
the visible view afterward. The renderer is spread-agnostic and cannot know its
single-page canvas is hidden, so the render decision belongs in the shell —
exactly as manual navigation (`goToPageNum`) already routes through `controller`
vs `renderer` by mode.

## Fix

Make `goToResult` **arm-only** and move the visible render to the shell's
`onNavigate` callback, routed by mode:

- **`pdf.ts`** — `goToResult` now only arms `pendingHighlight` and sets
  `_currentPage`; it no longer renders. A new `renderResult()` method wraps the
  internal `renderPage(_currentPage)` (which already applies the armed highlight
  as its final gen-guarded step) to render the current single-page view.
- **`types.ts`** — `ContentRenderer` gains an optional
  `renderResult?(): Promise<void>` companion to `goToResult`.
- **`shell.ts`** — the search `onNavigate` callback renders in both modes,
  routed like `goToPageNum`: spread → `controller.goToPage(renderer.currentPage)`;
  single → `renderer.renderResult?.()`.

A result click now triggers exactly one logical render via the correct surface —
the spread render in spread mode, the single-page render in single-page mode —
with a single set of `highlightRestores`, so `clearSearch()` restores only the
visible view.

This keeps the renderer spread-agnostic (no hidden 0×0-guard or spread-awareness
hacks) and mirrors the existing navigation routing. `epub.ts` / `image-archive.ts`
are unaffected — they don't implement `search`, and `renderResult` is optional.

## Tests

- `pdf.test.ts` — updated the `clearSearch()` cases now that `goToResult` is
  arm-only (a `renderResult()` call is interposed before highlight assertions),
  plus a new case pinning the contract split: `goToResult` alone produces no
  `.search-highlight` span; `renderResult()` then applies it.
- `shell.test.ts` — a new pair of cases asserting the result click renders via
  the correct surface per mode: in spread mode `renderPageInto` runs and
  `renderResult` does **not**; in single mode `renderResult` runs and
  `renderPageInto` does **not**.

Full suite: 457 tests pass; `tsc` typecheck and `vite build` clean.


natb1:
## QA Summary — #1719: print spread-mode search-result double-render fix

**PR #1748** | Branch: `1719-print-in-spread-mode-clicki`

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Print package unit-test suite is green | script-verifiable | ✅ PASS |
| 2 | Spread mode: clicking a search result highlights once on the visible spread | needs-browser | ✅ PASS |
| 3 | Spread mode: clearing search query removes highlight cleanly | needs-browser | ✅ PASS |
| 4 | Single-page mode: clicking a search result navigates and highlights (regression guard) | needs-browser | ✅ PASS |
| 5 | Search highlight visual appearance is correct in both modes | needs-human-judgment | ⏸ DEFERRED |

**Counts:** PASS 4 · FAIL 0 · SKIP 0 · Deferred 1

Walkthrough GIF: `tmp/qa-fix-walkthrough-1719.gif`

### Key findings

**Item 1 (vitest):** 457 tests, 28 files, all green.

**Item 2 (spread mode result click):**
- Spread mode active (`aria-pressed=true`, `spread-mode solo` class applied)
- Searched "Socrates" → 10 results; clicked first result
- Highlights: **2 spans in spread canvas, 0 in hidden single-page canvas** ✓
- Spread layout unchanged after click (no collapse to single-page)
- No console errors

**Item 3 (clear search in spread):**
- After clearing search: `allHighlights=0`, `spreadHighlights=0`, `singlePageHighlights=0`
- Spread mode still active; no stale spans anywhere in DOM ✓

**Item 4 (single-page mode regression):**
- Exited spread mode; searched and clicked a result
- 2 highlight spans on the **visible** single-page canvas, 0 in spread elements ✓
- Navigation and highlight work correctly in single-page mode ✓

### Disposition triage

One residue item entered the disposition workflow:

| Item | Kind | Class | Aesthetic | Verify | Rationale |
|------|------|-------|-----------|--------|-----------|
| 5 — visual highlight appearance | needs-human-judgment | **needs-human** | ✅ yes | n/a | Purely visual/aesthetic check; functional correctness (placement, routing, clear) fully machine-verified above |

`needs-main` items: none filed. `needs-main` items are filed as `blocked_by` follow-ups and dropped from the escalation set, so they do not park this PR — but there are none here.

**Non-fixing pass:** one `needs-human` aesthetic item escalates to office-hours for a visual sign-off. Screenshots from the walkthrough show the highlight ("SOCRATES") rendered orange, well-aligned over text glyphs in both spread and single-page modes — but final visual approval is a human call.

natb1:
## QA marked complete (manual override)

Per maintainer direction, the one deferred `needs-human` item (Item 5 — visual
search-highlight appearance) is **skipped** rather than escalated to
office-hours. There were no `opus-fixable` items to resolve — all four
machine-verifiable checks passed:

- Spread-mode result click renders once on the visible spread (0 highlights in
  the hidden single-page canvas, 2 in the spread canvas)
- `clearSearch()` in spread mode removes all highlights cleanly
- Single-page mode result click still renders + highlights correctly
- Full vitest suite green (457 tests)

`dispatch:qa-done` applied; QA phase complete.

natb1:
## Review: #1719 — print: spread-mode search result click double render

### Fixed

- **`print/test/viewer/shell.test.ts:839`** — Both new shell tests called `initViewer` with 5 arguments (`null` as store), missing the `fakeStore()` PositionStore stub and `null` uid. Fixed to pass `fakeStore(), null` matching the 6-arg signature used by all other call sites. (commit `7f899d41`)
- **`print/test/viewer/shell.test.ts:886`** — Same missing store/uid issue in the single-mode routing test. Fixed in same commit. (commit `7f899d41`)

### Informational

- **`print/src/viewer/types.ts:49`** — The updated `ContentRenderer` doc comment says renderers implementing search "must also implement `goToResult` and `clearSearch`, and implement `renderResult` to render the armed highlight" — but `renderResult` is declared optional on the interface. The comment and the interface are slightly inconsistent; the deferred finding below tracks the stronger fix.
- **`print/src/viewer/pdf.ts:658-660`** — `renderResult()` calls `renderPage(_currentPage)` where `_currentPage` initialises to `0`. If `renderResult()` is ever called before `goToResult()` arms it, it silently renders page 0. Not a bug in this PR's usage (shell only calls `renderResult` from the `onNavigate` callback, which fires after `goToResult`), but worth noting.
- **`print/test/viewer/shell.test.ts:879`** — The 'single mode' routing test uses `makeMockSpreadRenderer` (a renderer with `renderPageInto`) while intentionally staying in single mode. This is valid — the test is asserting that the spread-capable renderer does NOT use `renderPageInto` when not in spread mode — but the name `makeMockSpreadRenderer` may confuse a future reader.

### Deferred

- **`print/src/viewer/shell.ts:455`** — `renderer.renderResult?.()` silently no-ops if a renderer implements `search`/`goToResult` but omits `renderResult`. A future search renderer that forgets `renderResult` would produce broken single-page result clicks with no type error or runtime error. Tracked in #1817 (blocked by #1719).

### Security

No attack surface changes in non-test source. No CodeQL alerts open against this PR. No dependency changes.

natb1:
# Fix-checks summary — PR #1748

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `.claude/skills/dispatch-propagate/scripts/run-unit-tests.sh`
- **Failure excerpt:** `AssertionError: expected null not to be null` at
  `test/viewer/pdf.test.ts:787` — test
  `clearSearch() > concurrent renderPageInto() for the same page keeps one wrapper (spreadGen guard)`.
- **Why not caught:** This PR made `goToResult` arm-only (it no longer renders;
  the new `renderResult()` does the visible render and applies the armed
  highlight). Sibling tests were updated to interpose
  `await renderer.renderResult!()` before any DOM-highlight assertion, but the
  `concurrent renderPageInto()` test was missed — it asserted
  `.search-highlight.active` immediately after `goToResult`, so with arm-only
  `goToResult` the highlight was absent.
- **Fix:** Interposed `await renderer.renderResult!();` between the `goToResult`
  call and the `.search-highlight.active` assertion (and updated the adjacent
  comment to the arm-then-render pattern), matching the sibling tests. Test-only
  change to `print/test/viewer/pdf.test.ts`. Full suite: 558 tests pass.
  Commit `8b346960`.

PR #1751, opened 2026-06-15
print: guard search() against null pdfDoc before the page-fetch loop

natb1:
Closes #1730

## Problem

`search()` in `print/src/viewer/pdf.ts` reached the page-fetch loop without an
early `pdfDoc` null guard. It checked the trimmed query, then iterated
`for (let i = 1; i <= _pageCount; i++)` and called `pdfDoc!.getPage(i)` (a
non-null assertion). When `pdfDoc` was `null` while `_pageCount` was non-zero,
that assertion threw `TypeError: Cannot read properties of null`, surfacing as an
unhandled promise rejection.

The trigger is **`search()` after `destroy()`**: `destroy()` sets `pdfDoc = null`
but does not reset `_pageCount`, so the stale page count keeps the loop entering
while `pdfDoc` is already `null`. The `destroyed` flag checks inside the loop run
only *after* `await pdfDoc!.getPage(i)` resolves, so they cannot protect the
first dereference.

## Fix

Add `if (!pdfDoc) return [];` as the first statement after the trimmed-query
guard, mirroring the guard `getOutline()` already uses. `search()` now returns
`[]` immediately when no document is loaded, with no assertion thrown and no
change to normal-path behavior.

## Test

Adds a regression test to the `describe("search()")` block that calls
`init()` then `destroy()` (leaving a null `pdfDoc` with a stale `_pageCount` and
an empty `pageTextCache`) and asserts `search("the")` resolves to `[]` without
throwing. The empty cache is load-bearing: a prior `search()` would populate the
cache and let the loop skip `pdfDoc!.getPage(i)`, masking the bug. Reverting the
one-line source guard makes this test fail with a `TypeError`.

All 31 print PDF-viewer tests pass (30 pre-existing + 1 new).


natb1:
# QA Pass — Issue #1730 / PR #1751

## Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | search() returns [] immediately when pdfDoc is null | Shell-command (vitest) | PASS |
| 2 | No regression in normal search behavior when pdfDoc is non-null | Shell-command (vitest) | PASS |
| 3 | getOutline() and search() guard patterns are consistent | Deferred (needs-human-judgment) | DEFERRED |

**Counts: PASS 2 / FAIL 0 / SKIP 0 / Deferred 1**

## Test Results

All 31 print PDF-viewer tests pass, including the new regression test for the post-`destroy()` null-pdfDoc case.

```
Test Files  1 passed (1)
      Tests  31 passed (31)
   Duration  438ms
```

Command: `npx vitest run --root print test/viewer/pdf.test.ts`

## Bugs Found

None.

## Disposition Triage

One residue item entered triage (the deferred needs-human-judgment item):

| 3 | getOutline() and search() guard patterns are consistent | `opus-fixable` | n/a | Verifiable from source: both `search()` (line 549) and `getOutline()` (line 585) use the identical guard `if (!pdfDoc) return [];` before any pdfDoc use. The disposition Workflow reclassified from needs-human-judgment to opus-fixable — this is a code-structure consistency question, not a pixel-level aesthetic call. Grep confirms the patterns are structurally identical. |

`needs-main` items: none — Step 3.6 not run.

**Non-fixing pass (escalation):** The disposition Workflow classified item 3 as `opus-fixable` but `fix_plan.units` was empty (no deviation). Per Step 3.7, empty units with no deviation means planning produced nothing usable → escalating to office-hours.

> **Note for office-hours:** Source already confirms the patterns ARE consistent — `if (!pdfDoc) return [];` appears at line 549 (`search()`) and line 585 (`getOutline()`), both with identical early-return shape before any pdfDoc dereference. This escalation may be a false positive from the planning phase; a quick source review should confirm the acceptance criterion is met.

natb1:
# QA Complete — escalation overridden

Following up on the prior QA comment: the escalation has been **overridden** and QA is marked complete (`dispatch:qa-done` applied).

## Resolution

- **Item 1 (null guard returns []):** PASS — 31/31 vitest tests pass.
- **Item 2 (no normal-path regression):** PASS — all pre-existing `search()` tests green.
- **Item 3 (guard-pattern consistency, the opus-fixable residue):** **Resolved — no code change required.** The acceptance criterion is already met: `search()` (`print/src/viewer/pdf.ts:549`) and `getOutline()` (`pdf.ts:585`) use the identical guard `if (!pdfDoc) return [];` before any `pdfDoc` dereference. The disposition Workflow returned an empty fix plan, confirming there was nothing to fix — the prior escalation was a false positive from the planning phase.

No remaining items. Office-hours escalation cleared; QA done.

natb1:
## Review: #1730 — print: search() missing early null guard

**Security review:** no attack surface — diff is empty (no changed files detected). The fix for #1730 was already merged to `origin/main` via #1743 and followed up by #1746 (`print: capture pdfDoc into a local in search() to drop the non-null assertion`). This branch's working tree is identical to `origin/main` at merge base `5deac54a`.

### Fixed

_None._

### Required (security)

_None._

### Informational

- **Diff is empty — no changed files** (`N/A`) — the branch's fix commit (`cfe23271`) has already landed in `origin/main` through separate PRs; the effective diff against merge base is empty. No code review, review, or security finders ran over new code.

### Dismissed

_None._

### False-positive (security)

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1754, opened 2026-06-15
print: clear stale highlight state in goToOutlineEntry

natb1:
Closes #1745

## Problem

`goToOutlineEntry` in `print/src/viewer/pdf.ts` navigated to a new page without
clearing the search-highlight state that its sibling navigation methods
(`goToPage`, `next`, `prev`) all clear at the top of their bodies
(`pendingHighlight = null; unwrapHighlights();`). Two defects followed:

1. **Detached-node leak.** `unwrapHighlights()` drains `highlightRestores` (the
   text-layer divs mutated by a prior `applyHighlight`). Because
   `goToOutlineEntry` never called it, repeated outline navigations after a
   search accumulated detached-node references unboundedly — the same leak
   PR #1736 fixed for `goToResult`.
2. **Stale highlight re-applied on resize.** `pendingHighlight` stayed armed
   across the outline navigation. The resize handler re-renders the current
   page, and `renderPage` re-applies the highlight when
   `pendingHighlight.page === pageNum`. If the outline entry landed on the same
   page as the search result, the stale highlight reappeared.

## Fix

Add the two guard lines at the top of `goToOutlineEntry`, after the existing
`page === undefined` early return, mirroring the sibling methods. `goToResult`
is deliberately left untouched — it keeps `pendingHighlight` armed by design.

## Test

One regression test added to the existing `describe("clearSearch()")` block in
`print/test/viewer/pdf.test.ts`, which drives the real renderer end-to-end. The
outline entry deliberately resolves to the **same page** as the search result:
both `applyHighlight` call sites are gated on `pendingHighlight.page === pageNum`,
so a different-page entry would pass green even on unfixed code (vacuous). The
discriminating assertion runs immediately after `goToOutlineEntry` (before any
resize); a confirmatory resize step covers the resize criterion. Verified the
test fails at that assertion when the 2-line fix is reverted, then restored.

Full viewer suite passes (`test/viewer/pdf.test.ts`: 31/31; `test/viewer/`:
285/285).


natb1:
## QA — Issue #1745

**Result: PASS** (3 passed, 1 skipped, 0 failed)

### Items

| # | Title | Class | Result |
| 1 | Guard lines present inside `goToOutlineEntry` | script-verifiable | ✅ PASS |
| 2 | PDF viewer test suite passes (regression test + search-and-navigate) | script-verifiable | ✅ PASS |
| 3 | Broader viewer suite shows no regression | script-verifiable | ✅ PASS |
| 4 | Search then outline navigation works without visible regression | needs-browser | ⏭ SKIP |

### Details

**Item 1 — Guard lines grep**: Both `pendingHighlight = null;` and `unwrapHighlights();` are present in `goToOutlineEntry` after the `page === undefined` early return, mirroring the sibling methods.

**Item 2 — `pdf.test.ts` 31/31**: All tests pass including the new regression test in `describe("clearSearch()")`. The regression test arms a highlight on page 1, calls `goToOutlineEntry` to the same page, and asserts `.search-highlight` is null both immediately after navigation and after a simulated resize.

**Item 3 — Viewer suite 285/285**: No regressions across all 10 viewer test files. The stderr blob/network errors are pre-existing happy-dom noise from the epub test environment; all assertions passed.

**Item 4 — Browser walkthrough (SKIP)**: QA server started at `http://localhost:5172`. The PDF viewer loaded, search returned 9 results for "Socrates", and navigating to a result displayed an orange highlight (search-and-navigate golden path confirmed). The outline panel was empty in both seeded test PDFs (`plato-republic` and `plato-phaedrus`) — the QA environment uses stub PDFs without embedded TOC entries, so `goToOutlineEntry` could not be triggered via browser UI. The `goToOutlineEntry` code path is definitively covered by the unit tests in Item 2 (including the discriminating same-page regression test). No console errors observed during the walkthrough.

GIF recorded: `tmp/qa-fix-walkthrough-1745.gif`

### Acceptance criteria

- ✅ `goToOutlineEntry` calls `unwrapHighlights()` before rendering (grep + unit test)
- ✅ `goToOutlineEntry` sets `pendingHighlight = null` before rendering (grep + unit test)
- ✅ After outline nav, `highlightRestores` is empty (unit test — same-page discriminating assertion)
- ✅ After outline nav, resize does not re-apply stale highlight (unit test — resize step)
- ✅ Existing search behavior unaffected (unit tests pass + browser confirmed orange highlight on `goToResult`)

natb1:
## Review — #1745 (`1745-print-gotooutlineentry-does`)

### Fixed (cf888275)

- **`print/test/viewer/pdf.test.ts:604`** — `vi.useFakeTimers()` was called without a `try/finally` guard. If an assertion in the timer block threw, `vi.useRealTimers()` would never run, leaving subsequent tests with a broken timer environment. Wrapped the timer block in `try/finally` so `vi.useRealTimers()` always fires.

- **`print/test/viewer/pdf.test.ts:585`** — `fakeDoc.getOutline` was mutated inline and restored only at the bottom of the test body. If any assertion between mutation and restore threw, later tests would inherit the mutated mock. Restore moved into the `finally` clause of the same `try/finally` block added above.

- **`print/src/viewer/pdf.ts:511`** — `goToPosition()` was the only remaining navigation method missing the `pendingHighlight = null; unwrapHighlights();` guards at its top. This PR added those guards to `goToOutlineEntry`; `goToPosition()` (bookmark-position navigation) had the same omission. Added the two guard lines to mirror `goToOutlineEntry`, `goToPage`, `next`, and `prev`.

### False-positive (security)

- **`print/test/viewer/pdf.test.ts:84`** — `__fakeDoc` added to the `pdfjs-dist` mock was flagged as a potential shared-state exposure. This is test-only infrastructure: `pdfjs-dist` is mocked entirely in `vi.mock()` scope, `__fakeDoc` is only reachable from test code, and there is no production surface. Not a real issue.

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1769, opened 2026-06-16
print: goToPosition clears pendingHighlight and drains highlightRestores

natb1:
Closes #1744

## What

`goToPosition` is the bookmark / external-caller navigation entry point in
`print/src/viewer/pdf.ts`. Its sibling manual-navigation methods `goToPage`,
`next`, and `prev` each discard any active search highlight before rendering by
running `pendingHighlight = null; unwrapHighlights();` at the top of the method.
`goToPosition` was the lone navigation path that omitted both calls.

This change adds those two lines (plus the matching one-line comment) to the top
of `goToPosition`, mirroring its siblings.

## Why

When a search result was highlighted (via `goToResult`) and the user then jumped
via a bookmark, two defects appeared:

1. **Detached-node leak** — `renderPage` rebuilds the text layer, detaching the
   previously-highlighted divs, but the closure-private `highlightRestores` array
   kept referencing those now-detached nodes, growing without bound across
   navigations.
2. **Stale highlight reappears** — `pendingHighlight` stayed armed, so the next
   resize-triggered re-render re-applied the old search highlight on the
   bookmarked page.

`parsePositionPage` never throws and always returns a valid page, so
top-of-method placement has no early-return hazard — the discard is
unconditional, exactly as in `goToPage`.

This is the analogous fix to #1736, which fixed the same omission in the *arming*
path `goToResult`; the navigation path was explicitly filed as this follow-up.

## Test

Adds one regression test to the `describe("clearSearch()")` block in
`print/test/viewer/pdf.test.ts`, mirroring the #1736 `goToResult` drain test. It
arms a highlight, jumps via `goToPosition`, and asserts the previously-highlighted
div was restored (no leaked references) and no stale highlight survives. The test
fails on the pre-fix `goToPosition` and passes with the fix; the full `print`
suite stays green and the build typechecks.

### DIFF
--- a/print/src/viewer/pdf.ts
+++ b/print/src/viewer/pdf.ts
@@ -509,6 +509,10 @@ export function createPdfRenderer(onError?: (err: unknown) => void): ContentRend
     },
 
     async goToPosition(position: string): Promise<void> {
+      // Manual navigation discards any active search highlight so a stale
+      // highlight never reappears on the new page.
+      pendingHighlight = null;
+      unwrapHighlights();
       const page = parsePositionPage(position, _pageCount);
       _currentPage = page;
       await renderPage(page);
--- a/print/test/viewer/pdf.test.ts
+++ b/print/test/viewer/pdf.test.ts
@@ -597,4 +597,25 @@ describe("clearSearch()", () => {
     // Exactly one highlight is live in the DOM (the current result).
     expect(container.querySelectorAll(".search-highlight.active").length).toBe(1);
   });
+
+  it("goToPosition() drains the active highlight and disarms pendingHighlight", async () => {
+    const renderer = createPdfRenderer();
+    await renderer.init(container, "fake://source.pdf");
+
+    const results = await renderer.search!("the");
+    const r1 = results.find((r) => r.label === "Page 1")!;
+    await renderer.goToResult!(r1);
+
+    // Capture the text-layer div holding the highlight before navigation.
+    const div1 = container.querySelector(".search-highlight")!.parentElement!;
+    expect(div1.querySelector(".search-highlight")).not.toBeNull();
+
+    // goToPosition must unwrap the highlight and disarm pendingHighlight.
+    await renderer.goToPosition("2");
+
+    // The highlight span is removed from the previously-live text div.
+    expect(div1.querySelector(".search-highlight")).toBeNull();
+    // No highlight exists anywhere in the container.
+    expect(container.querySelectorAll(".search-highlight").length).toBe(0);
+  });
 });


natb1:
## QA Summary — #1744 / PR #1769

**Result: Clean pass — all 5 items passed, no residue.**

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | goToPosition() drains active highlight and disarms pendingHighlight (regression test) | script-verifiable | ✅ PASS |
| 2 | Full print suite stays green with the fix | script-verifiable | ✅ PASS |
| 3 | goToPosition clears pendingHighlight and calls unwrapHighlights before renderPage (source check) | script-verifiable | ✅ PASS |
| 4 | Bookmark jump after a search highlight leaves no stale highlight | browser walkthrough | ✅ PASS |
| 5 | Resize after a bookmark jump does not resurrect the search highlight | browser walkthrough | ✅ PASS |

### Details

**Script-verifiable lane:**
- `npx vitest run --root print test/viewer/pdf.test.ts` — 38 tests passed, including the new regression test that arms a highlight via `goToResult`, jumps via `goToPosition("2")`, and asserts the highlight div was restored and no `.search-highlight` survives.
- `npx vitest run --root print` — 538 tests passed across 31 files; blob/network stderr from happy-dom is a known environment warning, not a failure.
- Source grep confirmed both `pendingHighlight = null;` and `unwrapHighlights();` are present at the top of `goToPosition` in `print/src/viewer/pdf.ts`.

**Browser walkthrough (plato-phaedrus PDF, public seed, Windows Chrome → WSL2 QA server):**
- Searched "Socrates", clicked first result — `goToResult` armed the highlight (2 `.search-highlight` elements confirmed in DOM).
- Clicked saved bookmark "Page 1 / 1" — triggered `goToPosition` — verified 0 `.search-highlight` elements immediately after.
- Dispatched `resize` event with debounce wait — verified 0 `.search-highlight` elements; no stale highlight reappeared.
- No console errors, no network errors.

GIF: `tmp/qa-fix-walkthrough-1744.gif`

**Environment note:** The local Playwright acceptance runner could not install its browser (NixOS; `apt-get` not available — known limitation). CI is passing and is authoritative for the acceptance suite.

natb1:
## Review — #1744 (`goToPosition` highlight cleanup)

### Fixed

- **RV-003** `print/test/viewer/pdf.test.ts:614-618` — Concern that the test only checked `div1` for highlight removal was found to be a false positive: `container.querySelectorAll('.search-highlight').length === 0` already covers the full container. No code change needed.

### Informational

- **RV-001** `print/src/viewer/pdf.ts:511-513` — The comment explaining the discard is correct; a minor phrasing suggestion: noting it mirrors `goToPage` would add traceability. No change required.
- **SEC-001** `print/src/viewer/pdf.ts, print/test/viewer/pdf.test.ts` — No exploitable security vulnerabilities in the navigation/highlight management diff. Security surface reviewed: input-validation, red-team, auth, data-exposure, firebase. All clean.

### Dismissed

- **RV-002** `print/test/viewer/pdf.test.ts:605-608` — Non-null assertions (`!`) on `container.querySelector` results. Correct given test preconditions: `goToResult` guarantees the highlight exists before the assertions run. Adding null guards would obscure intent.

---

_No fixes applied. No follow-up issues filed._

PR #1699, opened 2026-06-15
Instrument apps with web-vitals -> GA4 events for field RUM

natb1:
Closes #1493

## What

Instrument every app with [web-vitals](https://www.npmjs.com/package/web-vitals)
Core Web Vitals reporting, routed through GA4 as a stable `web_vitals` event, so
GA4 becomes a machine-readable source of web field RUM (`/roadmap` already reads
GA4 via `fetch-analytics.sh`). This is the **producer** side; the reader side is
#1494 (blocked by this).

The five reporters (`onLCP`, `onCLS`, `onINP`, `onFCP`, `onTTFB`) are registered
inside `analyticsutil`'s `initAnalytics`, right after the existing
`setUserProperties(traffic_type)` call and before the page-view tracker is
returned. Each metric is emitted as one `logEvent(analytics, "web_vitals", …)`,
guarded by the same `classifyError` + `reportError` pattern the page-view tracker
uses.

## Implementation location — `analyticsutil`, not `firebaseutil/app-context.ts`

AC #1 names `firebaseutil/app-context.ts`, but the change lands in
`analyticsutil/src/index.ts`. This is deliberate and meets AC #1's intent (shared,
automatic, every app, no per-app change):

- `createAppContext` only receives `initAnalyticsSafe`'s `(path) => void`
  page-view tracker back; the Firebase `analytics` instance never leaves
  `initAnalytics`. There is no handle in `app-context.ts` to call `logEvent` for
  web-vitals without exposing or re-initializing the instance — strictly worse.
- `analyticsutil` is the package every app already uses for analytics (via
  `firebaseutil` → `createAppContext`), so wiring web-vitals there reaches all
  apps (landing, budget, print, audio, office-hours, fellspiral) with no per-app
  change.

`web-vitals` is added as a **direct dependency** of `analyticsutil` (not a peer,
unlike `firebase`): it has no app-owned singleton, so apps pick it up through the
npm-workspace hoist with no per-app dependency change.

## GA4 event contract (pinned with #1494)

One stable event, with the metric carried as a parameter (so the reader issues a
single Data API query and groups by `metric_name`):

- **Event:** `web_vitals`
- **Parameters:** `metric_name` (`LCP|CLS|INP|FCP|TTFB`), `metric_value`
  (`Math.round`; CLS scaled ×1000 per the standard recipe, others ms),
  `metric_rating` (`good|needs-improvement|poor`), `metric_id` (per-page-load id
  for dedup/aggregation).
[...]

PR #1776, opened 2026-06-16
roadmap: surface GA4 web-vitals field metrics in roadmap context

natb1:
Closes #1494

Surfaces the GA4 `web_vitals` field-RUM signal (the producer half landed in
#1493, PR #1699) in `/roadmap`, so the Engineering persona can weigh real-world
Core Web Vitals performance. This completes the field-RUM side of the
web-performance epic #260; the lab side (PageSpeed Insights, #1492) is
independent and not coupled here.

## What changed

- **`fetch-analytics.sh` — new block (d) Web Vitals field metrics.** Inside the
  per-app GA4 loop, a single `runReport` query grouped by the
  `customEvent:metric_name` / `customEvent:metric_rating` custom dimensions and
  the `customEvent:metric_value` custom metric, filtered to the `web_vitals`
  event. jq re-aggregates the grouped rows into a per-metric
  `eventCount`-weighted average and a `% good` rating share, emitting one line
  per metric (LCP/CLS/INP/FCP/TTFB). CLS is unscaled ÷1000 (the producer stored
  it ×1000 as an integer) and shown unitless; the others are labelled `ms`.
  Reuses the established curl + three-way error branch (`api_error_reason`) and
  the `gsub("[\r\n]"; " ")[0:200]` sanitization from the sibling blocks.

- **`roadmap-engineering.md` — Web Performance (Field RUM) lens.** A new thinking
  framework directs the Engineering persona to read the per-app web-vitals field
  metrics in the analytics context and weigh the averages and `% good` share as a
  real-world performance-health signal, treating absent data as absent rather
  than as good performance.

- **`roadmap/SKILL.md` — Analytics setup docs.** Documents the new field metrics
  and the **one-time GA4 custom-definitions registration** (event-scoped
  `metric_name` / `metric_rating` dimensions + `metric_value` metric in each
  property) required before the metrics become queryable, plus the
  not-retroactive / ~24–48h-latency caveat and the graceful-degradation
  behavior.

## Design notes

Two GA4 Data API constraints shaped the design:

1. **Custom event parameters are not queryable until registered** as custom
   definitions in each property, and registration is not retroactive. Until the
   operator registers them, `runReport` returns an API error — absorbed by the
   existing degradation contract (inline note, script still `exit 0`).

2. **`runReport` has no percentile aggregation** (only TOTAL/MIN/MAX/COUNT), so
   the issue's "p75" is infeasible through the Data API. This delivers
   **average + a good/needs-improvement/poor rating distribution** instead — the
   rating share maps directly to the Core Web Vitals standard. True p75 would
   require the GA4→BigQuery export, out of scope for a `fetch-analytics.sh`
   extension.

## Verification

`fetch-analytics.sh` has no test harness and is not in CI; live GA4 verification
is not chain-verifiable (no autonomous creds, definitions likely unregistered,
24–48h latency). Verified mechanically: `bash -n` passes; the new `jq -n`
request body emits valid JSON; the aggregation jq program run against synthetic
responses produces correct weighted averages, the CLS unscale, and `% good`
shares, and the zero-row path emits the degradation note.


natb1:
# QA Summary — Issue #1494 / PR #1776

## Results

**10 items executed** (9 script-verifiable, 1 needs-human-judgment)

| Result | Count |
|--------|-------|
| PASS   | 9     |
| FAIL   | 0     |
| SKIP   | 0     |

No browser walkthrough (no frontend component changed). No GIF.

## Items

1. **PASS** — `fetch-analytics.sh` parses without syntax errors (`bash -n`)
2. **PASS** — Web Vitals request body queries GA4 for `web_vitals` events (correct dimensions, metrics, date range)
3. **PASS** — Per-app field metrics aggregate correctly: weighted average, `% good` share, and CLS ÷1000 unscaling verified against synthetic data (CLS: 0.5 avg, 100% good; LCP: 14400 ms avg, 80% good)
4. **PASS** — Empty/missing `rows` response triggers the absence-note branch, not a crash
5. **PASS** — Absence-note string present in script at line 250
6. **PASS** — Script exits 0 with credentials unset (degradation contract preserved)
7. **PASS** — Producer header string `Web Vitals (field RUM, 30-day)` (script line 252) matches the exact string the Engineering persona is directed to find (roadmap-engineering.md line 32)
8. **PASS** — Engineering persona has the Web Performance (Field RUM) thinking framework naming all five metrics (LCP, CLS, INP, FCP, TTFB) and the absent-data caveat
9. **PASS** — `SKILL.md` documents the field metrics, all three custom-definition param names, the not-retroactive/24-48h latency caveat, and graceful degradation
10. **Deferred (needs-human-judgment)** — CLS scale factor (÷1000) matches the #1493 producer's emit encoding — see Disposition Triage below

## Disposition Triage

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR. (No `needs-main` items this pass.)

| 10 | CLS scale factor (÷1000) matches the #1493 producer's emit encoding | `opus-fixable` | n/a |

**Item 10 rationale (classifier):** The #1493 producer is vendored in this repo at `analyticsutil/src/index.ts`. The classifier found `const CLS_SCALE = 1000` (line 22) and `metric.value * CLS_SCALE` for CLS (line 80). The consumer divides by 1000 — the scale factors are balanced and the code is correct. However, the gated fix-plan phase returned no units (planning-failed: no usable fix plan emitted). Per protocol, an unfixed `opus-fixable` item escalates regardless.

**Note:** Independent verification during this run confirmed `CLS_SCALE = 1000` at `analyticsutil/src/index.ts:22` and `metric.name === "CLS" ? metric.value * CLS_SCALE : metric.value` at line 80. The human reviewer can fast-confirm and clear this item — the code is correct.

**Terminal behavior (non-fixing pass — planning-failed):** The gated fix-plan returned no units for the `opus-fixable` item. Protocol requires escalation. Item 10 escalates to office-hours; `dispatch:qa-done` is not applied.

natb1:
## Review — PR #1776

### Fixed

- **`fetch-analytics.sh:273` — GA4 averaging formula incorrect** (`code-review`): `customEvent:metric_value` returns per-cell SUM from the GA4 Data API, not per-cell average — the `* .count` multiplication double-counted event weight. Fixed to grand-sum ÷ total-event-count; updated the comment on line 227 to reflect the correct aggregation model. (68d431a9)
- **`SKILL.md:175` — GA4 setup docs missing aggregation type** (`code-review`): Setup instructions for `metric_value` now specify "Standard (SUM)" measurement type and integer unit — prevents silent misconfiguration where a user registers AVG and gets wrong averages from the API. (68d431a9)
- **`fetch-analytics.sh:256` — Display jq pipeline unguarded under `set -euo pipefail`** (`code-review`): The `printf | jq` pipeline at the else-branch ran without error absorption; a jq error (e.g. null metricValues fields) would abort the script, violating the exit-0 degradation contract. Fixed by null-guarding the `metricValues` field accesses (`// "0"` fallback before `tonumber`). (68d431a9)
- **`fetch-analytics.sh:277` — Dead CLS formula `/1000 * 1000` cancels to no-op** (`code-review`, `review`): `(($avg / 1000) * 1000 | round) / 1000` was an identity on `$avg` before the final `/1000` — the inner `/1000 * 1000` pair cancels. Replaced with `($avg | round) / 1000`. (68d431a9)

- **`fetch-analytics.sh:262` — `tonumber` on non-numeric GA4 metric values** (`red-team`): Verdict: **refuted** (1 upheld, 1 refuted — majority refuted). The finding correctly identifies the `set -e` abort mechanism, but the triggering premise does not hold: the GA4 Data API guarantees metric value fields are always numeric decimal strings; `"(not set)"` is a dimension sentinel only. The `elif \! jq -e '.rows // [] | length > 0'` guard already ensures the `tonumber` pipeline only runs on a valid, non-empty response. An attacker emitting non-numeric `metric_value` event parameters cannot cause the GA4 server-side aggregator to return a non-numeric string in a metric value field. The abort mechanism is real but the activation path is not reachable under normal or adversarial GA4 operation.

### Informational

- **`fetch-analytics.sh:231` — GA4 API aggregates `customEvent:metric_value` as SUM per row** (`review`): Consistent with the code-review-0 fix; the query correctly groups by metric_name + rating and the grand-sum formula is now the implementation.
- **`fetch-analytics.sh:233` — `limit: 100` is sufficient for 5 CWV × 3 ratings = 15 rows** (`review`): No change needed; the limit is adequate.
- **`SKILL.md:175` — GA4 standard properties have a 50 custom metric/dimension quota** (`review`): Worth noting for operators with dense GA4 properties, but not a blocker for this use case (2 dimensions + 1 metric = 3 of 50).

### False-positive (security)

- **`fetch-analytics.sh:260` — GA4 dimension values from client-emitted events** (`input-validation`): Client-controlled dimension strings are sanitized (`gsub("[\r\n]"; " ")[0:200]`) before display and the jq display pipeline does not execute shell code — no code injection path exists.

---
_No deferred findings. No security follow-ups. No deviation._

PR #1778, opened 2026-06-16
scaffolding: explicit per-app SCAFFOLD markers for firestore.rules; stop re-sorting workspaces

natb1:
Closes #1309

## What

Replace the heuristic `firestore.rules` text editing in
`scaffolding/firebase/internal/scaffold/` with explicit per-app
`// SCAFFOLD BEGIN <app>` / `// SCAFFOLD END <app>` markers, and stop
`AddWorkspace` from re-sorting the `workspaces` array on every scaffold.

## Why

Three confirmed defects in the scaffold's file-editing helpers:

1. **Brace counting included comments/strings** (`cleanup.go`):
   `RemoveFirestoreRules` skipped an app's block by counting `{`/`}` over every
   character. An unbalanced brace inside a comment or string would swallow
   subsequent blocks (other apps' rules, the catch-all marker) until rebalanced.

2. **Insert/remove detection asymmetry**: insert keyed on a whole-file
   `strings.Contains("match /<app>/")` while remove keyed on a line-prefix
   `match /<app>/`. A stray comment mentioning `match /myapp/` made insert skip
   while remove found nothing, so `Create` could report success having added no
   rules.

3. **`AddWorkspace` re-sorted** (`firebase.go`): `sort.Strings` reordered the
   whole `workspaces` array on every scaffold, discarding the deliberate
   `config`-first ordering in `package.json` and producing a noisy unrelated
   diff.

## How

- Insert wraps the generated block in `// SCAFFOLD BEGIN/END <app>` markers, and
  both insert's idempotency guard and remove's span detection now key on the same
  BEGIN marker — they are exact inverses (a byte-for-byte round-trip test guards
  this).
- Remove drops the `[BEGIN..END]` span inclusive plus the single trailing blank
  line, replacing the brace-counting loop. A three-case boundary keeps behavior
  safe for pre-marker apps: BEGIN present → span-remove; no marker but
  `match /<app>/` prefix lines present → clear error (rules exist without
  markers, manual removal required); neither → unchanged no-op. Existing app
  blocks (landing, budget, fellspiral, print, audio, office-hours) have no
  markers and cannot be retrofitted (their match blocks are non-contiguous), so
  marker-based removal governs only newly-scaffolded apps and the error branch
  surfaces the gap loudly rather than silently mis-editing.
- `AddWorkspace` appends without `sort.Strings`, preserving the existing order.

## Tests

`go test ./...` from `scaffolding/firebase` passes, including the byte-exact
insert→remove round-trip guard, a regression test that a stale `match /myapp/`
comment no longer makes insert skip, the no-marker error-boundary test, and an
`AddWorkspace` test asserting `config` stays first.

### DIFF
--- a/scaffolding/firebase/internal/scaffold/cleanup.go
+++ b/scaffolding/firebase/internal/scaffold/cleanup.go
@@ -8,10 +8,17 @@ import (
 	"strings"
 )
 
-// RemoveFirestoreRules removes all rule blocks for appName from firestore.rules.
-// Rules are identified by path pattern (match /<appName>/...) rather than markers.
-// Each removed block also strips immediately preceding comment lines and adjacent blank lines.
-// If no rules are found, it logs a note and returns nil (not an error).
+// RemoveFirestoreRules removes the SCAFFOLD-marked rule block for appName from
+// firestore.rules. The block is the contiguous span between the
+// "// SCAFFOLD BEGIN <app>" and "// SCAFFOLD END <app>" marker lines (inclusive),
+// plus the single blank line that immediately follows END.
+//
+// Three boundary cases:
+//  1. BEGIN marker present  → the marker span is deleted.
+//  2. No BEGIN marker, but lines with prefix "match /<app>/" present → returns an
+//     error, because the rules exist without SCAFFOLD markers and so must be
+//     removed manually (this is the case for pre-marker apps).
+//  3. Neither marker nor prefix → logs a note and returns nil (not an error).
 func RemoveFirestoreRules(repoRoot, appName string) error {
 	rulesPath := filepath.Join(repoRoot, "firestore.rules")
 	content, err := os.ReadFile(rulesPath)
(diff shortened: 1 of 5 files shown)


natb1:
# QA Summary — Issue #1309 / PR #1778

**Result: CLEAN PASS — all items passed.**

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Full Go test suite passes | shell-command | PASS |
| 2 | AddWorkspace no longer re-sorts (`sort.Strings` absent) | shell-command | PASS |
| 3 | Insert and remove keyed on same explicit per-app marker token | shell-command | PASS |
| 4 | Marker-less app removal fails loudly (three-case boundary) | shell-command | PASS |

**Counts: 4 PASS / 0 FAIL / 0 SKIP**

## Detail

**Item 1 — Full Go test suite:** `cd scaffolding/firebase && go test ./...` → `ok` for both `scaffolding/firebase` and `scaffolding/firebase/internal/scaffold`. Exercises the round-trip insert→remove guard, `TestInsertNotSkippedByStaleComment`, `TestAddWorkspaceAppendPreservesOrder`, "errors when rules exist without markers", and BEGIN/END marker assertions in `TestInsertFirestoreRules`.

**Item 2 — `sort.Strings` absent:** `grep -nE 'sort\.|"sort"' internal/scaffold/firebase.go` exits 1 (not found). Both the `sort.Strings` call and the `"sort"` import are gone from `firebase.go`.

**Item 3 — Same marker token in insert and remove:** Both `create.go` and `cleanup.go` reference `scaffoldBeginMarker`/`scaffoldEndMarker` (defined in `create.go:183-184`). No `depth++`/`depth--` lines remain in `cleanup.go` (brace-counting loop fully removed).

**Item 4 — Three-case boundary in `RemoveFirestoreRules`:** `cleanup.go:47` returns `fmt.Errorf("rules for %q exist in firestore.rules but have no SCAFFOLD markers; remove the block manually", appName)` for the marker-less-but-present case. The genuinely-absent case (`cleanup.go:51`) returns `nil` with a log note. Fails loudly per the PR's promise.

## Acceptance criteria

- [x] Rules insert/remove use explicit per-app span markers (no brace counting, no prefix/Contains asymmetry).
- [x] `AddWorkspace` appends without reordering existing entries.

natb1:
## Review — #1309 scaffolding: replace heuristic firestore.rules editing

CodeQL scan: clean (0 open alerts). No dependency changes.

---

### Fixed

_None._

### Required (security)

_None._

### Informational

- **`create.go:198` — idempotency prefix sensitivity** (`code-review`, `auth`, `firebase`, `review`): `InsertFirestoreRules` guards duplication with `strings.Contains(content, "    "+scaffoldBeginMarker(appName)+"\n")` — a hardcoded 4-space indent and a literal `\n` suffix. A rules file using tabs or CRLF line endings would bypass the check and double-insert. Works correctly for the project's current formatting but is fragile.

- **`rules_test.go` — missing test: BEGIN without END** (`code-review`, `review`): The error path at `cleanup.go:61–63` (SCAFFOLD BEGIN present, no matching SCAFFOLD END) is not exercised by the test suite. A malformed rules file hits a clear error message but that branch has no coverage.

- **`create.go:198` — idempotency check indentation hardcode** (`auth`, `firebase`, `review`): Same as the first item above; surfaced independently by multiple reviewers. The 4-space prefix assumption is load-bearing; a lint-reformatted file would break the guard.

### Dismissed

- **`create.go:183` — marker helpers defined in create.go, used in cleanup.go** (`review`): Observation that `scaffoldBeginMarker`/`scaffoldEndMarker` are defined in `create.go` but consumed in `cleanup.go`. Both files are in the same package (`scaffold`), so there is no visibility issue. File placement within a package is a style choice, not a correctness concern. Dismissed as non-actionable.

### False-positive (security)

- **`create.go:183–184` — appName unsanitized in marker string** (`input-validation`): `scaffoldBeginMarker`/`scaffoldEndMarker` concatenate `appName` directly. `appName` is validated upstream before these helpers are ever called; no untrusted input reaches them. False-positive.

- **`create.go:203–226` — appName as Firestore path segment** (`input-validation`): `fmt.Sprintf` interpolates `appName` into the generated rules block. `appName` is a scaffold-time developer-supplied identifier validated by `ValidateAppName` before `InsertFirestoreRules` is called. Not an injection surface. False-positive.

- **`create.go:189` — InsertFirestoreRules exported without ValidateAppName call** (`auth`): The function is exported but all callers in the scaffold pipeline validate `appName` before calling it. The scaffold tool is a developer CLI, not a public API surface. False-positive.

- **`cleanup.go:22` — RemoveFirestoreRules exported without ValidateAppName call** (`auth`): Same reasoning as above. Developer CLI; callers validate app name upstream. False-positive.

### Deferred

_None._

### Out-of-scope (security)

- **`create.go:211–212` — scaffolded template: `allow read: if true` on messages** (`auth`, `firebase`): The default rules template grants unauthenticated read to `/{app}/{env}/messages/{messageId}`. This is pre-existing template behavior not introduced by this PR (the template content is unchanged). Out of scope; no security follow-up filed (template-level policy, not a code defect).

---

No code changes applied (all findings were Informational, Dismissed, False-positive, or Out-of-scope).

PR #1624, opened 2026-06-14
office-hours: panel registry + responsive grid layout

natb1:
Closes #1465

## What

Refactors the office-hours dashboard from two hard-coded per-tier panel branches
into a **panel registry** composed into a **responsive CSS grid**, and wires the
already-scaffolded `queue-metrics` panel so it renders in both tiers.

`renderApp` previously composed its dashboard by hard-coding four panel renders
behind sequential `appendChild` calls — one branch for the demo tier, a
near-duplicate branch for the owner tier. That fixed composition is now a
registry of `Panel` entries rendered into a `.panel-grid`. This is the
foundational, no-new-data step for the multi-panel dashboard epic (#1464).

## How

Built in four commits:

1. **Panel registry + `renderApp` refactor + `.panel-grid` CSS.** A
   `definePanel<T>` factory binds each panel's `load`→`render` input type at the
   definition site and type-erases to a uniform `Panel`. `buildContext` resolves
   the per-tier data centrally; `renderApp` filters `PANELS` by `availableIn`,
   renders each into a `.panel-grid` (`grid-template-columns: repeat(auto-fit,
   minmax(min(100%, 22rem), 1fr))` — multi-column when wide, single column under
   ~22rem, no JS, no `@media`), and tags full-width panels with `panel-grid-full`.
   The four existing panels render unchanged; the history band spans full width.

2. **`renderQueueMetricsPanel` render fn + CSS + unit test.** A standalone render
   fn for the dispatch-queue snapshot (queue depth, net drain/day, runway), with
   an `.empty` placeholder for the null case, mirroring the existing per-panel
   render pattern.

3. **Wire queue-metrics live.** Adds `getOwnerQueueMetrics` (a single-doc `getDoc`
   read of `office-hours/{env}/metrics/dispatch-queue` — the first client read of
   this doc), extends the owner `ViewState` with `queueMetrics`, registers the
   panel, and loads it via `main.ts`'s `Promise.all`. The panel now renders in
   both demo and owner tiers — the acceptance criterion. The security rule already
   exists (`match /office-hours/{env}/metrics/{metricId}`), so no rules change is
   needed.

4. **Registry-composition + integration tests.** Asserts the `.panel-grid`
   container exists in demo/owner tiers and not in error, all registered panels
   render per tier, history spans full width, and the queue-metrics panel renders
   in both tiers (plus its empty placeholder for null owner metrics).

## Design notes

- **`renderApp` never renders a panel's `title`.** Each render fn already emits
[...]

PR #1802, opened 2026-06-16
office-hours: explicit never exhaustiveness guard in renderApp

natb1:
Closes #1387

## What

Add an explicit, self-documenting `never` exhaustiveness guard to `renderApp` in
`office-hours/src/app-view.ts`. After the existing `error` early-return and the
`demo` banner block, a new `else if (state.tier !== "owner")` branch asserts
`const _exhaustive: never = state` before the shared `buildContext`/grid path.

## Why

This is defensive hardening. The original behavioral bug the issue was filed for
— a bare `else` that silently absorbed the `error` tier (and any future
`ViewState` variant) into the error UI — was already removed when PR #1624
restructured `renderApp` to handle `error` via an explicit early-return and
share a panel-grid path for `demo`/`owner`.

The remaining intent is to make exhaustiveness **explicit at the dispatch site**.
Today the new variant guard is caught only incidentally by `buildContext`'s
`Extract<ViewState, { tier: "demo" } | { tier: "owner" }>` parameter. The
explicit `never` guard documents the exhaustiveness intent and survives a future
refactor that widens `buildContext`'s parameter: adding a fourth `ViewState`
variant without a branch becomes a compile error at the `const _exhaustive: never
= state` line.

## How it type-checks today

After the `error` early-return and the consumed `demo` branch, `state` is
narrowed to `{ tier: "owner" }` in the `else if` body, so `state.tier !== "owner"`
is statically `false` and `state` narrows to `never` — the assignment
type-checks. The `else if` body is unreachable at runtime for current tiers, so
there is no behavior change for `demo`, `owner`, or `error`.

## Verification

- `npm run build --prefix office-hours` — passes, no type errors.
- `npx vitest run --root office-hours` — 234 tests across 25 files pass; the
  existing `app-view.test.ts` error/demo/owner tier blocks confirm no behavior
  change.


natb1:
## QA Pass — #1387 `never` exhaustiveness guard in `renderApp`

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Exhaustiveness guard present in correct position | shell-command | ✅ PASS |
| 2 | TypeScript compilation passes with the never guard | shell-command | ✅ PASS |
| 3 | Existing unit tests remain green | shell-command | ✅ PASS |

**Counts: 3 PASS · 0 FAIL · 0 SKIP**

### Details

**Item 1 — Guard presence:** `grep -n "_exhaustive" office-hours/src/app-view.ts` found the guard at lines 141–144. It is positioned after the `error` early-return (lines 128–133) and the `demo` branch (lines 135–141), and before the shared `buildContext(state, now)` call (line 146). The guard form is `else if (state.tier \!== "owner") { const _exhaustive: never = state; throw new Error(...) }`, exactly as specified.

**Item 2 — TypeScript build:** `npm run build --prefix office-hours` passed with no type errors (`tsc && vite build` — 720 modules transformed, built in 2.33s). The pre-existing chunk-size warning is unrelated to this change.

**Item 3 — Unit tests:** `npx vitest run --root office-hours` — 234 tests across 25 files passed (2.06s). No behavior change for any current tier (error/demo/owner).

### Disposition

All acceptance criteria satisfied. No residue, no bugs found. No browser walkthrough required (this is a type-only change with no runtime behavior change for any current `ViewState` tier).

natb1:
## Review — PR #1802

Diff: `office-hours/src/app-view.ts` (+3 lines, exhaustiveness guard). Surface: `code`. CodeQL: no open alerts. No dependency changes.

---

### Fixed

_None._

### Required (security)

_None._

### Informational

- **`office-hours/src/app-view.ts:143`** — `String(_exhaustive)` in the error message where `_exhaustive` is typed `never`. TypeScript narrows to `never` statically, so this is fine at compile time. At runtime any rogue value reaching this branch (e.g. a future tier added without a branch) would serialize as `[object Object]` or similar rather than the tier name. Consider `(state as { tier: string }).tier` or equivalent for a more diagnostic message. No change required; surfaced for awareness. _(sources: review, red-team)_

### Dismissed

_None._

### False-positive (security)

- **`office-hours/src/app-view.ts:141-143`** — Claim: the synchronous `throw` from `renderApp` is a denial-of-service vector when an unhandled tier is encountered. Rationale: `renderApp` already threw (or the code would have been unreachable via `buildContext`); the guard makes the failure loud and traceable rather than silent. The throw is the correct behavior for a programming error, not an exploitable path. _(source: red-team)_

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1804, opened 2026-06-16
idbutil: cover negative-delta replace + document unreachable negative-delta eviction

natb1:
Closes #1417

## What

PR #1414 added eviction-on-replace-delta logic to the IndexedDB LRU blob cache
and a test (`replace evicts on the size delta, not gross incoming size`). That
test only exercises a **positive** delta — an entry grows and eviction is still
needed. The **negative-delta** branch, where a replacement is *smaller* than the
original (`delta < 0`), had no coverage.

This PR:

- Adds a regression test, `replace with a smaller blob shrinks the total and
  evicts nothing`, to the `LRU eviction` block in
  `idbutil/test/lru-blob-cache.test.ts`. It seeds an eviction (`a` at 180, `b` at
  90 with a 200-byte cap evicts `a`), then replaces `b` with a smaller 60-byte
  blob and asserts the total shrinks to 60, no further eviction occurs, and only
  `b` remains.
- Adds a comment at the guard/delta site in `evictIfNeeded`
  (`idbutil/src/lru-blob-cache.ts`) recording *why* a negative delta cannot
  trigger eviction for a sequential single-writer: with `delta < 0` and the
  `totalSize ≤ maxBytes` invariant, `totalSize + delta < maxBytes`, so the guard
  always returns early. The eviction body is reachable on a negative delta only
  under concurrent cap drift (the existing best-effort window).

## Behavior

No production logic changes — only a test addition and an explanatory comment.

## Verification

`npx vitest run --root idbutil lru-blob-cache` — all green (47 tests), including
the new negative-delta test alongside the existing eviction tests.


natb1:
# QA Summary — PR #1804 / Issue #1417

## Results

| # | Item | Result |
|---|------|--------|
| 1 | New negative-delta replace test passes | ✅ PASS |
| 2 | Test exercises the correct negative-delta shrink path | ✅ PASS |
| 3 | `evictIfNeeded` comment explains unreachability and drift condition | ⚠️ DEFERRED |
| 4 | No production logic changed — only comment and test added | ✅ PASS |

**Totals:** 3 PASS, 0 FAIL, 0 SKIP, 1 deferred

No browser walkthrough (pure library PR — no frontend components).

## Item Detail

**Item 1 (PASS):** `npx vitest run --project idbutil lru-blob-cache` — 29 tests green, including `replace with a smaller blob shrinks the total and evicts nothing`. Covers acceptance criterion 1.

**Item 2 (PASS):** Grep confirms the test seeds `maxBytes: 200`, puts `a=180` at t=1000, `b=90` at t=2000 (forcing eviction of `a`), then replaces `b` with `60` at t=3000 (delta=-30, no eviction). Assertions cover `getEntry("b")` not null, `getEntry("a")` null, `totalBytes === 60`, `entryCount === 1`. Matches the recommended scenario exactly.

**Item 4 (PASS):** `git diff origin/main -- idbutil/src/lru-blob-cache.ts` shows only comment-line additions; no executable statements changed. Confirms "no production logic changes."

**Item 3 (DEFERRED):** Whether the `evictIfNeeded` comment's prose adequately explains the unreachability reasoning and drift condition (acceptance criterion 2) was flagged `needs-human-judgment` by QA triage. See Disposition Triage below.

## Disposition Triage

Item 3 was submitted to the disposition Workflow for classification.

| 3 — `evictIfNeeded` comment quality | `opus-fixable` | n/a | Classifier: the added 9-line comment explicitly states (1) a negative delta never triggers eviction in the sequential single-writer case (because `totalSize + delta < maxBytes` given the invariant), and (2) the eviction body is reachable on a negative delta only via concurrent cap drift, with the `entry.key === key` skip protecting the replaced entry. Criterion is assessable from the diff. |

**Fix plan result: scope-deviation (`deviation=true`, `units=[]`).** The fix planner declined to produce a plan: "The sole finding is a deferred human-judgment assessment of comment prose; the `evictIfNeeded` comment and negative-delta test already exist in the PR diff and fully cover both acceptance criteria, so there is no actionable opus-fixable defect to fix."

Per the spec, `deviation=true` triggers the scope-deviation escape regardless of reason. A fixing pass cannot be taken; this pass escalates to office-hours.

`needs-main` items: none. No `blocked_by` follow-ups filed.

**Non-fixing pass — residue escalates to office-hours:** Item 3 (`opus-fixable`, deviation) escalates. The office-hours reviewer can confirm the comment satisfies AC2 (the deviation_reason suggests it does) and close the loop.

natb1:
## Review summary — #1417 / PR #1804

### Fixed

- **`idbutil/test/lru-blob-cache.test.ts:204`** — Two consecutive `await cache.getStats()` calls opened two IDB read transactions where one suffices. Replaced with a single `const stats = await cache.getStats()` and asserted `stats.totalBytes` / `stats.entryCount` against the captured value. (commit `6eaa2c40`)

### Dismissed

- **`idbutil/test/lru-blob-cache.test.ts:203`** — Reviewer noted that `expect(await cache.getEntry('a')).toBeNull()` follows the negative-delta replace of `'b'`, but `'a'` was already evicted earlier. This is deliberate: the assertion documents that the negative-delta replace did **not** restore `'a'` — it is a regression guard for the no-side-effect contract, not a check on the eviction trigger.

### Security

No open CodeQL alerts on this PR ref. No dependency changes (no `package.json` / `package-lock.json` touched). Surface: `code` — full security finder suite ran; no findings.

PR #1687, opened 2026-06-15
print: guard spread render path with spreadGen; fix single-page renderTask clear race

natb1:
Closes #1162

## Summary

`print/src/viewer/pdf.ts` already serializes single-page renders with a `renderGen`
generation token (added in #1151). This PR closes two related staleness defects that
remained in the same file.

### 1. Spread render path has no guard (primary scope)

`renderPageInto` (the spread/two-page renderer) called `renderPageToCanvas` and
`renderTextLayer` with no generation token and no `shouldAbort` hook. When
`cancelSpreadRenderTasks` fired (on every new spread render via the public wrapper,
and on `destroy`) while a `renderPageInto` was suspended at an `await`, the in-flight
call resumed and `spreadPages.push(...)` added a `SpreadPage` *after* the array was
cleared — leaving an orphaned spread render task untracked and uncancelled.

Fix: mirror the `renderGen` discipline with a new module-level `spreadGen`.
`cancelSpreadRenderTasks` bumps `spreadGen` before clearing `spreadPages`;
`renderPageInto` captures `gen` at entry, passes `() => gen !== spreadGen` as
`shouldAbort` to `renderPageToCanvas`, and re-checks `gen` after both the canvas and
text-layer awaits — cancelling its task and bailing before the `spreadPages.push`.

**`onTaskStart` is deliberately omitted for the spread path.** The single-page path
shares **one** canvas, so a superseding render must actively cancel the in-flight task
via `onTaskStart` to free that shared canvas. The spread path creates a **fresh canvas
per page** (`createPageWrapper()`), so a stale in-flight render paints only to a
soon-detached canvas and harms nothing — it just must not push into `spreadPages`. The
`shouldAbort` + post-await gen checks are sufficient. `gen` is capture-only (no
increment) because each public `renderPageInto` call is already preceded by a fresh
`cancelSpreadRenderTasks` bump via the wrapper, preserving left+right coexistence
within one spread.

### 2. Single-page `renderTask` clear was unconditional (additional scope)

When a render was superseded, `renderPage` unconditionally ran `renderTask = null`
after cancelling. If a superseding render B published its task via `onTaskStart` before
render A's continuation reached that line — possible when B's `getPage` resolves from
pdf.js's cache — A nulled the reference to B's *in-flight* task. A later render or
`destroy()` then found `renderTask === null`, skipped the cancel, and started a second
`page.render` on the shared canvas while B was still painting → "Cannot use the same
canvas during multiple render operations".

[...]

PR #1809, opened 2026-06-16
print: fix renderPageInto() wrapper leak on superseded-render path + regression test

natb1:
Closes #1727

## What

Fix a page-wrapper leak in the spread-mode `renderPageInto()` render path, and
add a concurrent-render regression test that locks it in.

## Background

#1727 reported that `renderPageInto()` applied `pendingHighlight` without a
gen-based race guard. The highlight-corruption aspect was already resolved by
two sibling PRs that merged after the issue was filed:

- #1687 added the `spreadGen` guard — the older of two concurrent same-page
  renders loses the gen race and bails before reaching `applyHighlight()`.
- #1747 made `applyHighlight()` call `unwrapHighlights()` first, so re-applying a
  highlight is idempotent.

Together those fix duplicate/corrupted highlight spans (acceptance criterion 2).
But writing the criterion-3 regression test for the concurrent spread path
surfaced a **separate, still-open defect**: a wrapper leak.

## The bug

`renderPageInto()` appends its page wrapper to the target **before** awaiting the
canvas render:

```ts
target.appendChild(wrapper);
const result = await renderPageToCanvas(pageNum, c, targetRect, wrapper, () => gen !== spreadGen);
if (!result) return;            // <-- wrapper left orphaned in the DOM
```

When a second concurrent `renderPageInto()` bumps `spreadGen`, the older call's
`renderPageToCanvas()` hits its `shouldAbort` check after its internal `getPage`
await and returns `null`. The `if (!result) return` early-return then leaves the
already-appended wrapper in the target — a blank duplicate page panel. The
`gen !== spreadGen` block immediately below already removes the wrapper for the
post-canvas-success supersession case, but the `!result` path was missing the
same cleanup.

## The fix

Mirror the existing cleanup on the `!result` path:

```ts
if (!result) {
  wrapper.remove();
  return;
}
```

This closes acceptance criterion 1 (no corruption — including no leaked blank
panel — from concurrent same-page renders).

## Test

Adds `concurrent renderPageInto() for the same page keeps one wrapper (spreadGen
guard)` to the `clearSearch()` block in `pdf.test.ts`. It fires two un-awaited
`renderPageInto(1, target)` calls into the same target and asserts exactly one
`.pdf-page-wrapper` survives — the load-bearing assertion that fails (sees 2)
when the one-line fix is reverted, confirmed as a real red→green discriminator.
It also checks the surviving render applied the armed highlight exactly once and
that `clearSearch()` restores the original text. This satisfies acceptance
criterion 3.

## Verification

- `npx vitest run --root print` — full print suite green.
- Red→green discriminator validated: reverting only the `wrapper.remove()` line
  makes the wrapper-count assertion fail.

Note: the production fixes referenced above landed in #1687 and #1747; this PR
adds only the missing one-line cleanup plus the spread-path regression test.


natb1:
# QA Summary — PR #1809 (Issue #1727)

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Concurrent spread-mode render does not duplicate highlight spans or corrupt text | script-verifiable | PASS |
| 2 | No regression across the print unit suite | script-verifiable | PASS |
| 3 | Spread-mode search highlight golden path on a public PDF | needs-human-judgment | deferred |

**PASS: 2 / FAIL: 0 / SKIP: 0 / Deferred: 1**

No walkthrough ran (no `needs-browser` items in plan). No QA server started.

## Deferred to office-hours

- **Item 3 — Spread-mode search highlight golden path on a public PDF** (`/view/plato-phaedrus`): Golden-path check that a search term is highlighted exactly once on the correct spread page with no doubled marks, no blank/orphaned panel, and clean text after clearing search. Requires human eye on the live spread view — classified `needs-human` (aesthetic).

## Bugs found

None.

## Disposition triage

| # | Title | Class | Aesthetic | Verify |
|---|-------|-------|-----------|--------|
| 3 | Spread-mode search highlight golden path on a public PDF | needs-human | yes | n/a |

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR. (No `needs-main` items on this pass.)

**Non-fixing pass** — no `opus-fixable` items, so residue escalates to office-hours. The single `needs-human` (aesthetic) item re-surfaces when `/office-hours` walks the spread-mode golden path manually.

natb1:
## Review — PR #1809

**Disposition summary:** 2 Informational, 1 Deferred (filed as #1816). No security findings. No code changes applied.

---

### Informational

_Observations surfaced for reference; no action required._

- **F01 — `print/test/viewer/pdf.test.ts:763-767`** — Test comment misdescribes which abort branch the superseded render exits through.
  The comment states that `p1` "after its canvas render, loses the gen race and removes its orphaned wrapper" — implying the `gen !== spreadGen` branch at `pdf.ts:451` fires. In practice, `p2`'s `cancelSpreadRenderTasks()` bumps `spreadGen` synchronously before any microtask drains, so `p1` resumes after `await pdfDoc!.getPage(pageNum)` with `shouldAbort()` already true and `renderPageToCanvas` returns `null` before canvas rendering begins. The actual path exercised is the new `!result` branch at `pdf.ts:447-449`, not the `gen !== spreadGen` branch at line 451. The test correctly covers the fixed line and would fail if the fix were reverted, but the comment will mislead future readers tracing the execution path.

- **F02 — `print/src/viewer/pdf.ts:447-449`** — Fix is correct and complete for the described bug.
  `wrapper.remove()` is correctly added to the `!result` early-return path. The three other gen-race early-return sites (lines 451–455, 458–464) already had `wrapper.remove()`. The 0×0 rect guard fires before wrapper creation so no leak is possible there. The regression test confirms the red→green discriminator. No action needed.

---

### Deferred

_Out-of-scope improvements filed as follow-up issues._

- **F03 — `print/src/viewer/pdf.ts:444`** — Exception paths after `wrapper.appendChild` leave an orphaned wrapper (pre-existing, not introduced by this diff). After `target.appendChild(wrapper)`, `renderPageToCanvas` can throw on non-cancellation errors (`getContext` null → `'Could not acquire 2D canvas context'`; non-`RenderingCancelledException` rejections), and `renderTextLayer` can rethrow non-`AbortException` errors — all leaving `wrapper` attached to `target`. This PR correctly fixes the normal-flow abort cases; the exception paths are pre-existing. Filed as #1816 (blocked by #1727).

---

_Fixed: None. Required: None. Dismissed: None. False-positive: None. Out-of-scope: None._
_CodeQL: 0 open alerts. Dependency audit: not applicable (no package.json changes)._

PR #1813, opened 2026-06-16
print: clearSearch() aborts an in-flight EPUB search

natb1:
Closes #1805

## Problem

The EPUB viewer's `clearSearch()` did not abort an in-flight `search()`, so
clearing the search box while a search was still running could let stale
highlights re-paint for the abandoned query.

`search()` (`print/src/viewer/epub.ts`) captures `epoch = ++_searchEpoch` at
entry, runs the spine loop (which can park on a slow `section.load()`), then
guards its highlight burst with `if (epoch !== _searchEpoch) return results;`.
`clearSearch()` only called `clearSearchHighlights()` — it never bumped
`_searchEpoch`. So emptying the search box removed the current highlights, but
the in-flight search then completed, saw an unchanged epoch, and re-painted
highlights for the cleared query. The results list stayed empty (guarded in
`search.ts`), but stale highlights remained visible.

## Fix

`clearSearch()` now bumps `_searchEpoch` before clearing highlights. Any
in-flight `search()` trips its existing post-loop guard and returns early
without a highlight burst. This reuses the monotonic epoch mechanism the
overlapping-search guard already honors — no new state or signal is introduced.
`destroy()`'s `_searchEpoch = 0` reset is unaffected.

## Test

Adds a regression test in `print/test/viewer/epub.test.ts`, inside the existing
`clearSearch` block, modeled on the overlapping-search test: it gates a
section's `load()`, starts a search that parks on the gate, calls
`clearSearch()` while parked, releases the gate, and asserts no highlight is
painted for the abandoned query's cfi. Verified fail-first — the test red-fails
against the unfixed `clearSearch()` and passes with the fix.

### DIFF
--- a/print/src/viewer/epub.ts
+++ b/print/src/viewer/epub.ts
@@ -348,6 +348,9 @@ export function createEpubRenderer(
     },
 
     clearSearch(): void {
+      // Stand down any in-flight search(): bumping the epoch trips its post-loop
+      // guard (search ~line 318) so it applies no highlights for the cleared query.
+      _searchEpoch++;
       clearSearchHighlights();
     },
 
--- a/print/test/viewer/epub.test.ts
+++ b/print/test/viewer/epub.test.ts
@@ -890,6 +890,37 @@ describe("createEpubRenderer", () => {
         renderer.clearSearch!();
         expect(mockRendition.annotations.remove).not.toHaveBeenCalled();
       });
+
+      it("aborts an in-flight search so its highlights never appear after clearSearch", async () => {
+        // One section whose load parks on a gate; find returns a match.
+        const section: FakeSection = {
+          href: "ch0.xhtml",
+          unload: vi.fn(),
+          find: vi.fn().mockReturnValue([{ cfi: "cfi-A", excerpt: "fox" }]),
+          load: vi.fn() as ReturnType<typeof vi.fn>,
+        };
+        let releaseFirst!: () => void;
+        const gate = new Promise<void>((resolve) => { releaseFirst = resolve; });
+        section.load.mockImplementationOnce(() => gate);
+        mockSpine.spineItems = [section];
+        mockBook.navigation.get.mockReturnValue({ label: "X" });
+
+        const renderer = await initRenderer();
+
+        // Start the search — it parks on the gated section.load().
+        const p = renderer.search!("fox");
+
+        // Clear the search while parked.
+        renderer.clearSearch!();
+
+        // Release the gate so the in-flight search resumes and hits the epoch guard.
+        releaseFirst();
+        await p;
+
+        // The aborted search must not have painted any highlights.
+        const highlightCfis = mockRendition.annotations.highlight.mock.calls.map((c) => c[0]);
+        expect(highlightCfis).not.toContain("cfi-A");
+      });
     });
   });
 });


natb1:
# QA Summary — #1805 (PR #1813)

**Branch:** `1805-print-epub-clearsearch-does`  
**Fix:** `clearSearch()` now bumps `_searchEpoch` to abort in-flight EPUB searches.

## Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Run the EPUB viewer unit tests | shell-command | ✅ PASS |
| 2 | Type-check the print package | shell-command | ✅ PASS |
| 3 | Browser smoke: EPUB search → clear still works end-to-end | walkthrough | ✅ PASS |

**Counts:** 3 executed, 3 PASS, 0 FAIL, 0 SKIP

## Detail

**Item 1 — Unit tests (`npx vitest run --root print test/viewer/epub.test.ts`)**  
48/48 tests pass. Includes the new regression test (gates `section.load()`, calls `clearSearch()` while search is parked, releases, asserts no highlight painted for the abandoned CFI) and the existing overlapping-search test. The `DOMException` blob-URL warnings in output are pre-existing happy-dom environment noise — they do not affect test results.

**Item 2 — Type-check (`npm run build --prefix print`)**  
Build succeeded cleanly. The large-chunk warning is pre-existing and unrelated to this change.

**Item 3 — Browser smoke (gutenberg-3296, "The Confessions of St. Augustine")**  
- Typed "God" into the search box; search returned multiple results and painted highlights in the EPUB body.
- Cleared the search box via the ✕ button.
- Result: highlights removed, results list cleared, document returned to clean reading state.
- No console errors, no 4xx/5xx network errors.
- GIF: `tmp/qa-fix-walkthrough-1805.gif`

> Note: The stale-highlight race (clearing mid-search) cannot be reliably reproduced in a manual browser walkthrough — it requires controlled timing. That regression is covered by the unit test in Item 1.

> Note: Local Playwright acceptance tests cannot run (known WSL/Nix Chromium version mismatch — CI is authoritative). CI is passing on PR #1813.

## Acceptance criteria

- [x] `clearSearch()` stands down an in-flight `search()` — verified by unit test (Item 1).
- [x] Regression test gates `section.load()`, calls `clearSearch()` while parked, asserts no highlight — passes (Item 1).
- [x] Existing tests including `"only the latest of two overlapping searches performs its highlight burst"` continue to pass — verified (Item 1).
- [x] No regression in the end-to-end search/clear UX flow — verified by browser smoke (Item 3).

## Disposition

No residue. Clean pass.

natb1:
## Review — PR #1813 (`clearSearch()` epoch bump)

No code changes applied — no Required findings, no Fixed items.

---

### Deferred

- **`print/src/viewer/epub.ts:302-318`** — epoch guard is checked only after the full spine loop completes. When `clearSearch()` bumps `_searchEpoch` mid-search, the in-flight `search()` continues iterating all remaining spine sections (`section.load()` + `section.find()` on each) before discovering the abort. For EPUBs with many sections this wastes potentially many section loads. The fix is correct and blocks highlight painting; it just does not exit the loop early. Follow-up: #1819

---

### Informational

- **`print/test/viewer/epub.test.ts:894-923`** — the new regression test does not assert that `section.unload()` is called after the aborted search completes. The production code wraps `section.load()` in a `try/finally` that calls `unload()`, so the resource is freed; asserting it in the test would make the cleanup contract explicit.

---

- **`print/src/viewer/epub.ts:318`** — aborted `search()` still returns its full results array to the caller. Non-issue: `search.ts` gates the results list on its own epoch and discards stale results; the array return is harmless.
- **`print/src/viewer/epub.ts:357-376`** — `destroy()` resets `_searchEpoch` to 0; a post-destroy `search()` call could race. Non-issue: the renderer is not reused after `destroy()`; the reset is intentional and pre-existing.
- **`print/src/viewer/epub.ts:350-355`** — `clearSearch()` bumps epoch then calls `clearSearchHighlights()` synchronously; a concurrent `search()` could see an old epoch snapshot. Non-issue: JavaScript is single-threaded; no true concurrency between the epoch bump and the guard check.

---

_No security findings. CodeQL: 0 open alerts. npm audit: no dependency changes in this PR._

PR #1815, opened 2026-06-16
print: EPUB search skips a section that fails to load instead of aborting

natb1:
Closes #1811

## Problem

The EPUB full-text search in `print/src/viewer/epub.ts` (`search()`) iterates
spine sections in a `try { await section.load(...); …find/push… } finally {
section.unload(); }` block with **no `catch`**. If one `section.load()` rejects —
a corrupt or missing resource, a network error, a malformed EPUB — the rejection
propagates out of the loop and aborts the **entire** search, discarding matches
already collected from earlier sections and surfacing nothing, even though most
of the book is searchable.

The resource-leak half of the original report (#1760) was already fixed in
shipped code (`unload()` lives in `finally`). This change addresses the remaining
resilience half: one bad section should degrade gracefully to "that section
omitted" instead of "no results at all".

## Change

- `print/src/viewer/epub.ts` — add a `catch` between the existing `try` and
  `finally` in the `search()` loop. A section that throws during `load()` or
  `find()` is caught, surfaced via `reportError` (matching the file's established
  non-fatal-error pattern, with the same `cause`-wrapping form used elsewhere in
  the file), and the loop continues over the remaining sections. `section.unload()`
  stays in `finally` unchanged. The stale-epoch guard and post-loop highlight burst
  are unaffected.
- `print/test/viewer/epub.test.ts` — add one test that builds three sections,
  makes the middle one's loader reject, and asserts that the search resolves,
  returns the matches from the surrounding sections, still calls `unload()` on the
  failing section, and reports the skipped section via `reportError`.

## Verification

`npx vitest run --root print` — the new test passes and all existing EPUB search
tests still pass.


natb1:
## QA Summary — Issue #1811 / PR #1815

**Result: PASS (all items)**

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Existing EPUB search test suite passes | shell-command | PASS |
| 2 | New test asserts failing section is skipped | shell-command | PASS |
| 3 | Catch block exists in EPUB search loop source | shell-command | PASS |
| 4 | Failing section's unload() runs via finally block | shell-command | PASS |

- PASS: 4 / FAIL: 0 / SKIP: 0
- No browser walkthrough needed (pure TypeScript/unit-test change, no browser component UI changed).

### Detail

**Item 1 & 2 — `npx vitest run --root print`:**
553 tests passed, 1 skipped, 0 failed. The new test "skips a section whose load() rejects, surfaces via reportError, and returns matches from remaining sections" passes, confirming all acceptance criteria: the search resolves despite a section rejection, `unload()` is called on the failing section, matches from surrounding sections are returned, and the error is surfaced via `reportError`.

**Item 3 — Catch block present:**
`print/src/viewer/epub.ts:317` contains `new Error("EPUB search: section failed to load", { cause: err })` inside the catch block.

**Item 4 — `section.unload()` in finally:**
`print/src/viewer/epub.ts:320` contains `section.unload()` in the `finally` block, confirming the resource is freed even when `load()` rejects.

### Acceptance criteria coverage

- [x] `section.load()` rejection for one section does not abort the search — verified by new unit test
- [x] `section.unload()` called for the failing section — verified by new unit test assertion + structural grep
- [x] Matches from surrounding sections are returned — verified by new unit test assertion
- [x] Unit test covers the skip path — new test present and passing
- [x] Existing EPUB search tests pass — 553 tests green

No deferred items, no bugs found, no escalation needed.

natb1:
## Review — PR #1815

Fix commit: `4e768927`

### Fixed

- **`print/src/viewer/epub.ts:317`** (code-review, review): The `catch` block's error message `'EPUB search: section failed to load'` was inaccurate — the try scope covers `section.find()` and `buildResult()` in addition to `section.load()`. Restructured so `load()` is in its own inner try/catch reporting `'EPUB search: section failed to load'`, and `find()`/`buildResult()` are in a separate catch reporting `'EPUB search: section failed during find'`.

- **`print/test/viewer/epub.test.ts:843`** (code-review, review): The `reportError` assertion was too weak (`toHaveBeenCalled()` only). Strengthened to `toHaveBeenCalledWith(expect.objectContaining({ message: 'boom' }))` to verify the forwarded error identity. Added `expect(s0.unload).toHaveBeenCalledTimes(1)` and `expect(s2.unload).toHaveBeenCalledTimes(1)` to confirm all three sections' `unload()` runs regardless of a sibling failure.

### Dismissed

- **`print/test/viewer/epub.test.ts:827`** (code-review): The test does not explicitly assert that `s1.find()` is never called after `load()` rejects. The results-array check (`["cfi-0", "cfi-2"]`) implicitly confirms no s1 matches were collected; an explicit `s1.find` call-count assertion would add coverage but is not a correctness gap.

### False-positive (security)

- **`print/src/viewer/epub.ts:307`** (red-team): Concern that the catch scope covering `section.find()` and `buildResult()` could suppress errors from crafted EPUB content. Both operations are pure in-memory reads over already-loaded DOM; neither writes state nor makes network calls — skipping them on failure is the same graceful-degradation behavior as skipping a failed load.

- **`print/src/viewer/epub.ts:311`** (red-team, code-review): Concern that silencing section errors hides diagnostic signal from users. `reportError` propagates every skipped-section error to the global error handler (matching the file's established non-fatal-error pattern), preserving visibility.

- **`print/src/viewer/epub.ts:314`** (red-team): Concern that passing the raw `Error` object to `window.reportError` exposes internal messages. `reportError` is the standard non-fatal-error surface in this file; the same pattern is used elsewhere. No confidential data flows through EPUB section load errors.

### Out-of-scope (security)

- **`print/src/viewer/epub.ts:311`** (red-team): `destroy()` nullifies `book` (line 374) without coordinating with an in-flight `search()`. If `destroy()` fires while `search()` is awaiting `section.load()`, the next iteration could reference a null spine. This race is pre-existing (not introduced by this diff) and independent of the section-skip change.

PR #1821, opened 2026-06-16
print: clean up renderPageInto() wrapper on render exception paths

### LINKED ISSUE #1816: print: renderPageInto() wrapper leaked on exception paths in renderPageToCanvas/renderTextLayer
In \`print/src/viewer/pdf.ts\`, \`renderPageInto()\` appends \`wrapper\` to \`target\` at line 444 and then calls \`renderPageToCanvas\` and \`renderTextLayer\`. The existing null-result and gen-race exit paths correctly call \`wrapper.remove()\`. However, thrown exceptions escape without cleanup:

- \`renderPageToCanvas\` at line 265 throws \`'Could not acquire 2D canvas context'\` when \`getContext('2d')\` returns null.
- \`renderPageToCanvas\` at line 275 rethrows a non-\`RenderingCancelledException\` render rejection.
- \`renderTextLayer\` at line 300 rethrows a non-\`AbortException\` error from \`tl.render()\`.

All three propagate out of \`renderPageInto()\` with \`wrapper\` still attached to \`target\`, leaking a DOM node. This is the same orphan-wrapper leak that PR #1809 addresses for the null-result and gen-race normal-flow exits; the exception paths were not introduced by #1809.

## Location

\`print/src/viewer/pdf.ts\` — \`renderPageInto()\`, post-\`target.appendChild(wrapper)\` block (line 444)

## Recommended fix

Wrap the post-append body in a \`try/finally\`. Use a \`committed\` boolean set to \`true\` just before \`spreadPages.push\` (line 465); in the \`finally\` block, call \`wrapper.remove()\` if \`!committed\`. This covers all throw paths without changing the existing normal-flow abort logic.

```typescript
const { wrapper, canvas: c, textLayerDiv: tlDiv } = createPageWrapper();
target.appendChild(wrapper);
let committed = false;
try {
  const result = await renderPageToCanvas(...);
  if (!result) { return; }
  if (gen !== spreadGen) { return; }
  const tl = await renderTextLayer(...);
  if (gen !== spreadGen) { tl?.cancel(); return; }
  committed = true;
  spreadPages.push({ renderTask: result.task, textLayer: tl });
  // ...
} finally {
  if (!committed) wrapper.remove();
}
```

Note: the existing \`wrapper.remove()\` calls in the null-result and gen-race branches become redundant once \`finally\` covers them — they can be removed in the same PR.

## Acceptance criteria

- [ ] A thrown non-cancellation error from \`renderPageToCanvas\` (e.g. simulated null canvas context) does not leave \`wrapper\` attached to \`target\`
- [ ] A thrown non-\`AbortException\` from \`renderTextLayer\` does not leave \`wrapper\` attached to \`target\`
- [ ] The normal-flow null-result path (\`renderPageToCanvas\` returns null) still removes \`wrapper\`
- [ ] The gen-race abort paths after canvas render and after text-layer render still remove \`wrapper\` (or are replaced by the \`finally\` block)
- [ ] \`spreadPages.push\` is only reached when no exception was thrown (\`committed = true\` invariant holds)
- [ ] Existing spread-render tests pass

natb1:
Closes #1816

## Summary

`renderPageInto()` in `print/src/viewer/pdf.ts` appends `wrapper` to `target`
before awaiting the canvas and text-layer renders. PR #1809 fixed the
normal-flow orphan-wrapper leaks (the `!result` null path and the two
`gen !== spreadGen` race paths). This PR closes the remaining **exception**-path
leaks: a thrown error from `renderPageToCanvas` ("Could not acquire 2D canvas
context", or a rethrown non-`RenderingCancelledException` render rejection) or
`renderTextLayer` (a rethrown non-`AbortException` error) previously escaped
`renderPageInto()` with `wrapper` still attached to `target`, leaving a stray
DOM node.

## Change

The post-`appendChild(wrapper)` body is now wrapped in a `try/finally` keyed on
a `committed` boolean. The `finally` removes the wrapper whenever the render did
not commit — covering both the three early-return branches and any thrown error.

- The three early-return branches drop their explicit `wrapper.remove()`; the
  `finally` handles them.
- The post-text-layer gen-race branch keeps `tl?.cancel()`; only its
  `wrapper.remove()` is removed.
- `committed = true` is set immediately before `spreadPages.push`, so a throw
  from `applyHighlight` after the push leaves the wrapper in place — by then it
  is owned by `spreadPages` and cleaned up by `cancelSpreadRenderTasks()`.

No behavior change to the throw sites themselves — only to `renderPageInto()`'s
cleanup.

## Tests

Two regression tests in the `clearSearch()` block of
`print/test/viewer/pdf.test.ts`, each driving a single `renderPageInto` so
`gen === spreadGen` holds and the new `finally` is the only wrapper-removal path:

- **Canvas-context throw leaves no wrapper** — `getContext('2d')` returns null;
  the public promise rejects with `Could not acquire 2D canvas context` and no
  `.pdf-page-wrapper` survives.
- **Text-layer throw leaves no wrapper** — `TextLayer.prototype.render` rejects
  with a plain `Error` (rethrown as non-`AbortException`); the public promise
  rejects and no `.pdf-page-wrapper` survives.

The existing #1809 concurrent-render test still passes.

## Verification

- `npx vitest run --root print` — 556 passed, 1 skipped.


natb1:
## QA — #1816 renderPageInto() exception-path wrapper cleanup

### Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | canvas-context throw leaves no wrapper | script-verifiable (vitest) | PASS |
| 2 | text-layer throw leaves no wrapper | script-verifiable (vitest) | PASS |
| 3 | normal-flow null-result and gen-race paths still clean up | script-verifiable (vitest) | PASS |
| 4 | `committed = true` immediately before `spreadPages.push` | script-verifiable (grep) | PASS |
| 5 | early-return branches no longer call `wrapper.remove()` directly | script-verifiable (grep) | PASS |

**PASS 5 / FAIL 0 / SKIP 0**

No browser walkthrough needed — all acceptance criteria are vitest-verifiable. No QA server started.

### Detail

- Full suite: `561 passed (32 test files)` — run from `print/` directory. (`npx vitest run --root print` hits a known local path issue with the `pdfjs-dist/build/pdf.worker.min.mjs?url` import; running from inside `print/` is correct and matches how CI runs it.)
- The two new regression tests (`renderPageInto() throwing on canvas context leaves no wrapper`, `renderPageInto() throwing on text layer leaves no wrapper`) both pass and confirm the `try/finally` covers the exception paths.
- `committed = true` is at line 458, immediately before `spreadPages.push` at line 459 — invariant holds.
- Only one `wrapper.remove()` remains in the file (`print/src/viewer/pdf.ts:464` — inside the `finally`). All three early-return branches correctly rely on the `finally` instead of calling `wrapper.remove()` directly.

### No residue

All items passed. No disposition triage, no follow-ups filed, no escalation.

natb1:
## Review findings — #1816

### Fixed

- **`print/test/viewer/pdf.test.ts:874`** — The two new exception-path tests inlined an identical `getBoundingClientRect` stub object. Extracted to a shared `makeTargetRect()` helper and applied to all three target stubs (both new exception-path tests and the existing spread-render test). _(fix commit: 212919b3)_

### Security

No open CodeQL alerts for this PR. `surface=code`, `deps=false` — no dependency audit run.

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

PR #1826, opened 2026-06-16
dispatch: make auto-merge the default (invert the config gate to opt-out)

natb1:
Closes #1820

## What

Inverts the `dispatch-auto-merge` config gate so that **auto-merge is ON by
default**. An absent `dispatch.config/auto-merge.json` now arms the reconciler
instead of leaving it inert. The config file becomes an **opt-OUT kill-switch**
for incident response rather than an opt-in enabler.

The three config states are now:

- **no file** → armed (the default)
- **file present, `.enabled == false`** → off (the kill-switch) — unchanged
- **file present, `.enabled == true`** → armed (redundant with the default) —
  unchanged

## Why

`dispatch-auto-merge` (tick Step 1d, #1739) squash-merges every PR that reaches
the dispatch ready state (`isDraft==false ∧ dispatch:reviewed ∧ CI passing ∧
mergeable==MERGEABLE`, plus a non-empty `Closes #N` guard). #1739 deliberately
shipped it inert behind an explicit opt-in as a trust ramp. It has since been
armed in production, so the opt-in default is now backwards: a fresh checkout or
new environment silently reverts to hand-merging until someone recreates the
config file. Auto-merging green, reviewed, conflict-free PRs is the intended
steady state, so absent config should mean ON.

The only behavior change is the no-config case (previously a no-op). The
`enabled`-when-present semantics are unchanged, and the kill-switch is retained
so auto-merging to `main` stays reversible during an incident.

## Changes

- **Unit 1** — invert the gate in `dispatch-auto-merge`: an absent config arms
  the reconciler; only an explicit `enabled:false` (file present) exits early.
  Header comment rewritten to describe the inverted default. Test 14 inverted
  (absent config → merges); Test 13 (`enabled:false` → no-op kill-switch)
  unchanged. Full dispatch script suite passes (2994/2994).
- **Unit 2** — docs updated to match: the Step 1d comment in
  `dispatch-select-tick` and the auto-merge schema in `dispatch-config-load` now
  describe the opt-OUT kill-switch, calling out that auto-merge inverts the
  generic absent→inert config convention (which still holds for every other
  config type). `auto-merge.example.json` stays `{"enabled": false}` as the
  canonical kill-switch example.

## Note on the production config

The `dispatch.config/auto-merge.json {"enabled": true}` added in production to
arm the feature is now redundant with the default ON behavior, but harmless to
leave in place. It is safe to delete.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-auto-merge
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-auto-merge
@@ -18,11 +18,12 @@
 # reviewed conflict-free PR auto-merges regardless of whether it closes a bug,
 # security, or enhancement issue.
 #
-# Config gate (off by default): read from dispatch-config-load auto-merge.
-#   - no config file  → `no-config` → exit 0 with no output (current human-merge
-#                       behavior preserved).
-#   - file present but .enabled != true → exit 0 with no output.
-#   - .enabled == true → arm auto-merging.
+# Config gate (ON by default): read from dispatch-config-load auto-merge. The
+# config file is an opt-OUT kill-switch (incident response), inverting the usual
+# absent→inert convention.
+#   - no config file  → `no-config` → armed (auto-merging is the default).
+#   - file present, .enabled == false → exit 0 with no output (kill-switch).
+#   - file present, .enabled == true → armed (redundant with the default).
 #
 # Merge: gh pr merge <n> --squash with explicit --subject (the PR title) and
 # --body (the PR body), so the non-TTY squash is deterministic and the
@@ -55,9 +56,13 @@ CFG=$("$SCRIPT_DIR/dispatch-config-load" auto-merge) || {
   echo "dispatch-auto-merge: error: dispatch-config-load auto-merge failed" >&2
   exit 1
 }
-[[ "$CFG" == "no-config" ]] && exit 0
-ENABLED=$(jq -r '.enabled' <<<"$CFG")
-[[ "$ENABLED" == "true" ]] || exit 0
+# Default ON: an absent config file arms auto-merging. The file exists only as
+# an opt-OUT kill-switch (incident response). When present, enabled:false turns
+# the reconciler off; enabled:true (the redundant explicit form) keeps it on.
+if [[ "$CFG" != "no-config" ]]; then
+  ENABLED=$(jq -r '.enabled' <<<"$CFG")
+  [[ "$ENABLED" == "false" ]] && exit 0
+fi
 
 # ---- Step 2: one fresh fetch of every open PR -------------------------------
 # Never proceed on a partial fetch — a truncated snapshot could merge a PR whose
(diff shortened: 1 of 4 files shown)


natb1:
# QA Summary — Issue #1820

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Absent config arms auto-merge (default ON) | shell-command | PASS |
| 2 | Kill-switch and explicit-on states preserved; full suite passes | shell-command | PASS |
| 3 | Main-health suppression and readiness predicate unchanged | shell-command | PASS |
| 4 | Documentation describes the config file as an opt-OUT kill-switch | shell-command | PASS |

**Totals: 4 PASS / 0 FAIL / 0 SKIP**

## Detail

**Item 1 — Absent config arms auto-merge:** Gate logic confirmed at line 62 (`if [[ "$CFG" != "no-config" ]]` — only disables when a file is present with `enabled:false`). Test 14 assertions in `test-dispatch-scripts.sh` match the inverted behavior (`merged #50`, merge log present, PR fetch present).

**Item 2 — Full test suite:** `bash test-dispatch-scripts.sh` → `2994/2994 passed, 0 failed`. All three config states covered: Test 1 (`enabled:true` → on), Test 13 (`enabled:false` → off, kill-switch preserved), Test 14 (absent → armed).

**Item 3 — OPEN_MB gate and readiness predicate unchanged:** `if [[ -z "$OPEN_MB" ]]` guard intact at `dispatch-select-tick:296`. The diff in `dispatch-select-tick` is comment-only (no gate logic touched).

**Item 4 — Documentation:** All three doc sites use opt-OUT kill-switch language (`dispatch-auto-merge:21-25`, `dispatch-config-load:220-227`, `dispatch-select-tick:294`). No stale "off by default", "absent config → no-op", "only true arms", or "reconciler is inert" wording found.

## Bugs found

None.

## Disposition triage

No residue — Step 3.5 was skipped.

natb1:
## Review findings — #1820

### Fixed

- **dispatch-config-load:219** — replaced hardcoded `(lines 14-18)` cross-reference with `(see "Stdout protocol")` to stay stable against future line insertions. Fix commit: `88eac4f9`.

### Informational

- **test-dispatch-scripts.sh:30524** — `write_auto_merge_config_enabled` writes `{"enabled":true}`, which under the new default-ON semantics is functionally identical to writing no config file. The test helper is harmless and the PR body already notes this redundancy; no change required.

### False-positive (security)

- **dispatch-auto-merge:62–65** — Config gate inverted from deny-by-default to allow-by-default. Flagged as a security posture change, but this is the explicit intent of #1820 with documented rationale (trust ramp reached, absent config should mean ON). Not a vulnerability.

- **dispatch-auto-merge:62–64** — Kill-switch check `[[ "$ENABLED" == "false" ]]` flagged as brittle if `jq` returns a non-string. `jq -r` on a JSON boolean always produces the literal string `false`; null/missing `.enabled` produces `null`, not `false`, so the gate is correct. Not a vulnerability.

### Out-of-scope (security)

- **dispatch-auto-merge:55–65** — `dispatch-config-load` subprocess inherits the caller's environment including any `DISPATCH_CONFIG_DIR` override. Pre-existing design; the diff did not introduce it.

- **dispatch-select-tick:296–303** — `dispatch-auto-merge` called with `|| true`, silently swallowing non-zero exits. Pre-existing pattern; the diff did not introduce it.

_No deferred follow-ups filed. No npm or CodeQL alerts._

PR #1847, opened 2026-06-18
print: PDF search matches queries spanning text-item / line boundaries

### LINKED ISSUE #936: print: implement PDF full-text search backend
Part of #935 (activating the print viewer's dormant search UI).

## Context

The search input, event wiring, contract, and UI tests already exist; no renderer implements the backend, so the input never appears. This issue implements it for the **PDF renderer** (`print/src/viewer/pdf.ts`) — the most tractable target, since PDF.js already exposes per-page text via `page.getTextContent()` (used for the text layer at `pdf.ts:127`).

`initSearch()` removes the `search-hidden` class once the renderer implements `search()` + `goToResult()` (`print/src/viewer/search.ts:36,45`); the contract and `SearchResult` shape are in `print/src/viewer/types.ts:43-46,6-17`.

## Scope and design

- Implement `search(query)`, `goToResult(result)`, and `clearSearch()` on the PDF renderer per the `ContentRenderer` contract.
- `search()`: extract each page's text via `getTextContent()`, find case-insensitive matches, return `SearchResult[]` with `location` = page-number string, `label` = "Page N", a `snippet` of surrounding text, and correct `matchStart`/`matchLength` offsets (honoring the `types.ts:4` invariant).
- Build the per-page text index lazily/incrementally so `init()` is not blocked on large PDFs.
- `goToResult()`: navigate to the result's page (reuse `goToPage`), resolving after the page is shown so the panel's `onNavigate()` refreshes position.
- `clearSearch()`: drop transient search state.
- No `shell.ts` change beyond confirming existing wiring.

## Out of scope

- EPUB search (sibling sub-issue).
- In-page highlighting of the match within the rendered text layer (this issue returns results + navigates to the page only).

## Acceptance criteria

- [ ] The PDF renderer implements `search()`, `goToResult()`, and `clearSearch()` satisfying the `ContentRenderer` contract in `print/src/viewer/types.ts`.
- [ ] Opening a PDF in the print viewer shows the search input in the info/navigation panel (the `search-hidden` class is removed).
- [ ] Typing a query returns matching results, each with a page label, a snippet, and a `<mark>`-highlighted match positioned via correct `matchStart`/`matchLength`.
- [ ] Every `SearchResult` satisfies `matchStart >= 0`, `matchLength > 0`, `matchStart + matchLength <= snippet.length`.
- [ ] Clicking a result navigates the viewer to the matching page and the panel position label updates.
- [ ] Clearing the input removes results and resets state via `clearSearch()`.
- [ ] Text extraction does not block `init()` on multi-page PDFs (lazy/incremental indexing).
- [ ] Unit tests cover PDF `search()` (match, no-match, case-insensitivity, snippet/offset correctness) and `goToResult()` page navigation.
- [ ] Existing `print/test/viewer/search.test.ts` UI tests still pass.

natb1:
Closes #936

## What

PDF full-text search (shipped in #1691) reconstructed each page's searchable
text by concatenating pdf.js `getTextContent().items` with an **empty** join.
A query spanning a line break — e.g. `"the"` and `"quick"` rendered on two
lines — was only searchable as `"thequick"` and never matched. This makes such
cross-line / cross-text-item queries match, and highlights them exactly.

## How

The match side and the highlight side previously agreed only *by accident* —
both used an empty join, so the stored `offset`/`length` mapped byte-for-byte
onto the text-layer divs. Inserting a separator on the match side alone would
drift every highlight by N characters. Instead, both sides now share a single
source of truth:

- A pure `reconstructPage(items)` returns both the reconstructed `text` **and**
  an index-aligned offset→item layout (`PageLayout { text, items: [{start,
  length}] }`). It inserts a single space `" "` **after** any item whose
  `hasEOL` is true; empty join everywhere else. The separator advances
  `text.length` but is not part of any item's `[start, start+length)` range.
- The highlight mapper (`offsetToDivRanges` → `offsetToItemRanges`) is now
  **separator-agnostic**: it consumes the precomputed layout and just intersects
  `[offset, offset+length)` with each item's range. Because separators live only
  in `reconstructPage`, the two sides cannot drift by construction. A match
  boundary landing inside a separator gap yields no segment for that gap.

`findMatches`, `buildSnippet`, and `encodeLocation`/`decodeLocation` are
unchanged — they operate on `layout.text` exactly as they operated on the old
`pageText`, so the `"page:offset:length"` location token and the
`SearchResult` invariant are preserved structurally.

A boundary guard in `applyHighlight` throws if `layout.items.length !==
tl.textDivs.length`, turning a future 1:1-alignment regression (e.g. someone
enabling marked content on one path) into a clear error rather than a silent
off-by-N highlight.

### Separator scope: why `hasEOL`-gated, not blanket

- **Line boundaries** (`hasEOL=true`) are exactly where pdf.js reliably omits
  whitespace, so a space there makes a user's natural cross-line query match.
  This is the new capability.
- **Intra-word splits stay unregressed:** pdf.js sometimes emits one word as
  multiple same-line items with no whitespace (`"world"` → `"wor"`+`"ld"`,
  `hasEOL=false`). These keep the empty join, so existing same-line / intra-word
  search is untouched.
- **On acceptance criterion 1 ("text-item boundary"):** a same-line word
  boundary already carries its space inside the item text, so it is matched
  today via the empty join. The only un-bridged text-item boundary is two
  distinct same-line words with **no** whitespace item between them — which is
  indistinguishable from an intra-word split (`hasEOL=false` for both), so
  bridging it would regress real intra-word matches. `hasEOL`-gated separation
  is the correct, distinguishable scope; the line-boundary case is the issue's
  motivating example.

## Tests

`print/test/viewer/pdf.test.ts` adds feature-proving cases against the four
acceptance criteria:

- **Cross-line match:** a new mock page reconstructing to `"alpha the quick"`;
  `search("the quick")` returns one `Page N` result with `decodeLocation` →
  `{offset:6, length:9}` (the old empty-join code searched `"alphathequick"`
  and found nothing).
- **Spanning-div highlight, exact coverage:** drives `goToResult`/`renderResult`
  and asserts exactly two `.search-highlight.active` spans reading exactly
  `"the"` and `"quick"`. The leading `"alpha",hasEOL:true` item is load-bearing
  — it places a separator before `matchStart`, so an unmirrored mapper would
  produce `"he"` and fail. This makes the test discriminating.
- **Offset invariant** on every result, and **intra-word same-line regression**
  (`"wor"`+`"ld"` → `search("world")` matches `{offset:0, length:5}`, no
  separator inserted).
- An `offsetToItemRanges` gap-boundary unit case.

Regression sweep: `npm test` (567 passed) and `npm run lint` (clean) from
`print/`; `test/viewer/spread-search-highlight.test.ts` and
`test/viewer/search.test.ts` stay green unedited.

Blast radius is contained to `print/src/viewer/pdf.ts` and its unit tests.


natb1:
## QA pass — #936 PDF search cross-line match

**Result: PASS ✓** — 2 items, 2 passed, 0 failed, 0 skipped.

### Items

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | PDF search unit + regression suite passes | shell-command | **PASS** |
| 2 | Changed source and test files exist as described | shell-command | **PASS** |

### Detail

**Item 1 — `npm test --prefix print`**

All 567 tests passed across 32 test files. This includes:
- New feature-proving cases: cross-line match (`search("the quick")` → page 4, offset:6, length:9), spanning-div highlight (exactly two `.search-highlight.active` spans reading "the" and "quick"), offset invariant on every result, and intra-word same-line regression (`search("world")` → offset:0, length:5, no separator).
- Renamed helper tests: `reconstructPage` and `offsetToItemRanges` unit cases, including the gap-boundary case.
- Pre-existing suites (`search.test.ts`, `spread-search-highlight.test.ts`) green unedited.

**Item 2 — symbol grep**

`print/src/viewer/pdf.ts` defines `reconstructPage`, `offsetToItemRanges`, and `pageLayoutCache`. The replaced symbols `buildPageText`, `offsetToDivRanges`, and `pageTextCache` are absent — the production implementation matches the PR description.

### Coverage note

All four acceptance criteria are exercised by unit tests in `pdf.test.ts` (criterion 1: cross-line match; criterion 2: spanning-div highlight exact coverage; criterion 3: offset invariant; criterion 4: pre-existing suites unedited). No browser walkthrough is required — the highlight logic operates on DOM simulated in jsdom; there is no interactive visual-layout behavior beyond what the tests cover. An optional manual check (open a real multi-page PDF, search a cross-line phrase) is deferred; it is explicitly optional in the plan and not an acceptance-blocking criterion.

natb1:
## Review — PR #1847 (#936)

### Fixed

- **F005** `print/test/viewer/pdf.test.ts:111` — Added a direct unit test in the `reconstructPage` describe block asserting the `hasEOL`-separator contract: `reconstructPage([{str:'foo',hasEOL:true},{str:'bar'}])` produces `.text === 'foo bar'` and `.items === [{start:0,length:3},{start:4,length:3}]`. This validates the headline feature at the unit level, not just through integration paths. (commit ce83e3f0)

- **F006** `print/src/viewer/pdf.ts:327,670` — Extracted `type TextContentItem = { str?: string; hasEOL?: boolean }` near the top of `pdf.ts`. Used as the `reconstructPage` parameter type and replaced the two duplicated inline casts at the render and search call sites. (commit ce83e3f0)

### Informational

- **F001** `print/src/viewer/pdf.ts:381-386` — The `applyHighlight` length-mismatch guard would fire if a future caller passed `includeMarkedContent:true` to `getTextContent()`, since that option adds extra items not reflected in `textDivs`. The existing `DETERMINISM` comment (lines 356–358) partially covers this; extending it to both call sites would be cheap reinforcement for future readers.

- **F002** `print/test/viewer/pdf.test.ts:121-130` — The `reconstructPage` test for no-`str` items (marked-content items) documents defensive handling of malformed input rather than a real pdf.js code path — marked-content items only appear with `includeMarkedContent:true`, which is never passed here. The test is still useful as a spec; the comment could note the condition for clarity.

- **F003** `print/src/viewer/pdf.ts:43-54` — Two consecutive `hasEOL:true` items produce a double-space gap (`"foo  bar"`) in the reconstructed text, which would prevent a single-space query from matching across a blank line. Out of scope for #936 (which targets normal cross-line boundaries), but worth noting if blank-line spanning is later requested.

- **F004** `print/src/viewer/pdf.ts:152-169` — Early-break condition `itemStart >= rangeEnd` confirmed correct. An item at exactly `rangeEnd` has an empty intersection so the guard is logically equivalent to the previous `cumulative >= rangeEnd` form; no change needed.

- **F007** `print/src/viewer/pdf.ts:149-151` — `rangeStart` is a pure alias for `offset`, used exactly once. Replacing with `offset` directly and removing the alias would tighten the function; low-priority style note.

- **F008** `print/src/viewer/pdf.ts:148,162` — `divIndex` in the return type reflects the old framing; `itemIndex` would better match the rename from `offsetToDivRanges` to `offsetToItemRanges`. However, `divIndex` accurately describes its actual use (indexing `tl.textDivs`) and the rename would require updating the test assertions too. Not an unambiguous improvement.

- **F009** `print/src/viewer/pdf.ts:314-344` — The render path recomputes `PageLayout` from scratch on every render while the search path caches in `pageLayoutCache`. The `applyHighlight` length-mismatch guard already catches divergence at runtime. Sharing the cache would eliminate dual computation but render may trigger for pages never searched and `textContent` is already in hand; left as-is.

### Security

No attack surface identified beyond the changed TypeScript. CodeQL returned no open alerts for this PR. No dependency changes (`deps=false`).

PR #1842, opened 2026-06-17
qa-fix: add 'already-satisfied' disposition class to auto-resolve already-met items

natb1:
Closes #1841

## Summary

Adds a fourth QA disposition class **`already-satisfied`** so the disposition
workflow can auto-resolve items whose acceptance criterion is already provably
met — without parking the PR in office-hours.

### Root cause (#1040 item 4)

The disposition axis conflated two cases under `opus-fixable`: a genuine **code
defect** Opus can fix, and an **already-satisfied** criterion Opus can confirm
from evidence at hand (passing tests, page text, code) with no code change. The
gated fix-plan phase handled both as "plan a fix"; for the already-satisfied
case it produced no units and signaled `deviation=true`, which the skill treats
as an escalation signal — so the item parked in office-hours even though no
human and no code change were needed.

### Fix

`already-satisfied` items are **partitioned out** of the `dispositions` array
into a new top-level `already_satisfied` return array and **dropped from the
residue as PASS**. Because `dispositions` no longer contains them, every
downstream consumer that filters/iterates it (the `opusFixable` fix-plan filter,
the needs-main join, and the escalation set) naturally excludes them. A run whose
only residue items are all `already-satisfied` yields an empty `dispositions` →
empty escalation set → clean autonomous pass (`dispatch:qa-done`).

## Changes

- **`qa-fix.js`** — `already-satisfied` added to `CLASSIFY_SCHEMA`; classify
  prompt documents the four-way axis and when to emit it; `applyQaDisposition`
  gains an explicit first branch returning `{ final_class: 'already-satisfied',
  verify: 'n/a' }` (no skeptic fan-out, symmetric with the aesthetic bypass); the
  aggregation step partitions the joined list into `dispositions` (opus-fixable /
  needs-main / needs-human only, input order) and `already_satisfied`
  (`{ id, title, kind, rationale }`); header JSDoc updated.
- **`dispatch-qa-disposition`** (normative bash kernel) — mirrors the JS kernel
  with an explicit first `already-satisfied` jq branch and updated header docs;
  the inverted-polarity warning is left intact.
- **`test-dispatch-scripts.sh`** — new `f7` already-satisfied fixture asserts
  the first-branch pass-through (`final_class=already-satisfied`, `verify=n/a`)
  and extends the output-order check to `f1..f7`.
- **`qa-fix/SKILL.md`** — documents the four-class axis and the
  `already_satisfied` return field (Overview + Step 3.5); the Step 4 PR comment
[...]

PR #1848, opened 2026-06-18
dispatch: guard closed issues at the router and retire squash-merged worktrees

natb1:
Closes #1845

## Problem

A worker session was spawned to run `/implement` on #1841 ~6.5 hours **after**
that issue was already closed and its PR (#1842) squash-merged. The dispatch tick
re-dispatched an already-completed issue — wasting a worker spawn and producing a
spurious `dispatch:office-hours` park on a closed issue.

Two independent gaps each allow this; either gate alone would prevent the
re-dispatch, so both are fixed (defense-in-depth).

## Fix 1 — Closed-issue guard in the router (Unit 1)

`dispatch-route` previously derived the phase from labels + open-PR presence only
and never checked `issue.state == CLOSED`, so a closed issue's lingering worktree
routed to `INVOKE /implement`.

- `dispatch-route` now fetches the issue state once — **after** the worktree
  cross-check, **before** the multi-minute `dispatch-provision-worktree` merge —
  and emits a new `STOP closed` directive when the issue is `CLOSED`. A transient
  `gh` failure logs and continues to route (no stranding), matching the existing
  gh-failure fallbacks.
- `dispatch-launch-worker` treats `STOP closed` as a benign no-park race
  (alongside `STOP done | STOP waiting | STOP wrong-worktree`): it releases the
  reservation and ticks, applying **no** `dispatch:office-hours` label. Without
  this, `STOP closed` would fall through to the `*)` arm and park the closed issue
  — the violation this fixes.

## Fix 2 — Retire squash-merged worktrees in the sweep (Unit 2)

`dispatch-sweep`'s merged-PR path called `worktree_in_sync`, whose
`git rev-list --count HEAD --not --remotes` over-counts a local-only merge commit
(left by an earlier `git merge origin/main`) once the remote head branch is
deleted on squash-merge — so a worktree byte-identical to `origin/main` was kept
as "not in sync."

- New sibling helper `worktree_merged_in_sync` in `lib-worktree-in-sync.sh`
  returns retire-able when the working tree is clean **and** the committed tree is
  identical to `origin/main` (`git diff --quiet origin/main HEAD`), **without** the
  reachability check. A bad-ref `git diff` error is logged distinctly from a real
  diff; both fall through to the safe "keep" default.
- `dispatch-sweep`'s merged-PR path now calls `worktree_merged_in_sync`. The
  parallel CLOSED-issue-no-PR path is left on `worktree_in_sync`: a closed-but-
  unmerged PR does not auto-delete its branch, so it never hits the squash-merge
  bug.

## Tests (Unit 3)

Added regression cases to `test-dispatch-scripts.sh`, each reproducing the bug
(failing on the pre-fix code), and extended two stubs the fixes require:

- **gh stub** — a `issue view <N> --json state -q .state` arm (defaults `OPEN`)
  for the router guard's state fetch.
- **sweep git shim** — a `diff --quiet origin/main HEAD` arm (defaults identical)
  for the new tree-identity helper.
- **closed issue → `STOP closed`** — the router emits the new directive instead of
  `INVOKE /plan-issue`.
- **closed issue → no office-hours park** — `STOP closed` is in the launcher's
  no-park benign-races arm (criterion #4).
- **squash-merged worktree retired despite non-zero rev-list** — the sweep logs
  `REMOVE_MERGED` when rev-list ≥ 1 but the tree is identical to `origin/main`.

Full suite: 3030/3030 passing under `bash`.


natb1:
# QA Summary — Issue #1845 / PR #1848

## Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Router emits `STOP closed` and exits 0 for a closed issue | shell-command | PASS |
| 2 | Launcher routes `STOP closed` to the no-park benign-races arm | shell-command | PASS |
| 3 | Sweep merged-PR path uses tree-identity (`worktree_merged_in_sync`) | shell-command | PASS |
| 4 | Helper `worktree_merged_in_sync` compares trees and omits reachability check | shell-command | PASS |
| 5 | Full dispatch-scripts regression suite passes under bash | shell-command | PASS |

## Details

**Item 1:** `dispatch-route` contains the closed-issue guard that fetches `gh issue view "$N" --json state -q .state`, checks `ISSUE_STATE == "CLOSED"`, emits `STOP closed`, and exits 0. Confirmed via `grep -Pzoq`.

**Item 2:** `dispatch-launch-worker`'s `case "$DIRECTIVE"` arm lists `"STOP done" | "STOP waiting" | "STOP wrong-worktree" | "STOP closed")` together in the no-park benign-races branch. Confirmed via `grep -Eq`.

**Item 3:** `dispatch-sweep`'s merged-PR path (lines ~177–179) calls `worktree_merged_in_sync`, not `worktree_in_sync`. The remaining `worktree_in_sync` call at line 145 is the intentional closed-issue-no-PR path, which the plan scoped to leave unchanged. Both calls are correct.

**Item 4:** `worktree_merged_in_sync()` uses `git -C "$path" diff --quiet origin/main HEAD` for tree-identity (no `rev-list` in its body). The three-case exit-code dispatch covers: 0 → retire-able, 1 → log `KEEP: tree differs`, >1 → log `ERROR: git diff failed — keeping`.

**Item 5:** Full suite result: **3030/3030 passed, 0 failed** under `bash`. This includes the new regression cases for #1845: closed-issue → `STOP closed`, and squash-merged worktree retired despite non-zero rev-list.

## No Disposition Triage

No residue — all items passed. Step 3.5 was skipped.

natb1:
## Review — #1845 (PR #1848)

### Fixed

- **`test-dispatch-scripts.sh:1739`** — Test 1d's grep regex anchored `"STOP closed"` as the final pipe-joined alternative, so adding any future alternative after it would silently break the test. Replaced with a looser pattern that asserts `"STOP closed"` appears bundled with at least one sibling benign-STOP alternative in a paren-closed case arm, tolerating added or reordered alternatives. _(fix commit: 0b4b6c59)_

### Informational

- **`test-dispatch-scripts.sh:1716`** — Test 1c asserts the `STOP closed` directive and exit code but does not assert that `dispatch-provision-worktree` was not called. A regression that moves the closed-issue guard after provisioning would still pass this test. Not a bug in the current code; a future hardening opportunity.
- **`lib-worktree-in-sync.sh:107`** — `worktree_merged_in_sync` compares the worktree tree against `origin/main` via `git diff --quiet origin/main HEAD`, but `dispatch-sweep` never fetches before calling it. If `origin/main` has advanced since the last fetch, the comparison runs against a stale remote ref. The helper returns `0` (retire-able) in this case — it may retire a worktree whose tree matches a now-stale `origin/main` but diverges from the actual current HEAD. The risk is bounded (the working tree is clean and the tree matches _some_ pushed state of main), but the stale-ref scenario is worth noting.
- **`test-dispatch-scripts.sh:7374-7376`** — The `diffrc` stub key derivation uses `echo "$ctx_path" | tr '/' '_'` to build a fixture filename. `echo` is used rather than `printf '%s'`, which is a minor inconsistency with the repo's `echo`-avoidance convention; in zsh the interpolation is safe here since the path contains no backslash escapes in practice.

### False-positive (security)

- **`dispatch-route:94-97`** — Red-team finding: the closed-issue guard fails open (continues routing) when `gh_retry` exhausts all retries and `ISSUE_STATE` is set to empty string. This is the intended behavior — failing open matches the existing gh-failure fallbacks in `dispatch-route`, preventing a transient GitHub outage from stranding issues. The alternative (fail closed / STOP on gh failure) would park every issue during an API degradation. Not a vulnerability.

### Out-of-scope (security)

- **`dispatch-route:94-101`** — TOCTOU window: issue state is fetched before `dispatch-provision-worktree` merges `origin/main` (~multi-minute operation), so a brief CLOSED→OPEN→CLOSED flip could be missed. This is a pre-existing design constraint of the router's linear guard placement and is not introduced by this PR. The window is narrow and the consequence is a spurious worker spawn (the pre-existing bug), not a security issue.

PR #1889, opened 2026-06-18
dispatch: stamp session artifact into a per-session sidecar for audit joins

natb1:
## What

Stamp the GitHub artifact each dispatch worker session acts on
(`repo/issue/pr/branch/base_sha`) into a per-session JSON sidecar, and surface it
in the token audit as a per-session `artifact` join record. This lets
`dispatch-token-audit` join a session to its real-world GitHub outcome — the
precondition for measuring efficacy against ground truth (the highest-ceiling
item of the runtime-efficacy-instrumentation epic #1858).

## Mechanism

A per-session sidecar written by trusted scripts, next to the transcript:
`<projectdir>/<session-id>.dispatch-stamp.json` holding
`{schema, session_id, repo, issue, pr, branch, base_sha, stamped_at}`.

- A `SessionStart` hook writes it at session birth from **local git only** (no
  network); `pr` starts `null`.
- `pr` is backfilled with the PR number when the PR opens (implement, via
  `dispatch-open-pr`) or read at session start (review-fix / qa-fix, which run on
  an already-open PR). plan-issue has no PR, so its sidecar `pr` stays `null`.
- The audit reads the sidecar by transcript-stem join: the sidecar
  `<id>.dispatch-stamp.json` sits next to `<id>.jsonl`, so `.sessions[].id` is the
  join key.

This was chosen over emitting the record into the transcript as a fenced JSON
block. A join key to GitHub ground truth must not come from spoofable transcript
content (the audit treats all transcript content as opaque, attacker-influenceable
data); a sidecar written only by the hook and `dispatch-open-pr` is authoritative.
Both the initial write (a hook) and the backfill (a scripted `jq` update) are
model-independent — the same scripted-durable-write precedent as
`dispatch-mark-complete` / #824.

## Units

1. `dispatch-stamp-session` writer/backfill script (git-only initial write,
   worker-branch-gated, idempotent, PR backfill that never fails its caller) +
   unit tests.
2. `SessionStart` hook (`stamp-dispatch-session.sh`) + `settings.json` wiring
   (`startup|resume` matcher). The hook reads `transcript_path` directly from
   SessionStart stdin (a documented common-input field), so no cwd-mangle fallback
   is needed. **To confirm on the next live worker run** (CI does not exercise the
   harness's SessionStart behavior): that `SessionStart:startup` actually fires for
   a detached `--bg` worker — the docs imply but do not explicitly state it. If it
   doesn't fire, the hook never runs, every sidecar is absent, and the audit reads
   `artifact: null` everywhere. QA check: the sidecar should exist at session start
   with `pr: null`, then carry the real PR number after `dispatch-open-pr`. The
   plan's documented fallback if startup proves unreliable is a scripted
   `dispatch-stamp-session` call at each phase skill's boundary.
3. PR backfill wiring: folded into `dispatch-open-pr` before it echoes the bare PR
   number (stdout contract preserved), plus one-line backfill calls in the
   review-fix / qa-fix preambles. The standalone preamble calls carry a
   `dangerouslyDisableSandbox` note since the sidecar under `~/.claude/projects` is
   outside the sandbox write-allowlist.
4. Audit reader: `aggregate-usage.sh` slurps each transcript's sidecar (absent →
   `[]`), `stage1.jq` emits `artifact: {repo,issue,pr,base_sha,branch}`,
   `stage2.jq` carries it into the per-session summaries. Sidecar-less sessions
   (subagents, router ticks) yield `artifact: null`.
5. `dispatch-token-audit/SKILL.md` documents the join (slice example + prose).

## Coordination with sibling #1860 (this PR ships first and sets the precedent)

The sidecar is the authoritative source for the overlapping join keys
(`repo/issue/pr/base_sha`). #1860's end-of-run outcome envelope — the model's
outcome self-report (findings/fixes/disposition) — should carry only its
non-overlapping outcome fields, not a second copy of the join keys. The audit
grows exactly one sidecar join-key reader here; #1860 adds its own reader for
genuinely different data. This precedent is documented in the audit SKILL.md.

## Tests

- `test-dispatch-scripts.sh` — covers Unit 1 (initial-write fields, branch no-op,
  backfill set/no-op, idempotent `pr` preserve) and Unit 3 (`dispatch-open-pr`
  backfills the sidecar while keeping stdout the bare PR number; missing-sidecar
  run is non-fatal).
- `test-aggregate-usage.sh` — the fixture's sidecar surfaces as `artifact` with the
  backfilled `pr/repo/issue`; a sidecar-less session yields `artifact: null`.

### DIFF
--- a/.claude/hooks/stamp-dispatch-session.sh
+++ b/.claude/hooks/stamp-dispatch-session.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# stamp-dispatch-session.sh — write a per-session dispatch sidecar at session
+# birth using local git only (no network). Records repo/issue/branch/base_sha;
+# `pr` starts null and is backfilled later by dispatch-stamp-session --backfill-pr.
+#
+# Bound to SessionStart:startup (fresh claude --bg worker sessions) and
+# SessionStart:resume (resumed sessions). Always exits 0 — never blocks session
+# start.
+set -uo pipefail
+trap 'echo "[stamp-dispatch-session] WARNING: unexpected error on line $LINENO" >&2; exit 0' ERR
+
+STDIN_JSON=$(cat 2>/dev/null) || STDIN_JSON=""
+SESSION_ID=$(printf '%s' "$STDIN_JSON" | jq -r '.session_id // empty' 2>/dev/null) || SESSION_ID=""
+TRANSCRIPT_PATH=$(printf '%s' "$STDIN_JSON" | jq -r '.transcript_path // empty' 2>/dev/null) || TRANSCRIPT_PATH=""
+
+if [[ -z "$SESSION_ID" || -z "$TRANSCRIPT_PATH" ]]; then
+  echo "[stamp-dispatch-session] session_id or transcript_path missing in hook payload; skipping stamp" >&2
+  exit 0
+fi
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)" || exit 0
+STAMP_SCRIPT="$SCRIPT_DIR/../skills/dispatch-propagate/scripts/dispatch-stamp-session"
+
+if [[ ! -x "$STAMP_SCRIPT" ]]; then
+  echo "[stamp-dispatch-session] dispatch-stamp-session not found or not executable at '$STAMP_SCRIPT'; skipping stamp" >&2
+  exit 0
+fi
+
+"$STAMP_SCRIPT" --session-id "$SESSION_ID" --transcript-path "$TRANSCRIPT_PATH" || true
+
+exit 0
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -133,6 +133,15 @@
             "command": "\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/restore-dispatch-skill.sh"
           }
         ]
+      },
+      {
+        "matcher": "startup|resume",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/stamp-dispatch-session.sh"
+          }
+        ]
       }
     ],
     "WorktreeCreate": [
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-open-pr
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-open-pr
@@ -61,6 +61,10 @@ usage() {
 # Used both for stripping a stray keyword before an extra number.
 KEYWORD_RE='close[sd]?|fix(?:e[sd])?|resolve[sd]?'
 
+# Resolve this script's own directory so the sibling dispatch-stamp-session
+# (PR backfill) is found regardless of the caller's cwd.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
 # --- argument parsing ---------------------------------------------------------
 
 PRIMARY=""
@@ -215,6 +219,12 @@ while :; do
     -q '.closingIssuesReferences[].number' | sort -n -u)"
 
   if [[ "$PARSED" == "$INTENDED_SORTED" ]]; then
+    # Stamp the now-known PR number into this session's dispatch sidecar so the
+    # token audit can join the session to its PR (#1861). The backfill is
+    # non-fatal (it exits 0 on any miss) and all its output goes to stderr; the
+    # `|| true` defends the set -e shell against any non-zero. The stdout
+    # contract is preserved: the bare PR number remains the ONLY stdout line.
+    "$SCRIPT_DIR/dispatch-stamp-session" --backfill-pr "$PR_NUM" 1>&2 || true
     echo "$PR_NUM"
     exit 0
   fi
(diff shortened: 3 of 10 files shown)


natb1:
## QA Pass — #1861 dispatch session-start stamping

### Results: 6/6 PASS, 0 FAIL, 0 SKIP

No browser component detected (pure scripts/hooks/docs PR). All items verified via shell commands.

### Items

| # | Title | Result |
|---|-------|--------|
| 1 | dispatch-stamp-session initial write + backfill unit tests | PASS |
| 2 | dispatch-open-pr backfills sidecar, keeps bare PR# on stdout | PASS |
| 3 | aggregate-usage.sh surfaces the artifact join field | PASS |
| 4 | review-fix and qa-fix SKILL.md preambles document the backfill call | PASS |
| 5 | dispatch-token-audit SKILL.md documents the artifact join | PASS |
| 6 | SessionStart hook exists and is registered in settings.json | PASS |

### Detail

**Item 1 + 2** (`test-dispatch-scripts.sh`): 3030/3030 tests passed. Confirms `dispatch-stamp-session` writes the correct sidecar fields (`repo`, `issue`, `branch`, `base_sha`, `pr: null`) on initial write; is a no-op on `main` and `office-hours-5`; backfills `.pr` correctly and is idempotent; and that `dispatch-open-pr` backfills the sidecar while keeping its stdout contract (bare PR number only).

**Item 3** (`test-aggregate-usage.sh`): 32/32 tests passed. Confirms the fixture's sidecar surfaces as `.artifact = {repo, issue, pr, base_sha, branch}` on the worker session, and sessions without a sidecar (router ticks) yield `artifact: null`.

**Item 4**: Both `.claude/skills/qa-fix/SKILL.md` and `.claude/skills/review-fix/SKILL.md` reference `dispatch-stamp-session --backfill-pr` in their preambles.

**Item 5**: `.claude/skills/dispatch-token-audit/SKILL.md` documents the per-session artifact join (lines 60–65), including a jq slice example and prose describing the join key and the #1860 precedent.

**Item 6**: `.claude/hooks/stamp-dispatch-session.sh` is present and executable; `.claude/settings.json` has it registered under `SessionStart` with the `startup|resume` matcher.

---

Clean autonomous pass — no residue, no bugs, no deferred items.

natb1:
## Review — PR #1889

### Fixed

- **`dispatch-stamp-session:252` (code-review-0 / review-0)** — Mode A's final `mv "$sidecar.tmp" "$sidecar"` had no error handler, unlike the `jq` build step above it. Added an exit-0 fallback mirroring Mode B's pattern. Also extended the idempotency check to preserve `base_sha` and `stamped_at` on resume (not just `.pr`), since re-deriving them on every SessionStart:resume was silently overwriting the original stamp-time values. Fix commit: `bccb9aed`.

- **`dispatch-stamp-session:118` (review-1)** — `^[0-9]+$` accepted `0` as a valid PR number; changed to `^[1-9][0-9]*$` to match the pattern already used for `--days` in `aggregate-usage.sh`. Fix commit: `bccb9aed`.

- **`test-dispatch-scripts.sh:370` (code-review-2)** — All six dispatch-stamp-session test cases used an HTTPS origin URL; the SSH normalization branch (`git@github.com:owner/name.git → owner/name`) had no test coverage. Added test case 1b with an SSH remote. Fix commit: `bccb9aed`.

### Informational

- **`dispatch-stamp-session:222` (code-review-1 / input-validation)** — `TRANSCRIPT_PATH` is used to derive the sidecar path without path-traversal validation. The hook payload originates from the Claude daemon's own JSON (not external input), so the threat model is weak, but callers with unusual paths should be aware. No change required for this PR.

- **`dispatch-stamp-session:145` (code-review-3 / review)** — Mode B's `find … | head -n1` silently picks the first match when multiple sidecars happen to share the same session ID prefix. In practice session IDs are UUIDs (no collisions), so this is not a real concern in operation. Noted for completeness.

- **`stamp-dispatch-session.sh:48` (code-review-4)** — The SessionStart hook matcher `startup|resume` lacks anchors. As an unanchored regex it could match a longer event name sharing those substrings, though no such event currently exists. No change required.

### Refuted (adversarial-verify dropped, 2-0 vote each)

- **`dispatch-stamp-session:145` (input-validation-0)** — Finding: `CLAUDE_CODE_SESSION_ID` interpolated into `find -name` without rejecting glob metacharacters. Refuted: `CLAUDE_CODE_SESSION_ID` is set exclusively by the Claude daemon and is always a UUID (`[0-9a-f-]`), which structurally cannot contain `*`, `?`, or `[`. An entity that could inject a glob metachar into the daemon-owned env var already has arbitrary write access to the sidecar directory — no privilege boundary is crossed.

- **`dispatch-stamp-session:133-145` (red-team-0)** — Same fnmatch surface, same conclusion from the red-team reviewer. Refuted on identical grounds: the threat model requires controlling the session environment, which already implies full access.

- **`dispatch-token-audit/SKILL.md:74-79` (red-team-1)** — Finding: artifact fields (`repo`, `branch`, `base_sha`) should carry opaque-data treatment like `tool_errors[].signature`. Refuted: `base_sha` is a 40-char hex SHA (uninjectible), `branch` is gated by the `^[0-9]+-` worker guard and `git-check-ref-format` constraints, and `repo` derives from the project's own `.git/config` (attacker write-access to that already means full compromise). The provenance is qualitatively different from `tool_errors` (which can contain arbitrary Bash error text). Defense-in-depth improvement for a future PR; not blocking.

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1897, opened 2026-06-19
Resolve nested workspace paths in dispatch CI change-detection

natb1:
Closes #1887

## Problem

`resolve_dirty_apps` in `.claude/skills/dispatch-propagate/scripts/lib.sh` mapped
a changed file to its workspace by the file's **first** path segment
(`top_dir="${file%%/*}"`). For a nested workspace path like `packages/ds`, an
edit to `packages/ds/base.css` collapsed to `top_dir=packages`, which matched
neither `all_apps` (keyed by the full workspace string `packages/ds`) nor
`shared_pkgs` (keyed by the short name `ds`). The file was silently dropped — the
`ds` workspace was never marked dirty and no consuming app retriggered.

## Fix

Resolve a changed file to its workspace by **longest-prefix match** against the
actual `workspaces[]` entries, then:

- mark the matched workspace dirty using its **full path** (`packages/ds` for a
  nested direct change; the flat name for flat workspaces — unchanged), and
- look up shared-package consumers by the workspace **basename** (`packages/ds` →
  `ds`), preserving today's flat behavior exactly (`${ws##*/}` == `ws` for flat
  workspaces).

A new top-level helper `_resolve_workspace_for_file` picks the longest workspace
key `ws` such that `file == ws` or `file` begins with `"$ws/"`. The trailing `/`
is load-bearing: it keeps `print` from matching `printer/x` and `packages/ds`
from matching `packages/ds-utils/x`. The root-config `case` arm
(`firebase.json`, `firestore.rules`, …) is untouched and still fans out to all
workspaces.

`get-changed-apps.sh` delegates to `resolve_dirty_apps` with no duplicated
resolution logic, so its change-detection picks up the fix for free.

## The one explicit trade

The basename bridge relies on the leaf directory name equalling the package's
`@commons-systems/<name>` short name — the same convention flat workspaces
already depend on. If a future nested workspace's leaf dir ever differs from its
short name, its shared-package retrigger would silently break. This is stated in
a code comment as a follow-up condition. Glob workspace entries (`packages/*`)
remain out of scope (a literal `packages/*` entry already hard-errors today at
the per-workspace `jq`) and are noted in the same comment.

## Tests

Added a `resolve_dirty_apps` block to `test-dispatch-scripts.sh` (runs in the
`hook-tests` CI job) using a synthetic fixture, since the real repo has no nested
workspace yet:

1. Nested direct change + shared retrigger — `packages/ds/base.css` → `blog`,
   `packages/ds`.
2. Flat regression — `landing/index.html` → `landing`.
3. Prefix-boundary — `packages/dsx/foo` → marks nothing.
4. Root-config fan-out — `package.json` → all workspaces.

Full suite passes locally: `Results: 3034/3034 passed, 0 failed`.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -685,6 +685,25 @@ kill_tree() {
   done <<< "$pids"
 }
 
+# Resolve the workspace key that owns a changed file via longest-prefix match.
+# Echoes the LONGEST key `ws` from the `all_apps` associative array such that
+# `file == ws` OR `file` begins with "$ws/" (the trailing "/" is load-bearing:
+# without it "print" would match "printer/x" and "packages/ds" would match
+# "packages/ds-utils/x"). Reads `all_apps` via bash dynamic scoping from the
+# caller. Echoes nothing if no workspace matches.
+# Args: $1 = changed file path
+_resolve_workspace_for_file() {
+  local file="$1" ws best=""
+  for ws in "${!all_apps[@]}"; do
+    if [ "$file" = "$ws" ] || [ "${file#"$ws"/}" != "$file" ]; then
+      if (( ${#ws} > ${#best} )); then
+        best="$ws"
+      fi
+    fi
+  done
+  echo "$best"
+}
+
 # Resolve which apps are affected by a set of changed files.
 # Reads changed file paths from stdin, one per line.
 # Outputs dirty app names to stdout, one per line (unsorted).
@@ -726,11 +745,10 @@ resolve_dirty_apps() {
   done
 
   declare -A dirty_apps
-  local file top_dir
+  local file
 
   while IFS= read -r file; do
     [ -z "$file" ] && continue
-    top_dir="${file%%/*}"
     case "$file" in
       firebase.json|firestore.rules|storage.rules|package.json|package-lock.json)
         # Root-level config changes affect all workspaces
@@ -739,15 +757,27 @@ resolve_dirty_apps() {
         done
         ;;
       *)
-        # Check if this is a shared package change
-        if [ -n "${shared_pkgs[$top_dir]+x}" ]; then
-          for app in ${shared_pkgs[$top_dir]}; do
+        # Longest-prefix match the file to its owning workspace key.
+        local ws short
+        ws=$(_resolve_workspace_for_file "$file")
+        [ -z "$ws" ] && continue
+        # Shared-package lookup keys on the workspace's leaf dir, which is
+        # bridged to the @commons-systems/<name> dependency short name in the
+        # shared_pkgs build above. This assumes the leaf dir equals the short
+        # name; if a future nested workspace's leaf dir ever differs (e.g.
+        # packages/foo providing @commons-systems/bar), its shared-package
+        # retrigger silently breaks — revisit if that convention is relaxed.
+        # A literal glob workspace entry (e.g. packages/*) is out of scope: it
+        # already hard-errors today at the jq on $repo_root/$app/package.json.
+        short="${ws##*/}"
+        if [ -n "${shared_pkgs[$short]+x}" ]; then
+          for app in ${shared_pkgs[$short]}; do
             dirty_apps["$app"]=1
           done
         fi
-        # Check if this is a direct app change
-        if [ -n "${all_apps[$top_dir]+x}" ]; then
-          dirty_apps["$top_dir"]=1
+        # Check if this is a direct app change (full matched path).
+        if [ -n "${all_apps[$ws]+x}" ]; then
+          dirty_apps["$ws"]=1
         fi
         ;;
     esac
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -713,7 +713,7 @@ resolve_dirty_apps() {
 
   # Discover all workspaces from root package.json
   declare -A all_apps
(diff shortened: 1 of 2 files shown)


natb1:
## QA — issue #1887 (PR #1897)

**Result: PASS — all items verified, no residue.**

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Test suite passes (covers #1887 acceptance criteria) | shell-command | PASS |
| 2 | Golden path: nested edit marks nested workspace dirty + retriggers dependents (AC 1+2) | shell-command | PASS |
| 3 | Regression + boundary: flat workspaces resolve as before; sibling-prefix does not leak (AC 3) | shell-command | PASS |
| 4 | End-to-end via get-changed-apps.sh against the real repo (AC 4) | shell-command | PASS |

**Counts:** 4 PASS, 0 FAIL, 0 SKIP

### Detail

- **Item 1** — `bash test-dispatch-scripts.sh` ran 3043/3043 tests, 0 failed. All four new `resolve_dirty_apps` assertions passed:
  - `resolve_dirty_apps: nested direct + shared retrigger` ✓
  - `resolve_dirty_apps: flat workspace regression` ✓
  - `resolve_dirty_apps: prefix boundary marks nothing` ✓
  - `resolve_dirty_apps: root-config fan-out marks all workspaces` ✓
- **Item 2** — `packages/ds/base.css` piped through `resolve_dirty_apps` with a synthetic 3-workspace fixture (`landing`, `blog`, `packages/ds`, where `blog` depends on `@commons-systems/ds`) → output `blog,packages/ds` (sorted). AC 1 and AC 2 satisfied.
- **Item 3** — `landing/index.html` → `landing` (flat regression clean); `packages/dsx/foo` → empty (trailing-`/` boundary prevents `packages/ds` from matching `packages/dsx/*`). AC 3 satisfied.
- **Item 4** — `get-changed-apps.sh --base origin/main` exits 0 with no output for this branch's `.claude/`-only diff. The script delegates to `resolve_dirty_apps` with no duplicated logic; AC 4 satisfied transitively.

No bugs found. No browser walkthrough (pure scripts change). No deferred items.

natb1:
## Review pass — #1897

### Fixed

- **`ws` not declared `local` in `resolve_dirty_apps`** (`.claude/skills/dispatch-propagate/scripts/lib.sh:722`) — Added `ws` to the `local workspace_list` declaration so the variable is scoped to the function from entry. *(commit aa19eac0)*
- **Test temp-dir uses `$root` instead of `$TMPDIR_TEST`** (`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:31581`) — Renamed the nested-workspace test's temp dir variable to `TMPDIR_TEST` and added `TMPDIR_TEST=""` reset after cleanup, so the EXIT trap covers cleanup on unexpected failure. *(commit aa19eac0)*

### Informational

- **Bash dynamic scoping for `all_apps` in `_resolve_workspace_for_file`** (`.claude/skills/dispatch-propagate/scripts/lib.sh:695–705`) — The helper reads `all_apps` via dynamic scoping from its caller (`resolve_dirty_apps`). When invoked in a subshell (`ws=$(...)`), this works correctly because bash exports associative arrays to subshells from the calling scope. No change needed, but worth noting for future refactors that move `_resolve_workspace_for_file` out of `lib.sh`.
- **Leaf-dir/package-name bridge assumption** (`.claude/skills/dispatch-propagate/scripts/lib.sh:764–772`) — `short="${ws##*/}"` assumes the workspace's leaf directory equals its `@commons-systems/<name>` short name. A code comment already flags this and the follow-up condition; documented correctly.
- **Missing test cases: changed file in both a nested workspace and a root-config simultaneously** (`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:31588`) — The four cases cover the core scenarios; edge combinations (e.g. a file that matches both a direct workspace and the root-config arm) are untested. The root-config arm is unconditional (`case` matches first), so no functional gap today.
- **`rm -rf "$root"` runs only at the end of the section** (`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:31607`) — Fixed by renaming to `TMPDIR_TEST` (above); the EXIT trap now covers the full section.
- **No test for glob workspace entries** (`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:31597–31600`) — Glob entries (`packages/*`) hard-error at the per-workspace `jq` today; no test covers that error path. The code comment documents this as out-of-scope.

### False-positive (security)

- **Unquoted `${shared_pkgs[$short]}` expansion** (`.claude/skills/dispatch-propagate/scripts/lib.sh:774`) — The `shared_pkgs` map is built from `jq -r` output over the repo's own `package.json`; workspace names with internal spaces would be malformed JSON and `jq` would reject them before they reached the map. No realistic injection path.
- **`short="${ws##*/}"` — path traversal via workspace key** (`.claude/skills/dispatch-propagate/scripts/lib.sh:772`) — The workspace key comes from the locked `workspaces[]` array in `package.json`, not from user input. A maliciously crafted `package.json` in the repo would require write access, which already grants arbitrary code execution.
- **`echo "$best"` — unsanitized workspace key output** (`.claude/skills/dispatch-propagate/scripts/lib.sh:704`) — Same trust boundary: the workspace key is a controlled repo value, not external input. The output feeds back into `resolve_dirty_apps`'s own logic.
- **`root=$(mktemp -d)` without `local`** (`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:31581`) — Test-script global scope; no function boundary involved. Fixed by the `TMPDIR_TEST` rename above.

PR #1898, opened 2026-06-19
dispatch: sanitize launcher environment so the tick survives env-bloated interactive shells (E2BIG)

natb1:
Closes #1879

## Problem

The `dispatch` launcher (`nix/packages/dispatch.nix`) is a `writeShellScriptBin`
wrapper that `exec`s `dispatch-tick` carrying the launching interactive shell's
**full inherited environment**. A long-lived shell running the direnv hook on
every prompt accumulates an unbounded base64 `DIRENV_DIFF`/`DIRENV_WATCHES` blob
(compounded by triplicated nix-profile `PATH` entries). Eventually argv +
environment approaches `ARG_MAX` (3.2 MB here), and the next `execve` the tick
reaches dies with `Argument list too long` (E2BIG) — surfacing as a cryptic
`rm: Argument list too long` with no signal about the real cause. The autonomous
chain silently dies.

The wrapper's very first real statement today, `TOPLEVEL="$(git rev-parse
--show-toplevel)"`, is itself a command substitution that `execve`s `git`
carrying the bloated environment — so it E2BIGs *before* any sourced library or
external sanitizer could run.

## Fix

Harden the launcher boundary so the tick runs under a sanitized environment,
immune to the interactive shell's accumulated bloat — and, if a clean run is
still impossible, fail with a clear, actionable error naming the env-size cause
instead of a downstream E2BIG.

- **Unit 1** — new sourced library
  `.claude/skills/dispatch-propagate/scripts/lib-sanitize-launch-env.sh` defining
  `sanitize_launch_env()`: re-unsets the `DIRENV_*` family, dedups `PATH`
  (exec-free, first-occurrence-wins), then a size guard that measures
  `env | wc -c` against `getconf ARG_MAX` minus a 128 KB safety margin and fails
  loud with an actionable message if the pruned environment still exceeds budget.

- **Unit 2** — wires both launchers (`nix/packages/dispatch.nix` and
  `nix/packages/office-hours.nix`). The decisive prune must be **exec-free and
  run first**, so the literal first statement after `set -euo pipefail` is the
  shell builtin `unset -v "${!DIRENV_@}"` (prefix-name expansion, future-proof
  across the whole `DIRENV_*` family). Empirically that alone drops the multi-MB
  blob and brings the environment far below `ARG_MAX`, so `git`, the `source`,
  and the final `exec` all then succeed. After `TOPLEVEL` resolves, the wrapper
  sources the library and runs the full `sanitize_launch_env` (defensive
  re-unset + PATH dedup + size guard) before `exec`.

- **Unit 3** — `test-sanitize-launch-env.sh` covers the bug (DIRENV prune + PATH
  dedup + a child `execve` succeeding), empty-prefix `set -u` safety, and the
  over-budget guard (non-zero exit + actionable stderr). Auto-discovered by
  `run-unit-tests.sh`'s `test-*.sh` glob.

## Design notes

- **Denylist, not allowlist.** Dropping the known unbounded `DIRENV_*` family and
  deduping `PATH` preserves everything else (cert/locale/gh-auth/git config)
  automatically. An allowlist re-exec that omitted `SSL_CERT_FILE` /
  `NIX_SSL_CERT_FILE` / `CURL_CA_BUNDLE` (NixOS sets these for TLS) would break
  `gh`/`git` HTTPS — the exact class of downstream-cryptic failure this issue
  exists to kill. A denylist also fails *loud* (via the size guard) rather than
  silent. It is also the only approach implementable purely in builtins before
  the first exec.

- **`office-hours.nix` included deliberately.** The sibling launcher has the
  identical latent bug (same `exec`-with-inherited-env shape, same env-bloated
  source shell). Sanitizing at every launcher boundary is the greenfield design
  and is one PR, so the fix is applied to both wrappers. The issue names only
  `dispatch`; `office-hours` is a called-out extension, not silent scope creep.

- **Scope boundary.** This handles the reported regime: total env near `ARG_MAX`
  while the wrapper itself still starts. If a *single* variable ever exceeds the
  fixed 128 KB per-string `MAX_ARG_STRLEN` cap, the shell→wrapper `execve` fails
  before the wrapper runs at all — a shell-side limit outside this repo's
  control, which the issue explicitly acknowledges.

## Verification

- `test-sanitize-launch-env.sh`: 11/11 passing locally (prune, dedup, child
  execve, empty-prefix safety, guard fires with actionable message).
- Both `writeShellScriptBin` wrappers evaluate to valid derivations with the new
  inline bootstrap + source lines (Nix `''${` escaping confirmed); `nix flake
  check` runs in CI.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib-sanitize-launch-env.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib-sanitize-launch-env.sh
@@ -74,7 +74,11 @@ sanitize_launch_env() {
         fi
       fi
     done
-    export PATH="$rebuilt"
+    # Guard the export: if dedup produced nothing (e.g. a malformed PATH of all
+    # separators like `:::`), leave the original PATH untouched rather than
+    # clobbering it to empty — an empty PATH would break the `env`/`wc` forks in
+    # step 3 with a confusing "command not found".
+    [[ -n "$rebuilt" ]] && export PATH="$rebuilt"
   fi
 
   # --- Step 3: size guard (acceptance criterion 3) ---------------------------
--- a/.claude/skills/dispatch-propagate/scripts/lib-sanitize-launch-env.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib-sanitize-launch-env.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+# Sanitize the inherited shell environment before a dispatch launcher execs the
+# chain (#1879).
+#
+# Symptom: when `dispatch` is launched from an env-bloated interactive shell
+# (e.g. direnv layering a large PATH and a pile of DIRENV_* vars, plus whatever
+# else accumulated in a long-lived terminal), the chain dies with
+# "Argument list too long" (E2BIG). execve(2) caps the combined size of the
+# argument vector AND the environment block at ARG_MAX; once the inherited
+# environment alone approaches that ceiling, the very next external command the
+# launcher runs fails to exec.
+#
+# This library defines one function, sanitize_launch_env(), that the Nix
+# launcher wrappers `source` to prune the environment back under the ceiling.
+#
+# WHY THE ORDER MATTERS (this is the whole point):
+#   The first two steps — unsetting the DIRENV_* family and deduplicating PATH —
+#   MUST be exec-free (pure bash builtins, no external commands at all). They run
+#   while the environment is still bloated, i.e. while execve would E2BIG. If
+#   step 1 or step 2 shelled out to `tr`/`awk`/`sed`/`printf`-in-a-subshell, that
+#   fork+exec would itself fail before it could shrink anything.
+#   Only AFTER those builtin-only steps have pruned the environment is it safe to
+#   run external commands. So the size guard (step 3) — which uses `env`, `wc`,
+#   and `getconf` — comes last, when execve works again. Measuring the
+#   environment BEFORE the prune would mean even `env | wc -c` could E2BIG.
+#
+# The wrappers also do an inline pre-source `unset` of the DIRENV_* family before
+# they can even source this file; step 1 here repeats that defensively/
+# idempotently. The inline pre-git bootstrap, any allowlist, and any `env -i`
+# rebuild are out of scope for this library — it is exactly one function.
+#
+# Sourced under `set -euo pipefail` by the launcher wrappers, so every construct
+# here is written to be safe under `set -u` (no unbound-variable trip) and not
+# to trip `set -e` (the `|| true` guards on best-effort steps).
+
+sanitize_launch_env() {
+  # --- Step 1: re-unset the direnv family (defensive / idempotent) -----------
+  # `${!DIRENV_@}` is bash prefix-name expansion: it lists the NAMES of all set
+  # variables beginning with `DIRENV_`. When none are set it expands to nothing,
+  # so `unset -v` is called with no operands — a harmless no-op, and crucially it
+  # does NOT trip `set -u` (validated: empty expansion, no unbound-variable
+  # error). `unset` cannot fail in any way we care about here, but the
+  # `2>/dev/null || true` keeps a stray non-zero from tripping `set -e`. This is
+  # a builtin — no fork, no exec — so it is safe while the env is still bloated.
+  unset -v "${!DIRENV_@}" 2>/dev/null || true
+
+  # --- Step 2: deduplicate PATH with an exec-free builtin loop ----------------
+  # Duplicate PATH components are pure bloat: they cost execve budget without
+  # adding any lookup capability. A direnv-layered interactive shell readily
+  # accumulates the same entry many times over. We rebuild PATH keeping only the
+  # first occurrence of each component (first-wins preserves lookup precedence),
+  # using ONLY bash builtins — no `tr`, `awk`, `sed`, or subshell `printf`,
+  # because this runs before the env is small enough for execve to succeed.
+  #
+  # `local IFS=:` makes word-splitting break PATH on `:`. `local -A seen` is an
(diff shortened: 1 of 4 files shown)


natb1:
## QA summary — #1879 (PR #1898)

### Items executed

| # | Title | Result |
|---|-------|--------|
| 1 | Unit test suite passes (all 3 cases) | **PASS** |
| 2 | dispatch.nix wires bootstrap + source + call in correct order | **PASS** |
| 3 | office-hours.nix wires the identical bootstrap + source + call | **PASS** |
| 4 | E2BIG repro: large DIRENV_DIFF pruned, env passes through | FAIL → already-satisfied |
| 5 | Size-guard fires on over-budget env with actionable error | FAIL → already-satisfied |
| 6 | No regression: normal inherited env passes cleanly | **PASS** |

**Counts:** 4 PASS, 2 FAIL (both classified already-satisfied), 0 SKIP

No browser walkthrough (pure scripts/Nix change, no browser component detected).

### Disposition triage

Step 3.5 ran on 2 residue items. Both classified `already-satisfied` by Opus — dropped as PASS. No items remain in the escalation set.

- **Item 4 — E2BIG repro:** The FAIL was a test-harness artifact: zsh escaped `!` as `\!` before the `bash -c` string, causing a `bad substitution` in the scaffolding, not the product. Item 1's unit tests already verified all three sub-criteria (case1a: DIRENV_DIFF gone, case1b: PATH deduped, case1c: child execve succeeds). Expected outcome provably met; no code defect.

- **Item 5 — Size-guard actionable error:** The `GUARD MISS` label came from grep being unable to run in the oversized test environment, not a guard failure. `sanitize_launch_env` did return 1 and did print the correct stderr naming `#1879` and the env-size cause. Item 1's unit tests (case3) verified all three sub-criteria as PASS. Expected outcome provably met; no code defect.

natb1:
## Review — #1879 dispatch chain dies with E2BIG

Reviewed 4 changed files (`lib-sanitize-launch-env.sh`, `test-sanitize-launch-env.sh`, `dispatch.nix`, `office-hours.nix`). Surface: `code`. Sources: code-review, review, input-validation, red-team, security-review, auth, data-exposure, firebase.

### Fixed (commit e9580733)

- **`lib-sanitize-launch-env.sh:77`** — PATH export guarded for all-separator input: when every PATH component is empty (e.g. `PATH=":::"`) the dedup loop produces an empty `rebuilt`; replaced `export PATH="$rebuilt"` with `[[ -n "$rebuilt" ]] && export PATH="$rebuilt"` so a malformed PATH is preserved rather than clobbered to empty, which would break step 3's `env`/`wc`/`getconf` execs. (sources: code-review, review)
- **`test-sanitize-launch-env.sh:49`** — `run_case` crash guard added: when a case subshell exits non-zero without printing any PASS/FAIL lines, the harness now counts one synthetic FAIL rather than a vacuous pass. (source: review)
- **`test-sanitize-launch-env.sh:99`** — case1 child-execve assertion fixed: `env >/dev/null; child_rc=$?` replaced with `if env >/dev/null; then child_rc=0; else child_rc=$?; fi` so `set -e` can't abort before the assertion captures the return code. (source: review)
- **`test-sanitize-launch-env.sh:134`** — case3 made ARG_MAX-dynamic: BIG size is now computed as `(ARG_MAX - 131072) + 65536` at runtime instead of the hardcoded 3,100,000 bytes that assumed ARG_MAX = 3,200,000; skip guard added for environments where `ARG_MAX ≤ 262144`. (sources: code-review, review)
- **`test-sanitize-launch-env.sh:152`** — same case3 dynamic-size fix (separate hunk). (source: review)

- **`lib-sanitize-launch-env.sh:64`** — Glob expansion via unquoted `$PATH` with `local IFS=:` (sources: code-review, input-validation) — 2/2 skeptics refuted. The mechanism is real: bash applies pathname expansion to colon-split PATH words. However, exploitation requires the attacker to both control a PATH entry containing glob metacharacters AND have filesystem entries matching that pattern — preconditions that already imply same-user write access. No privilege boundary exists here: `sanitize_launch_env` runs as the invoking user, the same user who owns any `.envrc` that could inject `PATH=/attacker/*`. `direnv` itself executes `.envrc` as arbitrary shell code unconditionally, so the glob path is strictly weaker than what the attacker already has. Finding dropped; underlying robustness concern (PATH component containing a metachar could be silently expanded to multiple entries) is a correctness gap, not a security one.

### Informational

- **`test-sanitize-launch-env.sh:7`** — Top-level `source` defines `sanitize_launch_env()` in the parent shell but bash does not export functions to child processes by default; the case bodies re-source the lib explicitly inside each `bash -c` subshell, so the top-level source is redundant but harmless. (source: code-review)
- **`test-sanitize-launch-env.sh:68`** — `assert_eq` helper is copy-pasted into each of the three case body functions. Subshells cannot inherit parent-scope functions without `export -f`, so the repetition is load-bearing, not just style. (source: code-review)

### False-positive (security)

- **`test-sanitize-launch-env.sh:169`** — `$SCRIPT_DIR` interpolated into `bash -c "..."` strings: `SCRIPT_DIR` is set by `cd "$(dirname "$0")" && pwd` inside the test harness, not user-supplied. Not attacker-controlled. (source: input-validation)
- **`nix/packages/dispatch.nix:9-11`** — Bypass `sanitize_launch_env` via bash rc files / `BASH_ENV`: the attack requires write access to `~/.bashrc` or ability to set `BASH_ENV`, which already grants same-user code execution through a simpler path. No new capability. (source: red-team)

### Out-of-scope (security)

- **`nix/packages/dispatch.nix:14-29`** — Arbitrary code execution via malicious `lib-sanitize-launch-env.sh` written at TOPLEVEL: write access to the repository directory implies pre-existing repository compromise. The threat model for an env-sanitizer wrapper does not include local write access to the repo. Not in scope for this PR. (source: red-team)

### Deferred

_None._

PR #1839, opened 2026-06-17
audio: client-side tag extraction + .commons-audio sidecar

natb1:
Closes #1039

## What

Gives the `audio` app's local-folder source real client-side tag extraction and a
`.commons-audio/index.json` sidecar — mirroring print's already-merged local-folder
sidecar architecture. Local tracks previously showed placeholder tags ("Unknown
artist" / "Unknown album" / duration 0) and had no persisted player state.

Cloud items are 100% unchanged (Firestore-backed); everything here is additive on
the local side. The sidecar lives inside the user's chosen folder, so it rides
their folder sync and works fully unauthenticated.

## How

Built in 8 units:

1. **`music-metadata` + extractor** — adds the pinned `music-metadata` library and
   `audio/src/local-metadata.ts`'s `extractAudioMetadata`, mapping
   title/artist/album/trackNumber/genre/year/**duration** from local file bytes and
   returning `{}` on any parse failure. Defines the shared `AudioTags` leaf type.
2. **`audio/src/sidecar.ts`** — audio's own sidecar (`.commons-audio/index.json`)
   with a fresh schema: a metadata cache keyed on **`localName`** (local items have
   `storagePath: ""`, so keying on storagePath would collide them all), plus
   local-only `playerState` (queue + `currentLocalName` + `positionSeconds` —
   identified by filename, not a numeric index, since the live queue interleaves
   cloud + local) and named `playlists`. Tolerant parse coerces each field
   independently; a single-flight write chain gives a no-clobber merge with disk
   writes gated by the `writable` flag.
3. **local-source readwrite migration** — prefers `readwrite` (needed to write the
   sidecar) with a read-only fallback (extraction still works in-memory; the write
   no-ops). local-source stays the sole permission owner and hands the directory
   handle to the sidecar at all three bind points before returning. Adds
   `resolveLocalBytes` (raw bytes for enrichment; `null` on a read miss — a
   deliberate skip-and-retry signal).
4. **cache-first list enrichment** — `listLocalTracks` overlays cached metadata
   immediately (cheap, no IO); an async pass extracts uncached files and persists
   them in one batched write, with write suppression when nothing is new and no
   `{}` cached for unreadable files (preserves retry). `toItem` stays synchronous.
5. **home page** — after the initial render (placeholders), runs the enrichment
   pass and re-renders the whole library region so rows show real
   artist/album/duration with no template change; the focus-rescan path enriches too.
6. **player-state persistence + restore** — persists the local-only queue subsequence
[...]

Issue #1039, opened 2026-06-02
audio: client-side tag extraction + .commons-audio sidecar (player-state local)


natb1:
## Context

Part of the audio local-first listening-library epic (epic A). Once a local folder is the
source of truth (A1), its tracks need real tags (artist, album, duration, format) and a
place to persist player-state and playlists — without round-tripping to Firestore, so the
data rides the user's own folder sync and works unauthenticated.

## Approach

- Extract tags client-side: ID3 (MP3), Vorbis comments (FLAC / Ogg), and MP4 atoms (M4A /
  AAC) for artist / album / duration / format — the fields `toAudioItem` already populates.
- Write / read a `.commons-audio/index.json` sidecar **inside the user's folder** holding
  the metadata cache and **player-state / playlists for local items**.
- Cloud items keep using Firestore; only local items use the sidecar.

## Acceptance criteria

- [ ] Local tracks show real extracted tags (artist, album, duration, format).
- [ ] Player-state and playlists for local items persist in the `.commons-audio/index.json`
      sidecar and ride the user's folder sync.
- [ ] Cloud items are unchanged (still Firestore-backed).
- [ ] The sidecar is created on first use and tolerates a missing / partial file.

## References

- Parent epic: A (audio local-first listening library).
- Builds on A1 (`LocalFolderMediaSource`).
- blocked_by A1.

PR #1918, opened 2026-06-19
acceptance gate: resolve nix Playwright browsers when PLAYWRIGHT_BROWSERS_PATH is unset

natb1:
Closes #1853

## Problem

The local QA acceptance gate scripts fall back to
`npx playwright install --with-deps chromium` when `PLAYWRIGHT_BROWSERS_PATH`
is unset. On NixOS, `--with-deps` shells out to `apt-get`, which does not exist,
so the install fails with `apt-get: command not found` and the gate reports a
misleading "BLOCKED — Playwright cannot install Chromium".

The browsers are in fact already provisioned system-wide via nix — `flake.nix`
lists `playwright-driver.browsers` in the devShell and the shellHook exports
`PLAYWRIGHT_BROWSERS_PATH`. Only that env var is missing in the non-interactive
dispatch shell, which never enters `nix develop` / direnv.

This was hit during QA of PR #1839 (#1039).

## Fix

A shared `ensure_playwright_browsers` helper in `lib.sh`. When
`PLAYWRIGHT_BROWSERS_PATH` is unset and `nix` is available (NixOS), it re-execs
the calling script under `nix develop --command`, whose shellHook exports the
exact nix-provisioned browsers path. When `nix` is unavailable (e.g. CI on
ubuntu, where `apt-get` exists), it returns so the existing `npx playwright
install --with-deps chromium` fallback runs unchanged.

Re-exec gets zero-drift for free: the re-run inherits the shellHook export,
which resolves against the flake's *locked* nixpkgs, so the resolved chromium
revision is exactly the one the version-sync guard validates — no new flake
output to maintain. An infinite-loop guard (`_DISPATCH_NIX_REEXEC`) turns a
still-unset var on the second pass into a clear, actionable error instead of a
loop. The approach matches the repo's existing `nix develop --command`
invocation pattern (`.github/workflows/unit-tests.yml`).

## Changes

- **`lib.sh`** — add the `ensure_playwright_browsers` helper next to
  `ensure_deps`.
- **`run-acceptance-tests.sh`** — call the helper once at the top, after
  sourcing `lib.sh`. This single guard covers both existing fallback branches
  (external-base-url and emulator).
- **`run-smoke-tests.sh`** — same wiring; it carries the identical
  apt-get-prone fallback and is hit by the same NixOS failure off-CI.

The existing `npx playwright install --with-deps chromium` blocks are left
unchanged — they remain the nix-less (CI) fallback.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -491,6 +491,32 @@ ensure_deps() {
   fi
 }
 
+# Ensure Playwright browsers are resolvable. When PLAYWRIGHT_BROWSERS_PATH is
+# unset and nix is available (NixOS), re-exec the calling script under
+# `nix develop --command` so the devShell shellHook exports the nix-provisioned
+# browsers path. When nix is unavailable (e.g. CI on ubuntu), return so the
+# caller's `npx playwright install --with-deps chromium` fallback runs.
+#
+# Call as: ensure_playwright_browsers "$0" "$@"
+ensure_playwright_browsers() {
+  if [ -n "${PLAYWRIGHT_BROWSERS_PATH:-}" ]; then
+    return 0
+  fi
+  if command -v nix >/dev/null 2>&1; then
+    if [ -n "${_DISPATCH_NIX_REEXEC:-}" ]; then
+      echo "ERROR: re-exec'd under 'nix develop' but PLAYWRIGHT_BROWSERS_PATH is still unset." >&2
+      echo "       Enter the dev shell ('direnv allow', or 'nix develop') and retry." >&2
+      return 1
+    fi
+    local flake_dir
+    flake_dir="$(git rev-parse --show-toplevel)"
+    export _DISPATCH_NIX_REEXEC=1
+    exec nix develop "$flake_dir" --command "$@"
+  fi
+  # nix unavailable (CI ubuntu): caller's npx fallback handles install
+  return 0
+}
+
 # Extract the app name from the app directory path.
 # Args: $1 = app directory (e.g. "hello" or "/path/to/hello")
 get_app_name() {
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -509,7 +509,14 @@ ensure_playwright_browsers() {
       return 1
     fi
     local flake_dir
-    flake_dir="$(git rev-parse --show-toplevel)"
+    flake_dir="$(git rev-parse --show-toplevel 2>/dev/null)" || {
+      echo "ERROR: could not determine repo root; ensure you are inside a git repository" >&2
+      return 1
+    }
+    [ -n "$flake_dir" ] || {
+      echo "ERROR: git rev-parse --show-toplevel returned empty string" >&2
+      return 1
+    }
     export _DISPATCH_NIX_REEXEC=1
     exec nix develop "$flake_dir" --command "$@"
   fi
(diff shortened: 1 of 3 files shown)


natb1:
# QA Summary — Issue #1853

## Items executed: 7 (all shell-command lane)

| # | Title | Result |
|---|-------|--------|
| 1 | Helper no-ops when PLAYWRIGHT_BROWSERS_PATH is set | PASS |
| 2 | Helper re-execs under nix develop when var unset and nix available | PASS |
| 3 | Infinite-loop guard fires with clear error and return 1 | PASS |
| 4 | Helper returns 0 when nix is unavailable (CI/ubuntu path preserved) | PASS |
| 5 | Helper wired into run-acceptance-tests.sh after source lib.sh | PASS |
| 6 | Helper wired into run-smoke-tests.sh after source lib.sh | PASS |
| 7 | ensure_playwright_browsers defined in lib.sh, syntax valid | PASS |

No browser component detected (pure script change). No browser walkthrough run. No bugs found.

## Notes on verification

- Item 3: The triage-generated Command used `$?` in double quotes, which expanded in the outer bash context (always 0). Re-run with single-quote isolation confirmed the function correctly returns 1 with the actionable error.
- Item 4: Restricting PATH to `/usr/bin:/bin` fails on NixOS (no bash there). Re-run stripping `/nix/var/nix/profiles/default/bin` and `/run/current-system/sw/bin` confirmed the function returns 0 and the npx fallback path is preserved.
- Item 2: The re-exec path confirmed: `exec nix develop <flake_dir> --command <script> <args>` fires correctly when `PLAYWRIGHT_BROWSERS_PATH` is unset and `nix` is on PATH.

natb1:
## Review — #1853 / PR #1918

10 findings across quality and security passes; 1 fixed.

---

### Fixed

- **input-validation-0** — `lib.sh:512` — `flake_dir` from `git rev-parse --show-toplevel` used without checking whether the command succeeded or returned an empty string. Added error-on-failure guard and non-empty check before the `exec` path. (commit `99b991bc`)

---

### Informational

- **code-review-0** — `lib.sh:509` — `ensure_playwright_browsers` uses `return 1` to signal fatal failure (when re-exec'd under `nix develop` but `PLAYWRIGHT_BROWSERS_PATH` is still unset). The `return 1` inside a sourced function is correct but callers that don't check the exit code will silently continue. Not a bug in the PR's scope; noted for callers.
- **code-review-1** — `lib.sh:501` — Function name `ensure_playwright_browsers` does not signal that it may `exec` (replace) the calling process. The re-exec behaviour is documented in the function comment, so this is an observation for readers, not an error.
- **review-1** — `lib.sh:506–509` — Guard variable `_DISPATCH_NIX_REEXEC` uses a generic dispatch-level namespace; if another script in the dispatch family independently adopted the same variable it would collide. No collision exists today.

---

### Dismissed

- **review-0** — `lib.sh:514` — Suggestion to add `--no-update-lock-file` to the `nix develop` call. The repo's `flake.lock` is version-controlled and checked out with the working tree, so `nix develop` operates on the locked revision without fetching updates. Dismissed: the flag is not needed when the lockfile is checked in and present.

---

### False-positive (security)

- **input-validation-1** — `lib.sh:506` — `_DISPATCH_NIX_REEXEC` read from inherited env with no validation. The variable is an internal re-exec guard set only by this same function; it is not a user-controlled value used for access decisions. No validation is needed.
- **red-team-0** — `lib.sh:505` — `command -v nix` resolves whatever binary is on PATH. In this dispatch context PATH is system-controlled; an attacker who can manipulate PATH already has arbitrary code execution by other means.
- **red-team-1** — `lib.sh:506–510` — `_DISPATCH_NIX_REEXEC` could be pre-set by a parent process to suppress the re-exec guard. An attacker with env-var write access could cause more direct harm; this variable only affects whether `nix develop` is re-exec'd, not access decisions.
- **red-team-3** — `run-acceptance-tests.sh:17` — `exec nix develop FLAKE --command "$0" "$@"` where `$0` is the script path. `$0` is set by the shell at invocation time and is not a user-supplied argument; it re-execs the same script. Not exploitable via this path.

---

### Out-of-scope (security)

- **red-team-2** — `lib.sh:511–514` — Further hardening of the flake directory path (e.g. asserting it falls within the expected repo root). Pre-existing concern not introduced by this diff; the `input-validation-0` fix already guards the command failure and empty-string cases. No follow-up filed: below the meaningfulness threshold for a red-team observation on internal tooling.

PR #1922, opened 2026-06-19
CI: bound preview-and-smoke/acceptance jobs and wrap playwright install with timeout+retry

natb1:
Closes #1899

## Problem

CI for a recent PR appeared to "run for an hour" but had actually hung for ~6
hours: 12 of 13 checks passed, but the `preview-and-smoke` job (PR Checks
workflow) never finished and was finally **canceled by GitHub's default 6-hour
job timeout**, not by any failure. The stall was inside `run-smoke-tests.sh`'s
`npx playwright install --with-deps chromium` — `--with-deps` shells out to
`apt-get`, and an apt fetch from `archive.ubuntu.com` hung indefinitely.

Two compounding gaps let a transient network stall burn ~6 hours:

1. **No `timeout-minutes` on any job** in `.github/workflows/pr-checks.yml`, so
   every job inherits GitHub's 6-hour hard cap.
2. **`npx playwright install --with-deps chromium` had no per-command timeout or
   retry** at any of its three call sites — an apt stall hangs forever.

This is a flaky-network reliability fix, not a code regression.

### Unit 1 — bounded `playwright_install_with_deps` helper in `lib.sh`

The three `playwright install --with-deps chromium` call sites were
byte-identical and all source `lib.sh`, so the fix is one shared helper that owns
the timeout + retry + skip-guard, with the three sites collapsing to a single
call.

`playwright_install_with_deps` wraps the install in a per-attempt
`timeout --kill-after=30` (so a `SIGTERM`-ignoring D-state apt blocked on network
I/O still gets `SIGKILL`) plus a small retry loop. It owns the
`PLAYWRIGHT_BROWSERS_PATH` skip-guard (nix-provided browsers skip the install)
and exposes env-tunable knobs with defaults that mirror the repo's existing retry
idioms (`npm-ci-with-retry.sh`, `gh_retry`):

- `PLAYWRIGHT_INSTALL_TIMEOUT` (default `300` seconds, per attempt)
- `PLAYWRIGHT_INSTALL_ATTEMPTS` (default `2` = 1 try + 1 retry)

Worst case is `2 × 300s = 600s`, which fits under both job timeouts below. As a
sourced function it `return`s non-zero on exhaustion (not `exit`), so each
caller's `set -euo pipefail` aborts the script and fails the job. All three call
sites are bare calls (the third wrapped in a `( cd … && … )` subshell to preserve
cwd discipline), so the non-zero return fires `set -e`.

### Unit 2 — `timeout-minutes` on both `pr-checks.yml` jobs

- `acceptance` → `timeout-minutes: 30` (builds, starts emulators, seeds, runs the
  full Playwright suite — needs the larger budget; comfortably exceeds the 600s
  worst-case install).
- `preview-and-smoke` → `timeout-minutes: 20` (only deploys a preview and runs the
  `@hosting` smoke grep — lighter, tighter cap).

The job-level cap is the backstop; the helper's per-command timeout is the
fast-fail.

### Scope note (deliberate)

AC1 scopes the `timeout-minutes` change to `.github/workflows/pr-checks.yml`
only. The root-cause analysis observed that *no* workflow carries
`timeout-minutes`, but a repo-wide sweep is intentionally out of scope for this
issue — other workflow files are not touched here.

## Verification

1. **AC1 (timeouts present)** —
   `grep -n 'timeout-minutes' .github/workflows/pr-checks.yml` shows both jobs
   carry the key (`30` for acceptance, `20` for preview-and-smoke).
2. **AC2 (per-command timeout + retry)** — `playwright_install_with_deps` in
   `lib.sh`: `timeout` wrapper + bounded retry loop + env-tunable knobs.
3. **AC3 (a simulated stall surfaces as a fast failure)** — ran the wrapper
   against the **real** command with `PLAYWRIGHT_INSTALL_TIMEOUT=1` and
   `PLAYWRIGHT_BROWSERS_PATH` unset; the helper logged the retry and returned
   non-zero (`exit=1`) within seconds, not hours.

`bash -n` passes on all three edited scripts.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -516,8 +516,10 @@ playwright_install_with_deps() {
       return 0
     fi
     echo "playwright_install_with_deps: attempt $attempt/$attempts failed or timed out after ${timeout_s}s" >&2
-    sleep 5
     attempt=$((attempt + 1))
+    if [ "$attempt" -le "$attempts" ]; then
+      sleep 5
+    fi
   done
   echo "playwright_install_with_deps: failed after $attempts attempts" >&2
   return 1
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -491,6 +491,38 @@ ensure_deps() {
   fi
 }
 
+# ---- playwright_install_with_deps — bounded, timed Playwright browser install -
+# Wraps `npx playwright install --with-deps chromium` (which shells out to apt-get
+# and can stall indefinitely on a flaky archive mirror — #1899) in a per-attempt
+# `timeout` plus a small retry loop, so a transient network stall fails fast and
+# retries instead of hanging until GitHub's 6-hour job cap. Skips entirely when
+# PLAYWRIGHT_BROWSERS_PATH is set (nix provides browsers). Must run from the app
+# directory (npx resolves the project's playwright). Tunables (env):
+# PLAYWRIGHT_INSTALL_TIMEOUT (default 300 seconds, per attempt),
+# PLAYWRIGHT_INSTALL_ATTEMPTS (default 2 = 1 try + 1 retry). Returns non-zero (so
+# the caller's `set -e` aborts) once attempts are exhausted.
+playwright_install_with_deps() {
+  if [ -n "${PLAYWRIGHT_BROWSERS_PATH:-}" ]; then
+    return 0
+  fi
+  local timeout_s="${PLAYWRIGHT_INSTALL_TIMEOUT:-300}"
+  local attempts="${PLAYWRIGHT_INSTALL_ATTEMPTS:-2}"
+  local attempt=1
+  while [ "$attempt" -le "$attempts" ]; do
+    if [ "$attempt" -gt 1 ]; then
+      echo "playwright_install_with_deps: attempt $attempt/$attempts" >&2
+    fi
+    if timeout --kill-after=30 "$timeout_s" npx playwright install --with-deps chromium; then
+      return 0
+    fi
+    echo "playwright_install_with_deps: attempt $attempt/$attempts failed or timed out after ${timeout_s}s" >&2
+    sleep 5
+    attempt=$((attempt + 1))
+  done
+  echo "playwright_install_with_deps: failed after $attempts attempts" >&2
+  return 1
+}
+
 # Extract the app name from the app directory path.
 # Args: $1 = app directory (e.g. "hello" or "/path/to/hello")
 get_app_name() {
--- a/.claude/skills/dispatch-propagate/scripts/run-acceptance-tests.sh
+++ b/.claude/skills/dispatch-propagate/scripts/run-acceptance-tests.sh
@@ -25,9 +25,7 @@ cd "$REPO_ROOT/$APP_DIR"
 # (require the production build / prerendered output, absent from the dev server)
 # — everything else runs.
 if [ -n "$EXTERNAL_BASE_URL" ]; then
-  if [ -z "${PLAYWRIGHT_BROWSERS_PATH:-}" ]; then
-    npx playwright install --with-deps chromium
-  fi
+  playwright_install_with_deps
   BASE_URL="$EXTERNAL_BASE_URL" npx playwright test --config e2e/playwright.config.ts --grep-invert "@hosting|@testonly|@build"
   exit 0
 fi
@@ -93,12 +91,10 @@ env "${BUILD_ARGS[@]}" npm run build
 
 cd "$REPO_ROOT"
 
-# Install Playwright browsers (skip if nix provides them via PLAYWRIGHT_BROWSERS_PATH)
-if [ -z "${PLAYWRIGHT_BROWSERS_PATH:-}" ]; then
-  cd "$REPO_ROOT/$APP_DIR"
-  npx playwright install --with-deps chromium
-  cd "$REPO_ROOT"
-fi
+# Install Playwright browsers (bounded timeout+retry; skips when nix provides them).
+# Subshell keeps the surrounding code running from $REPO_ROOT; under set -e a
+# non-zero subshell aborts the parent.
+(cd "$REPO_ROOT/$APP_DIR" && playwright_install_with_deps)
 
 # Generate temporary firebase.json in repo root with relative path to dist.
 # Firebase emulator resolves public dir relative to the config file location.
--- a/.claude/skills/dispatch-propagate/scripts/run-smoke-tests.sh
+++ b/.claude/skills/dispatch-propagate/scripts/run-smoke-tests.sh
@@ -35,10 +35,8 @@ for i in $(seq 1 30); do
   sleep 2
 done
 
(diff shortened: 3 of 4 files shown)


natb1:
## QA Summary — #1899

**Result: PASS (3/3)**

No browser component detected (pure CI script/config change). All items verified via shell-command lane.

### Items

| # | Title | Result |
|---|-------|--------|
| 1 | Every CI job in pr-checks.yml has a bounded timeout-minutes (AC1) | ✅ PASS |
| 2 | Every playwright install site routes through the bounded helper (AC2) | ✅ PASS |
| 3 | Simulated install stall fails fast with retry (AC3) | ✅ PASS |

### Details

**Item 1 — AC1 (timeout-minutes present):**
Both jobs in `.github/workflows/pr-checks.yml` carry `timeout-minutes`:
- `acceptance`: `timeout-minutes: 30`
- `preview-and-smoke`: `timeout-minutes: 20`

Both values are well below the 6-hour default and within the 20–30 min target.

**Item 2 — AC2 (helper consolidation):**
Exactly one literal `npx playwright install --with-deps chromium` exists — inside `playwright_install_with_deps` in `lib.sh`. All three former inline call sites in consuming scripts now delegate to the helper:
- `run-smoke-tests.sh:39` — bare `playwright_install_with_deps`
- `run-acceptance-tests.sh:28` — bare call inside the external-URL branch
- `run-acceptance-tests.sh:97` — subshell `(cd "$REPO_ROOT/$APP_DIR" && playwright_install_with_deps)` to preserve cwd

The helper contains: `timeout --kill-after=30` wrapper, bounded retry loop, `PLAYWRIGHT_BROWSERS_PATH` skip-guard, and `return 1` (not `exit`) on exhaustion.

**Item 3 — AC3 (simulated stall fast-fail):**
With a `sleep 600` stub replacing `npx` (simulating a hung apt fetch), `PLAYWRIGHT_INSTALL_TIMEOUT=1`, and `PLAYWRIGHT_INSTALL_ATTEMPTS=2`:
- Returned non-zero in **12 seconds** (not hours)
- Logged retry (`attempt 2/2`) and exhaustion (`failed after 2 attempts`)
- Skip-guard (`PLAYWRIGHT_BROWSERS_PATH` set) returned 0 immediately

### Counts
- Items executed: 3 (all shell-command lane)
- PASS: 3 | FAIL: 0 | SKIP: 0
- Deferred to office-hours: none
- Browser walkthrough: not started (no frontend component)

natb1:
## Review — PR #1922 (issue #1899)

### Fixed

- **CR-1** `.claude/skills/dispatch-propagate/scripts/lib.sh:519-520` — `sleep 5` executed unconditionally after every failed attempt, including the final one. With 2 attempts, the loop burned 5 extra seconds on every exhausted-retry path before exiting. Fixed by incrementing `attempt` before the sleep guard, so the backoff only runs when a retry will follow. _(commit `ad64717b`)_

### Informational

- **CR-2** `.github/workflows/pr-checks.yml` — Six other workflow files (`unit-tests.yml`, `prod-deploy.yml`, `firestore-deploy.yml`, `functions-deploy.yml`, `storage-deploy.yml`, `budget-etl-release.yml`) still lack `timeout-minutes` and inherit GitHub's 6-hour cap. Pre-existing and intentionally out of this PR's scope per the issue body.
- **CR-3** `.claude/skills/dispatch-propagate/scripts/lib.sh:511-513` — Log asymmetry: attempt start message emitted for attempts 2+ but not attempt 1. Not a correctness issue; the failure message includes the attempt number.
- **RV-4** `.github/workflows/pr-checks.yml:14` — The 30-minute acceptance cap could be exceeded by an all-app run touching a shared utility package (~60-90 min for 6 apps). The cap is strictly better than the previous 6-hour implicit default; future work can tune once CI history data is available.
- **RV-5** `.claude/skills/dispatch-propagate/scripts/lib.sh:515` — `sudo apt-get` under `timeout --kill-after=30` may spawn a new process group, leaving an orphaned apt-get on the next attempt. Known `timeout` limitation; the job-level `timeout-minutes` backstop covers this regardless.
- **RT-1** `.claude/skills/dispatch-propagate/scripts/lib.sh:508-509` — `PLAYWRIGHT_INSTALL_TIMEOUT`/`PLAYWRIGHT_INSTALL_ATTEMPTS` have no numeric validation. Accessible only via workflow `env:` blocks (trusted authors) or a developer's shell; the job timeout caps total wall time; the helper fails closed. Noting for completeness.
- **SR-4** `.github/workflows/pr-checks.yml:14,40` — `timeout-minutes: 30` and `timeout-minutes: 20` are positive security hardening, preventing indefinite GHA minute consumption from a stalled apt-get.
- **SR-5** `.claude/skills/dispatch-propagate/scripts/lib.sh:515` — `npx playwright install --with-deps chromium` runs `apt-get` as root (pre-existing pattern). Consolidating into a helper does not expand the trust model.
- **CI-3** `.github/workflows/pr-checks.yml:33,56` — `github.head_ref` embedded in Vite build as `VITE_GITHUB_BRANCH`. XSS risk depends on how app code consumes it; shell injection via `env:` assignment is safe. Pre-existing, outside this diff's scope.

### False-positive (security)

- **IV-1** — `PLAYWRIGHT_INSTALL_TIMEOUT` in `timeout` arg: double-quoting prevents word-splitting; non-numeric value causes `timeout` to exit non-zero under `set -e`. No injection path. _(3 independent reviewers)_
- **IV-2** — `PLAYWRIGHT_INSTALL_ATTEMPTS` in `[ ]` test: quoted variable; bash emits `integer expression expected` and exits non-zero on non-integer input under `set -e`. No injection path.
- **RT-2** — `PLAYWRIGHT_BROWSERS_PATH` skip-guard is intentional (pre-existing design, identical to the inline guards it replaces; not set from CI). _(3 independent reviewers)_
- **SE-1** — No hardcoded secrets in the new helper. All vars are operational tuning parameters.
- **DE-1** — Diagnostic stderr output references only numeric counters (attempt index, timeout). No credentials or PII echoed.
- **CI-5** — Workflow permissions (`contents: read`, `pull-requests: write`) are minimally scoped.

### Deferred

- **RV-2** `.claude/skills/dispatch-propagate/scripts/lib.sh:494-524` — `playwright_install_with_deps()` has no unit tests, breaking the repo convention for `lib.sh` retry helpers (cf. `gh_retry` tests in `test-dispatch-scripts.sh`). Does not block merge; filed as #1926 (blocked by #1899).

### Out-of-scope (security, pre-existing)

- **IV-3** `.github/workflows/pr-checks.yml:61,75-76` — `github.event.pull_request.number` interpolated inline in `run:` blocks. GitHub-controlled integer; not attacker-controllable. Recommend passing via `env:` for hardening consistency. Pre-existing, not introduced by this diff.
- **IV-4** `.github/workflows/pr-checks.yml:65-67` — `GITHUB_OUTPUT` heredoc delimiter injection risk (`EOF` in script stdout could prematurely terminate value). Impact limited to truncating a preview URL. Pre-existing.
- **RT-3** `.github/workflows/pr-checks.yml:7-9` — `pull_request` trigger with `pull-requests: write`. Fork PRs receive read-only token from GitHub regardless of declared permissions. Pre-existing.
- **DE-2** `.github/workflows/pr-checks.yml:60-63` — Deploy script output teed to `/dev/stderr`, visible in CI logs. Pre-existing, not modified by this diff.

PR #1923, opened 2026-06-19
De-flake budget transactions blur-filter test (auto-waiting assertion)

natb1:
Closes #1919

## Problem

The `acceptance` CI check intermittently failed on the budget app's e2e test
**"typing a category in the filter and blurring filters table rows"** (mobile
project variant) in `budget/e2e/transactions.spec.ts`. The flake surfaced on a
PR that does not touch the budget app, while the same branch passed the identical
check on an earlier run — a non-deterministic timing failure, not a real
regression.

## Root cause

The test fills the category filter with "Food", blurs, polls until the visible
row count drops, then took a **one-shot** `count()` snapshot and iterated each
visible row reading `data-category`, asserting `/^Food/`. The blur handler toggles
each row's `style.display` per render pass. Between the poll settling and the
per-row attribute reads, a second filter re-render pass can still be in flight, so
a non-Food row may be transiently visible when its `data-category` is read —
especially on the slower mobile project. `count()` and `getAttribute()` are
one-shot reads with no auto-wait.

## Fix

Replace the one-shot snapshot + per-row loop with an auto-waiting negative-locator
assertion:

```ts
const visibleNonFoodRows = page.locator(
  '.txn-row:not([style*="display: none"]):not([data-category^="Food"])',
);
await expect(visibleNonFoodRows).toHaveCount(0);
await expect(visibleRows).not.toHaveCount(0);
```

`expect(locator).toHaveCount` re-queries on every retry, so a transient non-Food
row from an in-flight second render pass is retried away rather than read once and
asserted. The race is removed by construction, not merely made less likely. The
`[data-category^="Food"]` CSS prefix matches the same set as the old `/^Food/`
regex and is a superset of the filter's own match, so a settled DOM always yields
count 0 — the assertion cannot false-fail; it only retries until the render
settles. This is the same auto-waiting `toHaveCount` pattern the sibling
"clearing the filter input and blurring restores all rows" test already uses.

## Verification

- **By construction (primary):** the documented one-shot read between poll-settle
  and the per-row loop is the race; an auto-waiting locator assertion eliminates
  it. A flaky test passing CI once does not prove a fix — the by-construction
  argument does.
- **Local typecheck:** the spec still compiles (`tsc --noEmit`, clean). The change
  uses only already-imported Playwright APIs — no new imports.
- **Scope:** only the one test block changed; no other test references the removed
  `countAfter` local, and no `budget/src/**` source or CI config changed.

### DIFF
--- a/budget/e2e/transactions.spec.ts
+++ b/budget/e2e/transactions.spec.ts
@@ -250,12 +250,17 @@ test.describe("transactions", () => {
     // count() is a one-shot snapshot with no auto-wait; poll until the
     // blur-triggered filter re-render settles before reading the count.
     await expect.poll(async () => visibleRows.count()).toBeLessThan(countBefore);
-    const countAfter = await visibleRows.count();
-    expect(countAfter).toBeGreaterThan(0);
-    for (let i = 0; i < countAfter; i++) {
-      const category = await visibleRows.nth(i).getAttribute("data-category");
-      expect(category).toMatch(/^Food/);
-    }
+    // Auto-waiting assertion: no visible row's data-category falls outside the
+    // "Food" prefix. expect(locator).toHaveCount re-queries on each retry, so a
+    // transient non-Food row from an in-flight second render pass is retried
+    // away rather than read once and asserted (the prior one-shot loop raced
+    // the re-render, especially on mobile).
+    const visibleNonFoodRows = page.locator(
+      '.txn-row:not([style*="display: none"]):not([data-category^="Food"])',
+    );
+    await expect(visibleNonFoodRows).toHaveCount(0);
+    // At least one Food row stays visible after filtering.
+    await expect(visibleRows).not.toHaveCount(0);
   });
 
   test("clearing the filter input and blurring restores all rows", async ({ page }) => {


natb1:
# QA Summary — Issue #1919 / PR #1923

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Auto-waiting assertion construct replaces the one-shot per-row loop | shell-command | PASS |
| 2 | Removed `countAfter` local has no surviving reader | shell-command | PASS |
| 3 | Spec still type-checks | shell-command | PASS |
| 4 | No app source or build regression | shell-command | PASS |
| 5 | New assertion is semantically equivalent to the old loop and eliminates the race | (disposition) | PASS (already-satisfied) |

No browser walkthrough was performed — the only changed file is `budget/e2e/transactions.spec.ts` (a test file); no app source changed and no `needs-browser` items were in the plan.

## Bugs found

None.

## Disposition triage

Step 3.5 ran. `result.dispositions` is empty — every residue item was classified `already-satisfied`. See **Assessed by Opus** below.

- **New assertion is semantically equivalent to the old loop and eliminates the race** — The judgment resolves cleanly from the code itself, with no defect to fix. Semantic equivalence verified: the CSS attribute prefix operator `[data-category^="Food"]` is the exact selector analog of the anchored regex `/^Food/`; wrapping it in `:not(...)` and asserting `toHaveCount(0)` on the resulting `visibleNonFoodRows` locator is the universal-quantifier equivalent of the old loop's "every visible row's category starts with Food"; and the added `await expect(visibleRows).not.toHaveCount(0)` preserves the old `expect(countAfter).toBeGreaterThan(0)`. Race elimination is structural: `expect(locator).toHaveCount` auto-waits and re-queries the DOM on each retry, so a transient non-Food row from an in-flight second mobile render pass is retried away, whereas the prior one-shot `getAttribute` loop read once and asserted, racing the re-render. No code change and no human eye required.

natb1:
## Review: PR #1923 (issue #1919)

**Surface:** test-only diff (`budget/e2e/transactions.spec.ts`)

---

### Fixed

_None._

### Required (security)

_None._

_None._

### Informational

- **CR-001** `budget/e2e/transactions.spec.ts:256-268` — inline comments verbosity: the second comment "At least one Food row stays visible after filtering" duplicates the assertion intent. No change required; minor readability note.
- **CR-002** `budget/e2e/transactions.spec.ts:264-268` — assertion order (negative before positive): logically sound. `toHaveCount(0)` (no non-Food rows) followed by `not.toHaveCount(0)` (at least one Food row) covers the full original intent.
- **CR-003** `budget/e2e/transactions.spec.ts:261-263` — style selector brittle against inline style format variations: if the app's `style.display` toggling uses `display:none` without a space, the `[style*="display: none"]` selector would not match. Monitor in CI; if it flakes differently, switch to a `data-hidden` or `aria-hidden` attribute. No change required now.
- **CR-004** `budget/e2e/transactions.spec.ts:265-268` — potential micro-race between the two new `toHaveCount` assertions: both use auto-waiting `toHaveCount` so each retries independently. In practice the blur handler is synchronous per frame, so settled non-Food=0 implies settled Food>0 immediately. No change needed; noted for future reference.

### Dismissed

_None._

### False-positive (security)

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

---

No code changes applied. The fix is correct by construction.

PR #1924, opened 2026-06-19
dispatch-emit-outcome: reject explicit empty --terminated-reason on non-escalated dispositions

natb1:
Closes #1907

## Summary

`dispatch-emit-outcome` cross-validates `--terminated-reason` against
`--disposition`: a reason is **required** when `disposition == escalated` and
**forbidden** otherwise. The forbidden-path guard tested emptiness
(`[[ -n "$TERMINATED_REASON" ]]`), which can't tell "flag not passed" from "flag
passed as the empty string". So `--disposition completed --terminated-reason ''`
exited 0 and silently serialized `terminated_reason: null` — swallowing the
forbidden argument instead of rejecting it. That is exactly the
"defensive fallback that buries an error" `.claude/rules/code-style.md` warns
against; the greenfield-correct behavior is to reject the contract violation
with exit 2.

## Changes

- **Fix the guard** (`dispatch-emit-outcome`): track flag presence with a
  `TERMINATED_REASON_SET` boolean set in the parser, and gate the forbidden-path
  check on that flag rather than on emptiness. An explicit
  `--terminated-reason ''` on a non-escalated disposition is now rejected
  (exit 2) with the existing descriptive message. The escalated-required check
  keeps its `-z` test (an empty reason on `escalated` still exits 2), and the
  post-validation jq-builder branch is unchanged (after validation, only a
  valid escalated reason is ever non-empty). The header contract comment and an
  inline comment now document the set-flag treatment.

- **Add coverage** (`test-emit-outcome.sh`): a new standalone unit test
  following the `test-run-typecheck.sh` convention, encoding the four acceptance
  criteria. The script was previously untested. `run-unit-tests.sh`
  auto-discovers `test-*.sh` and auto-triggers the PR-scripts suite on changes
  under the scripts dir, so no runner wiring is needed.

## Verification

`test-emit-outcome.sh` passes 8/8:

- `--disposition completed --terminated-reason ''` → exit 2 (the fix)
- `--disposition escalated --terminated-reason ''` → exit 2 (preserved)
- `--disposition escalated --terminated-reason 'some reason'` → exit 0 with
  `terminated_reason == "some reason"` (no regression)
- `--disposition completed` (flag omitted) → exit 0 with
  `terminated_reason == null` (no regression)

## Caller compatibility

The only live callers are the `review-fix` and `qa-fix` skills/workflows. Every
non-escalated path omits `--terminated-reason` entirely; only escalated paths
pass a non-empty string. No caller passes `--terminated-reason ''` on a
non-escalated disposition, so the stricter behavior breaks nothing in the
autonomous chain.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-emit-outcome
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-emit-outcome
@@ -27,7 +27,10 @@
 # --disposition: one of the ALLOWED_DISPOSITIONS set
 #   (completed, completed_with_fixes, escalated).
 # --terminated-reason: required-non-null when disposition == escalated;
-#   must be absent (or not provided) for any other disposition.
+#   forbidden for any other disposition. An explicit empty
+#   --terminated-reason '' on a non-escalated disposition is rejected
+#   (exit 2), not silently treated as absent — presence is tracked by a
+#   set-flag, not by emptiness.
 # --pr, --base-sha: optional; serialize as JSON null when absent.
 #
 # DIVERGENCE FROM dispatch-mark-complete:
@@ -63,6 +66,7 @@ FOLLOWUPS_FILED=""
 SUBAGENTS_LAUNCHED=""
 DISPOSITION=""
 TERMINATED_REASON=""
+TERMINATED_REASON_SET=false
 
 usage() {
   cat >&2 <<'EOF'
@@ -127,7 +131,7 @@ while [[ $# -gt 0 ]]; do
       DISPOSITION="$2"; shift 2 ;;
     --terminated-reason)
       [[ $# -ge 2 ]] || { echo "dispatch-emit-outcome: --terminated-reason requires a value" >&2; exit 2; }
-      TERMINATED_REASON="$2"; shift 2 ;;
+      TERMINATED_REASON="$2"; TERMINATED_REASON_SET=true; shift 2 ;;
     *)
       echo "dispatch-emit-outcome: unexpected argument: $1" >&2
       usage
@@ -203,13 +207,16 @@ fi
 
 # ---------------------------------------------------------------------------
 # terminated_reason cross-validation
+# "Provided" is tracked by TERMINATED_REASON_SET (set in the parser), NOT by
+# emptiness, so an explicit --terminated-reason '' on a non-escalated
+# disposition is rejected rather than silently treated as absent.
 # ---------------------------------------------------------------------------
 if [[ "$DISPOSITION" == "escalated" && -z "$TERMINATED_REASON" ]]; then
   echo "dispatch-emit-outcome: --terminated-reason is required when --disposition is escalated" >&2
   exit 2
 fi
 
-if [[ "$DISPOSITION" != "escalated" && -n "$TERMINATED_REASON" ]]; then
+if [[ "$DISPOSITION" != "escalated" && "$TERMINATED_REASON_SET" == true ]]; then
   echo "dispatch-emit-outcome: --terminated-reason must not be provided when --disposition is $DISPOSITION (only valid for escalated)" >&2
   exit 2
 fi
--- a/.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Tests for dispatch-emit-outcome — the terminated_reason cross-validation contract (#1907).
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd -P)"
+source "$SCRIPT_DIR/test-helpers.sh"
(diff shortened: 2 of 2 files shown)


natb1:
## QA Pass — #1907 (`dispatch-emit-outcome` empty-reason rejection)

### Results: 5 PASS / 0 FAIL / 0 SKIP

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Non-escalated disposition rejects explicit empty `--terminated-reason` | shell-command | ✅ PASS |
| 2 | Escalated disposition with empty `--terminated-reason` still rejected | shell-command | ✅ PASS |
| 3 | Escalated disposition with valid reason still succeeds (no regression) | shell-command | ✅ PASS |
| 4 | Non-escalated disposition with flag omitted still succeeds | shell-command | ✅ PASS |
| 5 | Cross-validation block carries the explanatory comment | (deferred) | ✅ PASS (Opus: already satisfied) |
| 6 | Unit test suite passes all 8 cases | shell-command | ✅ PASS |

No bugs found. No browser walkthrough (pure shell-script change — no frontend component).

Item 5 was classified `needs-human-judgment` by the triage plan, but assessed `already-satisfied` by the disposition Workflow:

**Assessed by Opus (already satisfied)**
- **Item 5 — Cross-validation block carries the explanatory comment**: The cross-validation block already carries a substantive comment (lines 210–212 of `dispatch-emit-outcome`) explaining that `TERMINATED_REASON_SET` tracks flag presence rather than emptiness, and that an explicit `--terminated-reason ''` is rejected rather than silently treated as absent. The 4th acceptance criterion is provably met by reading the code — no code change and no human decision needed.

No `needs-main` items, no `needs-human` items. Escalation set is empty.

natb1:
## Review — `dispatch-emit-outcome` silently accepts `--terminated-reason ''` (#1907)

**Surface:** `code` | **CodeQL:** no open alerts | **Deps:** unchanged

---

### Fixed

- **RE-3** — `.claude/skills/dispatch-propagate/scripts/dispatch-emit-outcome:27–31`
  Header comment said "absent (or not provided)" but now rejects empty string. Updated to "forbidden" with a clear description of the set-flag enforcement. _(Already present in the PR diff; no additional commit required.)_

- **RE-4** — `.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh:1–10`
  Stale "Exit 0 = all passed; non-zero = one or more failures" comment was inconsistent with `set -euo pipefail`, which aborts on a failed assertion before `report_results` runs. Removed. _(Already present in the PR diff; no additional commit required.)_

---

### Informational

- **CR-1** — `.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh:1–103`
  `test-helpers.sh` was not extracted to its own commit; both files changed together. Not a defect — the helper is a test-infrastructure dependency of the new suite.

- **CR-5** — `.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh:5–11`
  `set -euo pipefail` + `set +e`/`set -e` toggle around the SUT call is a well-known pattern but slightly fragile (a subshell error before the `set -e` restore would be swallowed). Acceptable for a test harness.

- **CR-6** — `.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh:52–55`
  Cases 3 and 4 extract JSON via `sed` + `jq` rather than a direct herestring. Functions correctly; a herestring would be marginally cleaner.

- **RE-2** — `.claude/skills/dispatch-propagate/scripts/dispatch-emit-outcome:71–100`
  `usage()` output does not list `--terminated-reason`. Minor documentation gap; the header comment above covers the contract.

---

### Dismissed

- **CR-2** — `.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh:20–23`
  _"COMMON array parsed as separate args with word splitting."_ Incorrect finding: the array is expanded as `"${COMMON[@]}"` (correctly quoted) throughout.

- **CR-3** — `.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh:31–33`
  _"`RC`/`OUT` variables declared outside `run_sut` without `local`."_ Correct by design — `RC` and `OUT` are read by the test cases immediately after `run_sut` returns; they are intentionally script-global.

- **CR-4** — `.claude/skills/dispatch-propagate/scripts/test-emit-outcome.sh:52–55`
  _"printf '%s\\n' piped to sed instead of herestring."_ The `.claude/rules/shell-json.md` rule applies to `gh --json` output piped through `echo`; plain text through `printf '%s\n'` is correct here.

- **RE-1** — `.claude/skills/dispatch-propagate/scripts/dispatch-emit-outcome:131–133`
  _"`TERMINATED_REASON_SET=false` could be true after shift if `$2` is empty."_ Incorrect — the `[[ $# -ge 2 ]]` guard fires before the assignment, so an absent second argument exits 2 before `TERMINATED_REASON_SET` is set. The `$2 == ''` case is exactly the path the fix is targeting; setting `TERMINATED_REASON_SET=true` is correct.

---

_No security findings. No follow-up issues filed. No deviations._

PR #1712, opened 2026-06-15
dispatch: validate owner/repo format in dispatch-read-plan and dispatch-write-plan

natb1:
Closes #1233

## Summary

`dispatch-read-plan` and `dispatch-write-plan` derive `owner/repo` (exported as
`GH_REPO`) from `remote.origin.url`, then guarded it with a weak
`[[ -n "$repo" && "$repo" == */* ]]` check. That guard catches empty / no-slash
values, but two malformed remotes slip through and surface only as an opaque
downstream `gh` error:

- **SSH-with-port** (`ssh://git@github.com:22/owner/repo.git`): the
  `${repo#*github.com[:/]}` strip matches the `:` before the port, leaving
  `22/owner/repo` — two slashes, so it passes `*/*`. `gh` then fails with a
  confusing "Repository not found".
- **Non-GitHub remote** (`https://gitlab.com/owner/repo.git`): no `github.com`
  match, so `repo` stays the full URL — has slashes, passes `*/*`, and `gh`
  sends a malformed API path.

This PR replaces the combined guard in both scripts with a non-empty resolve
check plus a strict format guard, so a malformed `owner/repo` is rejected with a
clear, URL-naming diagnostic **before** any `gh` call:

```bash
[[ -n "$repo" ]] || { echo "<script>: could not resolve owner/repo from remote.origin.url ('$url')" >&2; exit 1; }
[[ "$repo" =~ ^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$ ]] || { echo "<script>: unexpected owner/repo format from remote.origin.url ('$url'): $repo" >&2; exit 1; }
```

The regex strictly subsumes the old `*/*` slash check, so `*/*` is dropped
rather than kept. The live remote
(`https://github.com/natb1/commons.systems.git` → `natb1/commons.systems`) and
the `git@github.com:owner/repo.git` SSH form both still pass — the `.` is in the
character class, so `commons.systems` is not falsely rejected.

This is a companion hardening to PR #1232 (Unit 3 — repo resolution fix), whose
review pass added the empty/missing-URL guard.

## Changes

- `dispatch-read-plan`, `dispatch-write-plan`: split the combined guard into a
  non-empty check and a strict `^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$` format guard
  with a stable `unexpected owner/repo format …` diagnostic.
- `test-dispatch-scripts.sh`: two new cases in the existing
[...]

PR #1946, opened 2026-06-19
dispatch: extract shared GH_REPO derivation into gh_repo_from_remote helper

natb1:
Closes #1771

## What

`dispatch-read-plan` and `dispatch-write-plan` each carried a verbatim copy of
the same `GH_REPO` derivation-and-format-guard block: derive `owner/repo` from
`remote.origin.url`, guard it non-empty, guard it against
`^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$`, then `export GH_REPO`. The two copies
differed only in the per-script error prefix. This duplication was introduced by
PR #1712 (which added the format guard to both scripts) and flagged there as out
of scope.

This PR extracts the block into a single `gh_repo_from_remote <url> <caller>`
helper in the scripts' shared `lib.sh`, so the derivation logic lives in one
place.

## How

- **`lib.sh`** — add `gh_repo_from_remote(url, caller)`: pure string logic that
  prints `owner/repo` to stdout on success, and on an empty / non-GitHub /
  malformed result prints a `<caller>`-prefixed message to stderr and `return 1`
  (a sourced library function returns; it does not `exit`). The `<caller>`
  argument preserves the per-script error prefix so logs still identify the
  originating script. The URL-handling explanation that was duplicated inline
  now lives single-site in the helper's doc comment.
- **`dispatch-read-plan` / `dispatch-write-plan`** — source `lib.sh` after
  `set -euo pipefail`, then replace the inline block with
  `GH_REPO="$(gh_repo_from_remote "$url" <caller>)" || exit 1; export GH_REPO`.
  The `|| exit 1` turns the helper's `return 1` into the same script-level exit-1
  the inline guards produced.
- **`test-dispatch-scripts.sh`** — add a focused `gh_repo_from_remote` block
  asserting the helper directly: three success cases (HTTPS with/without `.git`,
  SSH scp-style) and three failure cases (empty URL, non-GitHub remote, malformed
  owner/repo), each checking the exit code and the caller-prefixed stderr
  diagnostic.

Behavior for valid HTTPS/SSH remotes and for the three failure cases is
unchanged.

## Verification

  3069/3069 passed, including the new `gh_repo_from_remote` assertions and the
  existing end-to-end `dispatch-write-plan / dispatch-read-plan` block (the
  existing-behavior-parity check for valid remotes).
- End-to-end smoke: `dispatch-read-plan 1771` still resolves `GH_REPO` through
  the new helper and fetches the plan comment.
- Failure-path spot check: `gh_repo_from_remote "" probe` and
  `gh_repo_from_remote "https://gitlab.com/a/b" probe` both print a
  `probe:`-prefixed diagnostic and return 1.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-read-plan
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-read-plan
@@ -20,6 +20,7 @@
 # so gh addresses the right repository from any caller cwd. A missing arg or a
 # non-numeric issue-num is a clear error, not a fallback (exit 2).
 set -euo pipefail
+source "$(cd "$(dirname "$0")" && pwd)/lib.sh"
 
 N="${1:-}"
 if [[ $# -ne 1 || -z "$N" ]]; then
@@ -37,15 +38,8 @@ common_dir="$(git rev-parse --path-format=absolute --git-common-dir 2>/dev/null)
   echo "dispatch-read-plan: not inside a git repository" >&2; exit 1; }
 url="$(git --git-dir="$common_dir" config --get remote.origin.url 2>/dev/null)" || {
   echo "dispatch-read-plan: no remote.origin.url in git config at $common_dir" >&2; exit 1; }
-# Resolve owner/repo from the remote so gh addresses the repo independent of cwd
-# (dirname(common_dir) is not a working tree in the bare-repo + worktrees layout).
-# Handles https://github.com/<owner>/<repo>(.git) and git@github.com:<owner>/<repo>(.git).
-repo="${url%.git}"; repo="${repo#*github.com[:/]}"
-[[ -n "$repo" ]] || {
-  echo "dispatch-read-plan: could not resolve owner/repo from remote.origin.url ('$url')" >&2; exit 1; }
-[[ "$repo" =~ ^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$ ]] || {
-  echo "dispatch-read-plan: unexpected owner/repo format from remote.origin.url ('$url'): $repo" >&2; exit 1; }
-export GH_REPO="$repo"
+GH_REPO="$(gh_repo_from_remote "$url" dispatch-read-plan)" || exit 1
+export GH_REPO
 
 MARKER='<!-- dispatch:plan -->'
 PLAN_AUTHOR_ID="${DISPATCH_PLAN_AUTHOR_ID:-$(gh api user --jq '.id')}"
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-write-plan
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-write-plan
@@ -19,6 +19,7 @@
 # so gh addresses the right repository from any caller cwd. A missing arg, a
 # non-numeric issue-num, or empty STDIN is a clear error, not a fallback (exit 2).
 set -euo pipefail
+source "$(cd "$(dirname "$0")" && pwd)/lib.sh"
 
 N="${1:-}"
 if [[ $# -ne 1 || -z "$N" ]]; then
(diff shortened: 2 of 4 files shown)


natb1:
## QA Pass — Issue #1771 / PR #1946

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Both scripts source lib.sh cleanly and the helper is in scope | shell-command | PASS |
| 2 | dispatch-read-plan reaches a real GH_REPO derivation on the golden path | shell-command | PASS |
| 3 | dispatch-read-plan passes its OWN caller name into the helper's failure message | shell-command | PASS |
| 4 | dispatch-write-plan reaches a real GH_REPO derivation on the golden path | shell-command | PASS |
| 5 | dispatch-write-plan passes its OWN caller name into the helper's failure message | shell-command | PASS |
| 6 | Full dispatch-scripts unit suite passes with the new helper and its test block | shell-command | PASS |
| 7 | Refactor is a faithful, behavior-preserving extraction | (assessed by Opus) | PASS |

**6 PASS / 0 FAIL / 0 SKIP** (shell-command lane) · No browser walkthrough (pure-script PR, no browser component detected)

### Deferred to office-hours

None. All items resolved.

### Bugs found

None.

### Disposition triage

- **Item 7 — Refactor is a faithful, behavior-preserving extraction (no drift from prior inline logic)**
  The Opus reviewer confirmed this criterion is provably met by the diff alone, not a human-only judgment call. `gh_repo_from_remote` reproduces the prior inline logic verbatim: identical URL transformations (`${url%.git}`, `${repo#*github.com[:/]}`), identical empty-check, identical format regex (`^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$`), and the same two error messages now parameterized via `$caller` — exactly the "caller name is the only per-script difference" the criterion specifies. Exit-code semantics preserved: helper `return 1` + `|| exit 1` at the call site reproduces the prior inline `exit 1`. The helper's new `local repo` introduces no `set -u` drift. The full test suite passes (3087/3087), including 10 dedicated `gh_repo_from_remote` assertions. Dropped as PASS.

natb1:
## Review findings

### Fixed

- **RF-04** `.claude/skills/dispatch-propagate/scripts/lib.sh:42-48` — Non-GitHub remote misclassified as malformed rather than "not GitHub": `gh_repo_from_remote` now detects when the `github.com[:/]` strip is a no-op (i.e. the URL is not a GitHub URL at all) and prints a distinct "remote is not a GitHub repository" error, rather than falling through to the format guard and printing the misleading "unexpected owner/repo format". The unit test and end-to-end write-plan test updated to assert the new diagnostic. Fix commit: aacc6ded.

### Informational

- **RF-01** `.claude/skills/dispatch-propagate/scripts/lib.sh:34-49` — `gh_repo_from_remote` does not document the SSH-with-port edge case (e.g. `ssh://git@github.com:22/…` yields `22/owner/repo` after stripping, which then fails the format guard with a partially-stripped path). No change needed — this is a working, intentional failure with a clear error message; the doc comment accurately describes the two canonical URL forms the helper handles.

- **RF-05** N/A — CodeQL scan: no open alerts on PR head.

- **RF-06** N/A — npm dependency audit: not applicable (no `package.json` / `package-lock.json` changes).

### Dismissed

- **RF-02** `.claude/skills/dispatch-propagate/scripts/lib.sh:34-41` — Redundant comments: the consolidated doc comment in the helper is accurate and not redundant — the per-script inline comments it replaces are gone. Nothing to change.

- **RF-03** `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:28807-28837` — Arithmetic-expansion assert_eq bypass: this finding was hallucinated; `TOTAL`/`PASS`/`FAIL` increments are standalone statements, not arguments to `assert_eq`. No code matches the described pattern.

PR #1947, opened 2026-06-19
budget: stand up React toolchain + app shell + home + content-grid

natb1:
Closes #1874

## What

Foundation PR of epic #1873: migrate the `budget` app from vanilla-TS string-render
route handlers to **React on `@commons-systems/ds`**. This PR stands up the
React/TSX toolchain, ports the app shell (nav, header, footer, content-grid) and the
`/transactions` view to React composing ds, replaces theme.css color/radius literals
with ds tokens, and keeps the not-yet-migrated routes (`/`, `/accounts`,
`/accounts/reconcile`, `/rules`) working through a `<LegacyRoute>` bridge that gives
the sibling migrations (#1875, #1876, #1877) a clean seam.

Outcome: a hybrid app — React shell + React `/transactions` + a `<LegacyRoute>`
bridge — that is behavior-identical to production budget, with SSG preserved for
first paint. The local-first data/crypto/file-sync pipeline is moved **verbatim**;
only its DOM glue became React state.

## Units

1. **React/TSX toolchain** — `react`/`react-dom`/`@commons-systems/ds` deps, `react-jsx`
   tsconfig, ds-style esbuild JSX in vite, `react`/`react-dom`/`react/jsx-runtime`
   monorepo dedupe.
2. **React shell + LegacyRoute bridge** — `main.tsx` `createRoot`, `App.tsx` composing
   ds `<Nav>` + header + footer + content-grid + a minimal client router, `AuthControls`
   in the Nav `end` slot, a `Hero` island, and `LegacyRoute` (string-render + hydrate
   per route). The data/crypto/FSA pipeline moved verbatim into `use-app-state.ts`;
   `main.test.ts` ported to `use-app-state.test.tsx` preserving every FSA /
   plaintext-downgrade / external-reload / startup-branch assertion.
3. **React `/transactions`** — `Transactions.tsx` (markup ported to JSX) and
   `CategorySankey.tsx`, an imperative chart island whose core reads serialized data
   from a **prop**, not a DOM JSON blob.
4. **Entanglement refactors** — blur-save, infinite scroll, and chart-update converted
   from imperative DOM hydration to React state/effects (`use-transaction-table.ts`),
   deleting `home-hydrate.ts`. A `TableErrorBoundary` surfaces synchronous
   data-integrity throws as a data-error region instead of crashing the tree. The
   chart re-renders from appended-transaction props instead of a custom DOM event.
5. **Token migration** — the four ds token layers imported; the local `theme.css`
   `:root` color overrides removed (ds supplies identical values); radius/spacing/border
   literals in the shell + transactions/chart region retagged to ds tokens (all swaps
   value-identical, so rendering is unchanged). Sibling sections left untouched.
6. **Prerender rewrite (SSG)** — a pure `AppShell` (composed by both `App.tsx` and the
   prerender) is `renderToStaticMarkup`'d with the synchronous `renderBudgetsContent`
   baked into `<main id="app">` and injected into `<div id="root">`.
7. **Tests** — parity tests ported to `@testing-library/react` before the entanglement
   refactor (the bug-fix lock), plus `LegacyRoute` render-once / hydrate tests and a
   prerender SSG-shell test; the stale `<app-nav>` smoke assertion updated to the ds
   Nav output.

## Parity preserved

The bug-fixes from #1266 (orphan normalized-member suppression), #1267 (chart
stale-listener cleanup, now effect teardown), and #578 (chart serialization degrades
unknown budget ids) are preserved and re-locked by ported tests. The reimbursement
cell stays `type="number"` to match the legacy markup exactly.

## Deferred greenfield decisions (justified in-line)

- **`createRoot`, not `hydrateRoot`** — the baked route is always `/` (budgets) but
  served for every path, real route bodies render async while `renderToStaticMarkup`
  is sync, and the chart is a client-only island — all guarantee a hydration mismatch.
  The baked HTML is a static first-paint snapshot the client replaces on mount.
  `hydrateRoot` becomes correct only once every route is a React, SSR-able tree.
- **`theme.css` stays a partial monolith** — only the shell + transactions/chart region
  is migrated; the optional split into per-feature CSS files is deferred so this PR does
  not reorder CSS the sibling PRs own.

## Reviewer notes

- `budget/scripts/prerender-jsx-preload.mjs` exposes `React` globally for the prerender
  process only: this `tsx` version compiles JSX with the classic runtime rather than the
  tsconfig `react-jsx` automatic runtime, so the shared JSX modules need `React` in scope.
  Contained to the build command.
- The eslint cleanup removed `eslint-disable` directives for `react-hooks`/`jsx-a11y`
  rules this repo's shared config does not register (they were reported as
  "rule not found" errors). Registering those plugins for React packages is a possible
  follow-up, not done here.

## Verification

`npm --workspace budget test` (935 passed / 7 skipped), `npm --workspace budget run lint`
(clean), and `npm --workspace budget run build` (tsc + vite + prerender SSG all pass)
are green. Screenshot parity against deployed production budget is the QA-phase
acceptance method.


natb1:
# Fix-checks summary — PR #1947

## Iteration 1

**Failed checks:** `acceptance` (68 budget e2e failures across 10 specs; all other
checks green).

**Outcome:** fixed

**Reproduced:** yes — diagnosed from the CI failure logs + a full source trace.
The e2e/acceptance suite cannot run locally (Playwright browser-version mismatch:
Nix chromium 1217 vs project 1208), so CI is the verification oracle; the root
causes were confirmed by reading the changed React-migration source against the
e2e contract.

**Reproduce command:** acceptance is CI-only locally; per-failure evidence pulled
via `gh run view --job 82301332389 --log-failed` and source inspection of
`budget/src/{pages/Transactions.tsx,AppShell.tsx,LegacyRoute.tsx,use-app-state.ts,use-router.ts,pages/accounts-reconcile-hydrate.ts}`.
Unit suite + lint + build reproduced clean locally
(`run-unit-tests.sh budget` → 942 pass, `run-lint.sh budget`, `npm run build --prefix budget`).

**Failure excerpt:**
```
Failed acceptance tests: budget   (68 failed / 383 passed)
navigation.spec.ts:27  expect(locator('app-nav')).toBeVisible — element(s) not found
transactions.spec.ts:6 expect(locator('main > h2')).toHaveText('Transactions') — element(s) not found
responsive.spec.ts:6   expect(locator('main')).toBeVisible — element(s) not found
accounts-reconcile.spec.ts:82  locator('#reconcile-past .reconcile-past-event') toHaveCount 1 — received 0
```

**Why not caught:** the unit tests run under jsdom and query route components in
isolation (e.g. `Transactions` `h2` directly), so they never asserted the
production DOM contract the e2e/acceptance suite encodes. Three real regressions
slipped through because they only manifest in the built + served app:
1. **Missing `<main>`** — the React `/transactions` page (`Transactions.tsx`)
   returned a bare Fragment; `LegacyRoute` still wraps the other routes in
   `<main id="app">`, so `main` / `main > h2` was absent only on /transactions.
2. **`<app-nav>` → ds `<Nav>`** — the migration legitimately replaced the custom
   `<app-nav>` element with ds `<Nav>` (`<nav class="cs-nav">`), but the e2e specs
   still targeted `app-nav`. Selector-only update (links verified working via
   `NAV_LINKS` + use-router's `a[href^="/"]` click interceptor).
3. **Reconcile re-render** — `accounts-reconcile-hydrate.ts` `triggerReload()`
   dispatches a same-path `popstate` to force a route re-render; the new
   `useRouter` no-ops a same-path `popstate`, so `LegacyRoute` never re-keyed and
   the finalized reconcile event never re-rendered.

**Fix:** commit `a6e71d0d` (merged via `3747bced`, pushed). Five edits:
- `budget/src/pages/Transactions.tsx` — wrap the page in `<main id="app">`.
- `budget/src/use-app-state.ts` — add a `popstate` listener in the mount effect
  that bumps `navEpoch` (re-keys `LegacyRoute`), with matching cleanup; mirrors the
  legacy popstate-driven route re-render the reconcile flow relies on.
- `budget/e2e/navigation.spec.ts`, `prerender.spec.ts`, `rules.spec.ts` — swap the
  `app-nav` selectors to the ds Nav (`nav.cs-nav` / `class="cs-nav"`).

Verified locally: budget unit suite (942 pass), lint clean, build (tsc + vite +
prerender SSG) green. Acceptance re-runs in CI.

**Deferred to the next CI signal (not addressed this round):** a separate
runtime-only cluster of `count > 0` assertions on chart-derived / filtered DOM —
the transactions sankey-filter tests (`.txn-row:not([style*="display: none"])`),
the budgets pie/trend/variance chart counts, and home-scroll. These share a
`first()-visible-then-count-0` signature and stem from the chart island
imperatively styling React-owned rows; they need runtime DOM (unavailable locally,
zero CI artifacts uploaded) to diagnose without guessing. Fixing the three
structural regressions above clears the bulk of the noise so the next acceptance
run isolates this cluster cleanly.

natb1:
# Fix-checks summary — PR #1947

## Iteration 1

**Failed checks:** `acceptance` (68 budget e2e failures across 10 specs; all other
checks green).

**Outcome:** fixed

**Reproduced:** yes — diagnosed from the CI failure logs + a full source trace.
The e2e/acceptance suite cannot run locally (Playwright browser-version mismatch:
Nix chromium 1217 vs project 1208), so CI is the verification oracle; the root
causes were confirmed by reading the changed React-migration source against the
e2e contract.

**Reproduce command:** acceptance is CI-only locally; per-failure evidence pulled
via `gh run view --job 82301332389 --log-failed` and source inspection of
`budget/src/{pages/Transactions.tsx,AppShell.tsx,LegacyRoute.tsx,use-app-state.ts,use-router.ts,pages/accounts-reconcile-hydrate.ts}`.
Unit suite + lint + build reproduced clean locally
(`run-unit-tests.sh budget` → 942 pass, `run-lint.sh budget`, `npm run build --prefix budget`).

**Failure excerpt:**
```
Failed acceptance tests: budget   (68 failed / 383 passed)
navigation.spec.ts:27  expect(locator('app-nav')).toBeVisible — element(s) not found
transactions.spec.ts:6 expect(locator('main > h2')).toHaveText('Transactions') — element(s) not found
responsive.spec.ts:6   expect(locator('main')).toBeVisible — element(s) not found
accounts-reconcile.spec.ts:82  locator('#reconcile-past .reconcile-past-event') toHaveCount 1 — received 0
```

**Why not caught:** the unit tests run under jsdom and query route components in
isolation (e.g. `Transactions` `h2` directly), so they never asserted the
production DOM contract the e2e/acceptance suite encodes. Three real regressions
slipped through because they only manifest in the built + served app:
1. **Missing `<main>`** — the React `/transactions` page (`Transactions.tsx`)
   returned a bare Fragment; `LegacyRoute` still wraps the other routes in
   `<main id="app">`, so `main` / `main > h2` was absent only on /transactions.
2. **`<app-nav>` → ds `<Nav>`** — the migration legitimately replaced the custom
   `<app-nav>` element with ds `<Nav>` (`<nav class="cs-nav">`), but the e2e specs
   still targeted `app-nav`. Selector-only update (links verified working via
   `NAV_LINKS` + use-router's `a[href^="/"]` click interceptor).
3. **Reconcile re-render** — `accounts-reconcile-hydrate.ts` `triggerReload()`
   dispatches a same-path `popstate` to force a route re-render; the new
   `useRouter` no-ops a same-path `popstate`, so `LegacyRoute` never re-keyed and
   the finalized reconcile event never re-rendered.

**Fix:** commit `a6e71d0d` (merged via `3747bced`, pushed). Five edits:
- `budget/src/pages/Transactions.tsx` — wrap the page in `<main id="app">`.
- `budget/src/use-app-state.ts` — add a `popstate` listener in the mount effect
  that bumps `navEpoch` (re-keys `LegacyRoute`), with matching cleanup; mirrors the
  legacy popstate-driven route re-render the reconcile flow relies on.
- `budget/e2e/navigation.spec.ts`, `prerender.spec.ts`, `rules.spec.ts` — swap the
  `app-nav` selectors to the ds Nav (`nav.cs-nav` / `class="cs-nav"`).

Verified locally: budget unit suite (942 pass), lint clean, build (tsc + vite +
prerender SSG) green. Acceptance re-runs in CI.

**Deferred to the next CI signal (not addressed this round):** a separate
runtime-only cluster of `count > 0` assertions on chart-derived / filtered DOM —
the transactions sankey-filter tests (`.txn-row:not([style*="display: none"])`),
the budgets pie/trend/variance chart counts, and home-scroll. These share a
`first()-visible-then-count-0` signature and stem from the chart island
imperatively styling React-owned rows; they need runtime DOM (unavailable locally,
zero CI artifacts uploaded) to diagnose without guessing. Fixing the three
structural regressions above clears the bulk of the noise so the next acceptance
run isolates this cluster cleanly.

## Iteration 2

**Failed checks:** `acceptance` (13 unique budget e2e tests across mobile/tablet/desktop;
all other checks green). The chart/filter cluster deferred from Iteration 1.

**Outcome:** fixed

**Reproduced:** yes — diagnosed from CI failure logs + a full source trace. The
e2e/acceptance suite is CI-only locally (the QA Vite dev server floods esbuild
target-transform errors and is not faithful to the built+prerendered app CI tests;
the production build path needs the Firebase emulator + the pinned Playwright
browser). Root cause confirmed by reading the React-keying remount paths against
the e2e contracts; unit suite + lint + build reproduced clean locally.

**Reproduce command:** acceptance is CI-only; per-failure evidence pulled via
`gh run view --job 82309086235 --log-failed`. Local verification:
`run-unit-tests.sh --app budget` (942 pass), `run-lint.sh --app budget` (clean),
`npm run build --prefix budget` (tsc + vite + prerender SSG green).

**Failure excerpt:**
```
accounts-charts.spec.ts:19  legend items count toBeGreaterThan(0) — got 0 (first() WAS visible)
budgets.spec.ts:48          .chart-scroll-wrapper svg rect count toBeGreaterThan(0) — got 0
budgets.spec.ts:122/135     variance details: details not open / svg not found after expand
budgets-trend-charts:33     bar count toBeGreaterThan(0) — got 0
transactions.spec.ts:89     #category-sankey svg .sankey-node count > 0 — got 0 (first() visible)
transactions.spec.ts:110/187/346  countBefore (visible .txn-row) toBeGreaterThan(0) — got 0
home-scroll.spec.ts:287     category/budget filter applies to scroll-loaded rows — poll never true
```

**Why not caught:** unit tests run under jsdom/happy-dom and never exercise the
mount→remount timing or the scroll-append→remount path the e2e suite encodes. Two
React-keying remount regressions:
1. **Startup double-mount race** — `App.tsx` rendered the route body eagerly at
   `navEpoch=0` (mount 1); `useAppState.initialize()` always ends with a
   `transition()` that bumps `navEpoch` 0→1, and the body is keyed
   `${path}:${navEpoch}`, so mount 1 was unmounted and replaced — wiping the
   just-hydrated chart SVG and table rows. Tests' `toBeVisible` resolved on mount 1,
   then the one-shot `count()`/`countBefore` landed in the remount gap → 0. (Explains
   the "first() visible then count() 0" signature across both legacy and React charts,
   plus countBefore=0.)
2. **Sankey filter reset on scroll-append remount** — `Transactions.tsx` rendered
   `<CategorySankey key={appended.length}>`, so a scroll-append remounted the whole
   island including its JSX control inputs; `buildCategorySankey` re-initialized its
   filter state to hardcoded defaults, un-applying the active filter to newly loaded
   rows (home-scroll category/budget polled assertions never settled).

**Fix:** commit `1ca9a0ef` (merged via `8ef800c3`, pushed). Two fixes, no
transition/navEpoch semantics changed (preserves Iteration-1's reconcile re-key):
- **A (gate first render):** `use-app-state.ts` adds an `initialized` flag flipped in
  `initialize().finally(...)`; `App.tsx` holds the route body until `app.initialized`
  (shell/nav/header/footer/hero still render immediately) → the body mounts exactly
  once at `navEpoch=1`, matching legacy's single post-initialize render.
- **B (rebuild, don't remount):** `Transactions.tsx` drops `key={appended.length}`;
  `CategorySankey.tsx` effect deps become `[chartData, categoryOptions, budgetOptions]`
  (rebuild on data change; teardown preserves #1267); `home-chart.ts buildCategorySankey`
  seeds its filter/mode/checkbox/weeks state FROM the persisted control elements
  (first-mount defaults are value-identical to the old hardcoded defaults), so a
  rebuild honors the active filter.

Verified locally: budget unit suite 942 pass, lint clean, build (tsc + vite +
prerender SSG) green. Acceptance re-runs in CI.

natb1:
## QA — #1874 (budget: stand up React toolchain + app shell + home + content-grid)

**PR #1947** | Branch `1874-budget-stand-up-react-toolc`

---

### Results

| Shell-command | 4 | 4 | 0 | 0 |
| Browser (blocked — see below) | 3 | — | 3† | 0 |
| Deferred to office-hours | 1 | — | — | — |

†Browser items could not run — the Vite dev server crashes during `optimizeDeps` prebundling (esbuild 0.28.1 destructuring transform issue). **This is a pre-existing infrastructure problem:** identical errors appear on `origin/main` — not introduced by this PR. Structural equivalents are fully covered by the passing vitest suite (see Assessed by Opus below).

---

### Shell-Command Lane — 4 / 4 PASS

1. **Test suite** — PASS (942 tests, 0 failures)
2. **Lint / TypeScript type-check** — PASS (clean)
3. **Production build** (`tsc + vite build + tsx scripts/prerender.tsx`) — PASS
4. **React file existence** (11 React source files) — PASS

---

### Deferred to Office-Hours

- **Item 8 — Screenshot parity (home page layout vs pre-migration):** Aesthetic pixel-level comparison vs production screenshots. Requires a human eye; cannot be machine-verified.

---

### Disposition Triage (Opus classifier — 4 items)

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set. No `needs-main` items were found this pass.

**Item 8 — Screenshot parity:** class `needs-human` · aesthetic · verify `n/a`
> Pure pixel-level cosmetic screenshot-parity comparison (layout, spacing, font vs pre-migration production). `browser_available` is false and `screenshot_path` is empty — cannot be resolved from the DOM or a screenshot. Requires a human eye.

Non-fixing pass: item 8 escalates to office-hours for the user-input walkthrough.

---

### Assessed by Opus — Already Satisfied (dropped as PASS)

**Item 5 — Header `<h1>` renders on home page** (kind: fail → already-satisfied)
> Criterion (`<h1>Budget</h1>` renders) is provably met by passing vitest: `prerender.test.ts` "contains the `<h1>Budget</h1>` heading" asserts the exact heading via `renderToStaticMarkup` of `AppShell`. The QA-server esbuild crash is pre-existing and identical on `origin/main`, not a defect of this PR. No code change and no human needed.

**Item 6 — Nav links render as anchor elements** (kind: fail → already-satisfied)
> Criterion (nav renders four real `<a>` hrefs: `/`, `/transactions`, `/accounts`, `/rules`) is provably met by passing vitest: `prerender.test.ts` "renders nav links as real anchor elements" asserts all four `<a href>` matches plus `aria-current=page` on `/`. The ds `<Nav>` emits real anchors in SSG. No code defect to fix.

**Item 7 — App shell walkthrough** (kind: fail → already-satisfied)
> Broader walkthrough criterion (nav works, budget list loads from seed, `/transactions` loads with rows) is backed by the full budget vitest suite — 54 files / 942 tests pass: prerender (shell + seed budgets-table + seed-data-notice), home-scroll.smoke (scroll/pagination), use-app-state (app state machine including transactions), and CategorySankey (chart island). Browser blockage is the same pre-existing esbuild crash present on `origin/main`. No code defect; criterion met by green tests.

natb1:
## Review — PR #1947 (`1874-budget-stand-up-react-toolc`)

21 findings after dedup (4 fixed, 12 informational, 2 dismissed, 3 false-positive). No deferred or out-of-scope filings.

---

### Fixed

- **code-review-0** `budget/src/pages/home-chart.ts:406-465` — `buildCategorySankey` attached input/change/focus/blur listeners to persistent control elements on every chart re-run without removing them in its teardown, causing listener stacking. Fix: switched to a shared `AbortController` signal passed to all `addEventListener` calls; `signal.abort()` in the returned teardown removes them all. (commit `0fd0ab58`)

- **code-review-1** `budget/src/use-app-state.ts:488` — `setWriteBackStatusListener` was called on mount to register a callback but the effect cleanup never called `setWriteBackStatusListener(null)`, leaving the listener registered after unmount. Fix: added `setWriteBackStatusListener(null)` to the effect cleanup alongside the existing `popstate` `removeEventListener` and `AbortController.abort()` calls. (commit `0fd0ab58`)

- **review-0** `budget/src/AppShell.tsx:48` — Both external footer `target="_blank"` links used `rel="noopener"` but omitted `rel="noreferrer"`. `noopener` prevents `window.opener` access; `noreferrer` additionally suppresses the `Referer` header, preventing the destination from learning the origin URL. Fix: updated both links to `rel="noopener noreferrer"`. (commit `0fd0ab58`)

- **review-1** `budget/src/pages/Transactions.tsx:1836` — `computeAllBudgetBalances(transactions, budgets, budgetPeriods)` was called on every render of `TransactionTable`, including on `clearedBalances` state updates that don't change its inputs, causing O(n×m) budget-balance scans unnecessarily. Fix: wrapped the call in `useMemo` keyed on `[transactions, budgets, budgetPeriods]`, placed above the early return to satisfy the rules of hooks. (commit `0fd0ab58`)

---

### Informational

- **code-review-2** `budget/src/use-app-state.ts:549` — `AppApi.onUploadLabelKeydown` is typed `(key: string) => void` (accepts a keyboard key string) but the wiring suggests it is called with a `KeyboardEvent` object directly, and the implementation accesses `.key` on it. The interface type does not match the call site.

- **code-review-3** `budget/src/pages/use-transaction-table.ts:243` — `registerAutocompleteListeners()` adds document-level `scroll`, `resize`, and `click` listeners guarded by a module-level `listenersRegistered` flag. The flag is never reset, so the listeners persist for the lifetime of the module and are not removed when the transaction table unmounts.

- **code-review-4** `budget/src/pages/use-transaction-table.ts:365` — `useInfiniteScroll`'s `useLayoutEffect([nextBefore])` captures `onAppend`, `onScrollError`, and `onScrollLoading` from the `handlers` object in its closure without listing them as dependencies. If those handlers change identity across renders, the effect will call stale versions.

- **review-2** `budget/src/pages/home-chart.ts:225-270` — `buildCategorySankey` reads `#sankey-controls` and its child inputs via `document.getElementById` rather than React refs. As a chart island in `CategorySankey.tsx` this works, but tightly couples the chart logic to DOM IDs that must exist in the rendered markup.

- **review-4** `budget/src/pages/use-transaction-table.ts:3281` — `useInfiniteScroll`'s `useLayoutEffect` depends on `[nextBefore]`. Each successful scroll batch advances `nextBefore`, which triggers cleanup (detaching the `IntersectionObserver`) and re-setup immediately. This is correct but means the observer is torn down and re-attached on every batch.

- **review-5** `budget/src/use-app-state.ts:544-554` — `onUploadLabelKeydown` is typed as `(key: string) => void` in `AppApi` but the implementation ignores the `key` parameter and reads `document.getElementById('upload-label-input')` directly; the type signature does not reflect the actual call contract.

- **review-6** `budget/src/pages/CategorySankey.tsx:1496` — `buildCategorySankey`'s `ChartTeardown` removes the `TRANSACTIONS_APPENDED_EVENT` listener and disconnects the `ResizeObserver` but (before this review's fix) did not remove the control-element event listeners — addressed by code-review-0 above. Noted here for traceability.

- **review-7** `budget/src/LegacyRoute.tsx:666-671` — `dangerouslySetInnerHTML={{ __html: html ?? '' }}` sets arbitrary HTML from legacy route render functions into a `<main>` element. The HTML originates from `renderBudgets`, `renderAccounts`, etc. — internal functions that produce trusted markup — so XSS is not a concern at current scope, but this bridge pattern limits React's DOM ownership over that subtree.

- **auth-0** `budget/src/LegacyRoute.tsx:47,66` — `LegacyRoute` injects server-rendered HTML strings via `dangerouslySetInnerHTML`. The HTML comes from legacy render functions (`renderBudgets`, etc.) that produce trusted markup, so this is not a vulnerability, but it is the XSS-relevant seam: any future route that generates HTML from user-controlled data would need DOMPurify or equivalent before reaching this injection point.

- **auth-2** `budget/src/pages/use-transaction-table.ts:353-396` — `useInfiniteScroll` calls `getActiveDataSource().getTransactions()` without an explicit authorization guard. The only gate is that `nextBefore` must be set (transaction data must have loaded), which implies a prior authenticated read — consistent with the app's auth-on-load model.

- **data-exposure-0** `budget/src/use-app-state.ts:339` — In the `onExport` error handler, `error.message` is forwarded to the nav UI via `showNavError(error instanceof Error ? error.message : 'Export failed')`. Export errors from file-system APIs or crypto operations may include path fragments or internal state in their messages. This surfaces only to the logged-in user in the local-first context.

- **data-exposure-1** `budget/src/pages/Transactions.tsx:262-264` — `TableErrorBoundary.componentDidCatch` calls `logError(error, { operation: 'transactions-table-render' })` for render-time `DataIntegrityError` throws. `logError` forwards to Sentry/console; `DataIntegrityError` messages may include budget IDs or field names. Acceptable in the local-first model.

---

### Dismissed

- **code-review-5** `budget/src/pages/home-chart.ts:256` — `buildCategorySankey` throws `'sankey-controls element not found'` when `#sankey-controls` is absent. From `CategorySankey.tsx` this element is always present in the same render scope, so the throw cannot fire in practice. The guard is an appropriate defensive check at the DOM boundary.

- **review-3** `budget/scripts/prerender.tsx:268-276` — `new URL(import.meta.url).pathname` derives the dist directory path. On POSIX this is correct; on Windows the URL would include a drive letter producing an invalid path. The prerender script is a Node.js build tool run in the monorepo's Linux CI and is not expected to run on Windows.

---

### False-positive

- **red-team-2** `budget/src/upload.ts:87` — `UploadValidationError` includes the raw file-supplied version value in its message (`'Unsupported version: <value>'`). This is a local-first app where the user supplies their own files; the error message is shown only to the file's author. The version field is a small integer — not a credential or path. Not a vulnerability.

- **red-team-3** `budget/src/use-router.ts:35` — The document-level click interceptor validates the target pathname against `knownPaths` before calling `history.pushState`. The concern was that `pushState` could be called with an unvalidated URL; in fact the origin check and `knownPaths` guard happen before any state change.

- **auth-1** `budget/src/use-router.ts:35-37` — `history.pushState` is called with the raw `href` attribute value from a clicked anchor after validating the origin and that the pathname is in `knownPaths`. The origin check prevents open redirect; `knownPaths` prevents navigation to unregistered paths. Not a vulnerability.

PR #1956, opened 2026-06-19
qa-fix.js: test the already-satisfied partition logic

natb1:
Closes #1844

## What

Adds the missing test coverage for the `already_satisfied` partitioning logic in
the `qa-fix` workflow (`.claude/workflows/qa-fix.js`).

That partition is load-bearing: it drops `already-satisfied` residue items out of
the actionable `dispositions` array (treated as PASS) into a separate
`already_satisfied` array. A regression — e.g. filtering on `d.status` instead of
`d.class` — would silently either resurface already-met criteria as action items
or suppress real residue. The logic lived only in inline JS with **no** test (the
bash `dispatch-qa-disposition` kernel annotates `final_class`/`verify` but never
partitions, so it does not cover this).

## How

**Unit 1 — extract the partition into a pure function.** Moved the inline
partition block into a self-contained `partitionDispositions(allDispositions)`
returning `{ dispositions, already_satisfied }`, placed near `applyQaDisposition`
and delimited by stable sentinel comments. The call site is now a single
destructuring call. No behavior change — the filter/map expressions are identical.

**Unit 2 — node probe + CI assertions.** `qa-fix.js` is a Workflow-tool script
(top-level `await`, injected globals), so it cannot be imported or run by node.
The new probe `qa-fix-partition-probe.mjs` instead resolves `qa-fix.js` via
`import.meta.url`, slices the `partitionDispositions` function text out from
between the sentinel comments, and evals just that slice against a fixture mirroring
the f7-fixture class vocabulary (`opus-fixable` / `needs-main` / `needs-human` /
`already-satisfied`). It fails loudly if a sentinel is not found exactly once.

The assertions are wired into `test-dispatch-scripts.sh`, immediately after the
`dispatch-qa-disposition` f7 block. This placement is deliberate:
`run-unit-tests.sh` has no suite mapping for `.claude/workflows/*`, so a PR
touching only `qa-fix.js` triggers no vitest run — the only test that runs
unconditionally on every PR is the hook-tests job, which invokes
`test-dispatch-scripts.sh` directly. The new block asserts:

- the already-satisfied id is **absent** from `dispositions`,
- the already-satisfied id is **present** in `already_satisfied`,
- a non-already-satisfied id **stays** in `dispositions` (inverse guard against a
  filter that drops everything),
- the `partitionDispositions(allDispositions)` call site is still present in
  `qa-fix.js` (call-site coverage, since slice-eval tests the function not its caller).

## Verification

- Probe runs standalone (exit 0): `{"dispositions":["p1","p2","p3"],"already_satisfied":["p4"]}`.
- Full `test-dispatch-scripts.sh` suite green; new asserts pass alongside the
  existing `dispatch-qa-disposition` cases.
- Regression caught: flipping `d.class` → `d.status` in `partitionDispositions`
  turns the new asserts red.
- Sentinel-guard fires: renaming a sentinel makes the probe exit non-zero with a
  clear "sentinel not found" message.


natb1:
## QA Summary — Issue #1844 / PR #1956

### Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Partition probe runs standalone and produces correct partition | shell-command | PASS |
| 2 | New assertions pass inside the real CI vector (full hook-test suite) | shell-command | PASS |
| 3 | Regression sensitivity — wrong-field filter is caught | shell-command | PASS |
| 4 | Sentinel-guard fires loudly when a sentinel is missing | shell-command | PASS |
| 5 | Acceptance-criteria assertions are present in the test suite | shell-command | PASS |
| 6 | No behavior change vs. the original inline partition | needs-human-judgment | deferred → assessed by Opus |

**Counts: PASS 5 / FAIL 0 / SKIP 0**

No browser walkthrough (no browser component — changed files are scripts and workflow only).

### Evidence

- **Item 1** — `node qa-fix-partition-probe.mjs` exits 0, stdout: `{"dispositions":["p1","p2","p3"],"already_satisfied":["p4"]}` — exactly matches PR description.
- **Item 2** — `test-dispatch-scripts.sh`: 3093/3093 passed, 0 failed. New `qa-fix partition (#1844)` block passes alongside the existing `dispatch-qa-disposition` f7 cases.
- **Item 3** — Mutated source (`d.class → d.status`) produces `{"dispositions":["p1","p2","p3","p4"],"already_satisfied":[]}` — p4 stays in dispositions, demonstrating the test assertions would turn red on the named regression.
- **Item 4** — Renamed START sentinel makes the probe exit 1 with: `qa-fix-partition-probe: START sentinel not found exactly once (count=0)`.
- **Item 5** — `grep -c` returns 3 (absent from dispositions, present in already_satisfied, present in dispositions).

### Disposition Triage

Step 3.5 ran (1 residue item). Every residue item was classified `already-satisfied` by Opus — `result.dispositions` is empty.

- **No behavior change vs. the original inline partition** — The partition function is a pure transformation of its single argument (no shared state). The filter/map expressions are identical to the original inline block: `already_satisfied = items with class === 'already-satisfied'` projected to `{id, title, kind, rationale}`; `dispositions = items with class !== 'already-satisfied'` kept whole in input order. The call site destructures the same two names downstream consumers use. The `qa-fix-partition-probe.mjs` fixture further asserts the same partition. Criterion is provably met by code-reading evidence; no code defect to fix and no human input needed.

### Verdict

Clean pass. All acceptance criteria satisfied:

- ✅ A unit test (probe + `test-dispatch-scripts.sh`) exercises the JS partition step with a residue list containing an `already-satisfied` item.
- ✅ Test asserts `dispositions` does NOT contain the already-satisfied item (p4 absent assertion).
- ✅ Test asserts `already_satisfied` DOES contain the already-satisfied item (p4 present assertion).
- ✅ Inverse guard: non-already-satisfied item stays in `dispositions` (p1 present assertion).
- ✅ Call-site coverage: grep assertion confirms `partitionDispositions(allDispositions)` is still invoked.
- ✅ Sentinel guard fires on rename (non-zero exit with clear error message).
- ✅ Regression caught: `d.class → d.status` mutation breaks the test assertions.

Applying `dispatch:qa-done`.

natb1:
## Review findings — PR #1956

### Fixed (3)

- **`test-dispatch-scripts.sh:26221-26225`** — partition tests only asserted one non-already-satisfied item (p1/opus-fixable) in `dispositions`; added presence checks for p2 (needs-main) and p3 (needs-human), a dispositions-count assertion (must equal 3), and an `already_satisfied` key-shape assertion; probe now emits `already_satisfied_keys`. _(code-review, review)_ — fix commit `e2aaf2a6`
- **`qa-fix-partition-probe.mjs:26221-26225`** — same fix, probe updated to emit key-shape output for the new assertion. _(review)_ — fix commit `e2aaf2a6`
- **`qa-fix-partition-probe.mjs:67`** — `new Function('return ' + fnSource)()` called without a non-empty guard on `fnSource`; added explicit check after `.trim()` that exits with a clear probe error before `new Function()` is reached when sentinels are present but the slice is empty. _(review)_ — fix commit `e2aaf2a6`

### Informational (2)

- **`test-dispatch-scripts.sh:26232`** — call-site guard uses `grep -c '= partitionDispositions(allDispositions)'` asserting count equals `"1"`; this would fail if the function is called more than once (benign for now, but worth knowing). _(code-review)_
- **`qa-fix-partition-probe.mjs:67`** — `new Function(...)` evaluates the sliced function in global scope with no try/catch; a syntax error in the slice produces an unhandled exception rather than a clean probe exit. Acceptable for a CI test helper — not a production path — but noted. _(code-review)_

### Dismissed (2)

- **`test-dispatch-scripts.sh:26218`** — `out` variable reused from the f7-fixture block; in shell scripts `out` is a local transient — reassignment is the normal pattern, not a scoping hazard. _(review)_
- **`test-dispatch-scripts.sh:26232`** — `|| true` suffix on `$(grep -c ...)` is redundant since `grep -c` exits 0 with count `0` when there are no matches. Harmless; not changed. _(review)_

### False-positive (4)

- **`qa-fix-partition-probe.mjs:21-25`** — relative `../../../workflows/qa-fix.js` path traversal flagged; probe is a CLI tool that reads its own sibling file by design, not untrusted input. _(input-validation)_
- **`qa-fix-partition-probe.mjs:67`** — `new Function(fnSource)` flagged as eval of attacker-controlled code; source file is checked into the repo and controlled by the same commit, not external input. _(input-validation, red-team)_
- **`test-dispatch-scripts.sh:26218`** — `out=$(node ...)` without exit-code check before consuming `$out`; the script runs under `set -e`, so a non-zero exit from the node invocation terminates the test run before `$out` is consumed. _(red-team)_
- **`qa-fix.js:195`** — sentinel comment flagged as injectable; the comment is authored in the same commit as the probe that reads it and is not derived from user input. _(red-team)_

### Security

No attack surface in the changed files that was not covered above — no auth/data-exposure/Firebase surface touched. CodeQL: 0 open alerts on this PR ref. Dependency audit: not applicable (no `package.json`/`package-lock.json` changes).

PR #1892, opened 2026-06-18
dispatch: phases emit a machine-readable per-run outcome envelope

natb1:
Closes #1860

## Summary

Dispatch phases now emit a machine-readable per-run **outcome envelope** at every
terminal path, and `dispatch-token-audit` reads it to compute review-fix / qa-fix
hit-rate by deterministic aggregation — no LLM classification step.

Previously, measuring a phase's yield meant LLM-classifying its prose summary with
parallel subagents: expensive, non-reproducible, and unreliable (a "Fixed"
disposition could lie when no diff materialized). With the envelope, hit-rate
becomes a standing metric folded straight out of the transcripts.

Scope is `review-fix` + `qa-fix`; the schema doc explains incremental adoption for
the remaining phases.

## What landed (7 units)

1. **Schema contract** — `.claude/docs/outcome-envelope.md`: the single source of
   truth. Carrier shape (`<!-- dispatch:outcome:v1 -->` marker + fenced JSON,
   mirroring the `dispatch:plan` convention), every field with type + nullability,
   the `disposition` enum (`completed` / `completed_with_fixes` / `escalated`) and
   per-path mapping, `findings_actionable` membership per phase, the hit-rate
   formulas with zero-denominator guards, `subagents_launched` semantics, the
   reader contract (tool_result carrier, last-wins, worker-sessions-only), and the
   adoption recipe.
2. **Emit script** — `dispatch-emit-outcome`: prints the canonical block to stdout,
   patterned on `dispatch-mark-complete` for arg parsing and clear-error
   validation, but **unconditionally** (no `CLAUDE_JOB_DIR` no-op) so the envelope
   always lands in the transcript, including on escalation paths where no
   completion marker is written. Builds JSON via `jq -n`.
3. **Workflow return extensions** — `review-fix.js` / `qa-fix.js`: additive
   computed fields (`findings_surfaced`, `findings_actionable`, `fixes_applied`,
   `followups_filed`, `subagents_launched`, `disposition`) so the SKILL bodies do
   no arithmetic. Adds an explicit subagent-fan-out counter to each workflow.
4. **SKILL wiring** — an emit call at all five terminal paths (review-fix:
   normal-complete, deviation; qa-fix: clean-pass, auto-fix-applied, escalate).
   Because the qa-fix workflow plans but does not execute fixes (the auto-fix loop
   runs in the SKILL thread), the SKILL recomputes `fixes_applied` /
   `followups_filed` / `subagents_launched` and overrides `disposition` on the
[...]

PR #1958, opened 2026-06-19
dispatch-emit-outcome: validate enum sets with bash arrays (no word-splitting)

natb1:
Closes #1903

## Summary

`dispatch-emit-outcome` validates its `--phase` and `--disposition` flags
against two hardcoded enum sets. Both sets were space-separated strings, and the
`_in_set` helper iterated them with an unquoted word-split on `$set`. For the
current single-word values this works, but it is a latent correctness hazard: a
future multi-word or metacharacter-bearing enum value (e.g. `"completed with
fixes"`) would split into separate tokens, none matching the candidate, so every
lookup would silently return false.

This converts both enums to bash arrays and rewrites `_in_set` to the
spread-args form, iterating a quoted `"$@"` — no word-splitting — while keeping
every current value validating exactly as before. The script already assumes
bash 4+ (it uses array idioms like `pr_arg=(...)` elsewhere).

Surfaced by the #1892 review pass; out of scope for that PR.

## Changes

- `ALLOWED_PHASES` and `ALLOWED_DISPOSITIONS` are now bash arrays.
- `_in_set` takes the candidate plus the set as spread args
  (`_in_set "$val" "${ARR[@]}"`) and iterates `"$@"` — no unquoted `$set`.
- Both call sites spread the array (`"${ALLOWED_*[@]}"`); their error messages
  use `${ALLOWED_*[*]}`, whose default-IFS join reproduces the old
  `review qa` / `completed completed_with_fixes escalated` text verbatim.
- `test-emit-outcome.sh` gains a **Suite 3: enum validation** block — every
  existing enum value still validates (exit 0), unknown values are rejected
  (exit 2) with the full set listed in stderr, and a direct `_in_set` idiom test
  proves a multi-word element matches.

## Verification

- `test-emit-outcome.sh`: 26/26 passed (Suites 1, 2, and new 3).
- `test-dispatch-scripts.sh`: 3089/3089 passed — no caller regressed.
- Happy-path smoke (`--phase review … --disposition completed`): exit 0, emits
  the `<!-- dispatch:outcome:v1 -->` marker + JSON.
- Rejection smoke (`--phase bogus`): exit 2, stderr lists `expected one of:
  review qa`.


natb1:
# QA Pass — Issue #1903 / PR #1958

## Results: 6/6 PASS, 0 FAIL, 0 SKIP

All items were `script-verifiable` (pure-scripts PR, no browser component).

| # | Title | Result |
|---|-------|--------|
| 1 | `ALLOWED_PHASES` and `ALLOWED_DISPOSITIONS` declared as bash arrays | PASS |
| 2 | `_in_set` no longer uses unquoted word-splitting | PASS |
| 3 | All existing enum values continue to validate (golden path) | PASS |
| 4 | Unknown enum value rejected with full set listed in stderr | PASS |
| 5 | Hypothetical multi-word enum value matches correctly | PASS |
| 6 | Existing `dispatch-emit-outcome` tests pass unchanged | PASS |

## Evidence

- `grep` confirmed `ALLOWED_PHASES=("review" "qa")` and `ALLOWED_DISPOSITIONS=("completed" "completed_with_fixes" "escalated")` at lines 51–52.
- `_in_set` iterates `"$@"` (line 170); no `$set` word-splitting remains.
- `test-emit-outcome.sh`: **26/26 passed** (Suites 1, 2, and new Suite 3).
  - Suite 3 covers: all 5 existing enum values validate (exit 0), unknown phase/disposition rejected (exit 2) with the full set listed, and the multi-word element match guarantee via the `_in_set` idiom test.

natb1:
## Review pass — #1903

**Surface:** `code` · **CodeQL alerts:** none · **No fixes applied**

---

### Informational

- `test-emit-outcome.sh:156` — Suite 3's multi-word safety test exercises a hand-copied `_in_set_local` function rather than the SUT's real `_in_set`. The test proves the _in_set idiom works but does not directly exercise the binary's `_in_set`. This is a known trade-off for a shell helper that cannot be sourced in isolation; the idiom mirror is the correct pattern here.
  _(sources: code-review, review)_

---

### Dismissed

- `test-emit-outcome.sh:164` — The `&& ... || ...` pattern for assertion dispatch is unusual, but it is correct in this context: `assert_eq` always succeeds (it only prints and increments counters; it never exits non-zero), so the `||` branch is unreachable and the idiom is equivalent to a plain `if`. No change needed.
  _(source: code-review)_

---

### Fixed

_None._

### Required (security)

_None._

_None._

### False-positive (security)

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1962, opened 2026-06-19
dispatch-token-audit: null-ify partial outcome envelopes at the reader

natb1:
Closes #1909

## Problem

`dispatch-token-audit`'s stage-2 jq `rate($num; $den)` helper
(`aggregate-usage.sh:377`) guards its denominator (`($den // 0) == 0` → `null`,
no divide-by-zero) but coerces the **numerator** with `($num // 0)`. A *partial*
outcome envelope — valid JSON that survives the stage-1 `try fromjson catch null`
guard but is missing a required count key — therefore yields a silently-wrong
`0` rate instead of signalling the gap. An envelope missing `findings_surfaced`,
for example, produces `hit_rate = (null // 0) / 3 = 0.0`, masking the missing
field.

The writer (`dispatch-emit-outcome`) already `_require`s all three count flags,
so a partial envelope is **not** reachable through the normal emit path today —
only via a future phase-adopter that hand-builds an envelope, or a corrupted /
hand-edited transcript. This is a reader-side trust-boundary fix: per
`.claude/rules/code-style.md` (prefer clear errors over defensive fallbacks), the
reader must surface the gap rather than fabricate a `0`.

## Fix (Option 1 — strict validation)

Null-ify a partial envelope at the **single stage-1 chokepoint** so it is treated
as `outcome: null` everywhere downstream:

```jq
| ( ($outcome_raw // "") | try fromjson catch null ) as $parsed
| ( if ($parsed | type) == "object"
      and ($parsed.findings_surfaced   | type) == "number"
      and ($parsed.findings_actionable | type) == "number"
      and ($parsed.fixes_applied       | type) == "number"
    then $parsed else null end ) as $outcome
```

A **type check** (not `has()`) validates each count is a number, catching both a
missing key *and* an explicit `"findings_surfaced": null`. One edit covers both
consumers automatically: per-session `outcome` / `outcome_rates`, and pooled
`by_phase_outcome` (whose `.outcome != null` filter already drops the null-ified
envelope).

`followups_filed` and `subagents_launched` stay `// 0`-coerced by design — they
do not feed `rate()`, and #1909 is scoped to the three rate numerators.

## Changes

- `aggregate-usage.sh` — the chokepoint edit, plus an expanded outcome-envelope
  comment block documenting the null-ification and the followups/subagents
  exclusion.
- `.claude/docs/outcome-envelope.md` — a Reader-contract bullet documenting that
  a partial envelope is treated as absent (no fabricated `0` rates).
- `test-aggregate-usage.sh` — a new **isolated** regression section (fresh
  `mktemp -d` projects root, one worker session with a `findings_surfaced`-less
  envelope) asserting `outcome == null`, `outcome_rates == null`, and no
  fabricated `by_phase_outcome` bucket. The shared `setup()` fixture is untouched,
  so existing count assertions are unchanged.

## Verification

- `test-aggregate-usage.sh`: 69/69 passed (66 prior + 3 new).
- `test-audit-aggregate-writer.sh`: 32 passed, unchanged.
- Reverted-fix check confirms the three new assertions genuinely flip under the
  fix (they fail against the pre-fix script), proving they exercise the bug.

### DIFF
--- a/.claude/docs/outcome-envelope.md
+++ b/.claude/docs/outcome-envelope.md
@@ -153,6 +153,13 @@ source 2 before calling the emit script. The emitted value is the total.
 - Only top-level worker sessions emit. The reader excludes subagent transcripts,
   so a subagent that happens to print the marker does not produce a stray
   envelope.
+- **Strict count validation**: an envelope that parses as JSON but is missing
+  any of the three rate-feeding counts — `findings_surfaced`,
+  `findings_actionable`, `fixes_applied` (each non-nullable here) — is null-ified
+  and treated as **absent**, exactly like a session with no envelope. The reader
+  never coerces a partial envelope into a zero-count object, so a partial
+  envelope can never produce a fabricated `0` rate. (`followups_filed` and
+  `subagents_launched` are not validated this way — they do not feed any rate.)
 
 ## Adoption recipe for a new phase
 
--- a/.claude/skills/dispatch-token-audit/scripts/aggregate-usage.sh
+++ b/.claude/skills/dispatch-token-audit/scripts/aggregate-usage.sh
@@ -268,6 +268,19 @@ def cmd_prefix(c):
 # tool_results in document order and take the last. fromjson is wrapped in
 # try/catch so a malformed envelope binds null and NEVER aborts the file —
 # mirroring the clear-error/files_failed philosophy.
+#
+# STRICT VALIDATION (#1909): after a successful fromjson, a PARTIAL envelope
+# (valid JSON missing a required rate-feeding count key) binds null and is
+# treated as ABSENT — never coerced to a zero-count object. The doc marks the
+# three rate-feeding counts (findings_surfaced, findings_actionable,
+# fixes_applied) non-nullable, so we validate each is a `number` (a type check
+# catches both a missing key AND an explicit null). Without this, rate()'s
+# `($num // 0)` would silently fabricate a 0 rate from a partial envelope. Per
+# .claude/rules/code-style.md (clear errors over defensive fallbacks), surface
+# the gap as outcome:null instead. This single stage-1 chokepoint propagates to
+# per-session outcome/outcome_rates and pooled by_phase_outcome downstream.
+# followups_filed/subagents_launched are intentionally NOT validated here —
+# they do not feed rate(), so they keep their `// 0` coercion downstream.
 | ( [ $msgs[]
       | select(.type=="user")
       | .message.content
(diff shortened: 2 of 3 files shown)


natb1:
# QA Summary — #1909 (PR #1962)

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Run the full aggregate-usage test suite (69/69) | shell-command | PASS |
| 2 | Run the sibling writer test suite (32/32) | shell-command | PASS |
| 3 | Verify the fix lives at the stage-1 chokepoint | shell-command | PASS |
| 4 | Verify the documentation was updated | shell-command | PASS |
| 5 | Verify a complete envelope still produces non-null rates | shell-command | PASS (covered by item 1 — `sessions[sess-worker].outcome.findings_surfaced == 8`) |
| 6 | Manual spot-check of the jq validation predicate logic | deferred | see below |

## Bugs found

None.

## Disposition triage

Item 6 was the only residue item (kind: `needs-human-judgment`). The Opus disposition Workflow classified it **already-satisfied** — dropped as PASS.

- **#6 — Manual spot-check of the jq validation predicate logic**: The item asks a reviewer to confirm the jq predicate semantics are sound. All three claims are provably correct by reading the landed code in `aggregate-usage.sh` (lines 290–296) and verified by the passing test suite (69/69, including the three #1909 partial-envelope null-ification assertions). (1) `(.findings_surfaced|type)=="number"` correctly catches both missing keys and explicit nulls because jq yields type `"null"` for both. (2) `($parsed|type)=="object"` rejects non-object `fromjson` survivors (bare scalars/arrays/null). (3) `followups_filed`/`subagents_launched` are intentionally not gated — they never feed `rate()` and keep their `// 0` coercion downstream. There is no code defect to fix and no human intent needed; the criterion is a positive logic-correctness confirmation already met by readable evidence (code + passing tests).

- ✅ AC #1: `aggregate-usage.sh` stage-1 chokepoint null-ifies partial envelopes — schema-check validates all three rate-feeding counts as `number`, and `test-aggregate-usage.sh` fixture asserts `outcome == null` for a `findings_surfaced`-less envelope.
- ✅ AC #3: 69/69 passed in `test-aggregate-usage.sh` (66 prior + 3 new); 32/32 in `test-audit-aggregate-writer.sh`.
- ✅ (AC #2 superseded by AC #1 — Option 1 implemented, not Option 2.)

natb1:
## Review findings — #1909

### Fixed

- **R5** — `test-aggregate-usage.sh`: new partial-envelope section lacked a cleanup trap for `PARTIAL_ROOT`; the EXIT trap from the P4 section would not clean it up on test failure mid-section. Added `trap 'rm -rf "$PARTIAL_ROOT" "$FAKE_WRITER_DIR"; teardown' EXIT INT TERM` immediately after `mktemp`. (commit `247c8dc7`)

### Informational

- **R4** — `test-aggregate-usage.sh:552`: `jq . "$partial_jsonl" >/dev/null` validates the fixture silently — errors would be swallowed. Consider `jq . "$partial_jsonl" >/dev/null || { echo 'ERROR: malformed test fixture'; exit 1; }` for a clearer failure signal.
- **R6** — `test-aggregate-usage.sh`: the test envelope uses a missing `findings_surfaced` key; an explicit `"findings_surfaced": null` (present but null) is also caught by the `(type) == "number"` check but has no dedicated assertion. A second assert case would prove the type-check covers both absence and explicit null.

### Dismissed

- **R1** — alleged shell parsing fragility in `aggregate-usage.sh`: the `printf '%s\n'` + `sed` extraction pattern is in the calling review-fix skill, not in the changed script. No issue in the diff.
- **R2** — jq `and` short-circuit concern in `aggregate-usage.sh:290–295`: jq does short-circuit `and`; this is idiomatic and not a real issue.
- **R3** — comment in `aggregate-usage.sh` references `.claude/rules/code-style.md` path: this follows established codebase convention for cross-referencing rules; path coupling is intentional.

### False-positive (security)

- **R7** — no security surface: all changed files are internal tooling scripts (`.claude/skills/`) and documentation (`.claude/docs/`), with no network-facing code, authentication logic, data exposure risk, or external API integration.

PR #1969, opened 2026-06-19
test-dispatch-scripts: extract make_malformed_remote_fixture helper for the 5b-5e malformed-URL cases

natb1:
Closes #1955

## What

The four malformed-URL format-guard test cases in
`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — 5b
(read-plan / SSH-with-port), 5c (write-plan / non-GitHub), 5d (write-plan /
SSH-with-port), 5e (read-plan / non-GitHub) — each opened with a byte-identical
~14-line bare-repo + worktree fixture setup block. Across the four cases the
setup differed in only two things: the `origin` remote URL and the local root
variable name.

This extracts that block into a single `make_malformed_remote_fixture` helper.
The helper takes the URL, builds the temp bare repo + seeded `main` + worktree,
and echoes the created root path for the caller to capture — matching the
suite's established `make_*` helper convention (`make_pr`, `make_pr_union` emit
their result via `printf` for `$(...)` capture). Each of the four call sites is
now a single capture line; its own (already distinct) assertion and inline
cleanup are unchanged.

## Why

Removes a four-way copy of identical fixture setup. The suite already defines
dozens of `make_*` / `*_setup` helpers, so a section-local fixture helper
matches the existing convention rather than introducing a new one.

## Out of scope

- The per-case assertion blocks and expected diagnostics — each case keeps its
  distinct script and expected error.
- The per-case cleanup (`worktree prune` + `rm -rf`) — left inline; the helper
  covers setup only.
- The unrelated "5." bare+worktree block (`bw_root`) — a real-`github.com` case
  with a fake-gh stub, not a malformed-URL case.

## Verification

- `bash -n` syntax check passes.
- Full suite: 3101/3101 pass, 0 failures (no assertions added or removed).
- All four malformed-URL cases still print PASS.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -28968,10 +28968,11 @@ rm -rf "$bw_root"
 
 # Build a bare-repo + worktree fixture whose origin remote is <url>, for the
 # malformed-URL format-guard tests (5b-5e). Creates a temp root with .bare/,
-# seeds one commit on main via a throwaway seed clone, adds a worktree at
+# seeds one commit on main via a throwaway seed repo, adds a worktree at
 # worktrees/42-foo, and echoes the root path for the caller to capture:
 #   root=$(make_malformed_remote_fixture <url>)
 make_malformed_remote_fixture() {
+  set -e
   local url="$1"
   local root seed
   root=$(mktemp -d)
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -28715,23 +28715,35 @@ fi
 git --git-dir="$bw_root/.bare" worktree prune 2>/dev/null || true
 rm -rf "$bw_root"
 
+# Build a bare-repo + worktree fixture whose origin remote is <url>, for the
+# malformed-URL format-guard tests (5b-5e). Creates a temp root with .bare/,
+# seeds one commit on main via a throwaway seed clone, adds a worktree at
+# worktrees/42-foo, and echoes the root path for the caller to capture:
+#   root=$(make_malformed_remote_fixture <url>)
+make_malformed_remote_fixture() {
+  local url="$1"
+  local root seed
+  root=$(mktemp -d)
+  git init --bare "$root/.bare" >/dev/null 2>&1
+  git --git-dir="$root/.bare" config user.email "t@t" 2>/dev/null
+  git --git-dir="$root/.bare" config user.name "t" 2>/dev/null
+  git --git-dir="$root/.bare" remote add origin "$url"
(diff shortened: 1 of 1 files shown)


natb1:
## QA Summary — Issue #1955

**Branch:** `1955-test-dispatch-scripts-extra` → PR #1969

### Results: 4/4 PASS, 0 FAIL, 0 SKIP

| # | Item | Result |
|---|------|--------|
| 1 | Helper `make_malformed_remote_fixture` defined exactly once | PASS |
| 2 | All four cases (5b–5e) call the helper, no inline fixture remains | PASS |
| 3 | Script syntax valid (`bash -n`) | PASS |
| 4 | Full test suite passes end-to-end | PASS |

### Detail

- **Item 1:** Exactly one `make_malformed_remote_fixture()` definition found in `test-dispatch-scripts.sh`.
- **Item 2:** Four real call sites confirmed (`mf_ssh_root`, `mf_gl_root`, `mf_ssh_w_root`, `mf_gl_r_root`). The grep used in the command also matched the docstring comment inside the helper itself — the actual 4 call sites are all present and correct.
- **Item 3:** `bash -n` exits 0, no syntax errors.
- **Item 4:** Suite exits 0, **3101/3101 passed, 0 failed**. All four malformed-URL cases specifically confirmed PASS:
  - `read-plan: SSH-with-port malformed URL exits 1` (5b)
  - `write-plan: non-GitHub remote malformed URL exits 1` (5c)
  - `write-plan: SSH-with-port malformed URL exits 1` (5d)
  - `read-plan: non-GitHub remote malformed URL exits 1` (5e)

### Disposition

No bugs found. No `needs-human-judgment` items. Clean autonomous pass.

natb1:
## Review: PR #1969

---

### Fixed

- **F1a** `make_malformed_remote_fixture` (lines 28974–28992): `errexit` is not inherited in command substitution — a failing `git commit`, `git push`, or `git worktree add` inside the helper previously fell through silently to `printf '%s' "$root"`, returning a half-built fixture to the caller. Added `set -e` as the first statement in the function body; since the helper is called exclusively inside `$(...)`, errexit is scoped to that subshell and does not affect the main test script. (fix commit b4b0dd14)

- **F1b** line 28971: Comment said "a throwaway seed clone" — the helper uses `git init` on a fresh directory and pushes into the bare repo, which is the opposite of a clone. Changed to "a throwaway seed repo". (fix commit b4b0dd14)

---

### Informational

- **F3** Refactoring fully covers all four duplicate cases (5b–5e) — no gaps.
- **F4** Placement and naming (`make_malformed_remote_fixture`) match the suite's established section-local `make_*` helper convention.

---

### Deferred

- **F2** The 5a bare+worktree block (`bw_root`) uses a byte-identical setup but was intentionally excluded (it is a real-github.com case with a fake-gh stub, not a malformed-URL case). If a future PR absorbs it, a more generic helper name (e.g. `make_bare_worktree_fixture`) would be appropriate. Filed as follow-up #1977.

PR #1975, opened 2026-06-19
blog: test destroy() teardown and onAuthStateChanged same-UID guard

natb1:
Closes #1716

## What

Adds three test-only assertions to `blog/test/create-blog-app.test.ts` covering
`destroy()` teardown paths and the `onAuthStateChanged` same-UID guard that
previously had no dedicated coverage. No source changes —
`blog/src/create-blog-app.ts` is untouched.

The existing suite only called `destroy()` in `afterEach` for cleanup, never
asserting on it. These tests make a future refactor that drops a teardown
registration or the same-UID early-return fail CI instead of silently
regressing.

## Tests added

- **Test A — router popstate listener removed by `destroy()`.** Enters via
  `/admin` (which replaces the prerendered `#posts`), calls `destroy()`, then
  fires a `/` popstate and asserts `renderHomeHtml` is not called. The `/admin`
  entry is deliberate: it leaves `#app` with no live `#posts`, so the home render
  guard does not short-circuit — the test discriminates and fails if
  `router.destroy()` is dropped from teardowns.
- **Test B — `headerObserver.disconnect()` called by `destroy()`.** Spies on
  `ResizeObserverStub.prototype.disconnect` before app creation and asserts it
  fires on `destroy()`.
- **Test C — same-UID guard.** Fires the captured `onAuthStateChanged` callback
  twice with the same UID and asserts `getPosts` runs only once — the second fire
  hits the `if (user?.uid === currentUser?.uid) return` early return.

The auth-*unsubscribe* teardown is already covered by in-file Cases 8–10
(#1715) and is not duplicated here.

## Verification

`npx vitest run --root blog` — all 392 tests pass, including the three new cases.
Test A's discrimination was confirmed by temporarily removing
`teardowns.push(() => router.destroy())` and observing the failure, then
restoring the source.

### DIFF
--- a/blog/test/create-blog-app.test.ts
+++ b/blog/test/create-blog-app.test.ts
@@ -513,6 +513,100 @@ describe("createBlogApp routing and panel behavior", () => {
     expect(unsubscribeSpy).toHaveBeenCalledTimes(1);
   });
 
+  // Case A — AC#1: router.destroy() teardown: popstate no longer handled after destroy().
+  // Discrimination: if teardowns.push(() => router.destroy()) is dropped, the popstate
+  // listener survives and renderHomeHtml is called after handle.destroy() — test fails.
+  it("router popstate listener removed by destroy()", async () => {
+    history.pushState({}, "", "/admin");
+    document.body.innerHTML = `
+      <div id="nav"></div>
+      <div class="page">
+        <header></header>
+        <div id="app"><div id="posts" data-prerendered="true">PRERENDERED HOME</div></div>
+        <aside id="info-panel"></aside>
+        <button id="panel-toggle"></button>
+      </div>
+    `;
+
+    handle = createBlogApp(makeConfig());
+    const app = document.getElementById("app")!;
+
+    // Admin route replaces #posts with admin markup.
+    await vi.waitFor(() => expect(app.querySelector("#posts")).toBeNull());
+
+    handle.destroy();
+    handle = undefined; // prevent afterEach double-destroy
+
+    renderHomeHtml.mockClear();
+
+    // Fire a popstate toward "/" — the router's listener must no longer be active.
+    history.pushState({}, "", "/");
+    window.dispatchEvent(new PopStateEvent("popstate"));
+    await new Promise((r) => setTimeout(r, 0)); // macrotask tick
+
+    expect(renderHomeHtml).not.toHaveBeenCalled();
+  });
+
+  // Case B — AC#2: headerObserver.disconnect() called by destroy().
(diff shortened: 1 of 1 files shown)


natb1:
# QA Summary — #1716

**Result: CLEAN PASS ✓**

All 4 acceptance-criterion checks passed. No residue, no escalation.

## Items executed

All items were `script-verifiable` (shell-command lane). No browser component detected — browser lanes skipped.

| # | Title | Result |
|---|-------|--------|
| 1 | Test A — router popstate listener removed by destroy() | PASS |
| 2 | Test B — headerObserver.disconnect() called by destroy() | PASS |
| 3 | Test C — same-UID guard calls getPosts once | PASS |
| 4 | All new tests in blog/test/create-blog-app.test.ts and full suite passes | PASS |

## Verification

Command: `npx vitest run --root blog`

```
 Test Files  25 passed (25)
      Tests  392 passed (392)
   Duration  1.86s
```

All 392 tests pass (389 pre-existing + 3 new cases added by this PR).

The three new tests are at:
- `blog/test/create-blog-app.test.ts:519` — "router popstate listener removed by destroy()"
- `blog/test/create-blog-app.test.ts:553` — "headerObserver.disconnect() called by destroy()"
- `blog/test/create-blog-app.test.ts:568` — "same-UID guard: onAuthStateChanged twice with same UID calls getPosts once"

No source changes — `blog/src/create-blog-app.ts` is untouched.

natb1:
## Review — PR #1975

### Fixed

- **`blog/test/create-blog-app.test.ts:561`** — Case B disconnect assertion tightened from `toHaveBeenCalled()` to `toHaveBeenCalledTimes(1)`, pinning it to the single `headerObserver` created by `createBlogApp`. (fix commit `df1e930b`)

### Informational

- **`blog/test/create-blog-app.test.ts:519-548`** — Case A inlines a full DOM scaffold identical to `scaffoldDom()` (including `data-prerendered` on `#posts`) rather than calling the shared helper. The inline version works, but any future change to `scaffoldDom()` would need to be mirrored here manually.
- **`blog/test/create-blog-app.test.ts:573`** — Case C re-declares all six `firebase` config fields inline, duplicating the same block already present in `makeConfig()`. Passing only the `onAuthStateChanged` override via `makeConfig({ firebase: { onAuthStateChanged: ... } })` would remove the duplication.
- **`blog/test/create-blog-app.test.ts:573-586`** — (same location as above, corroborated by a second reviewer) — the inline `firebase` block repeats the identical set of fields verbatim.

### Security

PR #1979, opened 2026-06-19
test: extract shared playwright stubs into group-local helpers

natb1:
Closes #1971

## What

The four `playwright_install_with_deps` unit tests in
`test-dispatch-scripts.sh` each defined byte-for-byte-identical `npx`,
`timeout`, and `sleep` PATH stubs. This extracts the shared stubs into two
group-local helpers and drops the unreachable `timeout`/`sleep` stubs from the
skip-guard test.

## Changes

- Added two group-local helper functions right after the
  `=== playwright_install_with_deps ===` banner:
  - `write_playwright_npx_stub` — writes the counting `npx` stub (shared by
    tests 1–3).
  - `write_playwright_hang_stubs` — writes the passthrough `timeout` and no-op
    `sleep` stubs (shared by tests 2–4).
- Test 1 (skip-guard) now calls only `write_playwright_npx_stub`. Its counting
  `npx` stub is load-bearing — the `0 npx calls` assertion is only meaningful if
  `npx` invocations are observable — but `timeout`/`sleep` are unreachable when
  the guard returns early, so they are removed.
- Tests 2 and 3 call both helpers; test 4 keeps its unique conditional `npx`
  stub (the `exit 1` then `exit 0` recovery path under test) inline and calls
  `write_playwright_hang_stubs` for the rest.

The helpers stay group-local rather than hoisted into `setup()` — other test
groups rely on the real `timeout`/`sleep` for hang protection.

## Verification

No behavioral change. `bash test-dispatch-scripts.sh` passes the full suite
(3124/3124, `FAIL: 0`), with all 8 `playwright_install_with_deps` assertions
green.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -6749,28 +6749,37 @@ teardown
 echo ""
 echo "=== playwright_install_with_deps ==="
 
-# 1. Skip-guard: PLAYWRIGHT_BROWSERS_PATH set → return 0, 0 npx calls.
-echo "Test: playwright_install_with_deps skip-guard → rc 0, 0 npx calls"
-setup
-cat > "$TMPDIR_TEST/bin/npx" <<'FAKE'
+# Group-local stub writers (not hoisted into setup() — other groups rely on
+# real timeout/sleep for hang protection). Each writes to $TMPDIR_TEST/bin,
+# already first on PATH.
+write_playwright_npx_stub() {
+  cat > "$TMPDIR_TEST/bin/npx" <<'FAKE'
 #!/usr/bin/env bash
 cf="$NPX_COUNT_FILE"
 c=0; [[ -f "$cf" ]] && c=$(cat "$cf"); c=$((c+1)); echo "$c" > "$cf"
 exit "${NPX_EXIT:-0}"
 FAKE
-chmod +x "$TMPDIR_TEST/bin/npx"
-cat > "$TMPDIR_TEST/bin/timeout" <<'FAKE'
+  chmod +x "$TMPDIR_TEST/bin/npx"
+}
+write_playwright_hang_stubs() {
+  cat > "$TMPDIR_TEST/bin/timeout" <<'FAKE'
 #!/usr/bin/env bash
 while [[ "$1" == -* ]]; do shift; done
 shift
 exec "$@"
 FAKE
-chmod +x "$TMPDIR_TEST/bin/timeout"
-cat > "$TMPDIR_TEST/bin/sleep" <<'FAKE'
+  chmod +x "$TMPDIR_TEST/bin/timeout"
+  cat > "$TMPDIR_TEST/bin/sleep" <<'FAKE'
 #!/usr/bin/env bash
 exit 0
 FAKE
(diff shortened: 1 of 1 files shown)


natb1:
## QA Pass — #1971

**Result: PASS (7/7 items)**

### Items executed

| # | Title | Result |
|---|-------|--------|
| 1 | Full test suite passes after refactor | PASS |
| 2 | All playwright_install_with_deps assertions green | PASS |
| 3 | Helper functions defined once, called by intended tests | PASS |
| 4 | Test 1 (skip-guard) defines no timeout/sleep stubs | PASS |
| 5 | Test 4 keeps unique conditional npx stub inline | PASS |
| 6 | Script remains syntactically valid | PASS |
| 7 | No other test groups affected (setup() unchanged, helpers group-local) | PASS |

### Counts

- PASS: 7
- FAIL: 0
- SKIP: 0

### Details

**Item 1 — Full suite:** `bash test-dispatch-scripts.sh` → 3125/3125 passed, FAIL: 0 (one new test was added since the issue was filed, now 4 tests in the group).

**Item 2 — playwright group:** All 8 assertions green:
- skip-guard → rc 0 ✓ / 0 npx calls ✓
- first-attempt success → rc 0 ✓ / 1 npx call ✓
- both-attempts fail → rc non-zero ✓ / 2 npx calls ✓
- first-fails-then-succeeds → rc 0 ✓ / 2 npx calls ✓

**Item 3 — Helper structure:** `write_playwright_npx_stub` defined at line 6800, called by tests 1/2/3 (lines 6827, 6844, 6862). `write_playwright_hang_stubs` defined at line 6809, called by tests 2/3/4 (lines 6845, 6863, 6890). One definition each — no duplicates.

**Item 4 — Test 1 no hang stubs:** The skip-guard test block calls only `write_playwright_npx_stub`; no `timeout`/`sleep` stub writes present (they are unreachable when the guard returns early).

**Item 5 — Test 4 inline npx:** Test 4 defines its own inline conditional `npx` stub (`exit 1` then `exit 0` recovery path) and calls `write_playwright_hang_stubs` for timeout/sleep — not `write_playwright_npx_stub`, preserving the unique behavior under test.

**Item 6 — Syntax:** `bash -n` passes with no errors.

**Item 7 — Group-local:** Both helpers are defined after the `=== playwright_install_with_deps ===` banner (line 6795). `setup()` at line 43 contains no `write_playwright_*` calls — other groups continue to use the real `timeout`/`sleep` for hang protection.

No bugs found. No residue.

natb1:
## Review — PR #1979

**Security:** Security review: no attack surface — test-only diff (every changed path is a test file).

### Informational

- **`test-dispatch-scripts.sh:6818`** — `write_playwright_npx_stub` and `write_playwright_hang_stubs` are defined at file scope and persist for all subsequent test groups in the script. No action required, but if future test groups accidentally call these helpers without re-defining them, stale stubs may mask failures. The current structure is intentional (per the PR description) and the test passes cleanly.

- **`test-dispatch-scripts.sh:6828`** — The `timeout` stub's argument-parsing loop (`while [[ "$1" == -* ]]; do shift; done; shift`) implements non-trivial flag-stripping logic that is duplicated between the stub here and the original inline definitions. The helper consolidation is correct; this is just an observation that the stub's logic is load-bearing and should be kept in sync if the real `timeout` interface changes.

### Fixed

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1964, opened 2026-06-19
test: unit-test playwright_install_with_deps() in test-dispatch-scripts.sh

natb1:
Closes #1926

## What

Adds a `playwright_install_with_deps` test section to
`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`.

`playwright_install_with_deps()` (`lib.sh:530–552`) wraps
`npx playwright install --with-deps chromium` in a per-attempt `timeout` plus a
retry loop, short-circuits when `PLAYWRIGHT_BROWSERS_PATH` is set, and returns
non-zero once attempts are exhausted. It was the only `lib.sh` retry helper
without stub-based unit tests — `gh_retry`, its structural sibling, already has
them — breaking the convention that every such helper carries them.

## Tests

Three stub-driven tests, one `setup`/`teardown` block each, mirroring the
`gh_retry` idiom (PATH-prepended fakes + a counter file, helper run in a
`source`d subshell):

1. **Skip-guard** — `PLAYWRIGHT_BROWSERS_PATH` set → `rc 0`, 0 `npx` calls.
2. **First-attempt success** — `npx` exits 0 → `rc 0`, exactly 1 `npx` call.
3. **Both attempts fail** — `npx` exits 1, `PLAYWRIGHT_INSTALL_ATTEMPTS=2` → `rc`
   non-zero, exactly 2 `npx` calls.

The `npx`/`timeout`/`sleep` stubs make the section hermetic and fast: no real
download, no real timing, runs offline. Tests 2 and 3 `unset
PLAYWRIGHT_BROWSERS_PATH` inside their subshell so the live Nix-provided value
cannot short-circuit the helper on the dev machine.

## Verification

`bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` →
`Results: 3099/3099 passed, 0 failed`, with the new
`=== playwright_install_with_deps ===` section showing its three `PASS` lines and
no pre-existing test regressed.

## Scope

Test-only. No change to `lib.sh`, the helper itself, any other test section, or
CI config.

PR #1980, opened 2026-06-19
test: assert timeout invocation and npx args in playwright_install_with_deps tests

### LINKED ISSUE #1972: playwright_install_with_deps tests do not verify timeout invocation or npx argument shape
## Finding

The `timeout` stub in the three `playwright_install_with_deps` test cases transparently execs its payload without logging invocations. If `lib.sh` dropped the `timeout --kill-after=30` wrapper and ran bare `npx`, all three tests would still pass. Similarly, the `npx` stub ignores `$@`, so changing `--with-deps` or the browser name would not be detected. The timeout-expiry case (stub exits 124, representing a stall) is not tested at all, even though stall recovery under a flaky apt-get mirror is the core motivation for the wrapper.

## Motivation

The existing tests validate call-count and exit-code propagation but do not assert that `timeout` is actually invoked with the correct arguments, or that the function retries correctly when `timeout` exits 124. These gaps mean structural regressions (e.g. dropping the `timeout` wrapper or changing the `--with-deps` flag) go undetected.

## Recommended Fix

- Add argument logging to the `timeout` stub: `echo "$@" >> "$TMPDIR_TEST/timeout-calls.log"` before the `exec "$@"` line.
- Assert in each existing test that `timeout-calls.log` contains at least one entry (confirming `timeout` was invoked).
- Add a fourth test case where the `timeout` stub exits 124 unconditionally to verify the retry-on-timeout path triggers a second `npx` attempt.

## Location

`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` lines 6622–6628, 6657–6663, 6692–6698

## Acceptance Criteria

- [ ] The `timeout` stub in all three existing test cases logs `$@` to a file in `$TMPDIR_TEST`.
- [ ] Each existing test asserts that the timeout log file is non-empty (i.e. `timeout` was called at least once).
- [ ] A new test case stubs `timeout` to exit 124 unconditionally and asserts that `npx` is called exactly twice (retry after stall) and the function's final exit code is non-zero.
- [ ] The `npx` stub can be extended (or a variant written) to capture `$@` so argument shape (e.g. `--with-deps`, browser name) can be asserted. At minimum, the first-attempt-success test asserts the expected arguments were passed.
- [ ] All existing tests continue to pass after the changes.

## Context

Deferred from PR #1964 review (issue #1926) — out of scope for that PR.

natb1:
Closes #1972

## Why

PR #1964 added four unit tests for `playwright_install_with_deps()` in
`test-dispatch-scripts.sh`, but review (deferred to this issue) found they
under-verify the wrapper's structure:

- The `timeout` stub transparently `exec`ed its payload without recording the
  invocation — so if `lib.sh` dropped the `timeout --kill-after=30 "$timeout_s"`
  wrapper and ran bare `npx`, every test would still pass.
- The `npx` stub ignored `$@`, so a regression to `--with-deps` or the browser
  name (`chromium`) would go undetected.
- The timeout-expiry path (the wrapper's `timeout` returns 124 because the child
  stalled) — the core #1899 motivation — was never exercised.

## What changed

`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`:

- **Tests 2/3/4** (the cases that reach `timeout`): the `timeout` stub now logs
  its full `$@` to `TIMEOUT_LOG_FILE`, and each test asserts the log is non-empty,
  proving the `timeout` wrapper is actually invoked. Test 2 additionally captures
  the `npx` stub's `$@` and asserts the exact argument shape
  `playwright install --with-deps chromium`.
- **New fifth test** — timeout-stall: the `timeout` stub runs `npx` (incrementing
  the counter) and then returns 124, modeling `timeout` killing a stalled child.
  With `PLAYWRIGHT_INSTALL_ATTEMPTS=2` it asserts the retry produces 2 `npx` calls
  and a non-zero return.

Test 1 (skip-guard) is unchanged — it never reaches `timeout`/`npx`.

## Verification

Full suite passes (`bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`,
exit 0, 3131 tests). Adversarially confirmed the new assertions bite: dropping
the `timeout --kill-after=30 "$timeout_s"` prefix from `lib.sh` makes the
timeout-log assertions fail (reverted before committing).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -6957,6 +6957,47 @@ timeout_calls=$( [[ -s "$TIMEOUT_LOG_FILE" ]] && echo nonempty || echo empty )
 assert_eq "first fails then succeeds → timeout invoked" "nonempty" "$timeout_calls"
 teardown
 
+# 5. Timeout expiry (exit 124, stall) → retry exhausted → rc non-zero, 2 npx calls.
+#    Proves the timeout-expiry path works: timeout starts npx (counter increments)
+#    then returns 124 (simulating the child stalling past the deadline).  With
+#    PLAYWRIGHT_INSTALL_ATTEMPTS=2 both attempts time out and the wrapper fails.
+echo "Test: playwright_install_with_deps timeout-stall → rc non-zero, 2 npx calls"
+setup
+cat > "$TMPDIR_TEST/bin/npx" <<'FAKE'
+#!/usr/bin/env bash
+cf="$NPX_COUNT_FILE"
+c=0; [[ -f "$cf" ]] && c=$(cat "$cf"); c=$((c+1)); echo "$c" > "$cf"
+exit "${NPX_EXIT:-0}"
+FAKE
+chmod +x "$TMPDIR_TEST/bin/npx"
+cat > "$TMPDIR_TEST/bin/timeout" <<'FAKE'
+#!/usr/bin/env bash
+[[ -n "${TIMEOUT_LOG_FILE:-}" ]] && echo "$@" >> "$TIMEOUT_LOG_FILE"
+while [[ "$1" == -* ]]; do shift; done
+shift                     # drop the <timeout_s> duration arg
+"$@"                      # run npx (increments NPX_COUNT_FILE); not exec
+exit 124                  # simulate timeout killing the stalled child
+FAKE
+chmod +x "$TMPDIR_TEST/bin/timeout"
+cat > "$TMPDIR_TEST/bin/sleep" <<'FAKE'
+#!/usr/bin/env bash
+exit 0
+FAKE
+chmod +x "$TMPDIR_TEST/bin/sleep"
+NPX_COUNT_FILE="$TMPDIR_TEST/npx-5"
+rc=0
+(
+  source "$TMPDIR_TEST/lib.sh"
+  unset PLAYWRIGHT_BROWSERS_PATH
+  export PLAYWRIGHT_INSTALL_ATTEMPTS=2
+  export NPX_COUNT_FILE="$TMPDIR_TEST/npx-5"
+  playwright_install_with_deps 2>/dev/null
+) || rc=$?
+[[ "$rc" -ne 0 ]] && rc_state="nonzero" || rc_state="zero"
+assert_eq "timeout-stall → 2 npx calls" "2" "$(cat "$NPX_COUNT_FILE")"
+assert_eq "timeout-stall → rc non-zero" "nonzero" "$rc_state"
+teardown
+
 # ============================================================================
 # dispatch-complete-phase tests
 # ============================================================================
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -6985,6 +6985,7 @@ cat > "$TMPDIR_TEST/bin/npx" <<'FAKE'
 #!/usr/bin/env bash
 cf="$NPX_COUNT_FILE"
 c=0; [[ -f "$cf" ]] && c=$(cat "$cf"); c=$((c+1)); echo "$c" > "$cf"
+[[ -n "${NPX_ARGS_FILE:-}" ]] && echo "$@" >> "$NPX_ARGS_FILE"
 exit "${NPX_EXIT:-0}"
 FAKE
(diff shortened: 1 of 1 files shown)


natb1:
## QA Pass — Issue #1972

**Branch:** `1972-playwright-install-with-dep` → PR #1980

### Results: 5/5 PASS, 0 FAIL, 0 SKIP

| # | Item | Result |
|---|------|--------|
| 1 | Full dispatch-scripts test suite passes | PASS |
| 2 | Timeout-stall (exit 124) retry test exists and asserts 2 npx calls + non-zero rc | PASS |
| 3 | Three timeout-reaching tests log `timeout $@` and assert log non-empty | PASS |
| 4 | First-attempt-success test asserts exact npx argument shape | PASS |
| 5 | Diff touches only the test file (no production script changes) | PASS |

### Details

**Item 1 — Full test suite:** `3131/3131 passed, 0 failed`. The new timeout-stall test (Test 5) and strengthened Tests 2/3/4 all pass alongside the existing suite.

**Item 2 — Timeout-stall test (line 6960–6998):** Test 5 stubs `timeout` to run `npx` then `exit 124`, sets `PLAYWRIGHT_INSTALL_ATTEMPTS=2`, and asserts `timeout-stall → 2 npx calls` and `timeout-stall → rc non-zero`. Directly exercises the core #1899 motivation.

**Item 3 — Timeout log plumbing (Tests 2/3/4):** `TIMEOUT_LOG_FILE` is set outside and exported inside the subshell for each of the three timeout-reaching tests. The `timeout` stub appends `$@` to the log. Each test asserts the log is `nonempty`, proving the `timeout` wrapper is actually invoked and not bypassed.

**Item 4 — npx argument shape (Test 2):** `NPX_ARGS_FILE` captures `$@` from the `npx` stub. Test 2 asserts `playwright install --with-deps chromium` exactly, covering the `--with-deps` flag and browser name regressions.

**Item 5 — Scope:** Only `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` changed. No production-script modifications.

### No browser walkthrough needed
Change is test-only (shell script assertions). No browser component detected.

### No residue
All acceptance criteria verified machine-autonomously. Clean pass.

natb1:
## Review summary — #1972

**Surface:** tests-only diff — no security finder wave launched.

---

### Fixed

- **code-review-1** `.../test-dispatch-scripts.sh:6984` — Test 5 (timeout-stall) had no `NPX_ARGS_FILE` instrumentation in its `npx` stub and no assertion on the argument shape. Extended to Tests 3, 4, and 5: added `NPX_ARGS_FILE` logging to each `npx` stub, exported it into each subshell, and asserted the exact two-line argument log (`playwright install --with-deps chromium\nplaywright install --with-deps chromium`) to verify the retry-path argument contract.

- **review-1** `.../test-dispatch-scripts.sh:6892` — The `assert_eq` for npx args in Test 2 used bare `cat "$NPX_ARGS_FILE"` as the actual value without a file-existence guard; a missing file would produce a confusing `cat: no such file` error in the failure message. Replaced with `$( [[ -f "$NPX_ARGS_FILE" ]] && cat "$NPX_ARGS_FILE" || echo '<file missing>' )` across all args assertions.

Fix commits: `06db708b`, `1f6d304e`

---

### Informational

- **code-review-0** `.../test-dispatch-scripts.sh:6889` — Test 2's timeout-invocation assertion only checks that `TIMEOUT_LOG_FILE` is non-empty (timeout was called), not the exact argument shape. The argument shape is already verified in a more targeted way via the separate `npx args` assertion; adding a full timeout-arg assertion would be redundant.

- **code-review-2** `.../test-dispatch-scripts.sh:6861` — The instrumented timeout stub body is inlined across Tests 2, 3, 4, and 5 rather than extracted to a shared helper. The stubs vary slightly between tests (transparent `exec` vs stall+return-124), making a single helper awkward; the duplication is acceptable at this scope.

- **review-2** `.../test-dispatch-scripts.sh:7015` — Test 5 asserts `timeout-stall → 2 npx calls` before `timeout-stall → rc non-zero`, which is the reverse order used in Tests 3 and 4. No logic error — both assertions are present; the ordering is cosmetic.

---

### Dismissed

- **review-0** `.../test-dispatch-scripts.sh:7005` — "Test 5 never exports `TIMEOUT_LOG_FILE` into the subshell" — dismissed; `TIMEOUT_LOG_FILE` is now exported in the subshell (added as part of this review), and the `timeout invoked` assertion is present. This finding was already resolved before the second review pass ran.

---

### False-positive (security)

- **SEC-1** — Security review: no attack surface — test-only diff (every changed path is a test file).

PR #1984, opened 2026-06-19
budget: drive export.ts from collectionRegistry

natb1:
Closes #1697

## Summary

`budget/src/export.ts` was the last manual-enumeration holdout among the three
snapshot paths (upload, store, export). It listed all 13 IndexedDB stores twice —
once in the `Promise.all` fan-out and once in the output object literal — and
imported 13 per-entity `Idb*` types plus 13 per-entity `*ToRawJson` functions to
do it. Adding a store to the registry forced two extra edits in this file.

The sibling `budget/src/idb.ts` already drives off `collectionRegistry`
(`Object.keys(collectionRegistry) as DataStoreName[]`). This PR brings
`export.ts` to the same convention, fulfilling the registry's single-edit promise
for the export path.

## What changed

All changes are confined to `budget/src/export.ts`:

- Removed the 13 `import type { Idb* }` lines and the 13 `import { *ToRawJson }`
  lines; replaced them with the `collectionRegistry` value plus the `IdbOf` /
  `RawOf` / `DataStoreName` types (keeping `CollectionRawData`).
- Added a generic per-store helper `rawCollection<K extends DataStoreName>` that
  fetches a store and maps its records through that store's `toRawJson` adaptor.
- Replaced the manual 13-store `Promise.all` + output literal with a
  registry-driven body: iterate `Object.keys(collectionRegistry)`, build the
  collections via `Promise.all` → `Object.fromEntries`, and spread them into the
  output object.

The file shrank from 79 to 38 lines.

## Output is byte-identical

Registry key order matches the prior output-literal order exactly, and
`Object.keys` → `Promise.all` → `Object.fromEntries` preserve insertion order, so
`...collections` emits the same keys in the same order after
`version`/`exportedAt`/`groupId`/`groupName`. The existing `export.test.ts`
round-trip + field-mapping suite (16 tests) passes unchanged.

## Completeness is now by-construction

The issue prose said `CollectionRawData` would "still enforce completeness at
compile time," but that is in tension with the binding acceptance criterion that
adding a registry entry requires no change to `export.ts`: per-key compile-time
completeness would raise a tsc error inside `export.ts` on every registry
addition — exactly the burden this issue removes. So after the refactor,
completeness is by-construction (we iterate every registry key at runtime), not a
per-key tsc check. The `Object.fromEntries(...) as CollectionRawData` cast is
intended, and the registry being the sole source of collection keys is what
guarantees completeness.

### One implementation note for review

The plan's verbatim helper `records.map(collectionRegistry[name].toRawJson)` does
not type-check: TypeScript hits the correlated-union limitation and surfaces the
union of all 13 `toRawJson` signatures rather than the `K`-specific one (the
`idb.ts` `stores[name].put(record)` pattern dodges this only because `put`
accepts `any`). A plain `as (idb: IdbOf<K>) => RawOf<K>` cast was rejected for
insufficient overlap, so the helper binds the adaptor through an
`as unknown as (idb: IdbOf<K>) => RawOf<K>` local. The assertion is erased at
runtime, so output stays byte-identical (confirmed by the passing tests). A
cleaner structural fix would require typing the registry in
`collection-registry.ts`, which is out of scope here.

## Verification

- `npx vitest run --root budget export.test` — 16 passed.
- `npm run build --prefix budget` (`tsc --noEmit -p budget`) — clean.

### DIFF
--- a/budget/src/export.ts
+++ b/budget/src/export.ts
@@ -1,52 +1,23 @@
 /** Serializes IndexedDB stores back to the upload JSON format. Inverse of the upload pipeline (parseUploadedJson + toParsedData in upload.ts). */
 import { getAll, getMeta } from "./idb.js";
-import type { CollectionRawData } from "./collection-registry.js";
-import type { IdbTransaction } from "./entities/transaction.js";
-import { transactionToRawJson } from "./entities/transaction.js";
-import type { IdbStatement } from "./entities/statement.js";
-import { statementToRawJson } from "./entities/statement.js";
-import type { IdbStatementItem } from "./entities/statement-item.js";
-import { statementItemToRawJson } from "./entities/statement-item.js";
-import type { IdbReconciliationNote } from "./entities/reconciliation-note.js";
-import { reconciliationNoteToRawJson } from "./entities/reconciliation-note.js";
-import type { IdbAccount } from "./entities/account.js";
-import { accountToRawJson } from "./entities/account.js";
-import type { IdbJournalEntry } from "./entities/journal-entry.js";
-import { journalEntryToRawJson } from "./entities/journal-entry.js";
-import type { IdbJournalLeg } from "./entities/journal-leg.js";
-import { journalLegToRawJson } from "./entities/journal-leg.js";
-import type { IdbReconciliationEvent } from "./entities/reconciliation-event.js";
-import { reconciliationEventToRawJson } from "./entities/reconciliation-event.js";
-import type { IdbBudget } from "./entities/budget.js";
-import { budgetToRawJson } from "./entities/budget.js";
-import type { IdbBudgetPeriod } from "./entities/budget-period.js";
-import { budgetPeriodToRawJson } from "./entities/budget-period.js";
-import type { IdbRule } from "./entities/rule.js";
-import { ruleToRawJson } from "./entities/rule.js";
-import type { IdbNormalizationRule } from "./entities/normalization-rule.js";
-import { normalizationRuleToRawJson } from "./entities/normalization-rule.js";
-import type { IdbWeeklyAggregate } from "./entities/weekly-aggregate.js";
-import { weeklyAggregateToRawJson } from "./entities/weekly-aggregate.js";
+import { collectionRegistry } from "./collection-registry.js";
+import type { CollectionRawData, IdbOf, RawOf, DataStoreName } from "./collection-registry.js";
+
+async function rawCollection<K extends DataStoreName>(name: K): Promise<RawOf<K>[]> {
+  const records = await getAll<IdbOf<K>>(name);
+  const toRawJson = collectionRegistry[name].toRawJson as unknown as (idb: IdbOf<K>) => RawOf<K>;
+  return records.map(toRawJson);
+}
 
 export async function exportToJson(): Promise<string> {
-  const [transactions, budgets, budgetPeriods, rules, normalizationRules, statements, statementItems, reconciliationNotes, accounts, journalEntries, journalLegs, reconciliationEvents, weeklyAggregates, meta] = await Promise.all([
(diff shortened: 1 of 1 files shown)


natb1:
# QA Summary — Issue #1697 / PR #1984

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | User exports full budget — round-trip tests | shell-command | ✅ PASS |
| 2 | Export wiring type-checks against registry types | shell-command | ✅ PASS |
| 3 | export.ts drives Promise.all from collectionRegistry (AC1) | shell-command | ✅ PASS |
| 4 | Serialization uses registry[name].toRawJson, not individual imports (AC2) | shell-command | ✅ PASS |
| 5 | 13 per-entity type and *ToRawJson imports removed (AC3) | shell-command | ✅ PASS |
| 6 | Adding a registry entry requires no export.ts edit (AC6) | shell-command | ✅ PASS |

**Counts: 6 PASS / 0 FAIL / 0 SKIP**

## Detail

1. `npx vitest run --root budget export.test` → 16/16 tests passed (round-trip + field-mapping suite).
2. `npm run build --prefix budget` → clean build, no tsc errors.
3. `collectionRegistry` imported from `./collection-registry.js`; `Object.keys(collectionRegistry)` feeds the `Promise.all` fan-out; no hand-written store-name array.
4. Only `toRawJson` reference is `collectionRegistry[name].toRawJson` inside `rawCollection`; no per-entity `*ToRawJson` calls.
5. Zero matches for `from "./entities/` or `import type { Idb` in `export.ts` — all 26 per-entity imports removed.
6. Zero matches for any of the 13 store-name string literals in `export.ts` — completeness is by-construction via registry iteration.

No browser walkthrough was needed: the change is a pure TypeScript refactor with no UI changes; all acceptance criteria are machine-verifiable via vitest and tsc.

natb1:
## Review

Surface: `code` (app_or_rules=true). No CodeQL alerts. No dependency changes. All findings reviewed.

### Dismissed

- **CR-1** `budget/src/export.ts:18` — Missing null/empty check on `records` from `getAll`: `getAll` returns an array (never null/undefined by the IDB wrapper contract); a null guard here would be defensive dead code. _Dismissed: invariant from the calling layer eliminates this case._
- **CR-2** `budget/src/export.ts:19` — Type assertion `as unknown as (idb: IdbOf<K>) => RawOf<K>`: The PR body documents exactly why this cast is necessary (TypeScript's correlated-union limitation; the tsc build is clean). _Dismissed: acknowledged in the PR as the correct trade-off for this scope._
- **R-1** `budget/src/export.ts:19` — Unsafe type cast in `rawCollection` helper: Same as CR-2 — a structural fix would require retyping the registry, which is explicitly out of scope. _Dismissed: same rationale as CR-2._
- **R-2** `budget/src/export.ts:25` — `Object.fromEntries(pairs) as CollectionRawData` cast may miss runtime type safety: The `as CollectionRawData` cast is load-bearing for the output object type; runtime correctness is guaranteed by the registry-driven iteration (every key is a known `DataStoreName`) and confirmed by the 16 passing tests. _Dismissed: runtime correctness is by-construction, not by cast._

### False-positive (security)

- **SEC-1** `budget/src/export.ts:25` — Prototype pollution via `Object.fromEntries` with dynamic keys from `collectionRegistry`: Registry keys are hardcoded compile-time constants (`DataStoreName`); no user-controlled input reaches `Object.fromEntries`. Not a vulnerability.
- **SEC-2** `budget/src/export.ts:19` — `as unknown as` cast could mask type errors allowing malformed data into export: TypeScript type erasure means the cast has no runtime security impact; the underlying `toRawJson` adaptor is the same function that was previously called explicitly, and the test suite validates output fidelity.

### Fixed

_None._

### Informational

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #1686, opened 2026-06-15
budget: remove dead Firestore CRUD layer, enforce override-ordering at write boundary

natb1:
Closes #1269

## What

`budget/src/firestore.ts` carried a full Firestore CRUD layer — ~24 functions
(`getTransactions`, `updateTransaction`, `upsertReconciliationNote`,
`createJournalEntry`, …) plus internal helpers — that **no application code
calls**. `main.ts` builds only `SeedDataSource` / `IdbDataSource` (it has no
`firestore` import); the live mutation path is `IdbDataSource` /
`FileSyncingDataSource` in `data-source.ts`. This PR deletes that dead layer and
enforces the one piece of write-validation that was still missing at the live
write boundary.

## Changes

**Unit 1 — override-ordering at the write boundary.** Added
`validateBudgetOverrideOrdering` to `entities/budget.ts` and call it as the first
statement of `IdbDataSource.updateBudgetOverrides`, before any `get`/`put`. Until
now the live path wrote overrides without a sort check, so the app could persist
a budget state the read path (`requireOverrides`, `entities/budget.ts:130-133`)
would later refuse to load. The new check uses `<=` (strictly ascending, rejects
duplicate dates), matching the read-path invariant. A new
`data-source.test.ts` case asserts an out-of-order update throws `RangeError` and
does not persist — modeled on the existing reimbursement-range test.

**Unit 2 — delete the dead CRUD layer.** Reduced `firestore.ts` from 494 lines
to 50, keeping only the live surface:
- `assertLegStateTransition` — the only live *value* export (called by
  `IdbDataSource.updateJournalLegCleared`).
- The `SerializedBudgetPeriod` interface.
- The `export type { … }` re-exports (entity types, enum *type* re-exports,
  `GroupId`) that the app imports across `data-source.ts`, `balance.ts`,
  `pages/*`, and tests.

Everything else — all CRUD functions, internal helpers, the `firebase/firestore`
value imports, and the enum **value** re-exports — is deleted.
`test/firestore.test.ts` is pruned in the same commit to the surviving
`assertLegStateTransition` block; its dead-function tests are removed (the
reimbursement write-validation they covered now lives in `data-source.test.ts`).

## Notes for review

- **Reimbursement-range validation is already at the write boundary** — this PR
  does not touch it. `IdbDataSource.updateTransaction` already calls
  `validateReimbursementRange` (moved there by PR #1345), tested at
  `data-source.test.ts:116-124`. Override-ordering was the only remaining gap.
[...]

PR #1985, opened 2026-06-19
budget: unit-test validateBudgetOverrideOrdering edge cases

natb1:
Closes #1702

## What

Adds a dedicated unit-test file, `budget/test/budget-entity.test.ts`, pinning
the loop-boundary invariants of `validateBudgetOverrideOrdering`
(`budget/src/entities/budget.ts:153-161`). PR #1686 introduced the function but
its only test covered the reversed-order case (via the IDB integration path in
`data-source.test.ts`); the trivial and duplicate-timestamp boundaries were
unasserted.

## Why

The guard `overrides[i].date.toMillis() <= overrides[i-1].date.toMillis()` makes
two semantic boundaries load-bearing but untested:

- The `<=` (rather than `<`) makes **equal** timestamps invalid.
- The loop starting at `i = 1` makes **empty** and **single-element** inputs
  trivially valid.

A future change to the guard operator or the loop bound could silently regress
either. These unit-level tests call `validateBudgetOverrideOrdering` directly
(no IDB harness) and lock the behavior in place.

## Tests added

In `describe("validateBudgetOverrideOrdering")`:

- empty array → does not throw
- single-element array → does not throw
- two overrides with equal timestamps → throws `RangeError`
- two overrides in strictly ascending order → does not throw (valid baseline)

The file mocks `firebase/firestore` with the same local `MockTimestamp` used by
`statement-entity.test.ts` / `data-source.test.ts`, so it has no real Firestore
dependency.

## Scope

Test-only. No production code changed; no edits to `data-source.test.ts`.

## Verification

`npx vitest run --root budget` — full budget suite passes (926 tests), including
the 4 new cases, with no regressions.

### DIFF
--- a/budget/test/budget-entity.test.ts
+++ b/budget/test/budget-entity.test.ts
@@ -42,11 +42,46 @@ describe("validateBudgetOverrideOrdering", () => {
     expect(() => validateBudgetOverrideOrdering(overrides)).toThrow(RangeError);
   });
 
+  it("throws RangeError for two overrides in strictly descending order", () => {
+    const overrides: BudgetOverride[] = [
+      { date: Timestamp.fromMillis(2000), balance: 100 },
+      { date: Timestamp.fromMillis(1000), balance: 200 },
+    ];
+    expect(() => validateBudgetOverrideOrdering(overrides)).toThrow(RangeError);
+  });
+
   it("does not throw for two overrides in strictly ascending order", () => {
     const overrides: BudgetOverride[] = [
       { date: Timestamp.fromMillis(1000), balance: 100 },
       { date: Timestamp.fromMillis(2000), balance: 200 },
     ];
     expect(() => validateBudgetOverrideOrdering(overrides)).not.toThrow();
+  });
+
+  it("does not throw for three overrides in strictly ascending order", () => {
+    const overrides: BudgetOverride[] = [
+      { date: Timestamp.fromMillis(1000), balance: 100 },
+      { date: Timestamp.fromMillis(2000), balance: 200 },
+      { date: Timestamp.fromMillis(3000), balance: 300 },
+    ];
+    expect(() => validateBudgetOverrideOrdering(overrides)).not.toThrow();
+  });
+
+  it("throws RangeError when a three-element array violates ordering at index 2 (descending)", () => {
+    const overrides: BudgetOverride[] = [
+      { date: Timestamp.fromMillis(1000), balance: 100 },
+      { date: Timestamp.fromMillis(2000), balance: 200 },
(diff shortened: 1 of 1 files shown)


natb1:
# QA Summary — Issue #1702

**PR:** #1985
**Branch:** `1702-budget-add-unit-tests-for-v`
**Result:** ✅ Clean pass — all items passed

## Items executed

All 5 items ran in the shell-command lane (script-verifiable, no browser).

| # | Title | Result |
|---|-------|--------|
| 1 | Empty-array case does not throw | PASS |
| 2 | Single-element array does not throw | PASS |
| 3 | Equal timestamps throw RangeError | PASS |
| 4 | Tests live in budget-entity.test.ts, not the IDB harness | PASS |
| 5 | Full budget suite still passes | PASS |

## Detail

- Items 1–3: filtered vitest runs confirmed each new `validateBudgetOverrideOrdering` test passes individually.
- Item 4: file-existence + grep check confirms tests are in `budget/test/budget-entity.test.ts` and `validateBudgetOverrideOrdering` does not appear in `budget/test/data-source.test.ts`.
- Item 5: full suite — **926 tests passed, 7 skipped, 0 failed** across 50 test files (2 skipped). No regressions.

No browser walkthrough was needed — this is a test-only PR with no UI changes.

natb1:
## Review: #1702 — `validateBudgetOverrideOrdering` unit tests

### Fixed

- **`budget/test/budget-entity.test.ts:37-43`** — The equal-timestamp test only exercised the equality half of the `<=` guard. Added a strictly-descending two-element test (`fromMillis(2000)` → `fromMillis(1000)`) to pin the full `<=` semantics. _(commit 082d9830)_
- **`budget/test/budget-entity.test.ts:25-52`** — All error and valid tests used 2-element arrays; the loop at `i=2` was never exercised. Added three 3-element tests: valid strictly-ascending `[1000, 2000, 3000]` (does not throw), descending violation at index 2 `[1000, 2000, 1500]` (throws `RangeError`), and equal-timestamp violation at index 2 `[1000, 2000, 2000]` (throws `RangeError`). _(commit 082d9830)_

### Informational

- **`budget/test/budget-entity.test.ts:42`** — The `RangeError` assertion uses `toThrow(RangeError)` and does not assert the error message. The production function throws `RangeError("budget overrides must be in strictly ascending date order")`. Asserting the message would make the test more precise, but this is not a correctness issue in the current scope.

### Deferred

- **`budget/test/budget-entity.test.ts:3-17`** — `MockTimestamp` is duplicated verbatim from `statement-entity.test.ts`. A future change to one file and not the other could silently diverge from the Firestore `Timestamp` contract. Tracked as follow-up #1989 (blocked by #1702).

### Security

PR #1986, opened 2026-06-19
budget: validate override ordering on read in idbToBudget

natb1:
Closes #1703

## Problem

`idbToBudget` (`budget/src/entities/budget.ts`) maps a stored `IdbBudget` row to
a domain `Budget` but did not re-validate that `overrides` are sorted by date
ascending. The domain interface documents this invariant —
`findOverrideInPeriod` assumes the ordering — and PR #1686 added
`validateBudgetOverrideOrdering` at the *write* boundary, but a row written
before that PR could still hold out-of-order overrides. On read, `idbToBudget`
silently produced a malformed `Budget` that corrupts period lookups instead of
surfacing a clear error.

This left the IDB budget read path inconsistent with the transaction read path:
`idbToTransaction` already calls `validateReimbursementRange(row.reimbursement)`
at the top of the function to guard against stale, pre-validator rows.

## Change

Add the missing read-path guard, mirroring the `idbToTransaction` placement:

- Extract the override-ordering check that lived inline in `requireOverrides`
  into a shared private helper `assertOverridesSortedAscending`, so the read-path
  invariant has a single source of truth.
- Call that helper from `idbToBudget` after mapping the overrides, so
  `getBudgets()` throws a clear `DataIntegrityError` on a stale out-of-order row
  instead of returning a malformed `Budget`.

The read guard throws `DataIntegrityError` — the established read-path convention
in `budget.ts` (`requireEnum`, `requireTimestamp`, `requireOverrides` all do).
The write-boundary validator `validateBudgetOverrideOrdering` (which throws
`RangeError`) is deliberately left untouched; the write(`RangeError`) /
read(`DataIntegrityError`) split is intentional and matches the existing
`requireOverrides` read parser.

## Test

Adds `getBudgets throws for a pre-existing out-of-order overrides record` in
`budget/test/data-source.test.ts`, analogous to the existing stale-reimbursement
read test. The full `budget` suite passes.

### DIFF
--- a/budget/src/entities/budget.ts
+++ b/budget/src/entities/budget.ts
@@ -112,6 +112,14 @@ function requireAllowancePeriod(value: unknown): AllowancePeriod {
   return requireEnum(value, ALLOWANCE_PERIODS, "allowancePeriod");
 }
 
+function assertOverridesSortedAscending(overrides: readonly BudgetOverride[]): void {
+  for (let i = 1; i < overrides.length; i++) {
+    if (overrides[i].date.toMillis() <= overrides[i - 1].date.toMillis()) {
+      throw new DataIntegrityError(`overrides not sorted by date ascending at index ${i}`);
+    }
+  }
+}
+
 function requireOverrides(value: unknown): BudgetOverride[] {
   if (value == null) return [];
   if (!Array.isArray(value)) {
@@ -127,11 +135,7 @@ function requireOverrides(value: unknown): BudgetOverride[] {
     const balance = requireNumber(entry.balance, `overrides[${i}].balance`);
     result.push({ date, balance });
   }
-  for (let i = 1; i < result.length; i++) {
-    if (result[i].date.toMillis() <= result[i - 1].date.toMillis()) {
-      throw new DataIntegrityError(`overrides not sorted by date ascending at index ${i}`);
-    }
-  }
+  assertOverridesSortedAscending(result);
   return result;
 }
 
@@ -207,16 +211,18 @@ export function budgetToIdbRecord(b: Budget): IdbBudget {
 // ── IdbBudget → Budget ────────────────────────────────────────────────────────
 
 export function idbToBudget(row: IdbBudget): Budget {
+  const overrides = (row.overrides ?? []).map(o => ({
+    date: msToTs(o.dateMs) as Timestamp,
+    balance: o.balance,
+  }));
+  assertOverridesSortedAscending(overrides);
   return {
     id: row.id as BudgetId,
     name: row.name,
     allowance: row.allowance,
     allowancePeriod: requireAllowancePeriod(row.allowancePeriod),
     rollover: requireEnum(row.rollover, ROLLOVERS, "rollover"),
-    overrides: (row.overrides ?? []).map(o => ({
-      date: msToTs(o.dateMs) as Timestamp,
-      balance: o.balance,
-    })),
+    overrides,
     groupId: null as GroupId | null,
   };
 }
--- a/budget/src/entities/budget.ts
+++ b/budget/src/entities/budget.ts
@@ -211,10 +211,15 @@ export function budgetToIdbRecord(b: Budget): IdbBudget {
 // ── IdbBudget → Budget ────────────────────────────────────────────────────────
 
 export function idbToBudget(row: IdbBudget): Budget {
-  const overrides = (row.overrides ?? []).map(o => ({
-    date: msToTs(o.dateMs) as Timestamp,
-    balance: o.balance,
-  }));
+  const overrides = (row.overrides ?? []).map(o => {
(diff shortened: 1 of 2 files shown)


natb1:
## QA Pass — Issue #1703 (PR #1986)

All acceptance criteria verified via the budget vitest suite.

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Override-ordering re-validation on read throws DataIntegrityError | ✅ PASS |
| 2 | No regression in existing data-source tests | ✅ PASS |
| 3 | Golden path — well-ordered budgets still load | ✅ PASS |

**Summary:** 3 passed, 0 failed, 0 skipped

### Vitest output

```
Test Files  49 passed | 2 skipped (51)
      Tests  923 passed | 7 skipped (930)
   Duration  3.58s
```

The new test `getBudgets throws for a pre-existing out-of-order overrides record` passes — `getBudgets()` throws `DataIntegrityError` on a stale out-of-order overrides row as required. The helper extraction in `requireOverrides` introduced no regression in any existing data-source or parse-path test. No browser walkthrough needed — this is a pure TypeScript/logic change (entity + test files only, no UI components touched).

natb1:
## Review — #1703

### Required (security) — Fixed

- **red-team-0** `budget/src/entities/budget.ts:115-121` — `assertOverridesSortedAscending` used a `<=` comparison on `toMillis()` results, but `NaN <= NaN` is `false`; an IDB record with a NaN `dateMs` would bypass the sort guard and silently corrupt period lookups in `findOverrideInPeriod`. Added a `Number.isFinite` check in `idbToBudget` before `msToTs` runs, throwing `DataIntegrityError` on non-finite `dateMs` — closing the bypass. Fix commit: 9ea404d4.

- **input-validation-0** `budget/src/entities/budget.ts:214-216` — alleged missing `requireNumber` validation on `o.dateMs`/`o.balance` in `idbToBudget`. Refuted: every IDB write path validates for finite numbers before storing (upload path via `parseISOTimestamp`/`requireNumber`; UI edit path via explicit `Number.isFinite` guard in `collectOverridesForBudget`). The NaN-bypass mechanism is real but not exploitable via any reachable write path; this is a robustness nit, not a security vulnerability. (2/2 skeptics: refuted)

- **input-validation-1** `budget/src/entities/budget.ts:215` — alleged unsafe `as Timestamp` cast where `msToTs` can return `null`. Refuted: `o.dateMs` is typed `number`, no IDB write path stores `null` for `dateMs`, and the stale-data scenario the PR addresses was an ordering bug — not a nullability bug. The cast is imprecise but not a correctness gap in practice. (1/1 skeptics: refuted)

- **red-team-1** `budget/src/entities/budget.ts:213-228` — alleged missing `requireNumber(o.balance, ...)` validation on read, unlike the Firestore path. Refuted on two grounds: (1) this asymmetry pre-existed the PR — the diff did not introduce it; (2) all application write paths already validate `balance` for finite numbers before storing. No exploitable path through a same-origin, single-user IDB store. (2/2 skeptics: refuted)

### Informational

- **code-review-1** `budget/src/entities/budget.ts:115-121` — `assertOverridesSortedAscending` (private, read-path, throws `DataIntegrityError`) and `validateBudgetOverrideOrdering` (exported, write-path, throws `RangeError`) serve different invariants and error conventions; the naming parallel is intentional but could be clarified with a comment.

- **code-review-2** `budget/test/data-source.test.ts:150-158` — the new test covers stale out-of-order overrides (dateMs 2000 then 1000) but not the equal-timestamp rejection (`<=` guard). An additional case with equal `dateMs` values would give full branch coverage.

- **review-0** `budget/src/entities/budget.ts:142` — `requireRawOverrides` still contains the ascending-order check inline (loop at line ~147) instead of delegating to `assertOverridesSortedAscending`; the extraction is partial.

- **review-1** `budget/src/entities/budget.ts:157` — `validateBudgetOverrideOrdering` duplicates the same ascending-order loop a third time (throwing `RangeError`). The write-boundary / read-boundary error-type split is intentional, but the loop logic itself is now triplicated.

- **review-2** `budget/test/data-source.test.ts:150` — the equal-timestamp (`<=`) rejection path is not exercised; seeds with identical `dateMs` values would cover the `<= 0` branch.

### False-positive (security)

- **red-team-3** `budget/src/entities/budget.ts:118` — claimed the error message leaks a zero-based array index. Not a security concern: this is a local-first, single-user PWA with no server-side log aggregation; the index aids local debugging and crosses no trust boundary.

### Out-of-scope (security)

- **red-team-2** `budget/src/entities/budget.ts:218` — the new `assertOverridesSortedAscending` call converts stale pre-#1686 IDB rows from silent corruption into a permanent `DataIntegrityError` read failure. Noted as a UX concern (no recovery path beyond devtools); out of scope for this PR's guard-addition change.

PR #1992, opened 2026-06-19
budget: migrate inline MockTimestamp test mocks onto shared timestampMockFactory

natb1:
Closes #1989

## Summary

`MockTimestamp` — the stand-in for the Firestore `Timestamp` used in `vi.mock`
factories — was defined inline and verbatim in 6 budget test files. Each copy was
a divergence hazard: update one and not the others and the mock silently drifts
from the production `Timestamp` contract.

This migrates all 6 onto the shared `timestampMockFactory()` already exported from
`budget/test/helpers.ts`, which is the established pattern in ~20 other budget
tests. Every budget test now obtains the firestore mock from one place.

## Changes

In each of the 6 files, the inline
`vi.mock("firebase/firestore", () => { class MockTimestamp {…} … })` block is
replaced with the one-liner `vi.mock("firebase/firestore", () => timestampMockFactory())`
and `timestampMockFactory` is added to the `helpers` import:

- `budget/test/budget-entity.test.ts`
- `budget/test/statement-entity.test.ts`
- `budget/test/upload.test.ts`
- `budget/test/export.test.ts`
- `budget/test/data-source.test.ts` — its existing `./helpers` import (for
  `makeParsedData`) was hoisted above the `vi.mock` calls and merged, so the
  hoisted factory can reference the imported binding.
- `budget/test/prerender.test.ts` — its inline class was a superset (extra
  `toDate`/`fromDate`); the shared factory already carries that superset.

Each file's other `vi.mock(...)` calls are left untouched.

## Why reuse `helpers.ts` rather than a new `_mocks/` module

The issue suggests (`e.g.`) a new `budget/test/_mocks/firebase-firestore.ts`.
Creating a second mock home would fragment the firestore mock across two files —
re-introducing the exact divergence this change eliminates. Reusing the existing
factory yields a single source of truth and is the reuse-first choice.

## Why the Date-based shape is a safe drop-in

The deleted inline class is seconds/nanoseconds-shaped; `helpers.ts`'s
`MockTimestamp` is Date-shaped. Safe because the 6 files only consume
`Timestamp.fromMillis(ms)`, `.toMillis()`, and (prerender) `.toDate()` — none
call `new Timestamp(seconds, nanoseconds)` or read `.seconds`/`.nanoseconds`, and
no production code under `budget/src/` reads them either. For integer ms,
`fromMillis(ms).toMillis()` round-trips identically.

## Verification

- `grep -rn "class MockTimestamp" budget/test/` → single hit (`helpers.ts`).
- `grep -rln 'vi.mock("firebase/firestore", () => {' budget/test/` → no hits.
- `npx vitest run --root budget` → 931 tests pass, no regressions (exercises both
  the 6 migrated files and the ~20 existing `timestampMockFactory` consumers).


natb1:
# QA Summary — #1989 (PR #1992)

## Result: PASS ✓

All acceptance criteria verified. No regressions.

## Items executed

| # | Title | Result |
|---|-------|--------|
| 1 | Shared helper module exports the firestore mock factory | PASS |
| 2 | No inline MockTimestamp definition remains anywhere in budget/test/ | PASS |
| 3 | The two named test files import the shared factory instead of defining the mock inline | PASS |
| 4 | Budget unit test suite passes with no regressions | PASS |

## Details

**Item 1** — `helpers.ts` has `class MockTimestamp` at line 7 (private) and `export function timestampMockFactory()` at line 16. The factory is the consumption interface; the class stays private to helpers. ✓

**Item 2** — `grep -rn "class MockTimestamp" budget/test/` returns exactly one hit: `budget/test/helpers.ts:7`. No test file defines `MockTimestamp` inline. ✓

**Item 3** — Both `statement-entity.test.ts` and `budget-entity.test.ts` import `timestampMockFactory` from `./helpers` and use `vi.mock("firebase/firestore", () => timestampMockFactory())`. No inline class remains. ✓

**Item 4** — `npx vitest run --root budget` passed 949 tests (7 skipped) across 55 test files (2 skipped). The `blob:` URL NotSupportedError in the output is a pre-existing happy-dom issue unrelated to this change. ✓

No browser walkthrough was needed — all acceptance criteria were verifiable with shell commands and the vitest suite.

natb1:
## Review — PR #1992 (#1989)

---

### Fixed

- **RF-001** `budget/test/data-source.test.ts:3` — Merge import lines: `timestampMockFactory` and `makeParsedData` are already combined into a single `import { timestampMockFactory, makeParsedData } from './helpers'` in the PR; no new commit needed.

---

### Informational

- **RF-002** `budget/test/prerender.test.ts` — The inline `MockTimestamp` in this file included `toDate()` and `fromDate()` methods not present in the other 5 files. Verify the shared `timestampMockFactory()` in `budget/test/helpers.ts` provides both methods. (PR body notes it does; this is a cross-check reminder.)
- **RF-003** `budget/test/` — The deleted inline classes used a seconds/nanoseconds constructor shape; `helpers.ts` uses a Date-based shape. No test guards against a future drift in round-trip behaviour. Consider adding a round-trip assertion to helpers.ts.

---

### Deferred

- **RF-004** `budget/test/helpers.ts` — `MockTimestamp` is now the single point of trust for 6 test files, but no contract test asserts its public API surface (`fromMillis`, `toMillis`, `toDate`, `fromDate`) matches the Firestore `Timestamp` contract. A future rename or removal in `helpers.ts` would silently break all consumers. → Filed as #2128 (blocked by #1989).

PR #2016, opened 2026-06-19
office-hours: select only idle sessions (waiting/done/idle), skip working

natb1:
Closes #2011

## What

The office-hours queue worker should only surface a session the human can
productively attend to — one that is **idle** (`state` waiting/idle) or
**complete** (`done`), never one that is **actively working**. Previously
`issue_live_session_id` matched the first live session of *any* status and
emitted it as `live` → attach, with two defects:

1. **Attached to working sessions.** A `state:"working"` (`status:"busy"`)
   session was attached to the human, interrupting active autonomous work.
2. **Blind to completed sessions.** `claude agents --json` is active-only by
   default, so a `done` session awaiting the human was invisible to both the
   selector's liveness query and the entry script's attach-time
   sessionId→job-id resolution.

This PR gates office-hours selection on the daemon's granular `state`
(`busy` ⟺ `state:"working"`), queries with `--all` so `done` is visible
**end-to-end** (selection *and* attach), and renames the now-misnamed `live`
disposition verb to `idle`.

## State → action contract

| `state` | action |
| --- | --- |
| `working` (busy) | **skip the issue** — being worked autonomously; don't attach, don't fresh-launch into its worktree |
| `waiting` / `done` / `idle` | **attach** (`done` requires `--all` to see) |
| `stopped` / `paused` / any unrecognized state | **exclude from attach** — skip the issue (fail-safe) |
| not in registry | sessionless → fresh-launch |

## Units

1. **`lib-claude-agents.sh`** — new `claude_sessions_with_name_all <name>` helper:
   runs `claude agents --json --all` **directly** (bypassing the
   `DISPATCH_AGENTS_SNAPSHOT` snapshot, which is captured without `--all` and
   lacks `done` rows), filters by exact `.name`, emits `sessionId<TAB>state`.
   Scoped to the office-hours selector only — the shared snapshot-backed helpers
   on the dispatch router hot path are untouched.
2. **`office-hours-select-target`** — `issue_live_session_id` classifies the
   matched session by `state`: attach (rc 0) for waiting/done/idle, skip (rc 3)
   for working/stopped/paused/unrecognized, UNKNOWN (rc 2), sessionless (rc 1).
   The enumeration loop `continue`s on both rc 2 and rc 3 so a working/stopped
   item never falls through to a fresh-launch into its occupied worktree. Verb
   renamed `live` → `idle`.
3. **`office-hours`** (entry script) — `attach_session` queries
   `claude agents --json --all` so a `done` session resolves to its job id
   end-to-end; dispatch case renamed `live|parked-router` → `idle|parked-router`.
4. **`office-hours/SKILL.md`** — disposition doc renamed `live` → `idle`.
5. **`test-dispatch-scripts.sh`** — new `--all`-faithful fake
   (`office_hours_state_fake_claude`) that returns a `done` row **only when
   `--all` is present in argv**, so a path that forgets `--all` fails the suite.
   Rewrote the OHST3*/OH1/OH2/OH3c/OH5 attach tests to the `idle` verb with
   attachable states, and added working-skip, done-attach (selector + entry
   end-to-end), and stopped-exclude cases. Full suite green (3175/3175).

## Notes for the reviewer

- **Stopped/paused open question — resolved, not escalated.** Excluded from
  attach (treated like working — skip the issue). Known consequence: such a
  session still name-matches in `worktree_has_live_session`, so its issue won't
  fresh-launch either and can wedge until the session is cleaned up. Handling
  that wedge is **out of scope** here — it would require changing the shared
  `worktree_has_live_session` / `claude_agents_list_all` helpers, which gate the
  dispatch router's worktree-occupancy decisions on the hot path; making `done`
  sessions newly "occupy" a worktree there would block phase-worker launches.
  This consequence is documented in the selector header.
- **`attach_session` `--all` addition (Unit 3).** Not enumerated in the issue's
  migration steps, but required for end-to-end `done` attach: without it the
  selector can *select* a `done` session but the entry script can't *resolve* its
  job id. The done-attach entry test (OH5b) is the end-to-end proof.
- **Assumption (not blocking):** the design treats `claude attach` on a finished
  (`done`) job as meaningful — the issue's contract treats `done` as an
  attachable "complete" state. Please confirm against real daemon behavior.


natb1:
# QA Summary — PR #2016 / Issue #2011

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Full dispatch shell test suite passes | shell-command | PASS (3175/3175) |
| 2 | Selector queries liveness with `--all` so done sessions are visible | shell-command | PASS |
| 3 | New `claude_sessions_with_name_all` helper exists and uses `--all` | shell-command | PASS |
| 4 | Disposition verb rename `live` → `idle` in selector and entry script | shell-command | PASS |
| 5 | Entry script `attach_session` queries `--all` | shell-command | PASS |
| 6 | SKILL.md disposition doc renamed `live` → `idle` | shell-command | PASS |
| 7 | Stopped/paused resolution is documented in the selector header | needs-human-judgment | (assessed by Opus) |
| 8 | State classification covers working-skip / done-attach / stopped-exclude | shell-command | PASS |

**Counts:** 7 PASS, 0 FAIL, 0 SKIP

No bugs found. No browser walkthrough (pure shell scripts change — no browser component detected).

---

## Disposition triage (Step 3.5)

One residue item (item 7, `needs-human-judgment`) was assessed by the disposition Workflow.

---

| # | Title | Rationale |
|---|-------|-----------|
| 7 | Stopped/paused resolution is documented in the selector header | The criterion is provably met by readable evidence. The header of `office-hours-select-target` (lines 29–41) contains the state→action table row `stopped / paused / any unrecognized state → EXCLUDE from attach -- skip the issue (fail-safe)`, plus an explanatory block stating they are treated like `working` (skip), naming the known consequence (the session still name-matches in `worktree_has_live_session` so its issue won't fresh-launch and can wedge), and explaining why handling that wedge is out of scope. The prose is present, accurate, and self-consistent — no code defect and no user intent required, only positive confirmation. |

All residue items dropped as PASS. Escalation set is empty — clean pass.

natb1:
## Review: PR #2016

### Fixed

- **code-review-1 / review-0** — `.claude/skills/dispatch-propagate/scripts/office-hours-select-target:173–193`
  Duplicate `sid` and `state` subshell pipelines (`cut -f1` / `cut -f2`) replaced with a single `IFS=$'\t' read -r sid state` pass; eliminates two unnecessary forks per session row. Separately, an OHST3h test was added to guard the continue-past-skip branch: a single issue with a `working` basename session and a `waiting` `office-hours-<N>` session must return `idle s-office-hours-<N>`, proving the loop doesn't stop at the first non-attachable row.
  Fix commit: `36fbe61c`

### Informational

- **code-review-0** — `.claude/skills/dispatch-propagate/scripts/lib-claude-agents.sh:326`
  `claude_sessions_with_name_all` projects `.state` from `claude agents --json --all` but the field's set of possible values isn't formally documented in the function's header comment. The code relies on whichever strings the daemon emits (`working`, `waiting`, `done`, `idle`, `stopped`, `paused`); a daemon version that renames or adds states would silently fall into the `*` catch-all. No code change required — the catch-all is intentionally conservative (fail-safe to `saw_skip`).

- **code-review-2** — `.claude/skills/dispatch-propagate/scripts/office-hours-select-target:194`
  `(( saw_skip )) && return 3` fires before `(( saw_unknown )) && return 2`, encoding a skip > unknown precedence. This means a worktree with one working session and one unqueryable session returns 3 (skip/no-fresh-launch) rather than 2 (unknown/no-fresh-launch). Both outcomes agree on the visible effect (issue is not fresh-launched), but the priority ordering isn't commented. No change required; the behavior is correct and consistent with the fail-safe intent.

- **review-1** — `.claude/skills/dispatch-propagate/scripts/office-hours:86–99`
  `attach_session` calls `exec claude attach "$job_id"` for the `idle` verb, including when the selected session state is `done`. Whether `claude attach` on a finished job is meaningful (e.g., shows output, exits cleanly) or silently no-ops is a daemon contract question noted in the PR itself as an open assumption. No code change; flagged for confirmation against real daemon behavior.

### Dismissed

- **review-2** — `.claude/skills/office-hours/SKILL.md:77`
  Line 77 was not re-wrapped when the `live` → `idle` text was updated; the added prose makes the line slightly long. Not a functional issue in a Markdown doc; no fix warranted.

### False-positive

- **input-validation-0** — `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:2873`
  Claim: the `office_hours_state_fake_claude` helper builds the JSON payload by bash string concatenation and could inject malformed JSON via `name` or `state` values. In test context the helper is only ever called with literal hardcoded strings (`"42-x:working"`, `"99-y:idle"`, etc.) — no user-supplied input reaches it. Not a real injection vector.

- **input-validation-1** — `.claude/skills/dispatch-propagate/scripts/office-hours-select-target:221`
  Claim: the sessionId emitted in `printf 'idle %s\n' "$IDLE_ID"` could carry IFS-splitting characters. `IDLE_ID` is sourced from jq `@tsv`-encoded output passed through `head -n 1 | cut -f1` (or with the fix, `IFS=$'\t' read -r sid state`); the value is already the first tab-delimited field, so any embedded tab would have been truncated. Not exploitable.

- **red-team-1** — `.claude/skills/dispatch-propagate/scripts/office-hours:86–99`
  TOCTOU: selector picks an `idle` session; session transitions to `working` between the selector's query and `attach_session`'s job-id lookup. The worst case is that the human attaches to a now-working session — the same outcome the original code always produced. The window is narrow and the consequence is not a security issue; fail-safe behavior would require a second state check, which is out of scope.

### Out-of-scope

- **red-team-0** — `.claude/skills/dispatch-propagate/scripts/office-hours-select-target:194–195`
  A stopped or paused session in a worktree permanently blocks that issue from the queue until the session is cleaned up. Documented in the selector header and called out in the PR notes as a known consequence out of scope for this change.

- **red-team-2** — `.claude/skills/dispatch-propagate/scripts/office-hours-select-target:172–174`
  When `claude_sessions_with_name_all` returns multiple rows for a single session name, `head -n 1` picks the first. In the daemon's current model a `name` field is not guaranteed unique. This is a pre-existing property of the name-based lookup strategy shared with `worktree_has_live_session`; changing it is out of scope here.

PR #2019, opened 2026-06-19
budget: migrate the budgets section (Sankey + charts) to React

natb1:
Closes #1876

Migrates the budget app's budgets section from vanilla TypeScript to React
composing `@commons-systems/ds`, following the migration pattern established by
#1874 and the `Transactions.tsx` / `CategorySankey.tsx` siblings (parent epic
#1873). The chart geometry stays as app-specific vanilla layout wrapped in
imperative islands; only the bridge, composition, and color/spacing literals
change.

## What changed, by unit

**Unit 1 — Tokenize the chart family + budgets CSS.** All chart color literals
now resolve from design-system tokens at render time. A shared
`chartSeriesColor(container, index)` helper (in `chart-util.ts`) returns
`--chart-${(index % 6) + 1}`, and every categorical chart draws its series from
it. The Sankey (`sankey-render.ts`) is indexed by its `topLevelIndex` walk, so it
draws `--chart-*` **in token order** — satisfying criterion 2's "Sankey in token
order" clause. The other categorical charts (area/pie/waterfall) each cycle the
same `1..6` palette in their own domain order; cross-chart color *identity* is not
attempted and is not required by the criteria — these charts render different
entities on different pages/populations (the Sankey is the transaction-category
tree on `/transactions`; area/pie are per-budget on `/`; waterfall is per-budget
variance), so there is no shared category domain to align. Sign-encoded balance
bars map to the semantic `--favorable`/`--unfavorable` (not chart tokens, to keep
the green/red parity); the trend chart's 4 fixed series map to `--chart-1..4`
(the cool-blue is gone); the balance dot maps to `--accent`. Budgets-section
spacing/radius literals in `theme.css` move to `--space-*`/`--radius-*` tokens;
chart geometry CSS is left intact.

**Unit 2 — React `Budgets.tsx` page + chart island.** A new `Budgets.tsx` accepts
`RenderPageOptions`, loads via a `useBudgetsData` hook (mirroring
`useTransactionsData`: parallel `dataSource.*` calls, `LoadState`, per-source
`logError` + rethrow, soft errors → inline state), and ports the legacy markup to
JSX at parity — every id/class the e2e/smoke specs query is preserved. The
bar/pie/area chart family is wrapped as a single imperative island over the
existing `hydrateBudgetChart` (now returning teardown), since that function
already orchestrates the three render fns plus the shared windowing/date controls.
Summary stats render as DS `Metric` cards. (There is no Sankey on the budgets page
itself — the budgets chart set is bar/pie/area/trend/waterfall.)

**Unit 3 — Table + overrides interactivity.** The blur-save / change-handler
logic moves into React hooks (`useBudgetTableInteractivity` /
`useOverridesTableInteractivity` in `use-budget-table.ts`), modeled on
`useTableInteractivity`: budget name/allowance/period/rollover →
`updateBudget`, overrides date/balance/add/delete → `updateBudgetOverrides`, with
optimistic delete + restore-on-failure. Select state is read via `.value` (the
`option[selected]` attribute is unreachable on React-rendered selects), seeding a
last-saved `WeakMap` at attach so the unchanged-guard works. `applyRollingAverage`
and the pure transforms stay in `budgets-hydrate.ts`, un-duplicated.

**Unit 4 — Route flip + legacy removal.** `App.tsx` now serves `/` (and the
unknown-path fallback) as the React `Budgets` page keyed by `BUDGETS_PATH:navEpoch`,
with `KNOWN_PATHS` rebuilt to preserve nav order. The legacy `path: "/"` →
`renderBudgets` entry and its three hydration specs are removed. Dead symbols
(`renderBudgets`, the legacy `hydrateBudgetTable`/`hydrateOverridesTable`) are
deleted after grep-confirming zero callers; `renderBudgetsContent` is kept (still
the SSG prerender renderer) and `hydrateBudgetChart` is kept (the Unit-2 island
calls it).

## Verification

- `npm test --prefix budget` (vitest): green. The only failures are the 3
  pre-existing `prerender.smoke.test.ts` cases reading a stale `dist/index.html` in
  the sandbox — they fail on `main` too and are unrelated to this change.
- `npx tsc --noEmit -p budget/tsconfig.json`: clean.
- `npm run lint --prefix budget`: clean.
- e2e (`budget/e2e/budgets*.spec.ts`) and smoke selectors were confirmed by grep
  to be still produced by `Budgets.tsx` and the kept chart render fns; Playwright
  e2e is authoritative in CI.

## Notes

- The SSG prerender path still uses the vanilla `renderBudgetsContent` string
  renderer (full React SSR is out of scope for this migration); the live runtime
  route is fully React.


natb1:
# QA Summary — #1876 budget: migrate the budgets section to React

**PR:** #2019 | **Branch:** `1876-budget-migrate-the-budgets`

## Results

| Item | Title | Result |
|------|-------|--------|
| 1 | Budgets page renders at parity in public mode | ✅ PASS |
| 2 | "Viewing example data" notice shows for unauthorized users | ✅ PASS |
| 3 | Full chart family renders (bar, pie, area/trend, waterfall) | ✅ PASS |
| 4 | Chart date-picker and area-chart weeks controls are interactive | ✅ PASS |
| 5 | Budget summary stats render as DS Metric cards | ✅ PASS (already-satisfied — see below) |
| 6 | Edit affordances are read-only in public mode | ✅ PASS |
| 7 | No budgets-section design-system CSS color literals outside packages/ds | ✅ PASS |
| 8 | Screenshot + functional parity vs production budget budgets is cosmetically nil | 🔶 DEFERRED (needs-human-judgment) |

**PASS: 7 | FAIL: 0 | SKIP: 0 | Deferred: 1**

## Bugs Found

None.

## Walkthrough Recording

`tmp/qa-fix-walkthrough-1876.gif`

## Detail

**Item 1 — Budgets page renders at parity:** Page loaded successfully at `http://localhost:5170/` in public/unauthorized mode. `<main id="app">` with `<h2>Budgets</h2>`, `#budgets-table`, `#budget-metrics` (3 Metric cards), `#budgets-area-chart`, `#budgets-pie`, `#budgets-chart`, `#overrides-table`, and `#budgets-chart-controls` all present. No `#budgets-error`. The app composes `@commons-systems/ds` Metric cards for summary stats (confirmed by `.cs-metric` count = 3).

**Item 3 — Full chart family:** All five chart types verified:
- **Area chart** (`#budgets-area-chart`): SVG present, 31 paths — Food, Housing, Vacation, Other series in warm `--chart-*` tokens
- **Bar chart** (`#budgets-chart`): SVG present, 36 rects — sign-encoded green (`--favorable`) / red (`--unfavorable`) balance bars, no cool-gray/blue
- **Pie chart** (`#budgets-pie`): SVG present, 4 arcs — Food 12.1%, Housing 9.3%, Vacation 8.1%, Not Budgeted 70.6%
- **Trend chart**: 2 SVGs inside `#budgets-chart` container (bar + aggregate trend rendered by `hydrateBudgetChart`)
- **Waterfall chart**: SVG rendered on expanding the Food row — categorical colors (`--chart-*`) + favorable/unfavorable for total bars
- No Sankey on this page (correct — Sankey lives on `/transactions`)

**Item 4 — Controls interactive:** Changed `#area-chart-weeks` from 3 to 6 via `input` event; area chart re-rendered (path count 31→29) with no console errors.

**Item 6 — Read-only in public mode:** All `.edit-name`, `.edit-allowance` inputs and `.edit-period` selects have `disabled` attribute. No `#add-override` button. No `.delete-override` buttons.

**Item 7 — No CSS color literals:** `grep -rniE '#[0-9a-f]{6}|rgba?|hsla?|steelblue|...'` across all changed budgets source files returned exit 1 (zero matches outside comments). ✅

## Note on Prerender Tests

6 Playwright acceptance tests in `prerender.spec.ts > static HTML output` failed because they read `dist/index.html` directly (requiring a production build), while the QA environment uses a Vite dev server. These tests are not tagged `@build` and are not excluded by the acceptance runner's filter. They fail identically on `main` in the same environment — pre-existing infrastructure issue, not a regression. CI is passing (`dist/index.html` is built there). All 9 `runtime behavior` prerender tests passed, including "seed view renders without Firestore network requests," "hydrated nav links are functional," and "upload transitions from seed to IDB source."

## Disposition Triage

| 8 | `needs-human` (aesthetic) | n/a | Visual "does it look right" parity against production — requires a human eye. |

`needs-main` items are filed as `blocked_by` follow-ups and dropped from the escalation set, so they do not park this PR. *(No `needs-main` items this pass.)*

**Non-fixing pass:** item 8 (needs-human, aesthetic) escalates to office-hours. No `opus-fixable` items existed, so no auto-fix lane ran.

| Item | Rationale |
|------|-----------|
| 5 — Budget summary stats render as DS Metric cards | Provably satisfied: `Budgets.tsx` imports `Metric` from `@commons-systems/ds` and renders the three summary stats (`12-Week Avg Weekly Credits`, `Total Weekly Budget`, `12-Week Avg Weekly Spending`) as `<Metric label value>` with `formatCurrency`, replacing the legacy `<dl>/.metric` ad-hoc markup. No code defect exists; dropped as PASS. |

## Deferred to Office-Hours

- **Item 8 — Screenshot + functional parity vs production:** Visual cosmetic check requiring human comparison against production. No objective failure was found — all charts rendered with correct token colors, correct semantic encoding (green/red balance bars), no console errors. The check is whether the rendered result *looks right* side-by-side with production; only a human can make that judgment.

natb1:
## Review: #1876 budget — migrate the budgets section to React

### Fixed (3)

- **`budget/test/pages/budgets-hydrate.test.tsx:17`** — Misleading comment claimed malformed-blob variance error-path tests "remain in `budgets-hydrate.test.ts`" (a file that no longer exists). Rewrote the comment and ported the 16 variance error-path cases into a new `describe` block driving the live exported `hydrateVarianceDetails` directly. (commit 31a4f16f)
- **`budget/src/pages/Budgets.tsx:333`** — `OverrideRow` key was `${o.budgetId}-${o.index}-${i}` where `i` is the allOverrides-sorted position — unstable across re-renders. Changed to the stable `${o.budgetId}-${o.index}` pair which is unique for the lifetime of the rendered data. (commit 31a4f16f)

### Refuted — adversarial-verify dropped (1)

- **`budget/src/pages/budgets-hydrate.ts:377`** (red-team) — CSS selector injection via unescaped `budgetId` in `collectOverridesForBudget`. Refuted: `budgetId` originates from either Firestore auto-generated 20-char alphanumeric IDs (no CSS metacharacters possible) or the `.benc` upload path. For the upload path, `storeParsedData` clears all IDB stores before writing — the attacker controls both the malicious ID and all DOM rows; no victim data is ever in the DOM alongside the attacker's row. Additionally, a `DOMException` from a malformed selector calls `showInputError`, not silent data corruption. Real robustness defect but not security-Required: no cross-trust-boundary exploitability. Skeptic rationale: no path by which an injected selector touches or leaks a victim's data.

### False-positive — security (2)

- **`budget/src/pages/use-budget-table.ts:254`** — `innerHTML` `dateStr` injection concern. `dateStr` is constructed from a JS `Date` formatted by `toLocaleDateString` — produces locale-specific decimal date strings with no HTML-injectable characters. Not exploitable.
- **`budget/src/pages/Budgets.tsx:191`** — `DataIntegrityError` thrown inside `BudgetTable` render flagged as data-exposure concern. The error is intentional — it surfaces missing variance data that is a programmer error, not a data leak. The render-phase throw is caught by the error boundary; no user data is exposed.

### Out-of-scope — security (1)

- **`budget/src/pages/budgets-waterfall-chart.ts:91`** — New `throw new Error(…)` replacing silent `d3.scaleOrdinal` null for a missing category color. Pre-existing chart code; not introduced by this PR's diff on the calling side.

### Informational (13)

- **`budget/src/pages/budgets-hydrate.ts:240`** — `scrollAbort` is a module-level variable shared across all calls to `hydrateBudgetChart`. If the page were ever simultaneously mounted multiple times, concurrent teardowns could race. Current architecture mounts once; tracked as a latent risk.
- **`budget/src/pages/budgets-hydrate.ts:341`** — `wireChartDatePicker` adds a `change` listener on `#chart-date-picker` but the teardown does not remove it. Only one chart-date-picker exists per page; the omission is safe in the current single-mount lifecycle but worth noting.
- **`budget/src/pages/Budgets.tsx:270`** — `ChartContainers` useEffect dependency array omits `averageWeeklySpending` while including `averageWeeklyCredits`. If those two values can diverge independently, a spending update won't re-run the chart island.
- **`budget/src/pages/use-budget-table.ts:134`** — `showInputError` on select validation failure clears the input error on the next `change` event, but does not restore the original value. A user who enters an invalid period/rollover sees an error but the select snaps to an unknown state.
- **`budget/src/pages/Budgets.tsx:421`** — `deferProgrammerError` in the data-load catch path defers the throw to a `setTimeout`, so the React error boundary catches it asynchronously. If multiple errors fire before the boundary handles the first, only the last may surface.
- **`budget/src/pages/use-budget-table.ts:43`** — `useBudgetTableInteractivity` and `useOverridesTableInteractivity` use `[authorized]` as the only dependency. If `authorized` transitions from `false → true → false`, the second detach/re-attach leaves listeners on the old container if the component unmounts between transitions.
- **`budget/src/pages/Budgets.tsx:266-270`** — `ChartContainers` useEffect re-runs `hydrateBudgetChart` whenever budgets/periods/trend data changes. The chart island teardown is called on each change, but rapid data updates could produce overlapping animation frames. Acceptable with the current stable-load pattern.
- **`budget/src/pages/Budgets.tsx:382-427`** — `useBudgetsData` has an empty `[]` dependency array, so it runs once at mount and never re-fetches if `options`/`dataSource` changes after mount. By design for this app's data model, but worth documenting.
- **`budget/src/pages/use-budget-table.ts:287-298`** — `useOverridesTableInteractivity` cleanup removes `blur` (capture-phase) and `click` listeners, but the `change` listener added for the delete button is not removed in cleanup. Likely safe if the container element is removed with the component, but a latent listener leak.
- **`budget/src/pages/budgets-pie-chart.ts:66-67`** — Non-null assertion `colorByName.get(name)!` in `sliceColor` will silently return `undefined` if a name is missing, which renders as the string `"undefined"` on canvas. The preceding loop always builds the map from the same data source, so this should not fire in practice.
- **`budget/src/pages/budgets-hydrate.ts:826-863`** — `hydrateBudgetChart` teardown clears `scrollAbort` and removes listeners. Reviewed: teardown is correct for the single-mount lifecycle. See also code-review-0.
- **`budget/src/pages/Budgets.tsx:309-314`** — `BudgetTable` throws `DataIntegrityError` during render if a budget has no entry in the `variances` map. This is an intentional integrity gate caught by the error boundary; not a silent failure.
- **`budget/src/pages/use-budget-table.ts:106-120`** — `savedSelectValue` WeakMap is seeded once at `useLayoutEffect` attach. If a select's value changes between mount and attach (unlikely given synchronous layout), the guard would use a stale seed.

Issue #888, opened 2026-05-29
dispatch: adopt the workflow tool for phase-skill fan-out


natb1:
## Context

Opus 4.8 (released 2026-05-28) ships a `workflow` tool: deterministic JavaScript orchestration that fans out subagents (`parallel`/`pipeline`), forces structured output via a `schema`, and supports adversarial-verify patterns. Several `/dispatch` phase skills already perform model-driven subagent fan-out described in prose — e.g. `/security-review-fix` Step 2 instructs "launch 9 subagents directly in a single message with 9 Agent tool calls", then merges, de-duplicates, and classifies their free-text findings by model judgment.

Prose fan-out depends on the model counting and sequencing the calls correctly on every run and honoring an implicit "once all N return" barrier — the same dropped-trailing-step failure class that leaked dispatch sessions in #824. It also leaves dedup and classification as single model passes with no independent check.

Porting the fan-out **core** of these skills to the `workflow` tool makes coverage deterministic (the script spawns exactly N agents every run), replaces model-judgment dedup with code, forces schema-validated findings, and adds an adversarial-verify stage the skills lack today.

## Scope boundary

This epic covers only the per-turn fan-out core that runs inside a single phase skill. It explicitly **excludes**:

- The lifecycle layer — the router, worker chain, selection lock, per-worktree dedup, concurrency budgeter, and office-hours queue. That layer is cross-session, multi-day, GitHub-as-ground-truth coordination; the `workflow` tool dies at end of turn and cannot model it.
- `/dispatch-qa` — a single self-verifying pass with negligible fan-out.

Each ported skill becomes a sanctioned `workflow` caller by instructing the call in its SKILL.md (the tool's opt-in rule recognizes a skill whose instructions tell it to call `workflow`).

## Sub-issues

- Port `/security-review-fix` review fan-out (strongest fit).
- Port `/review-fix` Fixed/Deferred fan-out.
- Port `/code-review-fix` Deferred-issue fan-out.

## Acceptance criteria

- [ ] Each sub-issue is filed, linked as a sub-issue, and merged.
- [ ] At least `/security-review-fix` runs its review fan-out through the `workflow` tool, with a deterministic agent count and schema-validated findings.
- [ ] The lifecycle layer (router, worker chain, selection lock, per-worktree dedup, concurrency budgeter, office-hours queue) is unchanged.

PR #2030, opened 2026-06-19
print: mount the viewer behind a React component with ds-based chrome

natb1:
Closes #1872

## What

Mounts the print viewer behind a real React component and rebuilds its chrome
(toolbar, search, outline, bookmarks, spread controls) on `@commons-systems/ds`
components and tokens. This kills the legacy string-render + `dangerouslySetInnerHTML`
+ `afterRenderView` + module-global bridge that `/view/:id` used to glue an HTML
shell string to the imperative `initViewer` orchestrator.

The render engines (`pdf.ts`, `epub.ts`, `image-archive.ts`) and the
`SpreadController` are unchanged — only the shell and chrome were the legacy
surface. The `.viewer-canvas-wrap` stays a childless, engine-owned node (the
AppNav mount-survival pattern), so React never reconciles the canvas subtree the
engines draw into.

## How (the 7 units)

1. **`useViewerController` hook** — the orchestration brain. Ports the full body
   of `initViewer` into hook form: an init/full-teardown effect, `syncNav()` → React
   state, all nav/goto/zoom/spread/keyboard/fullscreen/orientation dispatch,
   debounced position persistence with the `readFailed` clobber-guard, and a
   `navSignal` + `getRenderer()` panel-facing API.
2. **`<Viewer>` + `<ViewerToolbar>` + route rewire** — the big-bang switchover.
   `<Viewer>` renders the fixed shell as JSX and consumes the hook; the toolbar is
   ds `Button`/`Input`. `view.ts` gains a pure `resolveViewerProps` helper (the
   `mediaType`→renderer-factory switch + source resolver + store pick) and the
   `ViewFrame` `shell` variant becomes `ready`. `main.tsx` drops the
   `flushSync`/querySelector dance — init runs inside the hook's effect after
   React commits the canvas node. Deletes `shell.ts` and its now-dead
   `shell.test.ts`. ds `Button`/`Input` are wrapped in `forwardRef` (backward
   compatible) so the toolbar can attach the spread-toggle/goto refs the hook
   requires.
3. **`<SearchPanel>`** — ports `search.ts` (debounced input, results list with
   `<mark>` highlight, count, stale-result discard); ds `Input`, custom
   token-styled results list. Deletes `search.ts`.
4. **`<OutlinePanel>`** — ports `outline.ts` as a recursive React tree
   (async-loaded entries, local per-node expand/collapse). Deletes `outline.ts`.
5. **`<BookmarksPanel>` + `useBookmarks`** — ports `bookmarks.ts`: the toggle
   lives in the toolbar, the list in the panel, both sharing a `useBookmarks`
   hook (cloud-vs-local store selection from the controller's `uid`/`readFailed`/
   `mediaId`, `navSignal`-driven aria-pressed sync, recreated localStorage
   helpers the deleted shell had owned). Deletes `viewer/bookmarks.ts`.
6. **CSS tokenization** — `viewer.css` design-system literals (`var(--muted-fg,#888)`,
   3px control radii, rem font/padding/spacing) → ds tokens (`--border`/`--muted`,
   `--radius-none`/`-sm`, `--space-*`/`--text-*`), keeping viewer-specific geometry
   (panel 280px, tap-zone 20%, canvas sizing, spread gap, pdf text-layer
   positioning) and content-layer highlight/selection colors.
7. **e2e selector sweep** — `viewer.spec.ts` state-class assertions for the now
   conditionally-rendered panels (`search-hidden`/`outline-hidden`) become element
   presence/absence; the still-valid class assertions (`spread-hidden`,
   `outline-collapsed`, canvas-wrap `zoomed`, panel `collapsed`) and all
   `.viewer-*` selectors + aria-labels are preserved.

## Acceptance criteria

- ✅ Viewer mounted via a React component; render engines unchanged at full parity.
- ✅ Chrome built from `@commons-systems/ds` components.
- ✅ No viewer design-system CSS literals remain outside `packages/ds` — only
  viewer-specific geometry and content-layer highlight colors.
- ✅ Dialog/control radius exceptions kept (form controls 4px via `--radius-sm`);
  everything else square (`--radius-none`).
- Functional parity for open/read/search/outline/bookmark across pdf, epub, and
  image-archive — validated by the unit suite and the e2e run in CI.

## Verification

- `npm run test --prefix print` — 553 passing (35 files), including the rewritten
  search/outline/bookmark component tests and the new hook + `<Viewer>` tests.
- `npm run test --prefix packages/ds` — 18 passing (the `forwardRef` wrap is
  backward compatible; verified against the `audio` consumer too).
- `npm run build --prefix print` — `tsc` + `vite build` clean.
- `npm run lint --prefix print` — clean.
- `print/e2e/viewer.spec.ts` exercises pdf / epub / image-archive; the e2e run is
  authoritative in CI.

## Notes

- `loadError` from the hook is sticky (matches the old viewer, which wrote these
  messages into `position.textContent`); surfaced in the `.viewer-position` slot.
- The bookmark toggle stays in the `.viewer-nav` row (parity); the list section
  is conditionally rendered in the panel body.
- `aria-pressed` on the spread toggle is currently written twice: by React
  (`aria-pressed={spreadEnabled}` in `ViewerToolbar`) and by the unchanged
  `SpreadController.setAttribute`. This is benign — the hook sets `spreadEnabled`
  state on every spread transition including the `loadPreference()` auto-enter
  path, so the two writes always agree — but it is redundant; a follow-up could
  drop the controller's `setAttribute` now that the toggle's pressed state is
  React-driven.


natb1:
# Fix-checks summary — PR #2030

## Iteration 1

- **Failed checks**: acceptance (3 e2e: `viewer.spec.ts:309 "image archive: zoom controls hidden for PDF"` on mobile / tablet / desktop)
- **Outcome**: fixed
- **Reproduced**: no (local Playwright blocked by chromium version mismatch — Nix 1217 vs project 1208; cause confirmed from the CI DOM artifact + `packages/ds/src/core/Button.tsx` inline `display`)
- **Reproduce command**: `.claude/skills/dispatch-propagate/scripts/run-acceptance-tests.sh print` (not runnable locally; verified `npm run build --prefix print` stays clean instead — CI acceptance is authoritative)
- **Failure excerpt**: `expect(locator('.viewer-zoom-in')).not.toBeVisible()` failed — element `<button ... class="cs-btn cs-btn--secondary viewer-zoom-in zoom-hidden">+</button>` resolved as `visible`.
- **Why not caught**: Unit 7 rewrote the viewer chrome to route the zoom (and spread) controls through the ds `Button` component, which sets `display: "inline-flex"` as an **inline** style. An inline style outranks a class selector, so `.zoom-hidden { display: none }` could not hide a ds `Button` — the control stayed visible for renderers without zoom (PDF / image-archive). The unit suite asserts class presence, not visibility, so only the e2e visibility assertion caught it. The `spread-hidden` case (also a ds `Button`) was silently broken the same way but is only asserted via `toHaveClass`, so it never failed CI.
- **Fix**: `print/src/style/viewer.css` — added `!important` to the shared hidden-utility rule (`.zoom-hidden, .spread-hidden, … { display: none !important; }`) so the stylesheet beats the ds `Button` inline `display`. The toggle classes are preserved, so the e2e `toHaveClass(/spread-hidden/)` assertions still hold. Commit `9977310d`.

natb1:
## QA Pass — #1872 / PR #2030

**Pre-QA acceptance:** 164 passed, 10 skipped (@testonly) — clean gate.

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 7 | No DS CSS literals in viewer.css | shell | ✅ PASS |
| 8 | Print unit tests (553 tests) | shell | ✅ PASS |
| 1 | PDF viewer renders with full chrome | browser | ✅ PASS |
| 2 | PDF page navigation (prev/next/goto) | browser | ✅ PASS |
| 3 | PDF search panel opens and returns results | browser | ✅ PASS |
| 4 | EPUB outline panel shows TOC entries | browser | ✅ PASS |
| 5 | Bookmark toggle add/remove round-trip | browser | ✅ PASS |
| 6 | EPUB viewer loads with ds chrome | browser | ✅ PASS |
| 9 | Image-archive walkthrough | deferred | ⊘ DEFERRED |

**Pass: 8/8 runnable items. 1 planned deferral.**

### Notes by item

**#7 (CSS literals):** The remaining `rgba()` and hex values in `viewer.css` (lines 122, 133, 138, 399, 403, 421) are all content-layer highlight/selection colors — PDF text-layer blue, search highlight yellow/orange, bookmark yellow. These are explicitly kept per acceptance criteria ("only viewer-specific geometry and content-layer highlight colors").

**#8 (Unit tests):** 553 tests across 35 files, all passing in 7.92s. DNS-resolve noise from the epub test fixture is a known sandbox artifact, not a test failure.

**#1 (PDF viewer):** `/view/plato-phaedrus` renders with full ds chrome. Viewer DOM: `viewer`, `canvas-wrap`, `viewer-nav`, `viewer-prev`/`next` (class `cs-btn cs-btn--secondary`), `viewer-goto-input` (class `cs-input`), `viewer-search-input` (class `cs-input`), `viewer-bookmark-toggle` (class `cs-btn`). ds component classes confirm AC2.

**#2 (Navigation):** `/view/plato-republic` (3 pages). Next: 1→2, prev: 2→1, goto field entered `3`→Enter→page 3. All three nav paths working.

**#3 (Search):** Typed "Socrates" in the ds Input search field → 10 results listed with page labels and text snippets. Count label shows "10 results". The `viewer-search` element is present (conditional rendering, not CSS-class-hidden).

**#4 (Outline):** `/view/gutenberg-3296` epub: outline panel present with 5 entries (first: "Preliminary Confessions"). Clicked "The Pleasures of Opium" → viewer navigated Ch. 1/3 → Ch. 2/3. Expand toggle present.

**#5 (Bookmarks):** At Ch. 2/3: clicked bookmark toggle → `aria-pressed="true"`, bookmarks panel appeared with entry "Ch. 2/3 — p. 1/46". Clicked toggle again → `aria-pressed="false"`, panel disappeared (conditional rendering). Full add/remove round-trip confirmed.

**#6 (EPUB):** "The Confessions of St. Augustine" renders. ds toolbar visible with Contents (outline) and Search panels. Position indicator "Ch. 1/3 — p. 1/1". Shared chrome works at parity with PDF format.

### GIF walkthrough

`tmp/qa-fix-walkthrough-1872.gif` (40 frames)

### Disposition triage

One residue item classified by the disposition Workflow:

| 9 | Image-archive walkthrough deferred to CI | needs-human | n/a |

> **Planned-deferral** (item #9): The acceptance criterion for image-archive format parity is documented as non-assertable at merge time via the public-seed walkthrough — only `testOnly` seed items exist for image-archive, and the public-seed QA policy excludes them. Parity is measured downstream by the CI `@testonly` acceptance suite. The office-hours pass confirms CI has run and passed on this format.

`needs-main` items: none filed.

Escalating item #9 to office-hours for human confirmation that CI @testonly acceptance tests cover image-archive parity.

natb1:
## Review — PR #2030 (`1872-print-mount-the-viewer-subs`)

16 findings surfaced across quality and security passes. 4 applied as fixes (commit `e21f128f`), 5 informational, 3 dismissed, 4 out-of-scope pre-existing security findings.

---

### Fixed

- **`useViewerController.ts:170–200` — broken cancellation guard (shared `st.cancelled` ref)**
  `st` was `internals.current`, a single ref object reused across effect invocations. On StrictMode double-mount, the cleanup from the first invocation set `cancelled = true` on the same object the second invocation was reading, causing premature exits. Replaced with a per-effect-invocation `let cancelled = false` local variable, closing over it in all async callbacks and setting `cancelled = true` in the cleanup return. Guards added to the `setReadFailed`/`setLoadError` catch blocks.
  _(sources: code-review, review — fix commit `e21f128f`)_

- **`OutlinePanel.tsx:96` — `loadedRef.current = true` set before async resolves**
  The once-guard was set synchronously before `r.getOutline()` resolved, so a rejected load left the guard as `true` and the next `navSignal` change never retried. Moved the assignment into the `.then()` success callback; a rejection now leaves the guard `false` and the next tick retries.
  _(source: code-review — fix commit `e21f128f`)_

- **`useBookmarks.ts:93` — redundant `controller` dep in `currentBookmarked` useMemo**
  `controller` was listed alongside `controller.navSignal` as a dep. The object reference is stable per mount, so it added no signal value but triggered an extra memo on every render where `controller` was a new identity. Dep array trimmed to `[controller.navSignal, bookmarks]`.
  _(source: review — fix commit `e21f128f`)_

---

### Informational

- **`Viewer.tsx:46–49` / `useBookmarks.ts:87` — `readFailed` change causes `bookmarksStore` memo to recompute**
  When the position store read fails after mount, `controller.readFailed` flips to `true`. Because `controller` was in the `currentBookmarked` dep array (now removed), this caused an extra memo recomputation. The `readFailed` → memo relationship is worth tracking even after the dep-array fix; the memo is cheap but the causation may surprise.

- **`useBookmarks.ts:72–87` — empty dep array on the bookmarks-load `useEffect` with 'store identity is stable per mount' comment**
  The comment is accurate for the current mount lifecycle, but the effect closes over `bookmarksStore`. If the store identity ever changes (e.g., wrapping in a provider), the stale closure will silently use the old store. Low risk today; noted for future refactors.

- **`SearchPanel.tsx:160` — search results keyed by array index**
  Index keys cause React to misidentify elements when results change between searches, producing incorrect diff patches. Stable identity keys (e.g., `${hit.page}-${hit.idx}`) would be safer.

- **`OutlinePanel.tsx:53,127` — outline entries keyed by array index**
  Same concern as above. Outline entries are generally stable once loaded, so the risk of misidentification is lower — but a composite key (`${entry.dest}-${index}`) is a straightforward improvement.

- **`Viewer.tsx:55` — `viewerRef` typed as `RefObject<HTMLElement>` but cast to `React.RefObject<HTMLDivElement>` at JSX**
  The type annotation in `UseViewerControllerResult` is wider than the actual usage. Tightening the type to `RefObject<HTMLDivElement>` would make the mismatch visible at the interface boundary instead of at cast sites.

---

### Dismissed

- **`view.ts:190` — `resolveLocal` calls `reportError` then throws**
  `reportError` records the telemetry event; the throw propagates to the hook's outer catch, which also calls `reportError`. This looks like a double-fire, but the two call sites target different error classes (`notFound` vs. the generic catch-all). The outer catch is the correct deduplication point; the inner `reportError` is intentional and documented in the surrounding code. _Not a bug._

- **`viewer.css` — `var()` fallback values removed**
  All `var()` references now lack fallback values (e.g., `var(--bg)` instead of `var(--bg, #fff)`). In a design-system context where these tokens are always defined, fallbacks are noise. The DS tokens are a hard build-time contract; a missing token is a build failure, not a runtime default. _Stylistic — no change._

- **`view.ts:183–190` — `reportError` called twice for the same local-file-not-found failure**
  On closer inspection the outer catch maps to a different code path than the inner `resolveLocal` call; the two calls emit distinct error shapes. _False positive._

---

### Out-of-scope (pre-existing Firestore rules — not changed by this diff)

These findings are in `firestore.rules`, which the diff did not modify. They are below the threshold for filing a tracking issue (no `critical`/`high` CodeQL alert, advisory rating not `high`/`critical`), so no follow-up issues were filed.

- **`firestore.rules:284` — bookmarks write rule uses `keys().hasAll(...)` not `keys().hasOnly(...)`**
  An authenticated user can write a bookmarks document with arbitrary extra keys beyond `uid`, `mediaId`, and `bookmarks`. Tightening to `hasOnly` would close the over-wide write surface.
  _(sources: auth, firebase)_

- **`firestore.rules:279–281` — bookmarks read rule allows `getDoc` on any `{uid}_{mediaId}` path when the document does not yet exist**
  A non-existent document returns `null`, so the read itself is harmless. But the rule grants access without a document-presence check, which widens the read surface slightly.
  _(source: auth)_

- **`firestore.rules:278–286` — bookmarks array size and structure unconstrained**
  The write rule validates that the `bookmarks` key is present but applies no constraints on array size or element shape. A large or malformed `bookmarks` value would be stored as-is.
  _(source: firebase)_

Issue #1557, opened 2026-06-14
dispatch: `empty` tick disposition dead-ends the chain (no continuation backstop)


natb1:
## Context

The autonomous `dispatch-tick` chain is event-driven with per-disposition
continuation (no standing heartbeat — that was deliberately rejected in #1453).
Each terminal no-spawn disposition must therefore arm its own continuation:

- `propagate` → the spawned worker's Stop-hook re-ticks (+ a convergence reseed)
- `concurrency-cap` → `dispatch-schedule-reseed` armed inside `dispatch-select-tick`
- `drain` → the `dispatch-tick-recover` backstop (#1445)

The `empty` disposition arms **nothing**. `dispatch-select-tick`'s terminal
`empty)` branch releases the lock and `echo "empty"` with no reseed; `dispatch-tick`
then matches `empty` in the `sync-failed|resolver-failed|empty|concurrency-cap)`
case and bare-`exit 0`s. When `empty` is the final disposition the chain enters an
absorbing dead state until an unrelated external event (a worker Stop-hook or a
manual `dispatch`) revives it.

`empty` means: work exists in the queue but every reachable open leaf is claimed
(live Claude session ∪ reservation marker), parked on `dispatch:office-hours`, or
in the tick's `--exclude` set, **and the fleet is below target**. The block is
transient — it clears when a worker finishes, a stranded reservation is reclaimed,
or a park resolves — but nothing re-ticks to observe that. In particular a
reservation marker held by a dead session fires no Stop-hook, so the chain can
stay dead indefinitely.

This is the unpatched sibling of #1445, which gave the same continuation backstop
to `drain` but not to `empty`.

## Reproduction / observed (2026-06-13)

1. An autonomous tick resolved to `empty` at 20:14:34 — journal:
   `dispatch-tick: select-tick disposition 'empty'; nothing spawned` — and
   bare-exited. In the run-up every reachable open leaf logged
   `all open leaves reachable from N are worktree-conflicted (owned by a live
   session or parked on dispatch:office-hours)`.
2. The chain stayed dead ~2.5 min until a manual `dispatch` at 20:16:59 revived it.

- **Expected:** after an `empty` tick, the chain re-evaluates on a bounded cadence
  (reclaiming stranded reservations, observing cleared parks) and resumes when work
  unblocks, or escalates to `dispatch:chain-stalled` if persistently blocked on a
  human.
- **Actual:** bare exit, no continuation; chain dead until an external kick.

## Proposed fix

[...]

PR #2044, opened 2026-06-19
dispatch: durable periodic heartbeat timer so abnormal death + post-cap stalls self-recover

natb1:
Closes #2022

## Why

The autonomous dispatch chain is continued **reactively**: a finishing worker's
Stop hook launches the next headless tick, and the reseed scripts arm one-shot
timers when the budget is capped. There is **no always-on periodic tick** — every
continuation depends on a Stop hook firing, an armed reseed, or an `OnFailure`
recover. Two gaps leave the chain dead with no armed continuation:

1. **Abnormal worker death.** A worker reaped by cgroup-kill / OOM / SIGKILL
   never reaches its Stop hook, so it launches no replacement and leaves no trace.
2. **Post-cap recover dead-end.** When `dispatch-tick-recover` exceeds its
   consecutive-failure cap it files a `dispatch:chain-stalled` latch issue and
   **arms no reseed**. From that point the chain has zero armed continuation and
   stays dead until a human runs `dispatch` — the failure counter only resets when
   recover is next invoked, and with no reseed there is no next tick to invoke it.

The accumulated drain-to-zero incidents (#1445, #1557, #2013) show that targeted
reseeds do not cover abnormal death or the post-cap dead-end. A single always-on
periodic heartbeat closes both: it fires regardless of Stop hooks, reseeds, or
recover state, so abnormal deaths and post-cap stalls both self-recover within one
interval. This reverses the #1010 decision that rejected a periodic heartbeat in
favor of targeted reseeds.

The heartbeat **coexists with** — does not replace — the `dispatch:chain-stalled`
escalation: the heartbeat provides automatic liveness; the latch issue provides
human visibility and is intentionally never auto-closed.

## What

A durable `systemd --user` `dispatch-heartbeat.timer` fires a no-arg
`dispatch-tick` every 15 min (plus 2 min after boot, `Persistent=true`), installed
best-effort alongside the existing durable units. Because `dispatch-tick` is a
pure-bash sequencer with **no model session** and is self-suppressing
(`busy`/`concurrency-cap`/`empty` dispositions spawn no worker), a flat drumbeat is
safe and consumes zero model tokens when idle.

- **Unit 1 — `ensure_heartbeat_units` helper (`lib.sh`).** A new helper modeled on
  `ensure_daemon_service` / `ensure_recover_unit` that writes the
  `dispatch-heartbeat.{service,timer}` pair with the idempotent content-compare →
  atomic temp-write → `daemon-reload` → `enable --now` pattern. The service is
  `Type=oneshot` with `ExecStart=<main_worktree>/.../dispatch-tick`,
  `KillMode=process`, and `OnFailure=dispatch-tick-recover.service`; the timer is
  `OnBootSec=2min`, `OnUnitActiveSec=15min`, `Persistent=true`,
  `WantedBy=timers.target`. Best-effort throughout.

- **Unit 2 — wire into install sites.** `ensure_heartbeat_units "$MAIN_WORKTREE" ||
  true` added next to the existing `ensure_*` calls in `dispatch-spawn-tick` (the
  primary universal Stop-hook continuation path), `dispatch-schedule-reseed`,
  `dispatch-schedule-convergence-reseed`, `dispatch-schedule-rate-limit-resume`,
  and `dispatch-schedule-target-reseed`. In `dispatch-tick-recover` the call is
  placed **early** — before the post-cap `exit 0` — so a cold start that hits a
  stall still arms the heartbeat. Deliberately not added to `dispatch-spawn-sweep`,
  which installs no recovery/daemon units.

- **Unit 3 — regression tests.** A no-op `dispatch-tick` (the heartbeat's
  `ExecStart`) at `concurrency-cap` spawns no worker — confirming no model session
  and no tokens — and `ensure_heartbeat_units` writes the correct service/timer
  directives and is idempotent (a second call short-circuits with no
  `daemon-reload`/`enable`). Full dispatch suite stays green (3189/3189).

## Critical design decision — `ExecStart` runs `dispatch-tick` directly

The reseed timer is the proven timer→tick analog: it runs `dispatch-tick` directly
as a `systemd --user` unit's `ExecStart` with `KillMode=process` and
`OnFailure=dispatch-tick-recover.service`. Mirroring those exact properties
guarantees the heartbeat-spawned worker survives the oneshot service exiting — the
worker's first `claude` call attaches to the durable
`dispatch-claude-daemon.service` cgroup, not the heartbeat service's cgroup, and
`KillMode=process` is the degraded-path fallback for hosts without `systemd --user`
(the #1197 rationale). Running `dispatch-spawn-tick` from inside a `--user` oneshot
service would be an untested nested-`systemd-run` combination; its only added
benefit (fixed-unit-name dedup) is already covered by the selection lock, which
makes a concurrent tick exit `busy` cheaply.

## Known, accepted cost

A 15-min no-op heartbeat runs the full gh-API pipeline
(JIT/calendar/statements/retriage/auto-merge) on `empty`/`concurrency-cap`
dispositions (~96/day at idle). This is **not** model-token cost — `dispatch-tick`
has no model session and spawns no worker on a no-op. These gh-API operations are
idempotent and already run on every Stop-hook / reseed tick; the heartbeat only
adds an idle floor.


natb1:
## QA pass — issue #2022 / PR #2044

**Branch:** `2022-dispatch-periodic-heartbeat` · **Date:** 2026-06-19

No browser component detected (pure scripts PR). Steps 3b/3c skipped; no QA server started.

---

### Results

| # | Item | Classification | Result |
|---|------|----------------|--------|
| 1 | Wiring: `ensure_heartbeat_units` at all 6 install sites | script-verifiable | ✅ PASS |
| 2 | AC3 placement: `dispatch-tick-recover` arms heartbeat before COUNT > CAP gate | script-verifiable | ✅ PASS |
| 3 | AC4 static directives: timer carries `Persistent=true`, `OnBootSec`, `WantedBy=timers.target` | script-verifiable | ✅ PASS |
| 4 | AC6: #1010 coexistence rationale documented in `lib.sh` | needs-human-judgment | ✅ Already satisfied (Opus) |
| 5 | Regression test coverage present for AC5 and `ensure_heartbeat_units` idempotency | script-verifiable | ✅ PASS |
| 6 | `ExecStart` targets `dispatch-tick` directly (not `dispatch-spawn-tick`) | script-verifiable | ✅ PASS |
| 7 | AC2 (live): propagation resumes after abnormal worker death | needs-human-judgment | 🔵 Planned deferral |
| 8 | AC3 (live): chain auto-recovers after post-cap stall, stalled issue persists | needs-human-judgment | 🔵 Planned deferral |
| 9 | AC4 (live): heartbeat survives reboot/logout | needs-human-judgment | 🔵 Planned deferral |

**Counts:** 5 PASS · 0 FAIL · 0 SKIP · 4 needs-human-judgment (1 already-satisfied by Opus, 3 planned-deferral)

---

### Pass details

**Item 1 — Wiring.** All 6 expected files invoke `ensure_heartbeat_units "$MAIN_WORKTREE" || true`: `dispatch-spawn-tick`, `dispatch-schedule-reseed`, `dispatch-schedule-convergence-reseed`, `dispatch-schedule-rate-limit-resume`, `dispatch-schedule-target-reseed`, `dispatch-tick-recover`. `dispatch-spawn-sweep` correctly omits it (by design — it installs no recovery/daemon units).

**Item 2 — Placement.** `ensure_heartbeat_units` is at line 170 of `dispatch-tick-recover`; the `COUNT > CAP` gate is at line 279. The early call fires on the post-cap dead-end path (`exit 0` at line 328 is never reached by the later `ensure_daemon_service` at line 380), ensuring cold-start stalls arm the heartbeat.

**Item 3 — Timer directives.** `lib.sh` contains all four: `Persistent=true`, `OnBootSec=2min`, `OnUnitActiveSec=15min`, `WantedBy=timers.target`.

**Item 5 — Regression coverage.** 3/3 expected coverage markers found in `test-dispatch-scripts.sh`: the AC5 no-spawn-job assertion, the cold-path `ensure_heartbeat_units` content block, and the hot-path idempotency assertion.

**Item 6 — ExecStart.** `TICK_SCRIPT` is set to `.../scripts/dispatch-tick` (not `dispatch-spawn-tick`). `ExecStart="$TICK_SCRIPT"`, `KillMode=process`, and `OnFailure=dispatch-tick-recover.service` are all present in the `desired_service` heredoc.

---

### Disposition triage (Opus)

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR. There are no `needs-main` items this pass.

**Assessed by Opus (already satisfied — dropped as PASS):**

- **Item 4 — AC6:** The `#1010` coexistence rationale is present and complete in `lib.sh:1525–1530`. The comment references `#1010`, explains the heartbeat provides automatic liveness while the `dispatch:chain-stalled` escalation is kept for human visibility, states the two are not mutually exclusive, and explicitly notes the helper does not touch the escalation latch. Provably satisfies AC6 from the code alone — no defect, no human decision needed.

**Deferred to office-hours (planned-deferral — verify=n/a):**

These three items are acceptance criteria the issue documents as non-assertable at merge time. They require live runtime conditions (a running systemd-user host, timer firings over ≤15-min intervals, reboot cycles) that are out of scope for automated QA. They will be verified downstream by monitoring. The office-hours walkthrough should review and confirm the deferrals are appropriate, then clear them.

- **Item 7 — AC2 (live):** Propagation resumes within one heartbeat interval after abnormal worker death. *Reason: runtime timer-driven recovery over a 15-minute interval; not script-verifiable; live-system verification out of scope.*
- **Item 8 — AC3 (live):** Chain auto-recovers after a post-cap stall while the stalled issue persists. *Reason: runtime timer + GitHub-state behavior over a heartbeat interval; not script-verifiable; live-system verification out of scope.*
- **Item 9 — AC4 (live):** Heartbeat survives reboot/logout. *Reason: requires a live reboot/logout cycle on a systemd-user host; not script-verifiable; live-system verification out of scope.*

**Non-fixing pass:** Items 7, 8, 9 are planned-deferral `needs-human` items — they escalate to office-hours for confirmation that the deferral is appropriate. No code defects were found and no Opus-fixable bugs exist.

natb1:
## Review findings — PR #2044

### Fixed

- **SEC-2 — temp file created world-readable before atomic rename** (`lib.sh:ensure_heartbeat_units`): `mktemp` inherits the process umask; typical umasks (0022, 0002) produce 0644 or 0664 on the temp file, making unit content (including the full `ExecStart` path) readable to other users on a multi-user host before the rename. Fixed by wrapping each `mktemp` call with `umask 077` / restore (temp file created 0600) and adding `chmod 0644` before the `mv` so the installed unit file is world-readable (required for systemd). Commit a377e6f1.

### Informational

- **CR-5 — `OnUnitActiveSec=15min` drifts per activation, not wall-clock periodic**: `OnUnitActiveSec` measures from the unit's last activation, not a fixed wall-clock schedule; activations longer than 15 min shift the next tick. Accepted: the heartbeat's purpose is liveness (fire within one interval of a dead chain), not strict periodicity — drift is fine.

- **RV-1 — test suite stubs `systemctl` but stub-log assertions are brittle**: The `ensure_heartbeat_units` tests assert on presence/absence of stub-log entries (`daemon-reload`, `enable`). These assertions break if the log format changes. Noted as a known fragility in the test harness.

### Dismissed

- **CR-2 — missing idempotency guard when units directory does not exist**: Not a gap — `mkdir -p` is already called before any temp-file write; the concern does not reflect the actual code path.

- **CR-3 — temp-write lacks `fsync` before atomic rename**: Best-effort install design (explicitly stated in the PR and skill); adding `fsync` would be inconsistent with `ensure_daemon_service` / `ensure_recover_unit` peers.

- **CR-4 — `|| true` at call sites silently discards errors**: Intentional — `ensure_heartbeat_units` is a best-effort helper, consistent with all peer `ensure_*` call sites.

- **RV-2 — concurrent `ensure_heartbeat_units` callers race the unit writes**: Safe race — both writers produce identical content; the atomic rename means the last writer wins with correct content.

- **SEC-3 — `dispatch-tick` path not checked for existence before embedding in unit**: Not an injection attack — systemd simply fails to start the service if the path is missing. Best-effort design accepts this; no change needed.

### Dismissed (pre-existing / already present)

- **SEC-1 — `MAIN_WORKTREE` path embedded without validation**: Validation guards already present at function entry in the original branch code: rejects values containing newlines, spaces, double-quotes, and backslashes with a clear `WARNING` before any file is written. No code change needed.

- **CR-1 — inconsistent quoting of `main_worktree` expansions**: With the upfront validation rejecting all chars that could cause word-splitting (spaces, newlines, quotes, backslashes), the in-body expansions are safe. No change needed.

### Deferred

- **CR-6 — no cleanup of stale heartbeat units when installation path changes**: If the dispatch installation moves, old unit files pointing at the old path remain under `~/.config/systemd/user/` and keep firing at the missing path. Valid gap, not blocking the initial feature (path is stable in practice). Filed as #2056 (`blocked_by` #2022).

### Security (CodeQL)

No open CodeQL alerts for this PR.

PR #2047, opened 2026-06-19
dispatch: run the worktree sweep on a durable periodic systemd --user timer

natb1:
Closes #2023

## Problem

The worktree garbage-collector (`dispatch-sweep`, launched via
`dispatch-spawn-sweep`) fires **only** from a finishing worker's Stop hook
(`dispatch-stop.sh`, the `spawn_sweep` branches). So worktrees are reclaimed only
while workers are finishing. When the chain is idle or has drained to zero, no
Stop hooks fire, no sweep runs, and worktrees accumulate precisely when nothing
is cleaning them.

Folding the sweep into the tick is not viable: there is no persistent periodic
heartbeat in the dispatch system. The tick is itself activity-coupled (spawned by
Stop hooks, re-armed by one-shot self-rescheduling reseed timers). The only
durable units are the `OnFailure=` recover handler and the `Restart=always`
daemon supervisor — neither is periodic. When the chain drains to zero, the tick
stops firing too, so there is no heartbeat to fold into.

## Change

Introduce a durable, **recurring** `systemd --user` timer that fires the existing
`dispatch-spawn-sweep` on a steady cadence, installed and **enabled** idempotently.

- **`ensure_sweep_timer()` in `lib.sh`** — a new durable-unit installer modeled on
  `ensure_daemon_service` (path guards, `strip_unit_env_path`, env overrides for
  tests, atomic write gated on content-differs, byte-compare hot-path skip with an
  `is-active` gate, and a `daemon-reload` + `enable --now` tail). It writes
  `dispatch-sweep-periodic.service` (`Type=oneshot`, ExecStart through
  `dispatch-spawn-sweep`) and `dispatch-sweep-periodic.timer`
  (`OnBootSec=5min`, `OnUnitActiveSec=15min`). It enables the **timer** — modeling
  the helper on `ensure_recover_unit` instead would install a `.timer` that passes
  every content-compare test yet never fires, because a timer is inert until
  `systemctl --user enable --now`.
- **Wiring** — `ensure_sweep_timer "$MAIN_WORKTREE" || true` added alongside the
  existing `ensure_recover_unit` / `ensure_daemon_service` calls at the five
  launcher call sites (`dispatch-spawn-tick`, `dispatch-schedule-reseed`,
  `dispatch-schedule-target-reseed`, `dispatch-schedule-convergence-reseed`,
  `dispatch-schedule-rate-limit-resume`). Any active tick/reseed installs and
  enables the timer once; it then fires on cadence regardless of all future chain
  activity.
- **Tests** — `test-dispatch-scripts.sh` gains install/enable idempotency coverage
  for `ensure_sweep_timer` (units written, `OnUnitActiveSec=` present, no
  `Persistent=`, ExecStart points at `dispatch-spawn-sweep`, `enable --now` of the
  timer, hot-path skip on a second identical call) and an AC #4 test that runs the
  service's ExecStart with no live worker and asserts the recorded `systemd-run`
  argv contains `--unit=dispatch-sweep`. The full suite passes (3187/3187); this
  suite runs in the `hook-tests` CI job.

## No new overlap risk

`dispatch-spawn-sweep` is reused unchanged: it already enforces the 300s mtime
throttle and the fixed-unit (`dispatch-sweep`) dedup, so a timer fire
deduplicates against a concurrent Stop-hook fire. `OnUnitActiveSec=15min` sits
comfortably above the 300s throttle. No `Persistent=` (it only affects
`OnCalendar=` timers and would be a no-op here, or cause catch-up thundering).

## Assumption (named, not introduced by this PR)

The "chain idle → still reclaimed" guarantee holds only if the `systemd --user`
manager keeps running while no session is active — i.e. `loginctl enable-linger`
for the user. This is the **same** assumption `ensure_daemon_service` already
depends on for the durable daemon; it is not introduced here and is out of scope.
Manual verification on a real host (CI cannot run systemd): trigger any launcher
once, then `systemctl --user list-timers | grep dispatch-sweep-periodic` and
`journalctl --user -u dispatch-sweep-periodic.service` to observe a fire with the
chain idle.


natb1:
# QA Summary — PR #2047 (issue #2023)

## Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Worktree sweep runs on a periodic schedule independent of Stop-hook firing | shell-command | ✅ PASS |
| 2 | With the chain idle, eligible worktrees are reclaimed within one sweep interval | — | ⏸ Deferred (planned-deferral) |
| 3 | Periodic trigger respects existing 300s throttle and fixed-unit dedup | shell-command | ✅ PASS |
| 4 | Regression coverage asserts sweep invoked on timer path with no live worker | shell-command | ✅ PASS |

**Counts:** 3 PASS · 0 FAIL · 0 SKIP · 1 Deferred (needs-human-judgment)

## Test Suite

Full `test-dispatch-scripts.sh` run: **3187/3187 passed, 0 failed**.

New `ensure_sweep_timer` tests confirmed passing:
- Cold path wrote both `.timer` and `.service`
- `.timer` has `OnUnitActiveSec=`
- `.timer` has no `Persistent=`
- `.service` ExecStart points at `dispatch-spawn-sweep`
- Cold path ran `enable --now dispatch-sweep-periodic.timer`
- Idempotent path did not re-run enable
- Idempotent path did not rewrite units (no daemon-reload)
- Idempotent path did not rewrite `.service` (inode stable)
- AC#4: `dispatch-spawn-sweep` exits 0 with no live worker
- AC#4: stdout is `spawned`
- AC#4: timer path reached `dispatch-sweep` (unit + cwd + exec, no OnFailure)

## Bugs Found

None.

## Deferred to Office-Hours

**1 planned-deferral** (requires real systemd — out of CI scope, per PR body):

- **AC#2 — With the chain idle, eligible worktrees are reclaimed within one sweep interval**  
  _Planned-deferral_: requires real `systemd --user` with `loginctl enable-linger`; CI cannot run systemd. PR body explicitly names this as out of CI scope — the same assumption `ensure_daemon_service` already depends on for the durable daemon. Verified downstream by host observation:
  ```bash
  systemctl --user list-timers | grep dispatch-sweep-periodic
  journalctl --user -u dispatch-sweep-periodic.service --since '20 min ago'
  ```

## Disposition Triage (Opus-classified)

| 2 | With the chain idle, eligible worktrees are reclaimed within one sweep interval | `needs-human` | n/a (planned-deferral) | planned_deferral is true: the acceptance criterion is documented as non-assertable at merge time. Requires real systemd --user with loginctl enable-linger, which CI cannot run. No defect Opus can fix — per the planned-deferral rule this is needs-human, not opus-fixable. |

**Terminal behavior (non-fixing pass):** The `needs-human` residue (AC#2 planned-deferral) escalates to office-hours for human verification on a real host. No fixes were applied. No `needs-main` items were filed.

natb1:
## Review: PR #2047 (`dispatch: run the worktree sweep on a periodic timer`)

**Surface:** code (bash scripts + test file) | **Security:** no attack surface on app/rules (shell-only dispatcher scripts; no CodeQL alerts; no dependency changes)

---

### Fixed

| Finding | Location | Fix |
|---|---|---|
| Missing cold-path daemon-reload assertion in `ensure_sweep_timer` tests | `test-dispatch-scripts.sh:15696` | Added `grep -q 'daemon-reload' "$est_log"` check in the cold-path success branch, mirroring `ensure_daemon_service` coverage |
| Missing inactive self-heal test case (units match on disk but timer not active) | `test-dispatch-scripts.sh:15738` | Added inactive self-heal test with `STUB_IS_ACTIVE_RC=3`, asserting `enable --now` and `daemon-reload` ran but unit inodes stayed stable |
| Missing input-validation rejection guard tests (newline, space, double-quote, backslash) | `test-dispatch-scripts.sh:15661` | Added four rejection tests matching the `ensure_recover_unit` pattern |
| Missing daemon-reload and enable failure-path tests | `test-dispatch-scripts.sh:15661` | Added `STUB_RELOAD_RC=1` and `STUB_ENABLE_RC=1` cases verifying non-zero exit and WARNING to stderr |
| Idempotent hot-path `enable` check used bare `grep -q 'enable'` (matches `enable --now` and `is-active` log lines) | `test-dispatch-scripts.sh:15754` | Tightened to `grep -qF 'enable --now dispatch-sweep-periodic.timer'`; added positive `is-active` assertion |

Fix commit: `20a59c9a`

---

### Informational

- **`lib.sh:1535–1556`** — input validation guards (newline/space/double-quote/backslash) are production-correct but lacked test coverage; now covered by the Fixed items above.
- **`lib.sh:1648–1663`** — `cat "$SERVICE_PATH"` called twice (hot-path and write-block guard); minor duplication, no correctness concern.

---

### Security

No attack surface: diff is confined to shell-only dispatcher scripts and their tests. No app source, Firestore rules, dependency, or config changes. No open CodeQL alerts on this PR.

Issue #1733, opened 2026-06-15
dispatch: self-heal phase workers that die on transient API rate-limit (server overload) instead of parking on office-hours


natb1:
## Context

When a dispatch phase worker (a `--bg` session named `<N>-slug`) hits the
transient Anthropic API error

```
API Error: Server is temporarily limiting requests (not your usage limit) · Rate limited
```

the session ends mid-phase. This is a server-side overload (HTTP 529-class),
explicitly **not** the user's own usage/quota limit. It is self-healing: the
same work succeeds once the overload passes.

## Current behavior

The error ends the session, so the harness fires Stop →
`.claude/hooks/dispatch-stop.sh`. Because no `phase-completed` marker was
written and the final turn scheduled no `ScheduleWakeup`, the stop lands in
**Branch A** (marker absent), which parks the issue on `dispatch:office-hours`
with reason "phase exited before completion (mid-phase exit or context
compaction)".

Result: a transient, self-healing server overload permanently escalates the
issue to a human, even though re-driving it a minute later would succeed. There
is no live process to hand a `continue` to — once the `--bg` worker errors out
the session is gone (it moves to completed/stopped; `claude agents --json` lists
it only under `--all`), and there is no scriptable "send a prompt to a running
background agent" primitive.

## Proposed design

Recognize this specific transient error in `dispatch-stop.sh` Branch A and,
instead of parking on a human, back off and self-heal by resuming the dead
session with `continue`.

### 1. Detection (in `dispatch-stop.sh` Branch A)

Add a discriminator `session_died_on_rate_limit`, modeled on the existing
`session_scheduled_wakeup`: read the stopping session's transcript
(`$TRANSCRIPT_PATH`, already resolved) and match its tail for the rate-limit
signature. Match the literal `(not your usage limit)` phrasing so the user's
*own* usage-limit exhaustion is **excluded** — that must not auto-retry. The
discriminator is fail-safe: any doubt (empty/unreadable transcript, no match)
returns false and falls through to today's office-hours park unchanged.

### 2. Backoff (new `dispatch-schedule-rate-limit-resume`)

The Stop hook cannot sleep (it is the dying session's own hook), so it arms a
transient `systemd.user` timer, modeled on `dispatch-schedule-target-reseed`:

- Read the highest `dispatch:rate-limit-retry-<n>` counter label on the issue.
- At a cap (default 5) → park on `dispatch:office-hours`
  ("rate-limit retries exhausted after N attempts"), schedule no timer.
[...]

Issue #1590, opened 2026-06-14
dispatch-stop.sh Branch A re-parks a live idle-waiting phase worker on office-hours every poll cycle


natb1:
## Context

Found while diagnosing why #1403 (PR #1548) repeatedly parked on the office-hours
queue. The issue was not actually blocked — its `/review-fix` worker was alive
and healthy, just slow — yet it re-parked on `dispatch:office-hours` every few
minutes.

## Finding

`dispatch-stop.sh` Branch A — the park whose comment reason is `phase exited
before completion (mid-phase exit or context compaction)` (`dispatch-stop.sh:257`)
— fires whenever a worker session ends a turn without a `phase-completed` marker
and CI is ready. It does **not** check whether the worker is still a live session
before parking.

A long-running phase worker that polls a slow background task via `ScheduleWakeup`
goes idle between polls. The Stop hook fires on each idle / turn-end boundary; the
marker isn't written yet (the phase legitimately hasn't finished), so Branch A
re-parks the issue on `dispatch:office-hours`. When the scheduled wakeup fires the
worker resumes, the park is cleared, it polls again, idles again, re-parks. The
`dispatch:office-hours` label oscillates on the worker's poll cadence until the
phase completes.

## Reproduction (observed on #1403 / PR #1548, review phase, 2026-06-14 01:49–02:15Z)

- The `/review-fix` worker ran as a single continuous session (one transcript,
  01:49:45Z → 02:15:53Z) and polled its review Workflow via `ScheduleWakeup` at
  270s intervals ("Dedup/classify agent is running. Checking back in ~2 minutes.").
- `dispatch:office-hours` toggled on that same cadence — labeled at 01:51:52,
  01:57:01, 02:01:39, 02:06:37, 02:11:44, each with the reason `phase exited
  before completion (mid-phase exit or context compaction)`.
- The only sessions in the worktree were this review-fix session and the prior
  qa-fix session, so the re-parks were this live session's own Stop-hook firings,
  not separate dying workers.
- It self-resolved at 02:15:53Z: the worker completed, applied `dispatch:reviewed`,
  undrafted the PR, and wrote its marker. No human action was ever needed.

**Expected:** an alive, idle-waiting worker is not parked; the issue stays in its
active phase until the worker completes or actually dies.

**Actual:** the issue is re-parked on office-hours once per poll cycle, polluting
the office-hours queue with false parks; each park also calls `spawn_tick`, firing
redundant ticks.

## Impact

[...]

Issue #1573, opened 2026-06-14
office-hours rules-test: add authenticated non-member unfiltered list denial test for test-env usage-samples


natb1:
## Context

The test added in #1562 authenticates as \`owner@test.com\` — the same email in the \`sampleDoc.memberEmails\` fixture — confirming that even an authorized member is denied an unfiltered list. However, no test covers an authenticated stranger (non-member) attempting \`getDocs(collection(db, 'office-hours/test/usage-samples'))\`.

The parallel items test suite (line 131-137 of \`rules-test/test/firestore/office-hours.test.ts\`) only covers the owner case for unfiltered list denial too. A Firestore rule misconfiguration that grants list access to any authenticated user regardless of \`memberEmails\` membership would not be caught by the current suite.

## Recommended fix

Add a companion test asserting that an authenticated stranger (e.g. \`stranger@test.com\`) is also denied an unfiltered \`getDocs\` on \`office-hours/test/usage-samples\`. This closes the gap where a rule change granting \`list\` to any authenticated user bypasses the \`memberEmails\` check.

**File:** \`rules-test/test/firestore/office-hours.test.ts\`

## Acceptance criteria

- [ ] A new test \`"denies unfiltered stranger list query on test-env usage-samples collection"\` (or equivalent name) is added to the \`office-hours usage-samples\` describe block
- [ ] The test authenticates as \`stranger@test.com\` (not present in \`sampleDoc.memberEmails\`) and calls \`getDocs(collection(db, 'office-hours/test/usage-samples'))\` with no \`where\` filter
- [ ] The test asserts \`assertFails\` on that \`getDocs\` call
- [ ] The rules test suite passes with the new test included

natb1:
## Plan preface (clean-context execution)

This plan executes in a clean build context (no skill body loaded). The active
step is the **`implement` phase of `/dispatch-propagate`** for issue **#1573**,
in the worktree `/home/n8/natb1/commons.systems/worktrees/1573-office-hours-rules-test-add`
(branch `1573-office-hours-rules-test-add`). Build the single unit below via
`/implement-unit` (one commit), then run the terminal procedure verbatim:

1. **Open the draft PR** (`dangerouslyDisableSandbox: true` — calls `gh`). Write
   the PR body prose to `tmp/pr-body.md` first, then:

   ```bash
   PR_NUM=$(.claude/skills/dispatch-propagate/scripts/dispatch-open-pr \
     1573 \
[...]

PR #2058, opened 2026-06-19
dispatch: recover leaked phase marker from structural side-effects before office-hours park

natb1:
Closes #2025

## Summary

A dispatch phase signals completion by writing a per-session `phase-completed`
marker (via `dispatch-mark-complete`). The Stop hook
(`.claude/hooks/dispatch-stop.sh`) parks the issue on `dispatch:office-hours`
whenever that marker is **absent** ("Branch A"). Because "done" rests on the
model remembering to call `dispatch-mark-complete` as its last step, a
trailing-step leak (the model ends its turn one step early — #824) produces a
**false** office-hours park even though the phase's real work (PR opened,
`dispatch:*` label applied, commits pushed) actually completed. MEMORY records
recurring false parks from this (#1573/#1583), and the phase skills are saturated
with defensive prose to compensate.

This change makes Branch A verify the phase's structural side-effects from
durable state before parking. If the side-effects show the work completed, it
advances the chain (recovering the missing marker) instead of parking a human.
The office-hours park is reserved for the case where the marker is absent **and**
the work genuinely did not complete. It generalizes the existing #1733
rate-limit-death triage, which already special-cases one transient marker-absent
cause in the same Branch A.

## How it works

Durable GitHub state alone cannot distinguish "plan completed → advance" from
"implement crashed before opening a PR → park": both look like `dispatch:planned`
present + no PR. The hook needs to know **which phase the session was dispatched
to run**. That phase is not recorded durably at launch, but it does survive in
the session transcript's first user record, which carries a structured
`<command-name>/<skill></command-name>` tag.

Once the dispatched phase `D` is recovered, the completion check reuses the exact
logic Branch B already uses:

> phase `D`'s side-effects are present  ⟺  `dispatch-phase` has advanced past `D`
> ⟺  `D != CURRENT_PHASE`

Walking the cases (`CURRENT_PHASE` is derived from durable state, already computed
before Branch A):

- plan completed → `dispatch:planned` set → `CURRENT_PHASE=implement` ≠ plan → **advance**
- plan crashed (no label) → `CURRENT_PHASE=plan` == plan → **park**
- implement completed → draft PR exists → `CURRENT_PHASE` ≠ implement → **advance**
- implement crashed (no PR) → `CURRENT_PHASE=implement` == implement → **park**
- qa completed → `dispatch:qa-done` set → `CURRENT_PHASE=review` ≠ qa → **advance**
- review completed → `dispatch:reviewed` set → `CURRENT_PHASE=done` ≠ review → **advance**
- fix-checks fixed CI → `CURRENT_PHASE` ≠ fix-checks → **advance**
- fix-checks still failing → `CURRENT_PHASE=fix-checks` == fix-checks → **park**

This comparison form has two safety properties:

1. **It can never skip a required phase.** The completion labels
   (`dispatch:qa-done`, `dispatch:reviewed`) gate `dispatch-phase`'s forward
   progress. "Advance" only does `spawn_tick`, which re-derives the phase via
   `dispatch-route`; if a completion label is missing, `dispatch-phase` routes
   *back* to that phase. So advance means "don't park; let the chain re-derive" —
   it cannot leapfrog qa or review.
2. **It dissolves the fix-\* special-casing.** No per-phase side-effect table and
   no fix-\* exclusion are needed; the comparison handles fix-checks/fix-conflicts
   correctly (fixed → advance, no-progress → park). The existing Branch C
   attempt-counter logic (the marker-present fix-\* case) is untouched.

**One conscious tradeoff:** dispatched=qa, CI breaks mid-phase (e.g. an
intervening main merge), no `dispatch:qa-done` → `CURRENT_PHASE=fix-checks` ≠ qa →
this design **advances** (routes to fix-checks) rather than parking. Advancing is
the better behavior and aligned with the anti-false-park intent: it routes to the
autonomous fix-checks handler instead of pulling a human for a CI failure, and qa
re-runs once CI is green (gated by the absent `dispatch:qa-done`). No phase is
skipped, no human is falsely pulled. This is intentional and noted in the code
comment.

## Changes

- **`dispatch-recover-dispatched-phase`** (new script) — parses a transcript's
  first user record for the dispatched slash command (`grep -m1` on the
  `command-name` tag so a later sub-skill cannot shadow the top-level command,
  with a `.claude/skills/<skill>` path fallback) and maps it to a `dispatch-phase`
  value matching `dispatch-route` exactly. Positive-evidence-only: any doubt →
  no output, non-zero exit (the hook then parks). Mirrors the structure and
  fail-safe discipline of `dispatch-detect-rate-limit-death`.
- **`dispatch-stop.sh` Branch A** — inserts the recovery check after the #1733
  rate-limit self-heal and before the default office-hours park, preserving the
  #1590 idle-poller and #1733 precedence. When the recovered dispatched phase
  differs from `CURRENT_PHASE`, it runs the exact Branch B advance set
  (`clear_rate_limit_retry_labels`, `strip_office_hours_label`, `spawn_tick`,
  `spawn_sweep`, `self_close`) instead of parking.
- **`test-dispatch-scripts.sh`** — unit tests for the recovery script's
  skill→phase mapping and `grep -m1` first-match discipline, plus Branch A wiring
  tests covering false-park recovery (per-phase advance), genuine mid-phase exits
  still parking, recovery failure still parking, fix-checks no-progress parking,
  and the #1590 idle-poll gate retaining precedence.

## Verification

- `run-unit-tests.sh` passes (3227/3227), including the new cases.


natb1:
## QA Summary — Issue #2025 / PR #2058

### Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Unit test suite passes (primary gate) | shell-command | ✅ PASS |
| 2 | `dispatch-recover-dispatched-phase` is executable and well-formed | shell-command | ✅ PASS |
| 3 | Branch A recovery call placed after #1733 block and before default park | shell-command | ✅ PASS |
| 4 | Recovery is cheap — no model/network invocation | shell-command | ✅ PASS |
| 5 | Regression coverage for false-park-recovered and genuine-park cases | shell-command | ✅ PASS |
| 6 | Live end-to-end behavior in production | deferred | — |

### Deferred to Office-Hours

| # | Title | Note |
|---|-------|------|
| 6 | Live end-to-end behavior in production | Planned-deferral AC — production false-park rate drop only measurable via live dispatch traffic + operator monitoring across multiple ticks after merge; not reproducible in the unit harness (which fakes the recovery script, dispatch-phase, and label state). |

### Bugs Found

None.

### Disposition Triage

| 6 | Live end-to-end behavior in production | needs-human | n/a (planned-deferral) | The five script-verifiable ACs (#1–5) are all covered by QA items that PASSED. Item 6 covers downstream production observation — false-park rate across live dispatch traffic — which the unit harness (with faked stubs) cannot verify. Not a code defect; not opus-fixable. Deferred to office-hours for operator monitoring post-merge. |

**Outcome:** Item 6 (needs-human, planned-deferral) is a user-input blocker. Escalating to office-hours. `/office-hours` will walk through the production monitoring item once the PR merges and traffic is observed.

natb1:
## Review — PR #2058

**3 findings fixed · 4 informational · 2 false-positive · 0 deferred · 0 security follow-ups**

Fix commit: fd4b1e6c

---

### Fixed

- **`dispatch-recover-dispatched-phase:53-56`** (code-review + review) — Fallback skills-path grep was scoped to the entire transcript file, allowing a later sub-skill invocation to shadow the top-level dispatched skill. Fixed: restricted the fallback `grep` to the first JSONL record (`head -n1`) so only the dispatched session's first user message is searched. Two new fallback tests added (happy path + later-record-ignored).

- **`test-dispatch-scripts.sh`** (code-review) — The existing `#2025` advance-pair tests only covered the main chain phases; fix-phase pairs and a backwards-phase test were absent. Fixed: added `fix-checks→qa` and `fix-conflicts→implement` advance pairs to the loop, and a new test `#2025-7` asserting that a backwards phase (`qa→implement`) parks rather than advances. The backwards test is intentionally red until the `!=` → forward-only hook fix lands.

---

### Informational

- **`dispatch-stop.sh:406`** (code-review + red-team) — The recovery condition uses strict inequality (`!= CURRENT_PHASE`) rather than a forward-only phase-advancement check. A backwards dispatch (e.g. dispatched=qa, current=implement) would advance rather than park. This is a known conscious tradeoff documented in the PR body; the backwards-phase test `#2025-7` (added above) marks it red and will gate a follow-up fix.

- **`dispatch-recover-dispatched-phase:54`** (code-review) — The fallback grep pattern `\.claude/skills/[a-z-]+` matches `implement-unit` before `implement` when both appear in the first record. Now that the fallback is scoped to the first JSONL record, the exposure is narrowed; the pattern still has a prefix-match hazard but only within a single line. Low risk given the fix above.

- **`dispatch-stop.sh:408`** (review) — `strip_office_hours_label` is called in the recovery-advance path before `dispatch-apply-office-hours` has run in the same invocation. The label may not be present, making the strip a no-op. This is harmless (stripping an absent label is safe) but the comment could be clearer that this is defensive cleanup.

- **`dispatch-stop.sh:406`** (review) — The `[ -n "$CURRENT_PHASE" ]` guard correctly prevents advancing on `dispatch-phase` failure (network error → empty string). Documented behavior; no change needed.

---

### False-positive (security)

- **`dispatch-recover-dispatched-phase:40-56`** (input-validation) — Script accepts an arbitrary file path as `$1`. The caller (`dispatch-stop.sh`) constructs the path from `jq`-extracted JSON, not from user-controlled input; the attack surface is the dispatch system itself, not external callers. Not a real injection risk in context.

- **`dispatch-stop.sh:405`** (input-validation) — `TRANSCRIPT_PATH` extracted from Stop hook stdin via `jq -r '.transcript_path'` without path constraint validation. The Stop hook runs in a trusted context (Claude Code harness); the transcript path is harness-controlled. Not actionable.

natb1:
# Fix-checks summary — PR #2058

## Iteration 1

- **Failed checks:** `hook-tests`, `unit-tests` (both ran the same suite,
  `test-dispatch-scripts.sh` / `pr-scripts`)
- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`
- **Failure excerpt:**
  ```
  Test: #2025 stop recovery: dispatched=qa, current=implement (backwards) → genuine park, no advance
    FAIL: #2025 qa→implement backwards-park: office-hours applied to issue 123 + non-empty reason
      apply-log:
    FAIL: #2025 qa→implement backwards-park: self-close NOT invoked (no false advance)
  ...
  FAIL: test-dispatch-scripts.sh
  Failed suites: pr-scripts
  ```
- **Why not caught:** The PR's own review-fix commit (`fd4b1e6c`) added a
  forward-only regression guard to `test-dispatch-scripts.sh` whose accompanying
  comment states it stays RED "until that sibling hook fix lands in the same
  commit." The sibling hook fix never landed: `.claude/hooks/dispatch-stop.sh`
  still advanced the chain on any `DISPATCHED_PHASE != CURRENT_PHASE` inequality,
  so a *backwards* main-chain move (dispatched=qa re-deriving to implement after
  a PR regressed) was wrongly treated as a completion → false advance instead of
  the required office-hours park. Not a flake and not infra — a deterministic
  logic gap in this PR's own change.
- **Fix:** Added a `phase_advanced_past` helper to `dispatch-stop.sh` and
  replaced the bare `!=` comparison in the Branch A recovery block with it. The
  helper: equal phases park; any change touching an off-chain fix-* remediation
  phase (`fix-conflicts`/`fix-checks`) advances (preserving the documented
  "CI broke mid-qa → route to fix-checks, don't park a human" tradeoff); moves
  between main-chain phases (`plan<implement<qa<review<done`, the canonical order
  with the fix-* phases removed) advance only when strictly forward. Commit
  `3e6de4d9`. `test-dispatch-scripts.sh` now passes 3365/3365.

PR #2076, opened 2026-06-19
Test-integrity sensor: forbid deleting/skipping tests to pass CI

### LINKED ISSUE #2062: Test-integrity sensor: forbid deleting/skipping tests to pass CI
Part of epic #2061 (code-quality decay sensors).

## Problem
No rule or mechanical check forbids deleting or skipping tests to make CI green.
dispatch runs autonomous green-the-CI lanes (`qa-fix`, `fix-checks`) — exactly the
agent most tempted to weaken a failing test instead of fixing the code. This is
Osmani's flagship "ratchet" example and it is absent here (verified: empty grep for
test-deletion/`.skip(`/`xit(` prohibitions across `.claude/rules/` and the qa-fix /
fix-checks / implement-unit / review-fix skills).

## Proposed change
- Add `.claude/rules/test-integrity.md` (~5 lines): removing or weakening a test to
  pass CI is prohibited; fix the code or escalate.
- Back it with a mechanical guard: flag *net* test deletions / new `.skip(` / `xit(`
  in a green-fixing PR — either a grep check in the test wrapper or a `/review-fix`
  finder.

## Acceptance criteria
- [ ] `.claude/rules/test-integrity.md` exists and states the prohibition concisely.
- [ ] A mechanical guard flags net test deletions / skips introduced by a PR.
- [ ] The guard is diff-scoped (net change vs origin/main), not absolute.

natb1:
Closes #2062

## What

Adds a test-integrity sensor that makes weakening tests to pass CI mechanically
detectable and blocking — part of epic #2061 (code-quality decay sensors). Two
complementary backstops:

1. **Prose prohibition** — `.claude/rules/test-integrity.md`. A concise rule that
   auto-loads as a project instruction into every Claude Code session, including
   the autonomous `qa-fix`/`fix-checks` orchestrator sessions most tempted to
   delete or skip a failing test. It frames a red test as a signal: fix the
   underlying code or escalate to office-hours — never disable the test.

2. **Blocking CI guard** — `.github/scripts/check-test-integrity.sh`, run by a new
   always-on `test-integrity` job in `unit-tests.yml`. It diffs `origin/main...HEAD`
   restricted to TS/JS test globs and flags three signals, each measured **net**
   across the whole diff (a test moved between files nets to zero):

   - new disabling-skips (`it.skip("title")` / `xit` / `xdescribe`),
   - net test-declaration removals (also catches the `it(` → `it.skip(`
     conversion case),
   - whole test-file deletions.

   On a violation the script exits 1 with a loud stderr remediation block steering
   toward *restore-or-escalate*. Because dispatch's auto-merge readiness classifies
   the full status-check rollup, a `FAILURE` here blocks promotion-to-ready /
   auto-merge with no branch-protection change, and routes the PR back through
   `fix-checks` — which reads that remediation text.

## Why prose alone is not enough

The `/implement-unit` fixer subagent does not inherit project instructions, so the
prose rule cannot reach it. The mechanical guard is the enforcement; the prose
steers the orchestrator that *does* inherit it.

## Correctness notes

- The Signal 1 discriminator requires a string-literal first argument (a quote
  char right after `.skip(`), so Playwright runtime-conditional skips like
  `test.skip(testInfo.project.name !== "desktop", "x")` are **not** false-flagged.
- All `grep -c` counts are guarded against `set -e` (zero matches is the common
  case for a non-test PR and would otherwise abort the script and redden every PR).
- The base diff errors clearly rather than silently narrowing to `HEAD~1`
  (per `.claude/rules/code-style.md`); CI checks out with `fetch-depth: 0`.

## Tests

`.github/scripts/test-check-test-integrity.sh` (wired into the `hook-tests` job)
runs 16 hermetic assertions over synthetic git repos covering all six cases: a
Playwright conditional skip (no flag), an `it.skip` conversion (flag), a removed
declaration (flag), a whole-file deletion (flag), a cross-file move (no flag), and
a non-test-touching PR (clean exit 0 — exercises the `set -e`/grep-zero-match
guards). All pass; the guard also exits 0 silently on this branch.

## Out of scope (follow-up notes)

`.only(` focus detection and Go `*_test.go` skip semantics are deliberately out of
AC scope. A legitimate test consolidation that nets to fewer declarations will
false-positive and route to the rule's fix-or-escalate path — intentional; there
is no self-serve escape hatch.

### DIFF
--- a/.claude/rules/test-integrity.md
+++ b/.claude/rules/test-integrity.md
@@ -1,0 +1,12 @@
+# Test Integrity
+
+## A failing test means fix the code or escalate — never weaken the test
+
+When a test fails, fix the underlying code. Removing, skipping (`.skip`, `xit`,
+`xdescribe`, `test.skip`, `it.skip`), commenting out, or otherwise disabling or
+weakening a test to make CI green is forbidden. A red test is a signal, not an
+obstacle — suppressing it destroys the signal without fixing the problem.
+
+If the underlying code genuinely cannot be fixed within the current session,
+escalate to office-hours. There is no self-serve escape hatch: the test stays
+enabled, and the work is parked until a fix is viable.
--- a/.github/scripts/check-test-integrity.sh
+++ b/.github/scripts/check-test-integrity.sh
@@ -1,0 +1,174 @@
+#!/usr/bin/env bash
+# Check for test-integrity violations in the current PR vs origin/main.
+#
+# Three signals — all measured as NET changes across the whole diff so a
+# test moved between files within a single PR nets to zero:
+#
+#   Signal 1 — Disabling-skip additions (net > 0 → violation)
+#     Added lines matching (it|test|describe).skip("title"…) or (xit|xdescribe)(
+#     MINUS removed such lines.
+#     Playwright runtime-conditional skips like test.skip(testInfo.project…) are
+#     NOT flagged because their first arg is an expression, not a string literal —
+#     the discriminator is a quote character immediately after the opening paren.
+#
+#   Signal 2 — Test-declaration removals (net removed > 0 → violation)
+#     Removed lines matching (it|test|describe)[.each(...)]( MINUS added such
+#     lines. A net-negative count means fewer test declarations → flag.
+#     This also catches the common skip-CONVERSION: removing `it("x", …)` while
+#     adding `it.skip("x", …)` — the removed line matches; the added line does
+#     NOT (`.skip` breaks the bare-call pattern).
+#
+#   Signal 3 — Whole test-file deletion
+#     Any test file (*.test.ts, *.test.tsx, *.spec.ts, *.spec.tsx) fully deleted
+#     in this PR.
+#
+# Exit contract: 0 = clean, 1 = violation. Silent on clean pass.
+# On violation: cat >&2 <<EOF remediation block (mirrors check-playwright-version-sync.sh).
+#
+# Requires: git, grep, bash (no node/nix/npm needed — git diff + grep only).
+set -euo pipefail
+
+# Test file globs passed as pathspecs to git diff.
+TEST_GLOBS=(
+  '*.test.ts'
+  '*.test.tsx'
+  '*.spec.ts'
+  '*.spec.tsx'
+)
+
+# ---------------------------------------------------------------------------
+# Capture diff — exit clearly on failure (never silently narrow to HEAD~1).
+# ---------------------------------------------------------------------------
+DIFF=""
+if ! DIFF=$(git diff --unified=0 origin/main...HEAD -- "${TEST_GLOBS[@]}" 2>&1); then
+  cat >&2 <<EOF
+ERROR: check-test-integrity: 'git diff origin/main...HEAD' failed.
+
+This is a required blocking gate. The diff against origin/main must be
+computable. CI checks out with fetch-depth: 0 so origin/main is always
+present. If this is a local run, fetch first:
+
+  git fetch origin main
+
+Raw error from git:
+$DIFF
+EOF
+  exit 1
+fi
+
+# Empty diff — nothing touched test files. Clean pass.
+if [ -z "$DIFF" ]; then
+  exit 0
+fi
+
+# ---------------------------------------------------------------------------
+# Extract added/removed lines (excluding diff header lines +++ / ---).
+# The || true guards protect set -e when grep finds no matches.
+# ---------------------------------------------------------------------------
+ADDED_LINES=$(printf '%s\n' "$DIFF" | grep '^+' | grep -v '^+++' || true)
+REMOVED_LINES=$(printf '%s\n' "$DIFF" | grep '^-' | grep -v '^---' || true)
+
+# ---------------------------------------------------------------------------
+# Signal 1: Disabling-skip additions
+#   Pattern A: (it|test|describe).skip( followed by a string-literal first arg
+#              (quote char immediately after the open paren).
+#   Pattern B: (xit|xdescribe)(
+#
+# The quote-char discriminator ['"`] after the open paren excludes
+# test.skip(testInfo.project.name !== "desktop", "desktop only") — there,
+# the character after '(' is 't', not a quote.
+# ---------------------------------------------------------------------------
+SKIP_PAT_A='(it|test|describe)\.skip[[:space:]]*\([[:space:]]*["'"'"'`]'
+SKIP_PAT_B='\b(xit|xdescribe)[[:space:]]*\('
+
+SKIP_ADDED=0
+SKIP_REMOVED=0
+
+if [ -n "$ADDED_LINES" ]; then
+  SKIP_ADDED=$(printf '%s\n' "$ADDED_LINES" | grep -cE "$SKIP_PAT_A|$SKIP_PAT_B" || true)
+fi
+if [ -n "$REMOVED_LINES" ]; then
+  SKIP_REMOVED=$(printf '%s\n' "$REMOVED_LINES" | grep -cE "$SKIP_PAT_A|$SKIP_PAT_B" || true)
+fi
+
+SKIP_NET=$((SKIP_ADDED - SKIP_REMOVED))
+
+# ---------------------------------------------------------------------------
+# Signal 2: Test-declaration removals
+#   Matches: it( / test( / describe( / it.each(...)( / test.each(...)(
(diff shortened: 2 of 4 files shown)


natb1:
# QA Summary — #2062 (PR #2076)

## Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | Rule file exists and states prohibition concisely | shell | PASS |
| 2 | Guard test suite passes (16 hermetic assertions) | shell | PASS |
| 3 | Guard flags net disabling-skip (Signal 1) — end-to-end | shell | PASS |
| 4 | Guard is diff-scoped vs origin/main, not absolute | shell | PASS |
| 5 | CI workflow wires guard as always-run job and test into hook-tests | shell | PASS |
| 6 | Guard actually blocks auto-merge in live CI | — | DEFERRED |

**5 PASS · 0 FAIL · 0 SKIP · 1 deferred**

No browser walkthrough (no browser component detected — changes are scripts and CI config only).

- **AC #1** `.claude/rules/test-integrity.md` exists and states prohibition — **PASS** (item 1)
- **AC #2** Mechanical guard flags net test deletions / skips — **PASS** (items 2, 3; 16/16 hermetic assertions including skip-conversion, removed-declaration, whole-file-deletion cases)
- **AC #3** Guard is diff-scoped (net vs origin/main) — **PASS** (items 4; pre-existing skips on main don't trigger the guard)

## Deferred to office-hours

**Item 6 — Guard actually blocks auto-merge in live CI (planned deferral)**
Verifying the live red-CI / auto-merge-block path requires intentionally committing a test-weakening change to this PR, which contradicts the PR's own clean state and the test-integrity rule. Items 2–4 prove the guard's exit-1 behavior hermetically via synthetic repos; item 5 proves the job is wired unconditionally. Observing the actual auto-merge block is a production / CI observation deferred to a maintainer.

## Disposition triage (Opus assess)

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR. *(No `needs-main` items this pass.)*

| 6 | `needs-human` | n/a | Planned deferral — AC is non-assertable at merge time; Workflow correctly excludes from skeptic fan-out |

**Non-fixing pass:** The `needs-human` residue escalates to office-hours. No code defect exists — this is a deliberate planned deferral, not a bug.

natb1:
## Review — PR #2076 (`2062-test-integrity-sensor-forbi`)

### Fixed (6)

**code-review-0 · `.github/scripts/check-test-integrity.sh:102`**
DECL_PAT used `[^)]*` inside the `.each(...)` optional group, which stops at the first `)`. This caused `it.each([1,2])(` to be truncated and mis-counted, and template-literal `.each` forms to break entirely. Fix: replaced the optional `.each` group with an OR of the plain bare-call pattern and a separate `.each`-prefixed pattern so all `.each` forms are counted alongside plain calls. (fix commit: 73e08dc1)

**code-review-1 · `.github/scripts/check-test-integrity.sh:68–69`**
Signal 2 counted textual `it(`/`test(`/`describe(` occurrences on raw added/removed diff lines, so a PR that _comments out_ a test — replacing `it(` with `// it(` — counted one removed and one added declaration (DECL_NET = 0) and passed clean. Fix: strip full-line `//` comment lines from both ADDED_LINES and REMOVED_LINES before applying DECL_PAT, so commented-out declarations are not counted as live. (fix commit: 73e08dc1)

**code-review-2 · `.github/scripts/check-test-integrity.sh:81`**
SKIP_PAT_A was missing a `\b` word-boundary anchor before `(it|test|describe)`, allowing identifiers ending with those substrings (e.g. `emit.skip(`) to be incorrectly matched. Fix: added `\b` to match SKIP_PAT_B. (fix commit: 73e08dc1)

**review-1 · `.github/scripts/check-test-integrity.sh:163–170`**
The remediation block said "ESCALATE to office-hours if the test is _genuinely obsolete_ or the fix requires scope beyond this PR," but `.claude/rules/test-integrity.md` (the companion prose rule) does not grant a "test is obsolete" escape hatch. Fix: removed "genuinely obsolete or" so both channels give the same guidance. (fix commit: 73e08dc1)

**review-2 · `.github/scripts/test-check-test-integrity.sh:116–138`**
The test suite covered `it.skip(` (Signal 1 Pattern A) but had no cases for `describe.skip(`, `test.skip(`, `xit(`, or `xdescribe(`. Fix: added four synthetic-repo test cases (one per variant), each asserting exit 1 and Signal 1 in stderr. Test count is now 28/28. (fix commit: 73e08dc1)

**red-team-1 (Required, Upheld) · `.github/scripts/check-test-integrity.sh:68`**
Confirmed bypass: replacing `it('name', fn)` with `// it('name', fn)` produced DECL_NET = 0 because `\b` in DECL_PAT matches at the space before `it` in `// it(` — the removed active line and the added comment line each matched, netting to zero. All three signals missed it. `.claude/rules/test-integrity.md` explicitly names "commenting out" as forbidden alongside removing and skipping. Fixed by the comment-line stripping applied in the code-review-1 fix above (same commit). (fix commit: 73e08dc1)

---

### Required — Refuted by adversarial verify (1)

**red-team-0 · `.github/scripts/check-test-integrity.sh:102`**
Claim: DECL_PAT does not match `test.concurrent('name', fn)` or `test.concurrent.skip('name', fn)`, leaving a skip-conversion bypass for concurrent-modifier chains.
Verdict: **Refuted** (skeptic votes: 1 upheld / 1 refuted). Both technical claims are empirically correct — `test.concurrent(` and `test.concurrent.skip(` do not match the current patterns. However, `test.concurrent` does not exist anywhere in the 506-file test suite (confirmed by grep). Because there are no such declarations to delete or disable, neither attack vector has an exploit path in the current codebase. The gap is a real abstract pattern-coverage hole but does not cross the exploitability threshold; it is intentional scope — unlike the Playwright runtime-conditional exclusion which is documented in the script header as a deliberate design boundary, `.concurrent` coverage is a known limitation.

---

### Informational (2)

**review-0 · `.github/scripts/check-test-integrity.sh:102`**
DECL_PAT counts `describe(` alongside `it(` and `test(`. A legitimate refactoring that collapses multiple `describe` blocks into one will trip Signal 2 (net negative) even with no test loss. This is a known false-positive class; the prose rule's "escalate if fix requires scope beyond this PR" path handles it without a CI escape hatch. No change required.

**review-3 · `.github/scripts/test-check-test-integrity.sh:149–176`**
Case (b) uses a `*.test.ts` file but does not explicitly verify that `*.spec.tsx` and `*.test.tsx` globs are also exercised by the check. The `TEST_GLOBS` array covers all four extensions; the gap is test-suite coverage breadth, not a script defect. No change required.

---

### Dismissed (2)

**review-4 · `.github/workflows/unit-tests.yml:140–147`**
The `test-integrity` job's `actions/checkout` step is already pinned to the same commit hash (`11bd71901bbe5b1630ceea73d27597364c9af683`) used in the rest of the file — finding was a misread.

**review-5 · `.github/scripts/check-test-integrity.sh:68–69`**
Claim that a diff hunk containing a `+`/`-` prefix on an interior line would be misclassified. Unified diff format guarantees `+`/`-` at column 0 are added/removed markers; the `grep '^+'`/`grep '^-'` filters are correct by the format spec.

---

### False-positive — Security (2)

**red-team-2 · `.github/scripts/check-test-integrity.sh:108–114`**
Signal 2 counts declaration lines, not body content — an attacker could empty the body while keeping the declaration. This is a claimed gap but is not a test-integrity bypass: the test still runs and fails (or passes on empty assertions). It is a different failure mode (`always-green hollow test`) that is already traceable via test reports. Not a bypass of the signal the guard is designed to catch.

**red-team-3 · `.github/scripts/check-test-integrity.sh:82`**
Claim: SKIP_PAT_B misses `xtest`, Vitest's alias for `test.skip`. The codebase contains no `xtest` usage (confirmed by grep); this pattern is not used in this project's test suite. The scope decision to cover `xit`/`xdescribe` without `xtest` is consistent with the codebase's actual Vitest API surface.

---

### Out-of-scope — Security (1)

**red-team-5 · `.github/workflows/unit-tests.yml:140–147`**
The `test-integrity` job triggers only on `push` (branches-ignore: main), not on `pull_request` events. If the repo allowed external contributors to open PRs without push access, the gate would not run on their PRs until merge. Pre-existing pattern for all jobs in this file; not introduced by this change and the project's contributor model is first-party. No follow-up filed (below meaningfulness threshold for this codebase's access model).

PR #2077, opened 2026-06-19
Type-safety escape-hatch CI sensor: flag net-new any/ts-ignore/eslint-disable/as/!

natb1:
Closes #2065

## What

Adds the first code-quality decay sensor from epic #2061: a deterministic,
diff-scoped CI check that flags **net-new** type-safety escape hatches a PR
introduces versus `origin/main`. The existing sensors verify correctness
(type/lint/test/CI); nothing measured the type-level analog of deleting a test —
an agent reaching for `any`, `@ts-ignore`, `@ts-expect-error`, an
`eslint-disable`, a broad `as <Type>` cast, or a non-null `!` to satisfy the
checker. The autonomous CI-fixing lanes (`qa-fix`, `fix-checks`) are tempted to
do exactly that.

A casual escape hatch now fails a fast, zero-token CI check, reported at
`file:line` on the PR diff. A genuinely-needed one is kept via a one-line,
diff-visible `// type-safety-ok: <reason>` suppression marker — converting a
casual choice into a deliberate, reviewed one.

## How

Three units:

1. **Core scanner + git wrapper** — `.github/scripts/check-type-safety-escapes.sh`.
   An awk core reads a unified diff on stdin (`--scan-stdin`, the testable entry
   point), walks `+++ b/<path>` and `@@ +c,d @@` headers to track the new-file
   line number, and inspects only added (`+`) lines. Per line it checks the
   suppression marker first, then comment-directive rules against the full line
   (`@ts-ignore`/`@ts-expect-error`/`eslint-disable`), then code rules against a
   `//`-stripped line (`any` in type position, `as <Type>` cast, non-null `!`).
   The default invocation is a thin git wrapper:
   `git diff --no-color -U0 --diff-filter=d origin/main...HEAD -- '*.ts' '*.tsx' '*.js' '*.jsx' '*.mjs' '*.cjs'`
   piped into the core. It hard-errors if `origin/main` is unresolvable (no
   `HEAD~1` fallback — CI uses `fetch-depth: 0`) and self-noops (exit 0) when the
   scoped diff is empty. Mirrors `check-playwright-version-sync.sh` idioms.

2. **Fixture-based unit test** —
   `.github/scripts/test-check-type-safety-escapes.sh`, mirroring
   `test-firebase-auth.sh` (pass/fail counters, subshell isolation, mktemp+trap,
   expected-total guard). 15 cases pipe synthetic unified diffs into the
   `--scan-stdin` core: one positive per pattern (incl. `as any` asserted as a
   single finding, not double-counted), exclusions (`as const`, import aliases,
   `a !== b`, `!!x`, leading `!foo`, `many`/`Company`/`anything`, comment-only
   hatches), suppression (non-empty reason suppresses, empty reason still flags),
   the diff-scoped property (a pre-existing hatch on an unchanged line is not
   flagged), and awk line-number bookkeeping.

3. **CI wiring** — `.github/workflows/unit-tests.yml`: a new `type-safety-sensor`
   job (checkout with `fetch-depth: 0`, no setup-node, runs **unconditionally** so
   it can't be skipped — the script self-noops on an empty scoped diff), plus a
   `hook-tests` step running the fixture test. This is the **first**
   `.github/scripts/test-*.sh` wired into CI (the existing
   `test-cloudflare-purge.sh` / `test-firebase-auth.sh` are unwired) —
   establishing the wiring.

## Design notes

- **Diff-scoped, stateless, zero-token:** computed purely from git at HEAD — no
  persisted store — and only flags additions; pre-existing hatches on untouched
  lines never block. It's a plain CI check, never inside the headless
  dispatch-tick.
- **`--diff-filter=d`** (all changes except pure deletions), not `=AM`: covers
  renames/copies, so `git mv foo.ts bar.ts` while adding an `any` in the same
  commit can't slip through — the threat model is "agents route around the
  check."
- **False-positive control:** the `any`/`as`/`!` rules strip trailing `//`
  comments first; import/export alias lines are excluded from the `as` rule;
  `as const` and `!=`/`!==`/`!!`/leading-`!` are deliberately not matched.
  Residual FPs (string literals, block comments) are covered by the suppression
  marker.
- **Epic-wide convention:** `// type-safety-ok: <reason>` is the shared
  suppression-marker family for all six decay sensors (#2062–#2067); siblings
  should adopt the same `// <sensor>-ok: <reason>` shape. Recorded in the script
  header.

## Testing

- `.github/scripts/test-check-type-safety-escapes.sh` — 15/15 pass locally.
- Sanity-checked the core against hand-written diffs (one-of-each → one
  annotation per hatch, correct line numbers, exit 1; clean additions → exit 0;
  exclusions don't fire; suppression works) and the wrapper (empty scope → exit
  0; unresolvable origin/main → hard error).

### DIFF
--- a/.github/scripts/check-type-safety-escapes.sh
+++ b/.github/scripts/check-type-safety-escapes.sh
@@ -54,15 +54,22 @@ scan_diff() {
   awk '
     BEGIN { violations = 0 }
 
-    # File header for the new side: "+++ b/<path>" (note the trailing space,
-    # which distinguishes it from an added source line beginning "+++").
-    /^\+\+\+ / {
+    # Old-side header "--- a/<path>" arms the new-side header rule. A line
+    # beginning "+++ " is only a file header when it immediately follows a
+    # "--- " line; otherwise it is an added source line (e.g. "++ x; const y =
+    # z as any;") that must be scanned, not consumed as a header.
+    /^--- / { in_diff_header = 1; next }
+
+    # File header for the new side: "+++ b/<path>", only when armed by a
+    # preceding "--- " line.
+    in_diff_header && /^\+\+\+ / {
       path = substr($0, 5)            # strip "+++ "
       sub(/^b\//, "", path)           # strip the "b/" prefix
       if (path == "/dev/null") path = ""
+      in_diff_header = 0
       next
     }
-    /^--- / { next }                  # old-side header; ignore
+    { in_diff_header = 0 }            # any other line disarms the header rule
 
     # Hunk header: "@@ -a,b +c,d @@" or "@@ -a +c @@". The new-file start line
     # is the number after the "+".
@@ -84,8 +91,12 @@ scan_diff() {
     # Removed line: does NOT advance the new-file counter.
     /^-/ { next }
 
-    # Anything else (context line " ", "\ No newline...", etc.) advances the
-    # counter. With -U0 there are no context lines, but handle them anyway.
+    # "\ No newline at end of file" marker: not a real source line, so it must
+    # NOT advance the new-file counter.
+    /^\\/  { next }
+
+    # Anything else (context line " ", etc.) advances the counter. With -U0
+    # there are no context lines, but handle them anyway.
     { line++ }
 
     END { if (violations > 0) exit 1 }
@@ -95,12 +106,12 @@ scan_diff() {
       violations++
     }
 
-    function process(content, path, line,    s, code) {
+    function process(content, path, line,    rest, code) {
       # 1. SUPPRESSION (first). A non-empty reason after the marker suppresses
       #    the whole line. An empty reason does not.
       if (match(content, /\/\/[ \t]*type-safety-ok:/)) {
         rest = substr(content, RSTART + RLENGTH)
-        if (rest ~ /[^ \t]/) return    # non-empty reason -> suppressed
+        if (rest ~ /[^ \t\r\n]/) return    # non-empty reason -> suppressed
         # empty reason -> fall through and keep checking
       }
 
@@ -125,10 +136,9 @@ scan_diff() {
       # `as <Type>` cast. Excludes import/export alias lines and `as const`.
       # `as any` is owned by the any rule above, so exclude lowercase `any`
       # from this alternation.
-      if (!(code ~ /^[ \t]*(import|export)([^A-Za-z0-9_$]|$)/) &&
-          !(code ~ /[ \t]from[ \t]*['"'"'"]/) &&
-          !(code ~ /^[ \t]*[A-Za-z0-9_$]+[ \t]+as[ \t]+[A-Za-z0-9_$]+,?[ \t]*$/)) {
-        if (code ~ /(^|[^A-Za-z0-9_$])as[ \t]+(unknown|string|number|boolean|object|symbol|bigint|[A-Z_$])/) {
+      if (!(code ~ /^[ \t]*(import|export)[^{]*\{/) &&
+          !(code ~ /[ \t]from[ \t]*['"'"'"]/)) {
+        if (code ~ /(^|[^A-Za-z0-9_$])as[ \t]+(unknown|string|number|boolean|object|symbol|bigint|never|null|undefined|void|[A-Z_$])/) {
           emit(path, line, "as <Type> cast")
         }
       }
@@ -147,6 +157,11 @@ scan_diff() {
 if [ "${1:-}" = "--scan-stdin" ]; then
   scan_diff
   exit $?
+fi
+
+if [ -n "${1:-}" ]; then
+  echo "Unknown argument: $1" >&2
+  exit 1
 fi
 
 # Default path: thin git wrapper.
--- a/.github/scripts/check-type-safety-escapes.sh
+++ b/.github/scripts/check-type-safety-escapes.sh
@@ -1,0 +1,184 @@
+#!/usr/bin/env bash
+# Diff-scoped net-new type-safety escape-hatch sensor.
+#
+# Flags type-safety escape hatches that a PR *introduces* on added lines,
+# relative to origin/main. It does NOT flag escape hatches that already exist
+# on origin/main — only net-new ones in the branch's diff. The six hatches:
+#
+#   1. @ts-ignore                (comment directive)
+#   2. @ts-expect-error          (comment directive)
+#   3. eslint-disable[-line|-next-line]  (comment directive)
+#   4. `any` in type position    (: any | as any | <any | any[] | any>)
+#   5. `as <Type>` cast          (as unknown / as string / as Foo / ...; not `as const`)
+#   6. non-null assertion `!`    (foo!.bar, arr[0]! ; not !=, !==, !!x, leading !foo)
+#
+# Suppression: append `// type-safety-ok: <reason>` to an added line. A
+# NON-EMPTY reason suppresses the whole line (no rule runs on it). An empty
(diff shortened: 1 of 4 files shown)


natb1:
# QA Summary — #2065 Type-Safety Escape-Hatch Sensor

**Result: CLEAN PASS** — all 8 items passed, 0 failures, 0 skips.

No browser component (all changes in `.github/scripts/` and `.github/workflows/`). All items ran in the shell-command lane.

## Items executed

| # | Title | Result |
|---|-------|--------|
| 1 | Unit test suite passes (all 15 cases) | ✅ PASS |
| 2 | Sensor script exists and is executable | ✅ PASS |
| 3 | CI workflow contains `type-safety-sensor` job | ✅ PASS |
| 4 | Sensor flags a synthetic `any` escape hatch | ✅ PASS |
| 5 | Sensor exits 0 on a clean diff (no hatches) | ✅ PASS |
| 6 | Suppression marker (`// type-safety-ok: reason`) works | ✅ PASS |
| 7 | Diff-scoped: unchanged context-line hatches are not flagged | ✅ PASS |
| 8 | `type-safety-sensor` CI job is unconditional (no `if:` gate) | ✅ PASS |

**Counts:** 8 passed / 0 failed / 0 skipped

## Acceptance criteria verification

- **AC1** (sensor flags net-new escape hatches): verified — items 1 and 4 confirm all six patterns (`@ts-ignore`, `@ts-expect-error`, `eslint-disable`, `any`, `as <Type>`, non-null `!`) are flagged, and the unit test suite's 15-case fixture suite confirms all positives pass.
- **AC2** (diff-scoped vs origin/main, no pre-existing blocks): verified — items 1 and 7 confirm unchanged context-line hatches are invisible to the scanner.
- **AC3** (findings surface as PR check failure): verified — items 3 and 8 confirm the `type-safety-sensor` job is wired unconditionally in `unit-tests.yml` with no `if:` gate.

natb1:
## Review summary — PR #2077 / #2065

**Fixes applied** in commit 75de1cdb:

- **`+++ ` header misparse** (rt-1): armed `+++ ` detection on a preceding `--- ` line — previously an added source line beginning with `++ ` was consumed as a file-header, clobbering `path` and suppressing all findings in the next file. `.github/scripts/check-type-safety-escapes.sh:59–75`
- **`\ No newline` marker advancing line counter** (cr-2): added `/^\\ { next }` to skip backslash-prefix lines — the `\ No newline at end of file` marker was counted as a source line, skewing all subsequent annotation line numbers. `.github/scripts/check-type-safety-escapes.sh:91–94`
- **`rest` not declared local in awk `process()`** (cr-4 / rv-2): changed function signature from `(content, path, line,    s, code)` to `(content, path, line,    rest, code)` — `rest` was used but leaked as a global; `s` was declared but unused. `.github/scripts/check-type-safety-escapes.sh:106`
- **CRLF suppression bypass** (rt-2): expanded suppression guard from `[^ \t]` to `[^ \t\r\n]` — a suppression comment ending with `\r` (CRLF line ending) had a non-empty "reason" match, preventing bypass of the guard. `.github/scripts/check-type-safety-escapes.sh:110`
- **Missing type keywords in `as <Type>` pattern** (rt-6 / cr-7): added `never|null|undefined|void` to the alternation — `as never`, `as null`, `as undefined`, and `as void` are real type-safety escape hatches that were silently missed. `.github/scripts/check-type-safety-escapes.sh:138`
- **`import` / `export` exclusion over-matching** (rt-4 / rt-5 / rv-1): narrowed exclusion from `(import|export)([^A-Za-z0-9_$]|$)` to `(import|export)[^{]*\{` — only brace-list rename forms are now excluded, so `export const x = y as Foo` is correctly flagged while `export { X as Y }` is still excluded. Also removed the single-identifier rename exclusion that matched trailing-comma casts. `.github/scripts/check-type-safety-escapes.sh:136–139`
- **Sensor absent from `pr-checks.yml`** (rt-7 / sr-4): added `type-safety-sensor` job to `.github/workflows/pr-checks.yml` — the sensor previously ran only on `push` (unit-tests.yml) and was absent from the `pull_request` workflow, so it never blocked PR merges. `.github/workflows/pr-checks.yml`
- **Unknown-argument guard** (iv-3): added explicit `[ -n "${1:-}" ]` check with `exit 1` — an unrecognized flag previously fell through silently to the git-wrapper path. `.github/scripts/check-type-safety-escapes.sh:159–162`
- **Fixture suite expanded** (cr-5): 15 → 18 tests, covering: `export const x = y as Foo` is flagged, `export { X as Y }` is excluded, `\ No newline` marker does not skew the counter. EXPECTED guard updated to 18.

### Deferred (follow-up filed)

- **sr-1** — Suppression marker inside a string literal causes false negative: `// type-safety-ok: <reason>` inside a string on the same line as a real escape hatch suppresses the finding. D2 limitation (line-based scanning without string-literal context). → #2093
- **sr-2** — File paths with commas produce malformed GitHub Actions annotations: `::error file=<path>,line=N::` format uses comma as delimiter; a path like `src/weird,file.ts` would be split incorrectly. → #2092

### Informational

- **cr-6** (code-review): Comment-stripping (`//` first-occurrence) can truncate code before real escape hatches in lines with inline strings — accepted D2 limitation; suppression marker is the documented escape valve.
- **rv-3** (review): The `as <Type>` pattern uses `[A-Z_$]` for user-defined types — a lowercase user type (e.g., `as foo`) is missed. The current scan detects the high-risk cases (built-in types and capitalized custom types); fully lowercased custom-type casts are an accepted D3 gap.
- **rv-4** (review): URL strings containing `://` can cause comment-stripping to drop subsequent code on the same line — same D2 limitation as cr-6; suppression marker applies.
- **rv-5** (review): Multiple hatches on the same line (e.g., `@ts-ignore as Foo`) emit multiple annotations — intended behavior; no change needed.
- **iv-1** (input-validation): `path` in the annotation is derived from git's `+++ b/<path>` header, which `core.quotePath=true` sanitizes for the CI path; `--scan-stdin` has no git sanitization — accepted; the reviewer who pipes the diff controls the input.

### False-positive

- **iv-2** (input-validation): `rest` global bleed risk — fixed by cr-4/rv-2 above; the false-positive that claimed it was already a local was incorrect.
- **rt-9** (red-team): No shell-injection, workflow-command injection, or data-exfiltration vectors found — awk program is a fixed inline string, no `eval`, no shell interpretation of diff content.
- **sr-5** (security-review): `permissions: contents: read` is correctly minimal for the checkout-only type-safety-sensor job.
- **sr-6** (security-review): Temp file handling in the test script uses `mktemp -d` + `trap 'rm -rf "$TEST_TMPDIR"' EXIT` — best practice, no issue.

### Out-of-scope

- **rt-3 / sr-3** (red-team + security-review): `//` inside a URL string literal causes comment-stripping to truncate the line, missing an escape hatch after the URL — accepted D2 limitation documented in the script header.
- **rt-8** (red-team): CI runs the PR branch's own scanner version, so a PR that weakens the scanner can self-approve — inherent to all in-repo CI scripts; defense-in-depth pinning to `origin/main` is a separate, independent concern.
- **sr-7** (security-review): `# main` comment on the `DeterminateSystems/nix-installer-action` SHA pin is cosmetically misleading (suggests mutable-branch tracking) — cosmetic only; SHA is the authoritative pin and is correct.

_Verify pass: all verify agents returned 401 on a transient auth failure; findings were classified by the initial finder wave without adversarial confirmation. The 8 Required findings were re-evaluated manually; 6 were fixed in-scope._

PR #2079, opened 2026-06-19
Promote the jq-echo prose rule into a diff-scoped run-lint check

natb1:
Closes #2067

## What

Promotes the highest-recurrence prose rule — the **jq-echo control-char trap**
(`.claude/rules/shell-json.md`, first bit #1170) — from a convention an agent
must remember into a mechanical check whose failure message injects the
remediation into agent context. Part of epic #2061 (code-quality decay sensors).

## How

- **`lint-prose-rules.sh`** (new) — a standalone, **diff-scoped** linter. It
  self-computes `git diff origin/main...HEAD --unified=0 -- '*.sh'`, recovers
  `path:line` for each *added* line, skips comment lines, and flags the
  documented `echo "$VAR" | jq` anti-pattern. On a match it prints each offending
  `path:line: <content>` followed by the concrete fixes from `shell-json.md` (the
  `<<<` here-string, `--jq`, and `printf '%s'` forms) and a pointer to the rule,
  then exits non-zero — the **remediation-in-error** that puts the fix in front of
  the agent that tripped it.
- **`test-lint-prose-rules.sh`** (new, registered in the `hook-tests` CI job) —
  four ephemeral-repo cases, including the discriminating one that proves the
  check is diff-scoped (a pre-existing violation unchanged at HEAD is **not**
  flagged) and that an added comment mentioning the pattern is **not** flagged.
- **`run-lint.sh`** — wires in a `RUN_PROSE` flag (auto-detected when any changed
  file matches `*.sh`, plus an explicit `--prose`), mirroring the existing
  `RUN_RULES` delegation. So the check runs in the CI `lint` job and in
  `/fix-checks`.
- **`shell-json.md`** — notes the rule is now mechanically enforced.

## Why diff-scoped, not whole-repo

This is the greenfield ideal for a decay sensor (matching the net-new framing of
sibling sensors #2064 / #2065) and is required for correctness: the repo already
contains `echo "$VAR" | jq` sites in `#!/usr/bin/env bash` scripts (e.g.
`.claude/hooks/statusline.sh`), where `echo` does **not** munge escapes — so they
are not latent bugs, and a whole-repo scan would force churn on working code.

## Honest framing

The original #1170 recurrence was *interactive zsh Bash-tool* usage, which is not
a committed file at all. This linter catches the adjacent committed-`.sh` surface
as a **project convention / defense-in-depth** (a bash script can be sourced into
zsh; uniformity aids review) — not as a claim that it fixes a zsh-only bug in
those bash files. The failure message and the `shell-json.md` note say this
plainly.

## Deferred follow-ups (out of scope here)

1. A one-time whole-repo cleanup of the pre-existing `echo "$VAR" | jq` sites
   (`.claude/hooks/statusline.sh`, `.github/scripts/cloudflare-purge.sh`,
   `audio/scripts/upload-media.sh`, `run-acceptance-tests.sh`,
   `run-preview-deploy.sh`).
2. A closing-keyword adjacency check on the PR-body surface in `dispatch-open-pr`
   — the second prose-rule candidate (`issue-references.md`), which operates on
   PR-body / commit-message prose rather than committed files, so it has no
   file-grep surface in `run-lint.sh`.

Both will be filed as `blocked_by` follow-ups during review.

## Verification

- `test-lint-prose-rules.sh`: 10/10 assertions pass.
- `run-lint.sh --prose` correctly fails with the `path:line` + remediation on a
  net-new violation and passes when there is none; the no-op guard still reports
  "No lint targets matched … Nothing to check."

### DIFF
--- a/.claude/rules/shell-json.md
+++ b/.claude/rules/shell-json.md
@@ -38,3 +38,5 @@ PR_NUM=$(printf '%s' "$PR_JSON" | jq -r .number)
 A direct pipe (`gh … --json … | jq`) is also safe — gh's stdout reaches `jq`
 without an `echo` in between. For non-JSON raw text fed to `jq`, use
 `jq --raw-input`.
+
+This rule is mechanically enforced for net-new added lines in committed `.sh` files by `.claude/skills/dispatch-propagate/scripts/lint-prose-rules.sh`, run via `run-lint.sh` in CI; on a match the linter fails with the remediation inline. Enforcement is a project convention for committed scripts (uniformity aids review) — it does not retroactively rewrite pre-existing sites, and it is not a claim that the rule fixes a zsh-only bug in those files. (The original recurrence #1170 was interactive zsh Bash-tool usage, which is not a committed file at all.)
--- a/.claude/skills/dispatch-propagate/scripts/lint-prose-rules.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lint-prose-rules.sh
@@ -8,9 +8,10 @@ SCRIPTS="$(cd "$(dirname "$0")" && pwd)"
 source "$SCRIPTS/lib.sh"
 
 # Detect: echo "$VAR" | jq  (quoted variable form only — see .claude/rules/shell-json.md)
-# Single-quoted to prevent the shell from interpreting \$ as $ (which would make
-# \$ an ERE end-anchor, breaking the match).
-PATTERN='echo[[:space:]]+"\$[A-Za-z_][A-Za-z0-9_]*"[[:space:]]*\|[[:space:]]*jq'
+# Covers optional echo flags (echo -e / echo -n) and the braced ${VAR} form, both
+# of which are the same anti-pattern. Single-quoted to prevent the shell from
+# interpreting \$ as $ (which would make \$ an ERE end-anchor, breaking the match).
+PATTERN='echo([[:space:]]+-[a-zA-Z]+)?[[:space:]]+"\$\{?[A-Za-z_][A-Za-z0-9_]*\}?"[[:space:]]*\|[[:space:]]*jq'
 
 # Compute the unified-0 diff of added lines in .sh files against origin/main.
 # Run from REPO_ROOT so that relative paths in diff output are consistent.
@@ -58,8 +59,7 @@ while IFS= read -r line; do
     content="${line:1}"
 
     # Skip comment lines (first non-whitespace character is #)
-    trimmed="${content#"${content%%[! ]*}"}"
-    if [[ "$trimmed" == '#'* ]]; then
+    if [[ "$content" =~ ^[[:space:]]*# ]]; then
       LINE_NUM=$(( LINE_NUM + 1 ))
       continue
     fi
--- a/.claude/skills/dispatch-propagate/scripts/lint-prose-rules.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lint-prose-rules.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+SCRIPTS="$(cd "$(dirname "$0")" && pwd)"
+
+# shellcheck source=lib.sh
+source "$SCRIPTS/lib.sh"
+
+# Detect: echo "$VAR" | jq  (quoted variable form only — see .claude/rules/shell-json.md)
+# Single-quoted to prevent the shell from interpreting \$ as $ (which would make
+# \$ an ERE end-anchor, breaking the match).
+PATTERN='echo[[:space:]]+"\$[A-Za-z_][A-Za-z0-9_]*"[[:space:]]*\|[[:space:]]*jq'
+
+# Compute the unified-0 diff of added lines in .sh files against origin/main.
+# Run from REPO_ROOT so that relative paths in diff output are consistent.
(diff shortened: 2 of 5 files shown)


natb1:
## QA Summary — #2067

**Result: CLEAN PASS ✓** All 5 acceptance-criterion checks passed.

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Recurring prose rule enforced by mechanical check (AC#1) | shell-command | PASS |
| 2 | Failure message contains remediation instruction — static (AC#2) | shell-command | PASS |
| 3 | Remediation emitted at runtime on violation (AC#2 end-to-end) | shell-command | PASS |
| 4 | Prose rule notes mechanical enforcement (AC#3) | shell-command | PASS |
| 5 | Linter wired into CI — auto-detect and explicit `--prose` | shell-command | PASS |

### Details

- **Item 1** — `test-lint-prose-rules.sh` ran all four cases (10/10 assertions): net-new violation flagged, pre-existing violation not flagged, comment-line not flagged, clean added line passes. AC#1 confirmed: the jq-echo prose rule from `shell-json.md` (#1170) is now mechanically enforced.
- **Item 2** — `lint-prose-rules.sh`'s FAIL path contains all remediation elements: `FAIL: echo-into-jq` header, `jq -r .field <<<"$VAR"` here-string form, `gh ... --json field --jq .field` form, `printf '%s' "$VAR" | jq` form, and `See: .claude/rules/shell-json.md` pointer. AC#2 confirmed.
- **Item 3** — The harness's Test 1 assertions (`remediation pointer present`, `good-form hint present`) pass, confirming the remediation text reaches the operator's terminal on a genuine runtime violation. AC#2 confirmed end-to-end.
- **Item 4** — `shell-json.md` contains "mechanically enforced", `lint-prose-rules.sh`, and `run-lint.sh` in the enforcement note. AC#3 confirmed.
- **Item 5** — `run-lint.sh` has the `*.sh) RUN_PROSE=true` auto-detect case, `--prose` explicit flag, and delegation to `lint-prose-rules.sh`; `unit-tests.yml` has the `test-lint-prose-rules.sh` step in the `hook-tests` job. Linter is live in CI.

No browser walkthrough needed — purely script/CI changes, no frontend component.

No bugs found. No items deferred to office-hours.

natb1:
## Review: PR #2079

**Surface:** `code` (5 files changed, 259 lines). No open CodeQL alerts. No dependency changes.

---

### Fixed (commit `aea775e2`)

- **`lint-prose-rules.sh:61-62` — broken comment-skip logic** (`code-review`, `review`): `trimmed="${content#"${content%%[! ]*}"}"` does not correctly strip leading whitespace in bash (glob `[! ]` is the literal characters `!` and space, not "not space"). Indented comment lines were not skipped, and the linter would incorrectly flag its own `# Detect:` comment on line 10. Fixed by replacing both lines with `if [[ "$content" =~ ^[[:space:]]*# ]]; then`.

- **`lint-prose-rules.sh:10` — self-flagging comment** (`code-review`): Because comment detection was broken, the linter would flag its own `# Detect: echo "$VAR" | jq` comment on line 10 as a violation. Resolved by the comment-skip fix above.

- **`test-lint-prose-rules.sh:19` — flagged comment in test fixture** (`code-review`): The `# Produces the string: X=$(echo "$J" | jq -r .a)` comment on line 19 is an added line in the diff and would be flagged by the broken linter. Resolved by the comment-skip fix.

- **`test-lint-prose-rules.sh:108` — Test 3 passes accidentally** (`code-review`): `COMMENT_LINE` used a bare `$X` (no dollar sign preserved), so it matched nothing and the test vacuously passed — the comment-detection path was never exercised. Fixed: Test 3 now writes both an un-indented and an indented comment containing a real variable reference (e.g. `# do not use: echo "$X" | jq`) and asserts exit 0.

- **`lint-prose-rules.sh:13` — PATTERN misses `echo -e`/`echo -n` and braced `${VAR}` form** (`code-review`, `input-validation`): The original pattern `echo[[:space:]]+"\$IDENT"` silently allowed `echo -e "$VAR" | jq` (the most dangerous form — explicit escape interpretation) and `echo "${MY_VAR}" | jq`. Extended pattern: `echo([[:space:]]+-[a-zA-Z]+)?[[:space:]]+"\$\{?IDENT\}?"[[:space:]]*\|[[:space:]]*jq`. Tests 5 and 6 added to cover the flag form and braced form.

- **`test-lint-prose-rules.sh:100-114` — no test for indented comments** (`review`): Test 3 only checked an un-indented comment. Now also writes a `  # space-indented` comment and asserts exit 0, proving the ERE comment guard fires on both forms.

---

### Deferred

- **`lint-prose-rules.sh:5-8` — unnecessary `lib.sh` dependency** (`review`): The script declares `SCRIPTS` and sources `lib.sh` but calls no functions from it. `lib.sh` exports globals at source time, creating latent coupling. Tracked as follow-up #2086 (blocked_by #2067).

---

### Dismissed

- **`lint-prose-rules.sh:84` — violation entries echoed verbatim** (`input-validation`): Violation messages are written to stderr with `echo " $v" >&2`. The concern is that `CURRENT_PATH` (from diff `+++ b/...` headers) could contain shell metacharacters. Since `echo` with a double-quoted argument does not evaluate metacharacters, and these paths come from `git diff` output (controlled paths), this is not an actionable injection risk. Dismissed.

---

_No security findings (surface: code, no auth/data/firebase/Firestore-rules changes; 0 open CodeQL alerts; no dependency changes)._

PR #2081, opened 2026-06-19
Add knip dead-code / unused-export sensor as a gated CI check

natb1:
Closes #2066

Adds a dead-code / unused-export sensor as a **gated CI check** built on
[knip](https://knip.dev) (the issue's preferred tool — it covers unused files,
exports, types, *and* dependencies in one pass). The check fails a PR when it
introduces *newly* orphaned code, while tolerating the pre-existing backlog via a
documented config allowlist. Part of epic #2061 (code-quality decay sensors).

## How it works

knip has no native baseline feature, so its own config (`entry` / `ignore` /
`ignoreDependencies` / `@public` JSDoc tags) *is* the baseline. A whole-repo run
is clean today; a *newly* orphaned export is not in the allowlist, so knip
reports it and CI fails — **including cross-file orphans** (a PR that removes the
last consumer of an export in an unchanged file is caught, because that export
was not previously dead). The run is whole-repo (not diff-scoped) precisely so
that canonical cross-file case is covered, since knip computes cross-workspace
reachability.

The sensor is **advisory/blocking, never auto-deleting**: the flagged set
includes harness-invoked scripts and seed files, so automatic removal would be
unsafe. Auto-fix / `/review-fix` integration is intentionally out of scope (a
possible future enhancement once the config proves stable).

## Unit 1 — knip + curated config

- Adds `knip` to root `devDependencies`.
- Adds a root **`knip.jsonc`** (JSONC so every allowlist entry carries a one-line
  `//` justification — the AC's "documented allowlist"):
  - **Tree-ignores** `.claude/**` (harness-invoked Workflow/skill scripts —
    deleting e.g. `review-fix.js` would break the harness) and
    `scaffolding/firebase/templates/**` (copied, never imported).
  - **Per-workspace `entry` overrides** for tooling that has no static importer
    but is invoked by a runner/harness: `seeds/**`, `e2e/**`, `scripts/**`,
    `bin/**`. `entry` (not `ignore`) is used so the utilities these files consume
    stay "used" rather than being blinded.
  - **CSS-imported deps** (`@commons-systems/style`, `@commons-systems/ds`) are
    allowlisted per-name — knip 6 doesn't trace bare CSS `@import` into
    `node_modules`.
  - **Genuine residual dead code** is *baselined, not deleted* (some "unused"
    exports are intentional public API). Encoded via narrow per-export `@public`
    JSDoc tags (12, all on `budget/src` sources) plus per-name dep entries — each
    documented. Every allowlist entry is narrow enough that a brand-new orphan of
    the same kind still fires.

Each suppression was checked against the live `npx knip` output of the installed
version, and the genuinely-unused deps (`d3-scale*`, `@google-cloud/secret-manager`,
and a few `@commons-systems/*`) were grep-verified individually before being
baselined as deletion candidates rather than removed in this PR.

## Unit 2 — gated CI job

- **`run-dead-code.sh`** — runs `npx knip --max-issues 0` **once at repo root**
  (cross-workspace reachability requires a single whole-repo run, unlike
  `run-lint.sh`'s per-workspace loop). Installs root deps via `ensure_deps`.
- **`detect-changes.sh`** — adds a broad `deadcode=true` gate on any TS/JS or
  knip-config change (any such change anywhere can orphan code elsewhere, so the
  whole-repo run must trigger).
- **`unit-tests.yml`** — adds a `dead-code` job mirroring the `lint` job
  (checkout `fetch-depth: 0`, detect-changes, gated `setup-node`, gated runner).

## Verification

- **Clean baseline:** `npx knip` exits 0 on the branch.
- **Catches a new orphan:** a deliberately-unused `export const __probe = 1`
  added to a tracked source was flagged by name (exit 1); removed → exit 0.
- **Catches a cross-file orphan:** deleting the sole consumer of an existing
  export made knip flag it in its *unchanged* defining file (exit 1); reverted.
- **Gate:** `detect-changes.sh` emits `deadcode=true` for `.ts`/`.tsx`/knip-config
  changes and omits it for docs-only diffs.
- **CI job:** `run-dead-code.sh` exits 0 locally; the `dead-code` job mirrors the
  `lint` job's structure with the same pinned action SHAs.

## Follow-up (not this PR)

The baselined residual is deletion-candidate work for a future ratchet: the 12
`@public`-tagged `budget` exports (`parseFirestore*`, `renderLoadError`) and the
genuinely-unused deps. Shrinking the allowlist is follow-up, not this PR's scope.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/detect-changes.sh
+++ b/.claude/skills/dispatch-propagate/scripts/detect-changes.sh
@@ -32,6 +32,11 @@ fi
 if echo "$CHANGED" | grep -qE '^(firestore\.rules$|storage\.rules$|rules-test/|\.claude/skills/dispatch-propagate/scripts/|firebase\.json$|package\.json$|package-lock\.json$)'; then
   echo "rules=true" >> "$GITHUB_OUTPUT"
 fi
+# dead-code check runs knip at repo root — any TS/JS/knip-config change can
+# orphan code elsewhere, so trigger on any such change across the repo.
+if echo "$CHANGED" | grep -qE '\.(ts|tsx|js|jsx|mjs|cjs)$|^knip\.(json|jsonc|ts)$'; then
+  echo "deadcode=true" >> "$GITHUB_OUTPUT"
+fi
 # go-tests needs the Go toolchain. Set go=true when a changed file is under a
 # discovered Go module. list-go-modules.sh discovers module roots from go.mod
 # locations, so a new Go module needs no edit here.
--- a/.claude/skills/dispatch-propagate/scripts/run-dead-code.sh
+++ b/.claude/skills/dispatch-propagate/scripts/run-dead-code.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+SCRIPTS="$(cd "$(dirname "$0")" && pwd)"
+
+# shellcheck source=lib.sh
+source "$SCRIPTS/lib.sh"
+
+# Install root node_modules once (knip runs at repo root, not per-workspace)
+ensure_deps
+
+echo "=== Dead-code check (knip) ==="
+if (cd "$REPO_ROOT" && npx knip --max-issues 0); then
+  echo "PASS: dead-code check"
+else
+  echo "FAIL: dead-code check" >&2
+  exit 1
+fi
+
+echo "All dead-code checks passed."
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -109,6 +109,24 @@ jobs:
       - name: Run lint
         run: .claude/skills/dispatch-propagate/scripts/run-lint.sh
 
+  dead-code:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          fetch-depth: 0
+      - name: Detect changed file categories
+        id: changes
+        run: .claude/skills/dispatch-propagate/scripts/detect-changes.sh
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        if: steps.changes.outputs.deadcode == 'true'
+        with:
+          node-version-file: .node-version
+          cache: 'npm'
+      - name: Run dead-code check
+        if: steps.changes.outputs.deadcode == 'true'
+        run: .claude/skills/dispatch-propagate/scripts/run-dead-code.sh
+
   typecheck:
     runs-on: ubuntu-latest
     steps:
--- a/budget/src/entities/account.ts
+++ b/budget/src/entities/account.ts
@@ -95,6 +95,11 @@ export interface SeedAccount {
 
 // ── Firestore → Account ───────────────────────────────────────────────────────
 
+/**
+ * @public knip baseline (#2066): residual unused export, not deleted per the
+ * no-bulk-delete line — part of the symmetric per-entity parseFirestore* set.
+ * Deletion candidate; see PR notes.
+ */
 export function parseFirestoreAccount(docSnap: QueryDocumentSnapshot<DocumentData, DocumentData>): Account {
   const data = docSnap.data();
   return {
--- a/budget/src/entities/budget-period.ts
+++ b/budget/src/entities/budget-period.ts
@@ -98,6 +98,11 @@ function requireCategoryBreakdown(value: unknown): Record<string, number> {
 
 // ── Firestore → BudgetPeriod ──────────────────────────────────────────────────
 
+/**
+ * @public knip baseline (#2066): residual unused export, not deleted per the
+ * no-bulk-delete line — part of the symmetric per-entity parseFirestore* set.
+ * Deletion candidate; see PR notes.
+ */
 export function parseFirestoreBudgetPeriod(docSnap: QueryDocumentSnapshot<DocumentData, DocumentData>): BudgetPeriod {
   const data = docSnap.data();
   const periodStart = requireTimestamp(data.periodStart, "periodStart");
--- a/budget/src/entities/budget.ts
+++ b/budget/src/entities/budget.ts
@@ -166,6 +166,11 @@ export function validateBudgetOverrideOrdering(
 
 // ── Firestore → Budget ────────────────────────────────────────────────────────
 
+/**
+ * @public knip baseline (#2066): residual unused export, not deleted per the
+ * no-bulk-delete line — part of the symmetric per-entity parseFirestore* set.
+ * Deletion candidate; see PR notes.
+ */
 export function parseFirestoreBudget(docSnap: QueryDocumentSnapshot<DocumentData, DocumentData>): Budget {
   const data = docSnap.data();
   const name = requireString(data.name, "name");
--- a/budget/src/entities/journal-entry.ts
+++ b/budget/src/entities/journal-entry.ts
@@ -68,6 +68,11 @@ export interface SeedJournalEntry {
 
 // ── Firestore → JournalEntry ──────────────────────────────────────────────────
 
+/**
+ * @public knip baseline (#2066): residual unused export, not deleted per the
+ * no-bulk-delete line — part of the symmetric per-entity parseFirestore* set.
+ * Deletion candidate; see PR notes.
+ */
 export function parseFirestoreJournalEntry(docSnap: QueryDocumentSnapshot<DocumentData, DocumentData>): JournalEntry {
   const data = docSnap.data();
   return {
--- a/budget/src/entities/journal-leg.ts
+++ b/budget/src/entities/journal-leg.ts
@@ -94,6 +94,11 @@ export interface SeedJournalLeg {
 
 // ── Firestore → JournalLeg ────────────────────────────────────────────────────
 
+/**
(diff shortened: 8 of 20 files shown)


natb1:
## QA Summary — issue #2066 / PR #2081

### Items executed: 7

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Dead-code sensor runs in CI (whole-repo knip job) | shell | PASS |
| 2 | Clean baseline passes (no false positives on current tree) | shell | PASS |
| 3 | Curated config and allowlist are present and documented | shell | PASS |
| 4 | knip registered as a root devDependency | shell | PASS |
| 5 | Change-detection gate triggers on TS/JS or knip-config changes | shell | PASS |
| 6 | dead-code job wired into the unit-tests workflow | shell | PASS |
| 7 | Sensor flags newly orphaned code (diff-scoped detection) | deferred | planned-deferral |

**PASS: 6 · FAIL: 0 · SKIP: 0 · Deferred (planned-deferral): 1**

No bugs found. No browser walkthrough (no browser-component items in plan).

---

### Disposition triage

| 7 | Sensor flags newly orphaned code (diff-scoped detection) | needs-main | n/a |

**Rationale (item 7):** `planned_deferral=true` — the positive-detection case (knip exits 1 on a deliberately-orphaned export, then 0 after revert) requires mutating the working tree; it is non-assertable as a single idempotent command against the clean merged tree. The PR documents it was verified at author time. Filed as a `blocked_by` follow-up for downstream verification.

Non-fixing pass: no opus-fixable items were found, so residue is routed to follow-up (needs-main items) rather than an auto-fix lane.

---

### Filed follow-ups

- **#2083** — `qa-needs-main #2066: sensor-flags-newly-orphaned-code-diff-scoped-detection`
  (will receive `main-qa` + `dispatch:office-hours` labels; blocked_by #2066)

natb1:
# Fix-checks summary — PR #2081

## Iteration 1

- **Failed checks:** lint
- **Outcome:** needs-human
- **Reproduced:** yes
- **Reproduce command:** `.claude/skills/dispatch-propagate/scripts/run-lint.sh --app budget`
- **Failure excerpt:**

  ```
  budget/src/budget-seed-data.d.ts
    6:3  error  './entities/budget' import is restricted from being used by a
    pattern. Layering violation: 'budget' is a top-level app package and must
    not be imported by libraries, utils, or @commons-systems/ds ...
    @typescript-eslint/no-restricted-imports
  ✖ 1 problem (1 error, 0 warnings)
  ```

- **Why not caught:** Pre-existing defect inherited from `origin/main`, not
  introduced by this PR. The architecture-boundary lint rule landed on main in
  commit `e32dee3b` ("Architecture boundary lint: forbid importing top-level app
  packages"); its `@typescript-eslint/no-restricted-imports` `group: [app,
  "${app}/**"]` pattern matches *relative* imports whose final path segment
  equals an app-package name. `budget/src/budget-seed-data.d.ts:6` does
  `export type { ... } from "./entities/budget"` — a legitimate intra-package
  relative import — and the bare `budget` pattern matches its trailing segment, a
  false positive. Both `config/eslint.config.js` and
  `budget/src/budget-seed-data.d.ts` are byte-identical to `origin/main`
  (`git diff origin/main...HEAD` empty for both); this PR's only `budget/src`
  changes are comment-only `@public` JSDoc tags. The failure surfaced here only
  because the budget workspace is linted on this PR's run; main never surfaced it.
  The fix belongs on main (anchor the rule so relative imports never match, plus a
  relative-import regression test in `config/test/eslint-boundaries.test.ts`) —
  `config/` is a sandbox-protected guardrail path and the change is a design
  decision on an architecture boundary, out of scope for this dead-code PR. It is
  systemic: any open PR that re-lints `budget/src` after merging main hits the
  same error.
- **Required action:** Fix the boundary-lint false-positive on `main` — make the
  `no-restricted-imports` patterns in `config/eslint.config.js` not match
  relative imports (`./...`) whose last segment coincides with an app-package
  name, and add a relative-import regression case to
  `config/test/eslint-boundaries.test.ts`. Likely a `dispatch:main-broken`
  candidate. Then re-merge main into this branch and re-run CI. This PR's own
  dead-code sensor (the `dead-code` job) already passes.

natb1:
# Fix-checks summary — PR #2081

## Iteration 1

- **Failed checks:** lint
- **Outcome:** needs-human
- **Reproduced:** yes
- **Reproduce command:** `.claude/skills/dispatch-propagate/scripts/run-lint.sh --app budget`
- **Failure excerpt:**

  ```
  budget/src/budget-seed-data.d.ts
    6:3  error  './entities/budget' import is restricted from being used by a
    pattern. Layering violation: 'budget' is a top-level app package and must
    not be imported by libraries, utils, or @commons-systems/ds ...
    @typescript-eslint/no-restricted-imports
  ✖ 1 problem (1 error, 0 warnings)
  ```

- **Why not caught:** Pre-existing defect inherited from `origin/main`, not
  introduced by this PR. The architecture-boundary lint rule landed on main in
  commit `e32dee3b` ("Architecture boundary lint: forbid importing top-level app
  packages"); its `@typescript-eslint/no-restricted-imports` `group: [app,
  "${app}/**"]` pattern matches *relative* imports whose final path segment
  equals an app-package name. `budget/src/budget-seed-data.d.ts:6` does
  `export type { ... } from "./entities/budget"` — a legitimate intra-package
  relative import — and the bare `budget` pattern matches its trailing segment, a
  false positive. Both `config/eslint.config.js` and
  `budget/src/budget-seed-data.d.ts` are byte-identical to `origin/main`
  (`git diff origin/main...HEAD` empty for both); this PR's only `budget/src`
  changes are comment-only `@public` JSDoc tags. The failure surfaced here only
  because the budget workspace is linted on this PR's run; main never surfaced it.
  The fix belongs on main (anchor the rule so relative imports never match, plus a
  relative-import regression test in `config/test/eslint-boundaries.test.ts`) —
  `config/` is a sandbox-protected guardrail path and the change is a design
  decision on an architecture boundary, out of scope for this dead-code PR. It is
  systemic: any open PR that re-lints `budget/src` after merging main hits the
  same error.
- **Required action:** Fix the boundary-lint false-positive on `main` — make the
  `no-restricted-imports` patterns in `config/eslint.config.js` not match
  relative imports (`./...`) whose last segment coincides with an app-package
  name, and add a relative-import regression case to
  `config/test/eslint-boundaries.test.ts`. Likely a `dispatch:main-broken`
  candidate. Then re-merge main into this branch and re-run CI. This PR's own
  dead-code sensor (the `dead-code` job) already passes.

## Iteration 2

- **Failed checks:** lint
- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `.claude/skills/dispatch-propagate/scripts/run-lint.sh --app budget`
- **Failure excerpt:** (same as Iteration 1 — boundary-lint false positive on
  `budget/src/budget-seed-data.d.ts:6`, `./entities/budget`)
- **Why not caught:** Same root cause as Iteration 1. Iteration 1 escalated this
  to office-hours as a main-introduced defect outside this PR's lane; the
  maintainer then directed fixing it here, so the fix is applied in this branch
  rather than deferred to a separate main-broken issue.
- **Fix:** Anchored the architecture-boundary rule so it can no longer match
  relative imports. In `config/eslint.config.js`, replaced the gitignore-style
  `group: [app, "${app}/**"]` (a bare segment matches at any path depth, so
  `budget` matched the trailing segment of `./entities/budget`) with an anchored
  `regex: "^<app>(/|$)"` that matches only specifiers *starting* with the app
  name — which a relative import (`./…`, `../…`) never does. The app name is
  RegExp-escaped. Added two regression cases to
  `config/test/eslint-boundaries.test.ts` (`./entities/budget` and `../print`
  must not flag); all 21 config tests pass and full-workspace lint is green.
  Commit `90402382` (pushed after a clean merge of origin/main).

natb1:
## Review: PR #2081 — Dead-code / unused-export sensor (knip)

**7 findings across all reviewers (code-review, review, security domains). No code changes required.**

---

### Informational

- `f1` — `.claude/skills/dispatch-propagate/scripts/run-dead-code.sh:14`: `REPO_ROOT` + `cd` approach is correct; `ensure_deps` may not change directory, and knip must run at repo root for cross-workspace reachability. Consistent with `run-lint.sh`. _No action needed._

- `f2` — `.claude/skills/dispatch-propagate/scripts/detect-changes.sh:35`: Broad `.js/.jsx` gate on any repo path is intentional — since knip is a whole-repo cross-workspace check, any TS/JS/config change anywhere can affect the dead-code result. _No action needed._

- `r1` — `knip.jsonc`: The config is contributor-editable, so a future PR could add an allowlist entry suppressing real dead code. The existing inline justification comments are the right mitigation; review process is the guard. _No action needed — informational for reviewers._

- `r2` — `budget/src/entities/account.ts` (and 11 similar): `@public` is a knip-specific convention. The existing `// knip baseline (#2066): ...` comments already explain intent. _No action needed._

---

### Dismissed / False-positive

- `iv1` — `detect-changes.sh:35` (input-validation): No injection risk — the `CHANGED` variable contains git diff filenames; the grep checks for file extensions only, not executing filenames. _False-positive._

- `rt1` — `knip.jsonc` (red-team): `ignoreFiles: [".claude/**"]` is correct; harness scripts are legitimately excluded. A contributor adding malicious code to `.claude/` is visible in the diff and subject to code review — dead-code sensors are not the right defense for that threat. _False-positive._

---

### Deferred

- `sr1` — `.github/workflows/unit-tests.yml` (security-review): The `dead-code` CI job has no explicit `permissions` block; it inherits workflow-level defaults. Adding `permissions: contents: read` is a hardening improvement (defense-in-depth). Filed as follow-up #2090.

---

_No security vulnerabilities found. No code changes applied. CI is passing._

Issue #2024, opened 2026-06-19
dispatch: execute the plan's Verification section in /implement before opening the PR


natb1:
## Context

`/plan-issue` writes a plan to the issue's `<\!-- dispatch:plan -->` comment, and the plan schema requires a **Verification section** describing how to test the change end-to-end (`.claude/skills/plan-issue/SKILL.md`). But `/implement` never runs it: implement's terminal step opens the PR and self-judges deviation (`.claude/skills/implement/SKILL.md`), and the QA phase later re-derives its own checks from the issue's acceptance criteria via `dispatch-context-pack` rather than reading the plan. The plan's Verification section is authored and then orphaned — never executed.

Net effect: every PR is opened without executing the verification its own plan specified. The Plan-Execute-Verify loop is broken at the Verify leg — defects the plan's own test would catch slip through to the QA phase one step later (or to a human).

## Proposed change

Before `/implement` opens the PR (or as a final `implement-unit`-style pass), read the plan's Verification section via `dispatch-read-plan` and execute the deterministic steps it names. On a fixable failure, loop within the bounded implement budget; on an unfixable failure, escalate via `dispatch-mark-deviation` instead of opening a green-looking PR. This makes the plan a contract verified against, not a discarded artifact, and catches defects one phase earlier.

## Acceptance criteria

- [ ] `/implement` reads the plan's Verification section and executes its deterministic steps before opening the PR.
- [ ] A verification failure that implement can fix is fixed within the existing implement budget before the PR opens.
- [ ] A verification failure implement cannot fix escalates via `dispatch-mark-deviation` rather than opening the PR as complete.
- [ ] When the plan has no runnable Verification steps, implement proceeds unchanged (no false block).
- [ ] The behavior is covered by a test exercising both the pass and the escalate paths.

PR #2106, opened 2026-06-20
Harden dispatch-run-verification against silent exit-3 on empty/upstream-failed input

### LINKED ISSUE #2080: Harden dispatch-run-verification against silent exit-3 on upstream (read-plan) failure
## Context

`dispatch-run-verification` (added in #2024,
`.claude/skills/dispatch-propagate/scripts/dispatch-run-verification`) reads plan
markdown on stdin and exits **3** ("no auto-runnable verify blocks → proceed
unchanged") when it finds no `## Verification` section or no ` ```verify ` blocks.

It cannot distinguish *"the plan genuinely has no verify blocks"* (the legitimate
exit-3 case) from *"stdin arrived empty because the upstream producer failed."*
In the canonical `/implement` pipe:

```
dispatch-read-plan <N> | dispatch-run-verification
```

a pipeline's exit status is that of its **last** command. So if
`dispatch-read-plan` fails — the issue has no `<!-- dispatch:plan -->` comment
(exit 1, empty stdout), or it runs sandboxed and `gh` fails — its failure is
masked: `dispatch-run-verification` reads empty stdin and exits 3, and
`/implement` reads that as "proceed unchanged" and **silently skips
verification**. That is the exact failure mode #2024 exists to prevent,
reintroduced one layer up.

This is a latent robustness gap, not a live bug. #2024 mitigated the immediate
trigger by noting `dangerouslyDisableSandbox: true` on the verify step in both
`implement/SKILL.md` Step 3 and the `plan-issue` preface template, so the `gh`
path runs unsandboxed. But the empty-stdin-masquerading-as-exit-3 hazard remains
for any other upstream failure (most notably a missing plan comment).

Per `.claude/rules/code-style.md` (clear errors over defensive fallbacks), an
upstream failure must surface as a hard error, never be swallowed into the
proceed-unchanged signal.

## Proposed fix (greenfield)

Either or both of:

- Have `dispatch-run-verification` distinguish **empty/absent input** (an error →
  a non-zero exit that does NOT mean "proceed unchanged") from a **well-formed
  plan that legitimately has no verify blocks** (the existing exit 3). A plan with
  no content on stdin is malformed input, not a plan with zero checks.
- Set `set -o pipefail` (or an equivalent guard) at the `/implement` call site so
  an upstream `dispatch-read-plan` failure surfaces as a hard error instead of
  collapsing into exit 3.

## Acceptance criteria

- [ ] When `dispatch-run-verification` receives empty stdin, it exits non-zero
      with a distinct diagnostic and does NOT emit the exit-3 "proceed unchanged"
      signal.
- [ ] A well-formed plan with a `## Verification` section but no ` ```verify `
      blocks still exits 3 (the legitimate proceed-unchanged case is preserved).
- [ ] An upstream `dispatch-read-plan` failure in the `/implement` pipe surfaces
      as a hard error (via `pipefail` at the call site and/or the runner's
      empty-input check), so verification is never silently skipped.
- [ ] `test-dispatch-scripts.sh` gains cases covering empty-stdin and the
      preserved no-verify-blocks exit-3 path, and the suite stays green.

natb1:
Closes #2080

## Problem

`dispatch-run-verification` reads plan markdown on stdin and exits **3** ("no
auto-runnable verify blocks → proceed unchanged") when it finds no `## Verification`
section or no ` ```verify ` blocks. It could not tell that legitimate case apart
from *stdin arrived empty because the upstream producer failed*. In the canonical
pipe

```
dispatch-read-plan <N> | dispatch-run-verification
```

a pipeline's status is its last command's. When `dispatch-read-plan` fails (no
`<!-- dispatch:plan -->` comment → exit 1, empty stdout), the runner read empty
stdin and exited 3, and `/implement` read that as "proceed unchanged" — silently
skipping verification, the exact failure #2024 exists to prevent, reintroduced one
layer up.

`set -o pipefail` alone does **not** fix this: it makes a pipeline return the
rightmost non-zero exit, but the runner's exit 3 is itself non-zero, so the
pipeline yields 3 regardless of `dispatch-read-plan`'s status. The load-bearing fix
is the runner distinguishing empty/absent input from a well-formed zero-verify
plan.

## Change

**Unit 1 — runner rejects empty/absent input with a distinct exit 4.**
`dispatch-run-verification` now guards immediately after reading stdin: if the
plan has no non-whitespace content, it emits a stderr diagnostic and exits **4**
(a new code, NOT 3). Exit 3 stays reserved for a well-formed plan that legitimately
has zero verify blocks, so the #2024 acceptance (no false block when a plan defines
no checks) is preserved. The header comment and `usage()` exit-code list now
document exit 4. New unit tests cover empty stdin → 4 (with diagnostic), whitespace-
only stdin → 4, and assert the existing exit-3 preservation cases.

**Unit 2 — call sites route a non-{0,1,3} exit to a hard stop.** Both documented
call sites (`implement` Step 3 and the `plan-issue` preface template embedded in
every future plan) now run the pipe under `set -o pipefail` and add an explicit
branch: any exit other than 0/1/3 — exit 4 (empty/absent input) or an upstream
failure surfaced via `pipefail` — is an environment/upstream error, not a verify-
block failure. The worker does **not** enter the `/implement-unit` fix lane; it
routes straight to `dispatch-mark-deviation` and stops. `pipefail` is labeled as
belt-and-suspenders behind the runner's exit-4 guard (it only helps the residual
case where `read-plan` fails but `run-verification` still exits 0), not as the fix.

Per `.claude/rules/code-style.md`, an upstream failure now surfaces as a clear
error instead of being swallowed into the proceed-unchanged signal.

## Verification

The dispatch script unit suite (hermetic, PATH-shimmed gh/git, no network; runs in
CI via `.github/workflows/unit-tests.yml`) exercises the whole change and is green:
**3381/3381 passed** (up from 3378). Spot-checks:

- `printf '' | dispatch-run-verification; echo $?` → `4`
- `printf '## Verification\nmanual\n' | dispatch-run-verification; echo $?` → `3` (preserved)


natb1:
## QA — Issue #2080 (Harden dispatch-run-verification)

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Empty stdin exits 4 with distinct diagnostic | shell-command | PASS |
| 2 | Whitespace-only stdin also exits 4 | shell-command | PASS |
| 3 | Well-formed plan with no verify blocks still exits 3 | shell-command | PASS |
| 4 | Full dispatch script unit suite stays green | shell-command | PASS |
| 5 | Runner header and usage() document exit 4 | shell-command | PASS |
| 6 | /implement call site sets pipefail + routes exit 4 to dispatch-mark-deviation | shell-command | PASS |
| 7 | plan-issue preface template shows pipefail + exit-4 routing | shell-command | PASS |
| 8 | Belt-and-suspenders routing semantics coherent across SKILL.md call sites | (deferred) | → already satisfied |

**Counts:** PASS 7 · FAIL 0 · SKIP 0

### Acceptance criteria

All four acceptance criteria met:

- ✅ Empty stdin → exit 4 with distinct diagnostic (not exit 3) — confirmed by test suite (3381/3381 green, up from 3378).
- ✅ Well-formed plan with no `verify` blocks still exits 3 — existing suite cases confirmed by the run.
- ✅ Upstream failure surfaces as a hard error — `set -o pipefail` added at both call sites (`implement/SKILL.md` Step 3, `plan-issue/SKILL.md` preface template), routing exit 4 / any-other-non-zero to `dispatch-mark-deviation` + stop (not the fix lane).
- ✅ `test-dispatch-scripts.sh` gained empty-stdin and whitespace-only-stdin → exit 4 cases; suite is green.

- **Item 8 — Belt-and-suspenders routing semantics coherent across SKILL.md call sites** — Both `implement/SKILL.md` and `plan-issue/SKILL.md` map each exit code identically: 0 → proceed, 1 → fix lane (cap 2) else deviation+stop, 3 → proceed unchanged, exit 4 / any-other-non-zero → `dispatch-mark-deviation` + stop (explicitly NOT the fix lane). Only stylistic differences (marker wording, list vs prose); no disposition inconsistency. No defect — dropped as PASS.

### Disposition triage

Step 3.5 ran; every residue item was classified `already-satisfied` by the Workflow. No items escalate to office-hours.

natb1:
## Review — #2080 (PR #2106)

Security review: no attack surface — diff touches only shell scripts (no application source, config, dependency, or Firestore-rules changes; `app_or_rules=false`, `deps=false`). No CodeQL alerts open on this PR ref.

---

### Fixed

- **plan-issue/SKILL.md:394** — The plan preface template used a single deviation marker snippet for two logically distinct failure cases: (1) exit-1 after exhausting the 2-attempt fix cap, and (2) exit-4/upstream failure (or any other non-zero exit). The merged string made incident triage ambiguous — a reader could not tell whether verification ran and hit the cap, or never ran due to upstream failure. Split into two separate `dispatch-mark-deviation` blocks, mirroring the `implement/SKILL.md` pattern. (sources: code-review, review — fix commit `7e9c3395`)

---

### Informational

- **implement/SKILL.md:128** — The comment explains `set -o pipefail` as belt-and-suspenders "for the residual case where `dispatch-read-plan` fails but `dispatch-run-verification` still exits 0." The analysis is correct that the exit-3 mask is the primary failure mode and exit-4 is the real fix. However, when `dispatch-read-plan` exits 1 and emits partial non-empty content, and the verifier exits 0 on that partial input, `pipefail` routes the pipeline to exit 1 — which the routing table maps to the fix lane, not deviation. So the documented residual-case promise is true only for the narrower scenario where `dispatch-read-plan` exits ≥2 while the verifier exits 0. The comment could be tightened to clarify which upstream failure modes pipefail actually catches without colliding with the exit-1 fix-lane path. No change required for this PR; the exit-4 guard is load-bearing regardless. (source: code-review)

- **implement/SKILL.md:135** — The `set -o pipefail` and pipeline are presented as two separate lines in a single code block. In an agent Bash-tool context each code block typically maps to one `Bash()` call, so `pipefail` applies only to that block's scope — which is fine. No action needed. (source: code-review)

- **test-dispatch-scripts.sh:34702** — The whitespace-only stdin test (`printf '%s' "$ws" | "$RUN_VERIFY"`) checks only the exit code (4) and not the diagnostic message. The empty-stdin test immediately above checks both the exit code and a `"empty plan input"` substring. The asymmetry is a minor coverage gap, not a blocker. (source: code-review)

- **test-dispatch-scripts.sh:34707** — The "Preservation guard (criterion 2)" comment references "the exit-3 cases above" to justify no additional test, but the exit-3 coverage is in a different section several hundred lines earlier. The comment is slightly misleading; "above" means the whitespace/empty tests, which test exit-4, not exit-3. No functional gap. (source: review)

---

### Dismissed

- **implement/SKILL.md:135** — Concern that a bare `set -o pipefail` in a shell session affects subsequent commands in the same call. In the single-pipeline use in SKILL.md the code block is self-contained; in practice agents execute each bash snippet in its own subshell. Not a real hazard. (source: review)

---

### Deferred

- **dispatch-run-verification:136** — An unclosed verify fence (EOF before closing fence) exits 1 — same code as a genuine verify-block failure. The routing logic treats exit 1 as "fix lane," so a malformed plan wastes up to 2 fix attempts before deviating. Pre-existing; the PR's addition of exit 4 was a natural moment to address it. Filed as follow-up → #2118. (source: review)

PR #2114, opened 2026-06-20
dispatch: append a structured per-tick decision log from the headless control path

natb1:
Closes #2038

## What

Adds a structured, machine-queryable decision log to the headless autonomous
dispatch control path. Every tick now appends one JSONL line per decision site to
a fixed, size-rotated file under `~/.local/share/commons-dispatch/`
(`routing-decisions.jsonl`), recording which target the deterministic router
selected, the derived phase, the chosen disposition, the budget gate
(`target_n`/`effective_live`/`gap`), and any skip reason.

This is the control-path complement to the session-efficacy `dispatch.outcome.v1`
envelope: that records what each model worker produced; this records what the
deterministic router decided.

## Why

The only audit trail for routing decisions today is ephemeral journald output plus
the durable GitHub side effects those decisions produce. Diagnosing a
drain-to-zero or a mis-selection means reverse-engineering journald, which rotates
away. A durable, `jq`-queryable record of the control path's own decision inputs
makes that diagnosis direct.

## How

One shared sourced helper, three append sites, each funneling through the helper
so rotation and the non-fatal guarantee live in exactly one place:

- **Unit 1 — `lib-decision-log.sh`**: a write-only, non-fatal `decision_log_append`
  helper. `flock`-guarded single-generation 5 MB rotation; the whole body is
  swallow-failure wrapped so a failed append can never make a caller under
  `set -euo pipefail` die. Exposes only `decision_log_append` and the path var — no
  read-the-log code anywhere.
- **Unit 2 — `dispatch-select-tick`**: an EXIT trap emits exactly one
  `site:"select-tick"` record per run, covering every terminal branch (busy, sync-*,
  concurrency-cap, resolver-failed, transient, explicit, pr/issue, main-broken,
  jit-reminder, empty). A pre-existing EXIT trap was chained, not clobbered.
- **Unit 3 — `dispatch-materialize-spawn`**: an EXIT trap emits exactly one
  `site:"materialize-spawn"` record per run (`propagate`/`drain`/`notify`/`error`
  plus `mode`/`gap`/`spawned`/`stop_reason`). `propagate` and `drain` are emitted
  only here, not in `dispatch-select-tick`, so instrumenting this script is required
  to satisfy the AC's disposition coverage.
- **Unit 4 — `dispatch-stop.sh`**: an EXIT trap armed only after the worker
  discriminator emits one `site:"stop"` record per worker invocation
  (worker/issue/pr/marker/branch/disposition). Router-named jobs that exit at the
  discriminator produce no record. The `source` is guarded against the script's
  active `ERR` trap so a missing lib degrades to a true no-op.
- **Unit 5 — regression tests**: `test-dispatch-scripts.sh` now stages
  `lib-decision-log.sh` into each runner's scratch dir and asserts a record with the
  correct `site` and `disposition` for `propagate`, `drain`, `concurrency-cap`,
  `empty`, and a `site:"stop"` Branch-B self-close, plus a non-fatal-guarantee test
  pointing the log dir at an unwritable path.

## Design constraints (verified)

- **Write-only**: the log is never read back by the control path to drive a routing
  decision, so it cannot become a drift-prone side file. The helper contains no
  read/parse code.
- **Non-fatal**: a failed append never makes the tick return non-zero or abort. The
  append is best-effort observability beside the real work.

## Verification

- Full suite green: `test-dispatch-scripts.sh` passes 3401/3401, exit 0.
- A negative sanity check (temporarily un-staging the lib) confirmed the new
  assertions are real, not vacuously passing.
- No `$HOME` pollution: the real `~/.local/share/commons-dispatch/routing-decisions.jsonl`
  is absent after a full suite run (tests use the `DISPATCH_DECISION_LOG_DIR`
  override).


natb1:
# QA Summary — Issue #2038 (PR #2114)

**Pass: 5 | Fail: 0 | Skip: 0**

No browser component detected (pure script change). All items verified via shell commands.

## Items executed

### 1. Test suite passes (AC5 + AC3 non-fatal guarantee) — PASS
- Command: `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`
- Result: **3506/3506 passed, 0 failed**. New AC5 assertions (propagate, drain, concurrency-cap, empty, stop Branch-B) and the AC3 non-fatal guarantee test all passed.

### 2. AC1/AC2: Required decision-log fields wired in select-tick and stop.sh — PASS
- `dispatch-select-tick` emits: `ts`, `target`, `phase`, `disposition`, `skip_reason`, `target_n`, `effective_live`, `gap` — all AC1 fields present.
- `dispatch-stop.sh` emits: `worker`, `marker_present`, `marker_phase`, `current_phase`, `branch`, `disposition`, `issue`, `pr` — all AC2 fields present.
- EXIT traps verified: armed at top of select-tick (re-registered after CI-verdict-cache mktemp), armed after worker discriminator in stop.sh.
- Accumulators set at every terminal branch.

### 3. AC4: Log is observability-only — no control-path read-back — PASS
- Grep confirms `routing-decisions.jsonl` appears only in `lib-decision-log.sh` (path resolution).
- No `cat`, `read`, `jq`, `grep`, or `tail` of the log file in any control-path script.
- `lib-decision-log.sh` exposes only `decision_log_append` and the path var — no read/parse code.

### 4. AC3: Log path, env overrides, rotation, and flock present — PASS
- Default path resolves to `$HOME/.local/share/commons-dispatch/routing-decisions.jsonl`.
- `DISPATCH_DECISION_LOG_FILE` / `DISPATCH_DECISION_LOG_DIR` env overrides honored.
- Single-generation rotation at 5MB threshold, inside flock critical section.
- Entire append body wrapped in `{ ... } 2>/dev/null || true` + `return 0` — non-fatal guarantee.

### 5. Test isolation: suite never writes to the real log dir — PASS
- `sel_tick_setup`, `mat_setup`, and `stop_setup` each stage `lib-decision-log.sh` into `TMPDIR_TEST` and export `DISPATCH_DECISION_LOG_DIR="$TMPDIR_TEST/decisionlog"`.
- Each teardown unsets `DISPATCH_DECISION_LOG_DIR`.
- No test writes to the real `~/.local/share/commons-dispatch/` path.

## Clean pass

No bugs found. No deferred items. No follow-ups required.

natb1:
## Review findings — PR #2114

### Fixed

- **dispatch-select-tick:96** — `lib-decision-log.sh` sourced without `2>/dev/null || true`, unlike `dispatch-stop.sh`. If the lib is missing, the `source` fails loudly under `set -euo pipefail`. Changed to `source "$SCRIPT_DIR/lib-decision-log.sh" 2>/dev/null || true` in both `dispatch-select-tick` and `dispatch-materialize-spawn`, matching the guarded pattern. (fix commit: `37880eded5a9d518f500db103d023b4435fe43fa`)
- **lib-decision-log.sh:92** — `flock 9 || exit 0` has no timeout: a hung/stale lock holder could block the append critical section indefinitely. Changed to `flock -w 2 9 || exit 0` to bound the worst-case wait to 2 seconds. (fix commit: `37880eded5a9d518f500db103d023b4435fe43fa`)
- **dispatch-stop.sh:218 / dispatch-select-tick / dispatch-materialize-spawn** — `decision_log_append "$json"` called with no guard for the case where `decision_log_append` is undefined (e.g. sourcing the lib failed). Added `command -v decision_log_append >/dev/null 2>&1 && decision_log_append "$json" || true` in all three emit helpers, preserving the non-fatal/best-effort contract. (fix commit: `37880eded5a9d518f500db103d023b4435fe43fa`)
- **dispatch-select-tick:226** — `resolve_project_root` failure exit after the EXIT trap is armed had no `DLOG_DISPOSITION`/`DLOG_SKIP_REASON` set, so the emitted record carried an empty disposition. Added `DLOG_DISPOSITION="internal-error"; DLOG_SKIP_REASON="resolve-project-root-failed"` before the `exit 2`. (fix commit: `37880eded5a9d518f500db103d023b4435fe43fa`)

### Informational

- **lib-decision-log.sh:111** — `stat -c%s` is GNU/Linux-specific; BSD/macOS use `stat -f%z`. On macOS the stat call fails silently (caught by `|| return 0`), meaning `_decision_log_rotate` becomes a no-op rather than a hard error. No action needed for this Linux-only toolchain.
- **test-dispatch-scripts.sh:21061** — AC5 decision-log assertions are only added to stop Test 1 (Branch B self-close). Remaining stop tests (Tests 2–N) do not assert the log record. Coverage is partial; follow-up testing could extend AC5 coverage to other branch paths. No blocking concern for this PR.

### False-positive (security)

- **lib-decision-log.sh:68** — `DECISION_LOG_FILE` assembled from `DISPATCH_DECISION_LOG_FILE` / `DISPATCH_DECISION_LOG_DIR` env vars with no sanitization. The log is a write-only local observability sink: it is never read back or executed, and its directory is `~/.local/share/commons-dispatch/` (the same user's home). No exploitable attack surface.
- **lib-decision-log.sh:85,94,95,113** — Lockfile path derived from `DECISION_LOG_FILE` by appending `.lock`. Same write-only local sink context; no cross-user or privilege concern.

PR #2149, opened 2026-06-20
dispatch-write-phase-log: guard --phase/--attempt shift on missing value

### LINKED ISSUE #2132: `--phase` and `--attempt` flag handlers call `shift 2` without verifying `$# >= 2`
## Context

In `.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log`, the `--phase` and `--attempt` flag handlers call `shift 2` unconditionally when parsing CLI arguments.

## Bug

When either flag is the last argument on the command line (i.e. no value follows), bash's `shift 2` with only 1 argument remaining exits non-zero under `set -euo pipefail`. The script dies immediately with no error message. The intended behavior is `exit 2` with a usage message; instead the caller sees an opaque exit-1.

Example invocation that triggers it:

```
dispatch-write-phase-log --phase
```

## Recommended Fix

Guard both `shift 2` calls with a `$#` check before shifting:

```bash
--phase)
  [[ $# -ge 2 ]] || { echo 'dispatch-write-phase-log: --phase requires a value' >&2; exit 2; }
  PHASE="$2"; shift 2 ;;
--attempt)
  [[ $# -ge 2 ]] || { echo 'dispatch-write-phase-log: --attempt requires a value' >&2; exit 2; }
  ATTEMPT="$2"; shift 2 ;;
```

## Motivation

Opaque exits make dispatch failures harder to diagnose. A clear error message with exit 2 aligns with the script's existing usage-error convention and gives callers actionable output.

## Acceptance Criteria

- [ ] `dispatch-write-phase-log --phase` (no value) exits with code 2 and prints a message to stderr naming the missing value.
- [ ] `dispatch-write-phase-log --attempt` (no value) exits with code 2 and prints a message to stderr naming the missing value.
- [ ] Normal invocations with valid `--phase` and `--attempt` values continue to work correctly.

## References

- Surfaced during review of #2108 (out of scope for that PR).

natb1:
Closes #2132

## What

`dispatch-write-phase-log`'s `--phase` and `--attempt` flag handlers called
`shift 2` unconditionally. When either flag was the **last argument** (no value
following), only one positional remained, so `shift 2` returned non-zero. Under
the script's `set -euo pipefail` that aborted the whole script with an opaque
**exit 1 and no message**, instead of the **exit 2 with a usage message** that
every other argument error in the script uses.

## Fix

Guard each `shift 2` with a `$# -ge 2` check that emits a clear stderr message
and `exit 2`, matching the script's existing usage-error convention. With the
guard in place the prior `${2:-}` defaulting is unnecessary, so the assignments
become plain `$2`. The rest of the parse loop, the downstream `PHASE`/`ATTEMPT`
regex validation, and the upsert logic are unchanged.

## Test

New hermetic, network-free regression suite
`test-write-phase-log.sh` (auto-discovered by `run-unit-tests.sh`'s
`RUN_PR_SCRIPTS` glob, so it runs in CI's `unit-tests` job) covers all three
acceptance criteria:

- **AC1** — `… 2132 --phase` (trailing flag) exits 2 with `--phase` in stderr;
  aborts at the guard before any git/gh access.
- **AC2** — `… 2132 --phase qa --attempt` (trailing flag) exits 2 with
  `--attempt` in stderr; also pre-git/gh.
- **AC3 (no regression)** — a valid `--phase qa --attempt 2` invocation parses
  past the loop and performs the comment write, run inside a temp git repo with
  a PATH-shim `gh` stub and `DISPATCH_PLAN_AUTHOR_ID` set; asserts exit 0 and a
  recorded `POST` to `issues/2132/comments`.

7/7 assertions pass.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
@@ -40,8 +40,12 @@ PHASE=""
 ATTEMPT="1"
 while [[ $# -gt 0 ]]; do
   case "$1" in
-    --phase)   PHASE="${2:-}"; shift 2 ;;
-    --attempt) ATTEMPT="${2:-}"; shift 2 ;;
+    --phase)
+      [[ $# -ge 2 ]] || { echo 'dispatch-write-phase-log: --phase requires a value' >&2; exit 2; }
+      PHASE="$2"; shift 2 ;;
+    --attempt)
+      [[ $# -ge 2 ]] || { echo 'dispatch-write-phase-log: --attempt requires a value' >&2; exit 2; }
+      ATTEMPT="$2"; shift 2 ;;
     *)
       echo "dispatch-write-phase-log: unknown argument '$1'" >&2
       echo "usage: dispatch-write-phase-log <issue-num> --phase <p> [--attempt <n>]" >&2
--- a/.claude/skills/dispatch-propagate/scripts/test-write-phase-log.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-write-phase-log.sh
@@ -1,0 +1,109 @@
+#!/usr/bin/env bash
+# Tests for dispatch-write-phase-log flag-value guard (#2132).
+#
+# AC1: --phase with no value exits 2 with '--phase' in stderr
+# AC2: --attempt with no value exits 2 with '--attempt' in stderr
+# AC3: valid invocation (--phase qa --attempt 2) exits 0 and POSTs to issues/2132/comments
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd -P)"
+source "$SCRIPT_DIR/test-helpers.sh"
+SUT="$SCRIPT_DIR/dispatch-write-phase-log"
+
+# ============================================================================
+# AC1: --phase with no following value → exit 2, stderr contains '--phase'
+# ============================================================================
+
+echo "AC1: --phase with no value -> exit 2, stderr mentions --phase"
+
+set +e
+AC1_STDERR=$(printf 'body\n' | "$SUT" 2132 --phase 2>&1)
+AC1_RC=$?
+set -e
+
+assert_eq "AC1: exit code is 2" "2" "$AC1_RC"
+assert_contains "AC1: stderr mentions --phase" "--phase" "$AC1_STDERR"
+
+# ============================================================================
+# AC2: --attempt with no following value → exit 2, stderr contains '--attempt'
+# ============================================================================
+
+echo ""
+echo "AC2: --attempt with no value -> exit 2, stderr mentions --attempt"
+
+set +e
+AC2_STDERR=$(printf 'body\n' | "$SUT" 2132 --phase qa --attempt 2>&1)
+AC2_RC=$?
+set -e
+
+assert_eq "AC2: exit code is 2" "2" "$AC2_RC"
+assert_contains "AC2: stderr mentions --attempt" "--attempt" "$AC2_STDERR"
+
+# ============================================================================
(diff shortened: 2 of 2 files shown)


natb1:
# QA Summary — Issue #2132

## Items Executed

| # | Title | Result |
|---|-------|--------|
| 1 | AC1 + AC2 + AC3 — hermetic regression suite passes | **PASS** |
| 2 | Test is auto-discovered by the unit-test runner (CI parity) | **PASS** |
| 3 | Regression — bare flag does not regress to opaque exit 1 | **PASS** |
| 4 | Regression — guard is additive to existing value validation | **PASS** |

## Results

- PASS: 4
- FAIL: 0
- SKIP: 0

## Detail

**Item 1** — `test-write-phase-log.sh` ran directly; all 7 assertions passed:
- AC1: `--phase` with no value exits 2 with `--phase` in stderr ✓
- AC2: `--attempt` with no value exits 2 with `--attempt` in stderr ✓
- AC3: valid `--phase qa --attempt 2` invocation exits 0 and POSTs to `issues/2132/comments` ✓

**Item 2** — `run-unit-tests.sh --pr-scripts` discovered and ran `test-write-phase-log.sh`; "All unit test suites passed."

**Item 3** — `printf 'body\n' | dispatch-write-phase-log 2132 --phase qa --attempt` exits 2 with "dispatch-write-phase-log: --attempt requires a value" — confirms exit code is 2, not the prior opaque 1.

**Item 4** — `--phase` bare flag exits 2 with "requires a value"; `--attempt x` flows past the new guard to existing regex validation and exits 2 with "must be a non-negative integer, got: 'x'" — confirms the guard is additive, not masking downstream checks.

## Disposition

No residue. All acceptance criteria are machine-verified and passing. No bugs found, no human judgment needed.

natb1:
## Review: PR #2149 (`dispatch-write-phase-log` guard + regression test)

Surface: `code` (shell scripts in `.claude/skills/dispatch-propagate/scripts/`). CodeQL: no open alerts. Erosion: no findings. No dependency changes.

---

### Fixed

_None._

### Required (security)

_None._

### Informational

- **`test-write-phase-log.sh:106`** *(code-review)* — The `assert_contains` label `'AC3: gh called with POST to issues/2132/comments'` asserts a POST to that specific endpoint, but the stub logs all `gh` invocations to the same file. The assertion is correct today because only one POST is made, but if the SUT is extended to call more `gh` commands, the log could contain multiple entries — the assertion would still pass even if the POST target changed. FYI only; not a blocking concern for this PR.

- **`test-write-phase-log.sh:71-75`** *(review)* — The gh stub in AC3 exits 0 as soon as any argument equals `POST`, regardless of which `gh` sub-command is being invoked. If the SUT ever adds a non-create `gh` call that happens to pass `POST` as a literal argument, the stub would early-exit without logging to `$GH_LOG`, and the AC3 assertions could give a false pass. The stub is intentionally minimal for this hermetic test, so this is a known trade-off. Surfaced for awareness.

### Dismissed

_None._

_None._

### False-positive (security)

- **`test-write-phase-log.sh:90`** *(red-team)* — Flagged: test harness prepends caller-controlled `$TMPDIR_TEST/bin` to PATH before launching the SUT. Assessed false-positive: this is an intentional test isolation pattern — the temp directory is created by the test itself (`mktemp -d`), not by external input, and the test runs in a controlled environment with no privilege escalation surface. The PATH manipulation is confined to the test process and its children.

- **`dispatch-write-phase-log:43-48`** *(red-team)* — Flagged: `$# -ge 2` guard allows a caller to pass an empty string `""` as the flag value, which passes the guard and reaches the regex validation downstream. Assessed false-positive: the subsequent `PHASE` regex validation (`^[a-z]`) rejects an empty string with an existing usage error. Empty-string injection is caught at the validation step; the guard's sole purpose is to prevent `shift 2` from underflowing, which it does correctly.

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #2150, opened 2026-06-20
dispatch-write-phase-log: anchor REPLACE-vs-APPEND selector to a full-line match

natb1:
Closes #2136

## Problem

`dispatch-write-phase-log` upserts a per-`(phase, attempt)` handoff entry into an
issue's `<!-- dispatch:phase-log -->` comment. On write it picks REPLACE (section
already present → update in place) vs APPEND (new section) by testing whether the
inner marker `<!-- phase-log:<phase>:<attempt> -->` is already in the comment body.

The two predicates that should agree on "is this section present" disagreed:

- **Selector (line 97):** `grep -qF "$INNER_MARKER"` — a **substring** match
  anywhere in the body.
- **awk replace block (line 117):** `if ($0 == target)` — an **exact full-line**
  match.

If the inner-marker text appeared only as a substring embedded in section prose
(e.g. a phase-log entry that quotes the marker — plausible on phase-log-related
issues), `grep -qF` selected the REPLACE branch, but the awk state machine found
no line where `$0 == target` and emitted the body unchanged. The subsequent PATCH
then wrote the unmodified body back, **silently dropping the new phase-log entry**.

## Fix

Add the `-x` flag to the selector: `grep -qF` → `grep -qxF`. The `-x` flag anchors
grep to match only when the marker occupies its own whole line, aligning the
presence-test with the awk exact-line semantics. The REPLACE branch is now taken
only when awk can actually find and replace the section. Single-character change;
the awk block, POST/PATCH calls, and argument parsing are untouched.

## Verification

The plan's `verify` block (syntax check + selector-line text assertion +
behavioral test of the fixed predicate against substring-in-prose vs own-line
fixtures, mirroring awk's exact-line semantics without `gh`/network) passes.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
@@ -94,7 +94,7 @@ fi
 # Comment exists — fetch its body and upsert the (phase, attempt) section.
 BODY=$(gh api "repos/{owner}/{repo}/issues/comments/$CID" --jq '.body')
 
-if printf '%s' "$BODY" | grep -qF "$INNER_MARKER"; then
+if printf '%s' "$BODY" | grep -qxF "$INNER_MARKER"; then
   # REPLACE the existing section in place. The awk state machine drops the lines
   # from the target inner-marker line up to (but not including) the next line
   # beginning '<!-- phase-log:' (the next section's delimiter, which is KEPT) or


natb1:
# QA Summary — Issue #2136 / PR #2150

## Items Executed

| # | Title | Result |
|---|-------|--------|
| 1 | Line 97 uses `grep -qxF` (fix is present) | PASS |
| 2 | Script syntactically valid after edit | PASS |
| 3 | Marker embedded in prose selects APPEND | PASS |
| 4 | Marker on its own line selects REPLACE; awk replaces correctly | PASS |

## Summary

- Passed: 4
- Failed: 0
- Skipped: 0

## Details

**Item 1 — Line 97 uses `grep -qxF`:** `grep -qxF "$INNER_MARKER"` found at line 97; no un-anchored `grep -qF "$INNER_MARKER"` remains.

**Item 2 — Syntax check:** `bash -n` exits 0 — no parse errors.

**Item 3 — Substring-in-prose selects APPEND:** When `<!-- phase-log:qa:1 -->` appears only inside a prose line, `grep -qxF` returns non-zero → APPEND branch taken. The old `grep -qF` would have falsely matched; the fix prevents the silent-drop.

**Item 4 — Own-line marker selects REPLACE, awk replaces correctly:** When the marker occupies its own line, `grep -qxF` returns zero → REPLACE branch. The awk state machine substituted the `qa:1` section with new content, kept the outer `<!-- dispatch:phase-log -->` marker, kept the adjacent `<!-- phase-log:review:1 -->` section, and discarded the old body. No regression on the normal replace path.

All four acceptance criteria are satisfied. Clean pass — no residue.

natb1:
## Review summary — PR #2150

**Surface:** `code` | **Findings:** 2 (0 actionable, 0 fixed) | **Security:** CodeQL clean, erosion clean

### Informational

- **CR-1** `.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log:97` — Consider adding a comment explaining why the `-x` flag is needed (e.g., `# -x anchors to full-line match so grep aligns with awk "$0 == target" at line 117`). _(code-review)_
- **R-1** `.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log` — No issues found; the single-character fix is minimal and correct. _(review)_

### Fixed

_None._

### Security (Required / False-positive / Out-of-scope)

_None. CodeQL: 0 open alerts. npm: no dependency changes. Erosion: 0 findings._

### Deferred

_None._

---

PR #2152, opened 2026-06-20
type-safety-sensor: percent-encode commas in annotation file paths

natb1:
Closes #2092

## Problem

The diff-scoped type-safety escape-hatch sensor
(`.github/scripts/check-type-safety-escapes.sh`) emits GitHub Actions
annotations in the form:

```
::error file=<path>,line=N::<message>
```

GitHub's workflow-command format uses the comma as the key–value delimiter
between annotation properties. A TypeScript file path containing a comma — e.g.
`src/weird,file.ts` — is interpolated raw into the `file=` property, so GitHub
splits it at the comma: `file=src/weird` is truncated and `file.ts` becomes an
unrecognized key, causing the annotation to be silently dropped. The violation
then never surfaces in the PR check output.

This was surfaced during review of PR #2077 and deferred because comma-bearing
TypeScript paths are rare.

## Fix

Percent-encode commas in the path before building the annotation, in the awk
`emit` function:

```awk
gsub(/,/, "%2C", path)
```

`path` is a function-local parameter, so the `gsub` mutates only the local copy
— it does not affect the awk-global `path` used elsewhere. A comma path now
produces a well-formed `::error file=src/weird%2Cfile.ts,line=1::…` annotation.

Scope is commas only, exactly as the acceptance criterion specifies — other
GitHub property-encoding characters (`%`, `:`, CR/LF) are out of scope. The
change is backward-compatible: comma-free paths are emitted byte-for-byte as
before.

## Test

Added one positive fixture case to
`.github/scripts/test-check-type-safety-escapes.sh` asserting a comma path emits
`file=src/weird%2Cfile.ts`, and bumped the expected-total guard from 18 to 19.

```
Results: 19 passed, 0 failed
```

### DIFF
--- a/.github/scripts/check-type-safety-escapes.sh
+++ b/.github/scripts/check-type-safety-escapes.sh
@@ -102,6 +102,7 @@ scan_diff() {
     END { if (violations > 0) exit 1 }
 
     function emit(path, line, hatch) {
+      gsub(/,/, "%2C", path)
       printf "::error file=%s,line=%d::Net-new type-safety escape hatch (%s). Suppress with `// type-safety-ok: <reason>` if intentional.\n", path, line, hatch
       violations++
     }
--- a/.github/scripts/test-check-type-safety-escapes.sh
+++ b/.github/scripts/test-check-type-safety-escapes.sh
@@ -412,6 +412,25 @@ EOF
   fi
 )
 
+echo ""
+echo "=== Comma-in-path: file= comma is percent-encoded to %2C ==="
+(
+  F="${TEST_TMPDIR}/f"; cat > "$F" <<'EOF'
+--- a/src/weird,file.ts
++++ b/src/weird,file.ts
+@@ -1,0 +1,1 @@
++const x = y as Foo;
+EOF
+  scan_fixture "$F"
+  n=$(err_count)
+  if [ "$RC" -eq 1 ] && [ "$n" -eq 1 ] && \
+     grep -q '^::error file=src/weird%2Cfile.ts,line=1::' "$OUT_FILE"; then
+    pass "comma-in-path -> file=src/weird%2Cfile.ts (comma percent-encoded)"
+  else
+    fail "comma-in-path (rc=$RC count=$n): $(cat "$OUT_FILE")"
+  fi
+)
+
 # ---------------------------------------------------------------------------
 # RESULTS + expected-total guard (catches a crashed/early-exiting subshell).
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


natb1:
# QA Summary — Issue #2092 (PR #2152)

## Items executed

| # | Title | Result |
|---|-------|--------|
| 1 | Fixture test suite passes (including the new comma-path case) | PASS |
| 2 | gsub comma-encoding is present in the emit function | PASS |
| 3 | Comma-path fixture test case is present in the test file | PASS |
| 4 | EXPECTED count in test file is 19 | PASS |
| 5 | Comma-encoded annotation is well-formed when path contains a comma | PASS |

**Counts:** 5 passed, 0 failed, 0 skipped

## Bugs found

None.

## Disposition triage

No residue — Step 3.5 was skipped. Every item passed on the first run.

## Clean pass

All acceptance criteria satisfied:
- `emit()` in `.github/scripts/check-type-safety-escapes.sh` percent-encodes commas via `gsub(/,/, "%2C", path)`
- The existing fixture suite passes (19 passed, 0 failed)
- A new comma-in-path fixture test case is present and passes (`file=src/weird%2Cfile.ts`)
- The EXPECTED guard is bumped to 19

natb1:
## Review: PR #2152 — type-safety-sensor comma-in-path fix

**Surface:** `code` (2 shell scripts changed)
**Findings:** 4 surfaced, 0 actionable, 0 fixes applied

---

### Informational

- **R1** `.github/scripts/check-type-safety-escapes.sh:104` — `gsub` modifies a local copy of `path` (awk function parameters are by-value), but a reader unfamiliar with awk scoping rules might assume the awk-global `path` is mutated. Suggested: add an inline comment confirming the parameter is a local copy, e.g. `# local copy — gsub does not affect awk-global path`.

- **R2** `.github/scripts/test-check-type-safety-escapes.sh` — No test for a path that already contains `%` at baseline. If a future change introduced `%` encoding before the comma-encode step, `%2C` could double-encode to `%252C`. Suggested: add a fixture with a `%`-bearing path to confirm no double-encoding occurs.

- **R3** `.github/scripts/check-type-safety-escapes.sh:104` — Other GitHub annotation property-delimiter characters (`%`, `:`, CR/LF) are out of scope per the acceptance criterion, but this is not documented at the change site. Suggested: add a comment such as `# only commas per acceptance criterion; %, :, CR/LF encoding is out of scope`.

---

### False-positive (security)

- **S1** `.github/scripts/check-type-safety-escapes.sh`, `.github/scripts/test-check-type-safety-escapes.sh` — No exploitable attack surface: both changed files are CI scripts run in a trusted GitHub Actions context. They are not executed against user-supplied input in any way that affects runtime security. No action needed.

---

### Deferred
_None._

### Out-of-scope (security)
_None._

### Fixed
_None._

PR #2108, opened 2026-06-20
dispatch: carry a structured cross-phase phase-log handoff note between workers

natb1:
Closes #2039

## What

Carry a structured cross-attempt / cross-phase handoff note between dispatch
workers, so a retry or downstream phase inherits "what was tried and why it
failed" instead of re-deriving it from the PR diff.

Two deliverables, both **advisory** (an absent note degrades cleanly to current
behavior — durable phase signals stay the labels and per-session markers):

1. **Reflexion-style retry input** — a retry `qa-fix` worker now reads the most
   recent `tmp/qa-fix-summary-<N>.md` on entry (same-worktree, fast, rich) and
   feeds it into triage + the fix-planner, so it does not re-run already-resolved
   QA items or re-derive known failures.
2. **A `<!-- dispatch:phase-log -->` issue comment** (GitHub-state-keyed,
   mirroring the existing `<!-- dispatch:plan -->` pattern) that each phase
   (`implement` / `qa-fix` / `review-fix`) writes a terse
   "phase · attempt · what-failed · what-changed" entry to, and reads on entry —
   the durable, cross-worktree / cross-phase note that survives worktree reaping.

## How — idempotent upsert, not blind append

Phases re-execute on the routine retry / crash-recovery path, so a blind append
would stack duplicates and break the idempotency the plan scripts advertise. The
write is an **idempotent upsert keyed on (phase, attempt)** via an inner marker
`<!-- phase-log:<phase>:<attempt> -->`:

- A genuine new attempt → a real append (the cross-attempt handoff).
- A crash-rerun of the same logical pass → an in-place replace (no duplicate).

Dedup keys on the robust inner marker, never on fuzzy header text. The outer
marker (`<!-- dispatch:phase-log -->`) and the inner marker prefix
(`<!-- phase-log:`) are deliberately disjoint so the outer is never mistaken for
a section boundary.

## Units

1. **lib.sh `dispatch_marker_comment_id` helper + `dispatch-write-phase-log`** —
   the write path. The helper consolidates the author-id-filtered find jq;
   `dispatch-write-phase-log` upserts the per-(phase,attempt) section (awk state
   machine for in-place replace, `printf` for append). The awk replace/append
   logic was unit-tested without any `gh` round-trip (16/16 assertions: replace a
   middle section, append a new key, replace the last section, never consume the
   outer marker).
2. **`--phase-log` read section in `dispatch-context-pack`** — emits
   `=== PHASE-LOG #N ===` with the comment body, or the `phase-log: none`
[...]

PR #2153, opened 2026-06-20
implement + review-fix: skip phase-log write on crash-recovery re-entry

natb1:
Closes #2137

## Problem

PR #2108 added a cross-phase phase-log handoff note: each phase worker
(`/implement`, `/review-fix`) composes a terse "what-changed" digest and upserts
it into the issue's phase-log comment, keyed on `(phase, attempt)`. The upsert
replaces the prior entry in place.

On the **crash-recovery re-entry path** the digest was recomputed from the
current turn's *limited* context and upserted unconditionally — overwriting the
prior, richer `(phase, 1)` entry with a thinner restatement. Safe against
duplication, but a semantic regression in exactly the recovery scenario the
feature targets: the re-entry that was supposed to preserve the original
attempt's failure/success context overwrote it.

The fix already existed as a precedent in `review-fix/SKILL.md`: the
outcome-envelope emit already skips entirely on re-entry. The phase-log write a
few paragraphs earlier simply lacked the same gate.

## Change

Gate the phase-log write so it runs **only when the implement/review steps
actually executed this turn**; on the crash-recovery / idempotent re-entry path,
skip the write entirely so the prior entry is preserved verbatim.

- `implement/SKILL.md` — Step 5 phase-log write is now gated on the preamble's
  re-entry branch. Critically, the discriminator keys on *"did Steps 1–4 run this
  turn"*, **not** on PR presence: `PR_NUM` is set on both paths at Step 5, so a
  `[ -n "$PR_NUM" ]` gate would skip the write always and break the normal path.
- `review-fix/SKILL.md` — Step 7 phase-log write reuses the existing
  *"only when the Workflow ran this session"* condition, reading identically to
  the outcome-envelope gate.
- The false-safety *"so the repeat is safe"* rationale is replaced in both files
  with the correct reasoning: on re-entry there is no fresh outcome data, so
  recomposing would overwrite the prior entry with a thinner restatement;
  therefore the write is skipped, not repeated.
- The orthogonal failure-vs-success instruction (*"Compose it unconditionally
  (no 'only on failure' branch)"*) is preserved in both files.

## Verification

The plan's deterministic `verify` block passes — it confirms the false-safety
claim is gone from both files, the failure-vs-success instruction is retained,
and each file carries its re-entry skip gate. Behavioral confirmation (re-enter
each skill with a PR / `dispatch:reviewed` already present and confirm the prior
phase-log entry is unchanged) is manual, per the plan.

### DIFF
--- a/.claude/skills/implement/SKILL.md
+++ b/.claude/skills/implement/SKILL.md
@@ -63,11 +63,15 @@ a prior attempt or earlier phase left. Treat the sentinel `phase-log: none` as
 empty. When non-empty, feed it into the Step 2 unit-build context so a rebuild or
 re-plan sees why a prior attempt diverged. An absent note leaves Step 2 unchanged.
 
-If a PR already exists, the build + PR already happened: capture its number as
-`PR_NUM`, **skip Steps 1–4**, and go straight to the Step 5 marker write (this
-covers a same-tick crash between PR-open and marker-write). If no PR exists, run
-all steps. (`dispatch-route` normally routes a PR-bearing issue to fix-checks/qa/review,
-not implement, so this is a same-tick crash-recovery edge.)
+If a PR already exists, the build + PR already happened: this is the **re-entry
+branch**. Capture its number as `PR_NUM`, **skip Steps 1–4**, and go straight to
+Step 5 — on re-entry only the **marker** write runs; the Step 5 **phase-log write
+is skipped** (this covers a same-tick crash between PR-open and marker-write).
+Note that the preamble took the re-entry branch (a PR already existed at entry) —
+Step 5 keys its phase-log gate on this. If no PR exists, run all steps (the
+preamble did NOT take the re-entry branch). (`dispatch-route` normally routes a
+PR-bearing issue to fix-checks/qa/review, not implement, so this is a same-tick
+crash-recovery edge.)
 
 ## Steps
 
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — Issue #2137 / PR #2153

## Results

| Shell-command | 3 | 3 | 0 | 0 |
| Browser | — | — | — | — |

**Total: 3 PASS, 0 FAIL, 0 SKIP**

No browser component detected (changed files: `.claude/skills/implement/SKILL.md`, `.claude/skills/review-fix/SKILL.md` only). Browser walkthrough skipped.

## Items checked

1. **False-safety rationale removed and re-entry skip documented in both skills** — PASS  
   Plan verification script: `OK: phase-log re-entry guards present in both skills`

2. **implement/SKILL.md Step 5 phase-log write gated on Steps 1–4 running, not PR presence** — PASS  
   Found "only when Steps 1–4 ran this turn" at line 229.

3. **review-fix/SKILL.md Step 7 phase-log write reuses "Workflow ran this session" gate** — PASS  
   Found at lines 616 and 718.

4. **Non-recovery path still writes; recovery path leaves prior entry unchanged** — assessed by Opus (see below)

**Item 4 — Non-recovery path still writes the full phase-log entry; recovery path leaves the prior entry unchanged**

The gating logic is present and provably correct in both files. `implement/SKILL.md` Step 5 writes the full phase-log entry on the executed path and skips on the crash-recovery re-entry path, gating on the re-entry flag (not PR presence) so the prior `(implement,1)` entry is preserved verbatim. `review-fix/SKILL.md` Step 7 writes only when the Workflow ran this session and skips on re-entry, mirroring the outcome-envelope skip-on-re-entry precedent. No code defect exists and no human input is needed — acceptance criterion is met as shipped (commit `0bfc1759`).

## Bugs found

None.

## Deferred to office-hours

None — all items PASSed or were assessed by Opus as already satisfied.

natb1:
## Review — PR #2153 (issue #2137)

**Security:** no attack surface — docs-only diff (no executable, config, dependency, or Firestore-rules changes).

### Informational

- **RF-01** `.claude/skills/review-fix/SKILL.md` — Phase-log gate should use a dedicated boolean flag, not absence of `result`. Checking `result` absence is implicitly correct but can be ambiguous if the Workflow returns null on failure rather than on re-entry. _Recommended:_ introduce a named flag (e.g. `WORKFLOW_RAN_THIS_SESSION=true`) set at the Workflow invocation point and gate on it.

- **RF-02** `.claude/skills/review-fix/SKILL.md:Step 7` — No explicit handling for the `result is null` case (Workflow failure, not re-entry). The current text only mentions `result` being absent on re-entry, but a null return on first run should not silently skip the phase-log write. _Recommended:_ add a handling clause distinguishing re-entry (intentional skip) from Workflow failure (should error/park).

- **RF-03** `.claude/skills/review-fix/SKILL.md` — The re-entry detection (preamble label check) and the Step 7 gate phrase ("only when the Workflow ran this session") are stated in separate sections without a cross-reference. They are implicitly the same condition but readers must infer it. _Recommended:_ add a one-line cross-reference tying them together.

- **RF-04** `.claude/skills/implement/SKILL.md:Step 5` — The phrase "preamble did NOT take the re-entry branch" appears three times without being tied to a concrete variable name. _Recommended:_ name the boolean explicitly in the skill text (e.g. `REENTRY=true`) so implementors have an unambiguous gate condition.

- **RF-06** `.claude/skills/implement/SKILL.md:Step 5`, `.claude/skills/review-fix/SKILL.md:Step 7` — No guidance on what to do if `dispatch-write-phase-log` fails on the normal path. _Recommended:_ clarify whether a write failure is fatal (stop, park) or non-fatal (continue to marker write).

### Deferred

- **RF-05** `.claude/skills/implement/SKILL.md`, `.claude/skills/review-fix/SKILL.md` — Missing behavioral verification procedure for the re-entry skip. The Verification section in the PR checks static text only; no test confirms that on re-entry the prior phase-log entry is actually preserved, or that the normal path writes it. Filed as #2163 (blocked by #2137).

### Fixed

_None._

### Dismissed

_None._

PR #2158, opened 2026-06-20
Fix awk REPLACE dropping blank separators in dispatch-write-phase-log

natb1:
Closes #2139

## Problem

`dispatch-write-phase-log` upserts a per-`(phase,attempt)` handoff section into
the issue's single `<!-- dispatch:phase-log -->` comment. When a section's inner
marker already exists, the awk REPLACE state machine drops the old section and
substitutes the new one. While `skip=1` it dropped **every** line that did not
begin `<!-- phase-log:` — including the single blank separator line the APPEND
path places between sections.

Effect: each idempotent re-upsert of a **non-last** section ate one blank line
before the following section. Crash-recovery re-runs re-upsert the same marker
repeatedly, so multi-section comments progressively collapsed — sections ran
together with no whitespace.

## Fix

In the REPLACE awk block, when the skip-mode encounters the next section's
`<!-- phase-log:` header, emit `print ""` before resetting `skip` and printing
the header — restoring exactly one blank separator before the following section.
The separator is emitted lazily, right before the next header, and never at EOF,
so:

- **Middle/non-last replace** → `repl\n\n<next-header>` — one blank, identical
  to the original.
- **Last-section replace** → ends `repl\n` — no trailing blank, byte-for-byte
  identical to the APPEND output, with no dependence on shell/GitHub
  trailing-newline normalization (per `.claude/rules/code-style.md`).

The `print repl; skip = 1` target block and the APPEND path are untouched.

## Tests

New self-contained `test-write-phase-log.sh` (auto-discovered by
`run-unit-tests.sh`). It uses a PATH-shim with stateful `gh`/`git` stubs that
round-trip the real comment bytes through the real awk, and asserts byte-for-byte
equality with `cmp`:

1. Non-last re-upsert is idempotent (blank separator preserved).
2. Three-section middle re-upsert, twice (all separators intact).
3. Last-section re-upsert is byte-for-byte identical to its APPEND (no trailing
   blank added).
4. APPEND still emits exactly one blank separator between sections.

Confirmed non-vacuous: against the pre-fix awk, cases 1 and 2 fail.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
@@ -98,9 +98,13 @@ if printf '%s' "$BODY" | grep -qF "$INNER_MARKER"; then
   # REPLACE the existing section in place. The awk state machine drops the lines
   # from the target inner-marker line up to (but not including) the next line
   # beginning '<!-- phase-log:' (the next section's delimiter, which is KEPT) or
-  # EOF, substituting the new section. The outer marker '<!-- dispatch:phase-log -->'
-  # begins '<!-- dispatch:' and is deliberately disjoint from the '<!-- phase-log:'
-  # terminator, so it is never treated as a section boundary.
+  # EOF, substituting the new section. When a next section header is found, a
+  # single blank separator line is re-emitted before it so the output is
+  # byte-for-byte identical to the original (one blank between sections). For the
+  # last section (no next header), the machine reaches EOF and emits no trailing
+  # blank — matching the APPEND output. The outer marker '<!-- dispatch:phase-log
+  # -->' begins '<!-- dispatch:' and is deliberately disjoint from the
+  # '<!-- phase-log:' terminator, so it is never treated as a section boundary.
   REPLACEMENT="$NEW_SECTION" TARGET="$INNER_MARKER" \
     awk '
       BEGIN { repl = ENVIRON["REPLACEMENT"]; target = ENVIRON["TARGET"]; skip = 0 }
@@ -108,6 +112,7 @@ if printf '%s' "$BODY" | grep -qF "$INNER_MARKER"; then
         if (skip) {
           # Inside the target section being dropped: a new inner marker ends the skip.
           if (index($0, "<!-- phase-log:") == 1) {
+            print ""          # restore the single blank separator before the next section
             skip = 0
             print
           }
--- a/.claude/skills/dispatch-propagate/scripts/test-write-phase-log.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-write-phase-log.sh
@@ -1,0 +1,280 @@
+#!/usr/bin/env bash
+# test-write-phase-log.sh — tests for dispatch-write-phase-log REPLACE path.
+# Verifies that the awk state machine preserves blank separator lines between
+# phase-log sections on repeated upserts (issue #2139).
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+source "$SCRIPT_DIR/test-helpers.sh"
+
+# ---- assert_files_eq: byte-for-byte file comparison using cmp -s -----------
+# Uses files rather than command substitution to avoid trailing-newline stripping.
+assert_files_eq() {
+  local label="$1" expected_file="$2" actual_file="$3"
+  TOTAL=$((TOTAL + 1))
+  if cmp -s "$expected_file" "$actual_file"; then
+    PASS=$((PASS + 1))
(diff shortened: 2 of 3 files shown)


natb1:
# QA Summary — Issue #2139 (PR #2158)

## Results

| 1. Self-contained idempotency test passes | script-verifiable | PASS |
| 2. Fixed script is syntactically valid | script-verifiable | PASS |
| 3. New test file exists at the stated path | script-verifiable | PASS |
| 4. Full unit suite auto-discovers and passes the new test | script-verifiable | PASS |
| 5. The `print ""` re-emit is gated to the header branch and cannot fire at EOF | needs-human-judgment | PASS (already-satisfied, see below) |

**Counts:** 4 script-verifiable PASSed, 0 FAILed, 0 SKIPped; 1 needs-human-judgment item dropped as PASS by Opus.

No browser walkthrough (pure shell/awk change — no frontend component detected).

## Findings

No bugs found. All acceptance criteria verified:
1. Non-last re-upsert produces byte-for-byte identical output (test 1: PASS).
2. Three-section middle re-upsert twice preserves all separators (test 2: PASS).
3. Last-section re-upsert adds no trailing blank (test 3: PASS).
4. APPEND still emits exactly one blank separator between sections (test 4: PASS).

Full unit suite: 3483/3483 passed (including test-write-phase-log.sh, auto-discovered).

- **Item 5 — The `print ""` re-emit is gated to the header branch and cannot fire at EOF:**
  The structural guarantee is provable from the code, not a judgment call. In `dispatch-write-phase-log` lines 112–119, `print ""` (line 115) is reachable only via `if (skip)` → `if (index($0, "<!-- phase-log:") == 1)`. On a last-section replace, the target header sets `skip=1`, then every remaining line is dropped via `next` until EOF; no post-section line begins with `<!-- phase-log:`, so the blank-re-emit branch is never entered — byte-for-byte identical to APPEND. Mechanically verified by test 3 (`cmp -s` identity assertion). Acceptance criterion 3 is decided by both code structure and passing tests — no subjective judgment remains. Dropped as PASS.

natb1:
## Review: #2139 (PR #2158)

### Fixed

- **`test-write-phase-log.sh:189` — PATH prepend accumulates on every `run_write` call** (`code-review`, `red-team`): `export PATH="$BIN:$PATH"` inside `run_write` prepends `$BIN` on every invocation, growing PATH unboundedly. Guarded with `[[ ":$PATH:" != *":$BIN:"* ]]` so `$BIN` is prepended only once. Fix commit: `6a579a81`.

- **`test-write-phase-log.sh:1` — test file not wired into CI** (`review`): `test-write-phase-log.sh` was not invoked by any CI job. Added a step to the `hook-tests` job in `.github/workflows/unit-tests.yml` to run it, matching the existing pattern for `test-dispatch-scripts.sh`. Fix commit: `6a579a81`.

### Informational

- **`test-write-phase-log.sh:270-275` — `blanks` uninitialized in test case 4 awk** (`code-review`): The blank-counting awk uses `blanks` without initializing it. In awk, uninitialized numeric variables default to 0, so this is not a bug — the count is correct. Noted for clarity.

- **`test-write-phase-log.sh:148,150` — gh stub single-comment GET returns raw file content rather than JSON** (`review`): The stub handler for `repos/owner/repo/issues/comments/100` returns the raw body file content directly (not JSON-encoded). This works because `dispatch-write-phase-log` uses `--jq .body` which extracts `.body` from JSON, but the stub short-circuits this by returning the text directly when `JQ_FILTER == ".body"`. Functional for the test purpose, but structurally unconventional.

### Dismissed

- **`test-write-phase-log.sh:185,189` — duplicate PATH prepend finding** (`review`): Same issue as `code-review-0` above — already fixed by the guarded prepend. Dismissed as duplicate.

### Deferred

- **`test-write-phase-log.sh:200,230` — no test for section content with embedded blank lines** (`review`): The awk skip state drops all lines not starting with `<!-- phase-log:` while `skip=1`. No test case exercises section body content containing a blank line. A future modification to the skip-termination logic that accidentally treated embedded blanks as terminators would go undetected. Filed as #2164 (blocked by #2139).

### Out-of-scope (security)

- **`dispatch-write-phase-log:114-118` — injected content in phase-log comment could survive re-upserts** (`red-team`): An actor with write access to a GitHub issue comment could inject a crafted `<!-- phase-log:` header into the comment body, causing the REPLACE state machine to misidentify section boundaries. Pre-existing in code the diff did not change; out of scope for this PR.

- **`dispatch-write-phase-log:95` — write-access actor can overwrite phase-log comment body** (`red-team`): An actor with GitHub write access to the repository could PATCH the phase-log comment to arbitrary content. Pre-existing trust-boundary behavior; out of scope for this PR.

PR #2170, opened 2026-06-20
dispatch-write-phase-log: reject issue-num 0 via ^[1-9][0-9]*$ validation

natb1:
Closes #2134

## What

Tighten the `issue-num` argument validation in
`.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log` from
`^[0-9]+$` to `^[1-9][0-9]*$`, so `0` (and `00`) are rejected.

## Why

GitHub issue numbers start at 1 — issue #0 does not exist — and the script's own
error message already promises "must be a positive integer", but the old regex
accepted `0`. The sibling script `dispatch-context-pack` already validates the
same kind of argument with `^[1-9][0-9]*$`, so the two were inconsistent. This
makes the guard match its own error message and the project's established
pattern, closing a latent inconsistency surfaced as a backlink from #2108.

All current callers derive the issue number from a branch name (always ≥ 1), so
this never triggered in practice; the change is a guard tightening with no
behavioral effect on valid input.

## Changes

- `dispatch-write-phase-log:33` — regex tightened to `^[1-9][0-9]*$`; the
  error message is unchanged (the edit makes it accurate).
- `test-write-phase-log.sh` — new regression case asserting issue-num `0` exits
  2 with a "positive integer" message. The validation fires before any git/gh
  calls, so the case needs no temp repo or `gh` stub.

## Verification

`run-unit-tests.sh --pr-scripts` passes, including the new 0-rejection
assertions in `test-write-phase-log.sh`.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
@@ -30,7 +30,7 @@ if [[ $# -lt 1 || -z "$N" ]]; then
   echo "usage: dispatch-write-phase-log <issue-num> --phase <p> [--attempt <n>]   (entry body on STDIN)" >&2
   exit 2
 fi
-if [[ ! "$N" =~ ^[0-9]+$ ]]; then
+if [[ ! "$N" =~ ^[1-9][0-9]*$ ]]; then
   echo "dispatch-write-phase-log: issue-num must be a positive integer, got: $N" >&2
   exit 2
 fi
--- a/.claude/skills/dispatch-propagate/scripts/test-write-phase-log.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-write-phase-log.sh
@@ -40,6 +40,20 @@ assert_eq "AC2: exit code is 2" "2" "$AC2_RC"
 assert_contains "AC2: stderr mentions --attempt" "--attempt" "$AC2_STDERR"
 
 # ============================================================================
+# issue-num 0 → exit 2, stderr mentions 'positive integer' (#2134)
+# ============================================================================
+
+echo ""
+echo "issue-num 0: exit 2, stderr mentions positive integer"
+
+set +e
+ZERO_STDERR=$(printf 'body\n' | "$SUT" 0 --phase plan 2>&1)
+ZERO_RC=$?
+set -e
+assert_eq "issue-num 0: exit code is 2" "2" "$ZERO_RC"
+assert_contains "issue-num 0: stderr mentions positive integer" "positive integer" "$ZERO_STDERR"
+
+# ============================================================================
 # AC3: valid invocation → exit 0, gh stub records a POST to issues/2132/comments
 # ============================================================================
 


natb1:
# QA Summary — Issue #2134 / PR #2170

## Items executed

All items ran in the **shell-command lane** (scripts-only change, no browser component).

| # | Title | Result |
|---|-------|--------|
| 1 | issue-num 0 is rejected with a positive-integer error | PASS |
| 2 | Leading-zero / multi-zero issue-nums are rejected | PASS |
| 3 | Non-numeric and empty issue-nums still rejected | PASS |
| 4 | Valid positive issue-num still works (golden path) | assessed by Opus (already satisfied) |
| 5 | Regex matches dispatch-context-pack's pattern (AC3) | PASS |
| 6 | Full PR-scripts unit suite passes | PASS |

**Counts:** 5 PASS, 0 FAIL, 0 SKIP, 1 assessed-already-satisfied.

## Bugs found

None.

## Disposition triage

Step 3.5 ran (one residue item: the `needs-human-judgment` triage of item 4).

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR. (No needs-main items this pass.)

- **Item 4 — Valid positive issue-num still works (golden path):** Golden-path confirmation is already provably met by the passing test suite (3545/3545, including the AC3 test that verifies a valid invocation proceeds past the validation gate). There is no code defect to fix and no human input needed. The triage agent's `needs-human-judgment` framing only concerned the impossibility of full end-to-end verification without a gh stub or network connection, but the validation-gate behavior under test is positively confirmed by the existing passing tests.

## Result

**Clean pass.** All acceptance criteria are met. `dispatch:qa-done` applied.

natb1:
## Review findings — PR #2170

### Fixed
_No code changes committed (working tree was clean after the review pass)._

- **R1** (`test-write-phase-log.sh:337`) — Test assertion order: exit code checked before stderr message. Workflow classified as Fixed; the file already had the correct order, so no edit was needed.
- **R4** (`dispatch-write-phase-log:33`) — Inline comment on the `^[1-9][0-9]*$` regex to note it rejects `0` and leading zeros. Workflow fix agent reported this as Fixed but did not materialize the edit; the line remains uncommented.

### Deferred
- **R2** (`test-write-phase-log.sh:327`) — No regression test for leading-zero input (e.g., `"00"`). → #2183
- **R3** (`test-write-phase-log.sh:327`) — No regression test for non-numeric input (e.g., `"abc"`). → #2183

### Dismissed
- **R5** (`dispatch-write-phase-log:33`) — Bash `=~` with unquoted RHS pattern: correct and conventional in bash; not fragile in practice. Dismissed.

### False-positive (security)
- **R6** — No upper-bound on issue-num could theoretically cause a loop. Not an actual vulnerability — the script passes the value to a GitHub API call, not a loop counter. False-positive.

### Informational
- **R7** — Test captures exit code and stderr atomically via `set +e` / `set -e` around a single pipeline; no race concern in practice.

### Security
No CodeQL alerts. No npm dependency changes. No attack surface findings.

PR #2178, opened 2026-06-20
test-generate-phase-model-policy.sh: source shared test-helpers.sh

natb1:
Closes #2165

## What

`test-generate-phase-model-policy.sh` defined its own `assert_eq`,
`report_results`, and `PASS`/`FAIL`/`TOTAL` counters inline, duplicating the
shared helpers in
`.claude/skills/dispatch-propagate/scripts/test-helpers.sh` that every sibling
test in the skill family already sources (e.g.
`test-dispatch-phase-model.sh`).

This replaces the inline `--- test helpers ---` block with a single
`$SCRIPT_DIR`-relative `source` of the shared file, matching the existing idiom
the script uses for `dispatch-config-load`.

## Why

The inline copies were a maintenance liability: they could not use the shared
file's extra helpers (`assert_contains`, `assert_exit_nonzero`,
`assert_file_contains`, `assert_file_exists`), they diverged independently, and a
behavioral divergence was already present — the inline `assert_eq` used `[[` and
single-quoted the expected/actual in FAIL output, while the shared `[`-based
version does not quote. Sourcing the shared helpers removes the duplication and
the divergence.

## Behavior preserved

The shared `report_results` calls `exit 1` when `FAIL > 0`, whereas the inline
one did not. The script's final two lines remain `report_results` then
`exit "$FAIL"`. On an all-pass run `FAIL=0`, so the shared `report_results` does
not exit and `exit "$FAIL"` exits 0 — preserving the green-path behavior. The
final `exit "$FAIL"` line is kept.

## Verification

The script's own 22-assertion suite passes with exit 0 after the change.

### DIFF
--- a/.claude/skills/dispatch-token-audit/scripts/test-generate-phase-model-policy.sh
+++ b/.claude/skills/dispatch-token-audit/scripts/test-generate-phase-model-policy.sh
@@ -16,30 +16,7 @@ CONFIG_LOAD="$SCRIPT_DIR/../../dispatch-propagate/scripts/dispatch-config-load"
 
 # --- test helpers -----------------------------------------------------------
 
-PASS=0
-FAIL=0
-TOTAL=0
-
-assert_eq() {
-  local label="$1" expected="$2" actual="$3"
-  TOTAL=$((TOTAL + 1))
-  if [[ "$expected" == "$actual" ]]; then
-    PASS=$((PASS + 1))
-    echo "  PASS: $label"
-  else
-    FAIL=$((FAIL + 1))
-    echo "  FAIL: $label"
-    echo "    expected: '$expected'"
-    echo "    actual:   '$actual'"
-  fi
-}
-
-report_results() {
-  echo ""
-  echo "================================"
-  echo "Results: $PASS/$TOTAL passed, $FAIL failed"
-  echo "================================"
-}
+source "$SCRIPT_DIR/../../dispatch-propagate/scripts/test-helpers.sh"
 
 # --- fixtures ---------------------------------------------------------------
 
--- a/.claude/skills/dispatch-token-audit/scripts/test-generate-phase-model-policy.sh
+++ b/.claude/skills/dispatch-token-audit/scripts/test-generate-phase-model-policy.sh
@@ -145,4 +145,3 @@ fi
 rm -rf "$RT_DIR"
 
 report_results
-exit "$FAIL"


natb1:
## QA Summary — Issue #2165 / PR #2178

**Pass: 4 / Fail: 0 / Skip: 0**

All acceptance criteria verified. No browser walkthrough needed (pure shell script change).

### Items executed (shell-command lane)

| # | Title | Result |
|---|-------|--------|
| 1 | No inline assert_eq/report_results/PASS/FAIL/TOTAL declarations | PASS |
| 2 | Sources test-helpers.sh via $SCRIPT_DIR-relative path | PASS |
| 3 | Running the script exits 0 and all tests pass | PASS |
| 4 | Sourced helpers produce equivalent output | PASS |

### Detail

**Item 1 — No inline helpers:** `grep` found no `assert_eq()`, `report_results()`, `PASS=`, `FAIL=`, or `TOTAL=` definitions in the file. The inline block is gone.

**Item 2 — Source line present:** Line 19 of the script reads `source "$SCRIPT_DIR/../../dispatch-propagate/scripts/test-helpers.sh"` and the target file exists at `.claude/skills/dispatch-propagate/scripts/test-helpers.sh`.

**Item 3 — Script exits 0:** `bash test-generate-phase-model-policy.sh` produced 22/22 passed, 0 failed, exit=0.

**Item 4 — Sourced helpers define symbols and produce correct output:** `test-helpers.sh` defines both `assert_eq()` (line 9) and `report_results()` (line 76). Running the script emits 22 `PASS:` lines and `Results: 22/22 passed, 0 failed` — consistent with prior behavior, no regressions.

### Bugs found

None.

### Deferred to office-hours

None.

### Disposition triage

Not applicable — residue list was empty (all items PASSed).

natb1:
## Review — PR #2178

_Surface: tests-only diff — no executable, config, dependency, or Firestore-rules changes._

### Fixed (1)

- **Dead `exit "$FAIL"` after `report_results`** (`.claude/skills/dispatch-token-audit/scripts/test-generate-phase-model-policy.sh:148`) — `report_results()` in the shared `test-helpers.sh` already calls `exit 1` on failure and returns normally on success; the trailing `exit "$FAIL"` was unreachable on the failure path and redundant on the success path. Removed to match sibling scripts (`test-dispatch-phase-model.sh`, `test-write-phase-log.sh`, `test-emit-outcome.sh`), all of which end with bare `report_results`. Fix commit: 26ed0bf8.

### Informational (1)

- **Failure output format changed silently** (`.claude/skills/dispatch-token-audit/scripts/test-generate-phase-model-policy.sh:19`) — The old inline `assert_eq` printed `expected: '$expected'` and `actual: '$actual'` (values single-quoted in output). The shared `test-helpers.sh` version does not quote them. No behavior change on passing runs; failure messages will look slightly different. Documented here for awareness; no action needed.

### Dismissed (1)

- **`source` of missing file** (`.claude/skills/dispatch-token-audit/scripts/test-generate-phase-model-policy.sh:19`) — Dismissed. `set -euo pipefail` (line 11) already causes a hard, loud exit when `source` returns non-zero for a missing file. An explicit existence guard is not needed and no sibling test script uses one.

### Security

PR #2179, opened 2026-06-20
Verify re-entry phase-log skip via a --reentry flag on dispatch-write-phase-log

natb1:
Closes #2163

## What

Make the crash-recovery **re-entry phase-log skip** a machine-verifiable script
behavior instead of an untested SKILL.md prose paragraph.

PR #2153 fixed a real bug: on the re-entry path (a PR already existed at skill
entry, so Steps 1–4/1–6 were skipped), `implement` and `review-fix` must **not**
re-write the phase-log entry — recomposing the handoff digest from the thin
re-entry context would clobber the prior pass's accurate entry. That fix made
each skill *skip the writer call* on re-entry. But the skip guarantee lived
entirely in prose; a future skill edit could drop the gate with no test catching
it. That gap is what #2163 was filed to close.

## How

Move the decision out of prose and into the script (per
`.claude/rules/design-proposals.md` and the push-logic-into-scripts lens):

- **Unit 1** — add a `--reentry <true|false>` flag to `dispatch-write-phase-log`.
  When `--reentry true`, the script short-circuits to a pure no-op (no STDIN read,
  no `gh` call, exit 0), preserving any existing phase-log comment verbatim. The
  flag is additive — default `false` preserves today's behavior exactly.
- **Unit 2** — both skills now **always** call the writer: `--reentry false` on
  the normal path (digest on STDIN, as before) and `--reentry true </dev/null` on
  the re-entry path. The gate key is unchanged ("did Steps 1–4/1–6 run this turn,"
  not PR-presence); the script now *enforces* the skip.
- **Unit 3** — `test-phase-log-reentry.sh`, a behavioral test with a stateful `gh`
  stub. It asserts AC1 directly: simulate a re-entry, assert the stored comment is
  byte-identical and no write verb fired. Also covers the normal write, the
  default-flag backward-compat path, and the invalid-value guard.

The test is auto-discovered by `run-unit-tests.sh --pr-scripts` (already in CI on
changes under `.claude/skills/dispatch-propagate/scripts/*`), so CI produces a
clear pass/fail — no CI wiring change needed.

## Deviation from the issue's written solution

The issue's text assumes "add a test/QA procedure" without touching the skills.
This PR additionally refactors the gate into the script, which is what makes the
guarantee genuinely machine-verifiable (rather than a `grep` of static prose, which
the issue explicitly calls inadequate). The `dispatch-write-phase-log` change is
additive and backward-compatible, so it fits in one PR. This was raised and
accepted at plan time as the faithful realization of the acceptance criteria.

## Verification

```
test-phase-log-reentry.sh   15/15 passed
test-write-phase-log.sh     11/11 passed  (no regression)
```

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-write-phase-log
@@ -2,7 +2,7 @@
 # dispatch-write-phase-log — upsert a per-(phase,attempt) handoff entry into the
 # issue's sentinel phase-log comment.
 #
-# Usage: dispatch-write-phase-log <issue-num> --phase <p> [--attempt <n>]
+# Usage: dispatch-write-phase-log <issue-num> --phase <p> [--attempt <n>] [--reentry <true|false>]
 #        (free-text entry body on STDIN)
 #
 # Each dispatch phase (implement / qa / review) writes a short
@@ -17,6 +17,11 @@
 # duplicate). This mirrors dispatch-write-plan's "re-running replaces, never
 # stacks a second comment" idempotency at the per-section granularity.
 #
+# Re-entry no-op: with --reentry true the script skips the write entirely and
+# preserves the existing phase-log comment verbatim — no STDIN read, no gh call,
+# exit 0. The decision of whether a phase's Steps 1–4 actually ran this turn
+# (i.e. whether to pass --reentry true) lives in the calling skill, not here.
+#
 # The repo is resolved explicitly (GH_REPO from the common dir's remote.origin.url)
 # so gh addresses the right repository from any caller cwd. A missing/non-numeric
 # issue-num, a malformed --phase, a non-numeric --attempt, or empty STDIN is a
@@ -27,7 +32,7 @@ source "$(cd "$(dirname "$0")" && pwd)/lib.sh"
 N="${1:-}"
 if [[ $# -lt 1 || -z "$N" ]]; then
   echo "dispatch-write-phase-log: an issue-num argument is required" >&2
-  echo "usage: dispatch-write-phase-log <issue-num> --phase <p> [--attempt <n>]   (entry body on STDIN)" >&2
+  echo "usage: dispatch-write-phase-log <issue-num> --phase <p> [--attempt <n>] [--reentry <true|false>]   (entry body on STDIN)" >&2
   exit 2
 fi
 if [[ ! "$N" =~ ^[0-9]+$ ]]; then
@@ -38,6 +43,7 @@ shift
 
 PHASE=""
 ATTEMPT="1"
+REENTRY="false"
 while [[ $# -gt 0 ]]; do
   case "$1" in
     --phase)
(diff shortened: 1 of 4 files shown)


natb1:
# QA Summary — Issue #2163

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Both test scripts exist and are executable | shell-command | PASS |
| 2 | Re-entry behavioral test passes (AC1 + AC2 + AC3 + AC4) | shell-command | PASS |
| 3 | Pre-existing writer test still passes (regression / backward compatibility) | shell-command | PASS |
| 4 | Script implements the --reentry true short-circuit before any gh/git call | shell-command | PASS |
| 5 | Both SKILL.md files call the writer on both the normal and re-entry paths | shell-command | PASS |
| 6 | SKILL.md prose keys the flag on "did Steps 1-4/1-6 run this turn," not on PR-presence | needs-human-judgment | → Opus: already satisfied |

**Counts:** PASS: 5, FAIL: 0, SKIP: 0

No bugs found.

## Verification detail

- `test-phase-log-reentry.sh`: **15/15 passed** — AC1 (re-entry verbatim no-op: exit 0, stored body byte-identical, no POST/PATCH), AC2 (normal write: POST fires, inner marker + content present), AC3 (default flag backward compat: same as AC2), AC4 (invalid value: exit 2, stderr mentions --reentry).
- `test-write-phase-log.sh`: **11/11 passed** — no regression in the existing upsert/append/replace behavior.
- `dispatch-write-phase-log`: `REENTRY` var present, `--reentry` arg parsed with a value guard, `^(true|false)$` validation with exit 2 on bad value, short-circuit before git/gh resolution at line 79–81.
- Both `implement/SKILL.md` and `review-fix/SKILL.md`: each carry `--reentry false` on the normal path and `--reentry true </dev/null` on the re-entry path, and each references `test-phase-log-reentry.sh` by path.

- **Item 6 — SKILL.md prose keys the flag on "did Steps 1-4/1-6 run this turn," not on PR-presence**: `implement/SKILL.md` (lines 256–262) instructs passing `--reentry true` only on the crash-recovery re-entry path (Steps 1–4 skipped this turn) and carries an explicit anti-PR-presence warning ("Gate on whether Steps 1–4 ran this turn, not on PR presence … a `[ -n "$PR_NUM" ]` gate would skip the write ALWAYS and break the normal path"). `review-fix/SKILL.md` (lines 647–652) mirrors this for Steps 1–6: "Gate on whether Steps 1–6 ran this session … NOT on label or PR presence." Both read coherently — no code defect, no human input needed. Dropped as PASS.

natb1:
## Review — PR #2179 (issue #2163)

**Surface:** `code` (shell scripts + SKILL.md docs) | **Security surface:** no app source, no Firebase rules, no npm deps changed | **CodeQL:** 0 open alerts | **Erosion:** no findings

---

### Fixed
_None._

### Required (security)
_None._

### Informational

| ID | Location | Note |
|----|----------|------|
| F1 | `test-phase-log-reentry.sh:1–30` | `test-helpers.sh` functions (`assert_eq`, `assert_contains`, `assert_file_contains`, `report_results`) are referenced but their existence is not verified in this review pass. The test suite passed QA (15/15), so they are present — no action needed, noted for awareness. |
| F2 | `dispatch-write-phase-log:75–83` | When `--reentry true` exits 0 without reading STDIN, callers using process substitution (not a pipe) hold an open descriptor. Both SKILL.md files already show `</dev/null` on this path; the script usage comment could reinforce it but this is not a bug. |
| F3 | `test-phase-log-reentry.sh:90–93` | The gh stub's `BIN` path is baked in at write time (heredoc interpolation), so any `PATH` change after `write_gh_stub` would not be reflected. This is intentional for hermeticity and is not a defect. |
| F4 | `test-phase-log-reentry.sh:230–245` | AC1 uses `cmp -s` for byte-for-byte comparison — correct and sufficient for the re-entry no-op guarantee. |
| F5 | `dispatch-write-phase-log:43–68` | When `--phase` is omitted alongside `--reentry true`, the `--phase` validation fires before the `--reentry` short-circuit (correct ordering — phase is always required). Not explicitly tested, but expected behavior. |

### Dismissed

| ID | Finding | Rationale |
|----|---------|-----------|
| F6 | Missing `--reentry` flag in `review-fix/SKILL.md` re-entry path | Misread: SKILL.md correctly shows both `--reentry false` and `--reentry true` call forms. |
| F7 | Shell injection via `--reentry` value | Not possible: value is validated by `[[ \! REENTRY =~ ^(true|false)$ ]]` and only used in a string comparison, never interpolated into a command. |

### False-positive (security)
_None._

_None._

### Deferred
_None._

### Out-of-scope (security)
_None._

---

No code changes applied. `dispatch:reviewed` applied.

PR #2018, opened 2026-06-19
budget: migrate the accounts section to React

natb1:
Closes #1875

Migrates the budget **accounts** section — the accounts list page (income
statement, cash-flow summary, divergence warning, three charts) and the
reconcile flow — to React composing `@commons-systems/ds`, following the
`Transactions.tsx` template established in #1874. Phase 3 of the budget → React
sub-epic (#1873).

This is a **port, not a rewrite**: all compute, reconciliation, entity,
chart-rendering, and data-source logic is reused unchanged. Only the rendering
layer (vanilla string-render + imperative hydrate) is replaced by React
components, the design-system CSS literals are tokenized, and the routes are
flipped from legacy to React.

## Units

1. **Tokenize the account chart series colors** — the net-worth and cash-flow
   chart renderers now read their series colors from the DS `--chart-1..6`
   palette at runtime via `readThemeVar`, dropping the hardcoded
   teal/red/blue/purple hex literals (including the banned cool-blue 12-week
   average series). Legend swatches read the same tokens. Chart geometry literals
   are unchanged.

2. **Accounts list page → React (`Accounts.tsx`)** — extracts pure view-model
   helpers into `account-view-model.ts`, adds a `useAccountsData` load hook,
   composes DS `Metric`/`Card`/`Badge` for the headline figures and status
   indicators, and renders the aggregate-trend / net-worth / cash-flow charts as
   imperative islands with ported scroll-sync / date-picker / resize wiring. The
   income-statement / cash-flow tables stay as budget `<table>` markup (DS has no
   Table component). Remaining accounts-list `theme.css` literals are tokenized.

3. **Reconcile flow → React (`AccountsReconcile.tsx`)** — faithfully
   re-implements the behavior-critical reconcile flow as React state and
   handlers over the unchanged `reconciliation.ts` / `reconcile-hints.ts` logic:
   URL query-param sync / deep-linking, suggested-cleared hints, cleared-checkbox
   optimistic-persist-and-revert, confirm-all, the balanced-vs-mismatch dialog,
   adjustment-entry creation with the orphan-preventing re-entrancy guard, and
   finalize. Composes DS `Select`/`Button`/`Input`/`Card`/`Metric`.

4. **Route the React pages, remove legacy, reconcile tests** — mounts both pages
   at `/accounts` and `/accounts/reconcile` via `isAccounts`/`isAccountsReconcile`
   branches mirroring `isTransactions` (reconcile checked first so the sub-path
[...]

PR #2181, opened 2026-06-20
budget: AccountsReconcile URL sync via @commons-systems/router navigate

natb1:
Closes #2031

## What

Route the `/accounts/reconcile` page's URL query sync through the shared
`navigate(href, { replace: true })` primitive from `@commons-systems/router/react`
instead of calling `history.replaceState` inline.

`budget/src/pages/AccountsReconcile.tsx`'s file-local `replaceQueryParam` helper
poked `history.replaceState` directly to mirror institution/account/period
selector state into the query string. Every other React page in the budget app
routes navigation through a shared abstraction; this aligns the reconcile page
with that convention. Surfaced during review of PR #2018 (port of #1875).

## Change

- Import `navigate` from `@commons-systems/router/react` (already a budget
  dependency via `@commons-systems/router/hydrate`).
- In `replaceQueryParam`, the only changed line is the write: the helper still
  builds the next URL from `new URL(location.href)` and deletes-on-empty /
  sets-otherwise; the final `history.replaceState(null, "", url)` becomes
  `navigate(url.pathname + url.search + url.hash, { replace: true })`. Unrelated
  query params and the clear-on-empty semantics are preserved.
- Reword the two helper comments that named `replaceState`; keep the
  "NOT pushState / NOT a popstate dispatch" intent notes (still true — `navigate`
  with `replace` uses `replaceState` and fires no `popstate`).

## Why it preserves behavior

`navigate(href, { replace: true })` performs the same `replaceState` write. Its
extra step is a trailing `emit()` that notifies `useLocation()` subscribers — and
no budget-reachable component subscribes (the only `useLocation` consumers are the
router package's own tests), so `emit()` is an observable no-op for the budget
tree. Budget's own `useRouter` listens to `popstate`, which `navigate` does not
fire, so a query-only write stays inert to the page router exactly as before.

The query read path is untouched — initial state is still seeded from
`parseReconcileQuery(location.search)`.

## Verification

- `run-unit-tests.sh --app budget` — build (`tsc && vite build`, typechecks the
  new import) + 957 unit tests pass, including
  `AccountsReconcile.test.tsx` and the accounts-reconcile smoke test, which assert
  the observable `location.search` written on select changes.
- `run-lint.sh --app budget` — clean.

### DIFF
--- a/budget/src/pages/AccountsReconcile.tsx
+++ b/budget/src/pages/AccountsReconcile.tsx
@@ -9,7 +9,8 @@
 // The imperative DOM mutations of the legacy hydrate become React state:
 //   - the URL query (institution/account/period) is React state, initialized
 //     from parseReconcileQuery(location.search) — the deep-link read — and written
-//     back with history.replaceState on a Select change. We do NOT dispatch
+//     back via the shared navigate(…, { replace: true }) abstraction on a Select
+//     change. We do NOT dispatch
 //     popstate: useRouter strips the query, so a query-only popstate is inert, and
 //     App re-keys only on a data transition. The component owns its query.
 //   - per-leg checked state is React state (the optimistic toggle); on a persist
--- a/budget/src/pages/AccountsReconcile.tsx
+++ b/budget/src/pages/AccountsReconcile.tsx
@@ -24,6 +24,7 @@
 // behaviorally identical and keeps one code path. The two legacy triggerReload()
 // sites map to: a Select change (query state change) and post-finalize (nonce bump).
 import { useEffect, useRef, useState, type ReactNode } from "react";
+import { navigate } from "@commons-systems/router/react";
 import { Button, Card, Input, Metric, Select } from "@commons-systems/ds";
 import { classifyError } from "@commons-systems/errorutil/classify";
 import { deferProgrammerError } from "@commons-systems/errorutil/defer";
@@ -62,13 +63,13 @@ interface ReconcileQuery {
 }
 
 // Faithful port of accounts-reconcile-hydrate.ts replaceQueryParam: write/clear a
-// single query param via history.replaceState (NOT pushState — matches legacy and
-// avoids back-button query cycling).
+// single query param via the shared `navigate(…, { replace: true })` abstraction
+// (NOT pushState — matches legacy and avoids back-button query cycling).
 function replaceQueryParam(name: string, value: string | null): void {
   const url = new URL(location.href);
   if (value === null || value === "") url.searchParams.delete(name);
   else url.searchParams.set(name, value);
-  history.replaceState(null, "", url);
+  navigate(url.pathname + url.search + url.hash, { replace: true });
 }
 
 /**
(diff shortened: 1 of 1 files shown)


natb1:
# QA Summary — #2031 (PR #2181)

**Result: PASS — all items satisfied**

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | `AccountsReconcile.tsx` no longer calls `history.replaceState` inline | shell-command | PASS |
| 2 | Selecting an account writes institution+account to the URL | browser walkthrough | PASS |
| 3 | Selecting a period writes period to the URL and resolves the reconcile view | browser walkthrough | PASS |
| 4 | Changing the account clears the previously selected period from the URL | browser walkthrough | PASS |
| 5 | Deep-link reload resolves the same reconcile view | browser walkthrough | PASS |
| 6 | Back button does not cycle through query-only states | browser walkthrough | PASS |
| 7 | No behavior change — reconcile interactions still function after the swap | needs-human-judgment → assessed by Opus | PASS (already satisfied) |

**Counts:** 6 PASS (lanes), 1 PASS (already-satisfied by Opus), 0 FAIL, 0 SKIP

## Walkthrough GIF

`tmp/qa-fix-walkthrough-2031.gif` (4 frames, 171 KB)

## Lane findings

- No console errors on the reconcile page. Two pre-existing React DOM-nesting warnings appeared on the `/accounts` page (in `IncomeStatementTable` — unrelated to this change).
- All network requests returned 200.

**Item 7 — No behavior change — reconcile interactions still function after the swap**

The disposition Workflow classified this item **already-satisfied**: behavioral equivalence after the `replaceState`→`navigate` swap is provable from code. `navigate({replace:true})` performs the same `history.replaceState` write (the `{}` vs `null` state arg is immaterial) plus `emit()`. `emit()` notifies only `location-store` subscribers; `AccountsReconcile`'s router (`budget/src/use-router.ts`) never subscribes to that store and keys path state on `location.pathname` only (query stripped). A query-param-only write leaves pathname constant — no re-render, no popstate fired. The checkbox-toggle, live-balance-update, and reconcile-dialog paths depend solely on local React state and are untouched by the diff. No defect to fix, no human input needed.

natb1:
## Review — PR #2181 (issue #2031)

**Surface:** `code` (1 changed file, app source) | **CodeQL:** no open alerts | **Dependency audit:** skipped (no `package.json`/`package-lock.json` changes) | **Erosion:** clean

---

### Fixed

- **`AccountsReconcile.tsx:12`** — Module-level design comment still said "back with `history.replaceState` on a Select change" after the PR replaced that call. Updated to "back via the shared `navigate(…, { replace: true })` abstraction on a Select change", consistent with the two comment sites already updated by this PR. (sources: code-review, review) — fix commit `72b717ba`

### Informational

- **`AccountsReconcile.tsx:72`** — `navigate()` in `location-store.ts` passes `{}` as the history state (`history.replaceState({}, '', href)`), whereas the previous direct call used `null`. In practice this is a no-op difference (no code reads `history.state` for this route), but it is a subtle behavioral delta worth noting for the record. No change required.

### False-positive (security)

- **`AccountsReconcile.tsx:72`** — Red-team finding: `replaceQueryParam` now calls `navigate(url.pathname + url.search + url.hash, { replace: true })` instead of `history.replaceState`. The concern was a potential open-redirect or unexpected navigation. Dismissed: `navigate` here receives a same-origin relative URL built from `location.href` with only `searchParams` mutations; no user-controlled path segment is ever injected. The PR body's "Why it preserves behavior" section explains the equivalence.

---

_Deferred: none. Out-of-scope security: none._

PR #2185, opened 2026-06-20
lib.sh dispatch_marker_comment_id: propagate gh_retry pipeline failure instead of masking it as no-comment

natb1:
Closes #2138

## Problem

Inside `dispatch_marker_comment_id` (`lib.sh`), the comment id was captured from
a single `gh_retry … | jq` pipeline. The pipeline's exit status reflects only
the last command (`jq`), which — on a `gh_retry` failure — reads the now-empty
input, emits nothing, and exits `0`. So a hard `gh` API failure was silently
swallowed: `cid` was left empty and the `[[ -z "$cid" || "$cid" == "null" ]]`
guard returned `0`, handing the caller an empty string **indistinguishable from
"no matching comment exists."**

The first affected caller, `build_phase_log_section` in `dispatch-context-pack`,
would then emit the `phase-log: none` sentinel on a real API failure — a hard
error masked as a clean degrade, contradicting `code-style.md` ("prefer clear
errors over defensive fallbacks") and inconsistent with the body-fetch on the
very next line (which already does `gh_retry … || return 1`).

## Fix

Split the single pipeline into the two-step, error-propagating form the sibling
`gh_api_array` already uses: capture `gh_retry`'s output with `|| return 1`,
then run `jq` on that captured output with `|| return 1`. A `gh_retry`/`jq`
failure now surfaces as a **non-zero return**, while the genuine absent case
still returns `0` with empty output.

No caller change is needed — callers are already wired for this
(`cid=$(dispatch_marker_comment_id …) || return 1`), so this converts a masked
failure into the clear, pack-aborting error the caller is already prepared to
propagate.

## Changes

- `lib.sh`: two-step error-propagating capture in `dispatch_marker_comment_id`;
  header comment records the new contract.
- `test-dispatch-scripts.sh`: hermetic regression test covering both criteria —
  a `gh_retry` failure returns non-zero (the bug), and the genuine absent case
  (valid JSON, no match) still returns `0` with empty output.

## Verification

- Hermetic contract check (the plan's ```verify block) passes.
- Full suite `test-dispatch-scripts.sh`: 3550/3550 passed, 0 failed.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -171,7 +171,9 @@ gh_api_array() {
 # plan scripts), fetches comments via gh_retry, and applies the same
 # author-id-filtered first(...) selector. Returns non-zero with a stderr message
 # on an unresolvable author id (clear error, not a fallback). Echoes nothing
-# (empty) when no matching comment exists.
+# (empty) when no matching comment exists. A gh_retry or jq pipeline failure
+# returns non-zero (a clear error), distinct from the empty-output absent case —
+# mirroring gh_api_array's error propagation rather than silently swallowing it.
 dispatch_marker_comment_id() {
   local n="$1" marker="$2"
   local author_id="${DISPATCH_PLAN_AUTHOR_ID:-$(gh api user --jq '.id')}"
@@ -179,10 +181,13 @@ dispatch_marker_comment_id() {
     echo "dispatch_marker_comment_id: could not resolve a numeric comment author id (got: '$author_id')" >&2
     return 1
   fi
+  local raw
+  raw=$(gh_retry gh api --paginate "repos/{owner}/{repo}/issues/$n/comments") \
+    || return 1
   local cid
-  cid=$(gh_retry gh api --paginate "repos/{owner}/{repo}/issues/$n/comments" \
-    | jq -r --arg m "$marker" --argjson author_id "$author_id" \
-        'first(.[] | select((.body | startswith($m)) and (.user.id == $author_id)) | .id)')
+  cid=$(printf '%s\n' "$raw" | jq -r --arg m "$marker" --argjson author_id "$author_id" \
+      'first(.[] | select((.body | startswith($m)) and (.user.id == $author_id)) | .id)') \
+    || return 1
   if [[ -z "$cid" || "$cid" == "null" ]]; then
     return 0
   fi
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — Issue #2138 / PR #2185

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | `dispatch_marker_comment_id` failure propagation + regression test (AC1, AC3) | shell-command | PASS |
| 2 | lib.sh header comment + two-step implementation present (option 1) | shell-command | PASS |
| 3 | dispatch-context-pack body-fetch (line 195) consistent with option 1 (AC4) | shell-command | PASS |

## Test suite results

`bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — **3554/3554 passed, 0 failed**

Including the new `#2138` regression block:
- PASS: `dispatch_marker_comment_id` returns non-zero when `gh_retry` fails
- PASS: `dispatch_marker_comment_id` returns 0 with empty output when no comment matches

## Acceptance criteria status

- **AC1** ✅ — `gh_retry`/`jq` failure surfaces as non-zero return (option 1 implemented via two-step `|| return 1` capture)
- **AC2** — N/A (option 2 not chosen)
- **AC3** ✅ — Genuine absent case still returns 0 with empty output
- **AC4** ✅ — Body-fetch at `dispatch-context-pack:195` uses `|| return 1`, consistent with option 1

## Disposition

Clean autonomous pass — no residue, no escalation.

natb1:
## Review — PR #2185 (`2138-lib-sh-dispatch-marker-comm`)

### Fixed

- **Missing success-path test (Criterion 2a)** — `test-dispatch-scripts.sh:36095`
  The test block covered Criterion 1 (gh_retry failure → non-zero) and Criterion 3 (no-match → rc=0 empty) but never exercised the happy path where a matching comment is found. Added Criterion 2a: stubs `gh_retry` to emit a single-element JSON array with a matching body and author id, asserts rc=0 and the printed id equals the expected value. _(fix commit: `253eb30a`)_

- **Missing jq-failure test (Criterion 2b)** — `test-dispatch-scripts.sh:36074`
  The `cid=$(… jq …) || return 1` guard on the `cid=` assignment had no test. Added Criterion 2b: stubs `gh_retry` to print invalid JSON (`'not json'`), asserts non-zero exit. Full suite passes 3556/3556. _(fix commit: `253eb30a`)_

### Dismissed

- **`printf '%s\n' "$raw" | jq …` vs `jq <<<"$raw"`** — `lib.sh:188`
  Both forms are safe against the zsh-echo control-char trap. The pipe form is a cosmetic non-issue; no change warranted.

### Informational

- **No diagnostic on jq failure** — `lib.sh:188`
  When `jq` exits non-zero (e.g. malformed JSON from `gh_retry`), the function returns 1 via `|| return 1` but emits no stderr message. Callers propagate the error cleanly; a diagnostic would be helpful for debugging but is not required for correctness.

### Security

Security review: `surface=code`, no CodeQL alerts (0 open alerts on PR ref), no npm dependency changes, no erosion findings.

### Follow-ups

_None._

PR #2198, opened 2026-06-20
print: cap EPUB/PDF full-document search results and surface truncation

natb1:
Closes #1810

## Problem

The print viewer's EPUB full-document search (shipped in #1688) accumulated
every match across the whole spine with no upper bound: `search()` pushed every
`buildResult()` into `results` and then highlighted all of them in a single
synchronous burst. Searching a common token (e.g. `"the"`) in a large EPUB built
an unbounded `SearchResult[]` and painted thousands of epub.js SVG annotations at
once, which could jank or freeze the viewer.

The PDF renderer already capped at 200, but **silently** — the shared
`SearchPanel` had no truncation affordance, so a capped PDF search read as
"exactly 200 results" with no hint that more matched.

## Change

- Introduce a shared `MAX_SEARCH_RESULTS = 200` constant and a `SearchResponse`
  envelope (`{ results, truncated }`) in `types.ts`. The truncation fact now
  travels *with* the results it describes rather than living in renderer state,
  so the panel never has to read a stateful getter after the await and overlapping
  searches don't race.
- Cap EPUB search with the same check-before-push pattern PDF uses, and stop
  scanning further sections once the cap trips. The existing
  `finally { section.unload() }` still unloads the section where the cap trips
  mid-match-loop; the highlight burst is bounded automatically because it iterates
  only the (now-capped) `results`.
- Rework PDF's push-then-break into the precise check-before-push shape so
  `truncated` is only `true` when a genuine (MAX+1)th match is encountered — a
  document with *exactly* MAX matches reports `truncated: false`.
- `SearchPanel` surfaces truncation in the existing `.viewer-search-count` span:
  `First N results shown — refine your search` when truncated, otherwise the
  plain `"N results"` text. Because the panel is shared, this also fixes PDF's
  previously-silent cap — a deliberate, in-scope consequence.

All in-repo `search()` callers and existing tests are updated to the new envelope
shape in the same change so the tree stays compiling and green.

## Tests

New behavior tests cover the issue's acceptance criteria:

- **EPUB cap (AC2):** a section returning `MAX_SEARCH_RESULTS + 1` matches yields
  `results.length === MAX_SEARCH_RESULTS`, `truncated === true`, the section is
  still unloaded once, and the highlight burst paints exactly
  `MAX_SEARCH_RESULTS` annotations.
- **PDF cap:** matches exceeding the limit yield the capped count and
  `truncated === true`.
- **Panel affordance (AC3):** `truncated: true` renders the truncation message;
  `truncated: false` renders the plain count.
- Not-truncated companion cases for both renderers.

## Verification

- `run-lint.sh --app print` — PASS
- `run-unit-tests.sh --app print` (tsc build + vitest) — 561/561 PASS

Manual / QA (not auto-run): load a large EPUB, search a common token, confirm the
viewer does not jank, the highlight count is bounded, and the panel shows the
truncation affordance; repeat for a large PDF to confirm the panel now surfaces
the 200-cap.

### DIFF
--- a/print/src/viewer/SearchPanel.tsx
+++ b/print/src/viewer/SearchPanel.tsx
@@ -45,11 +45,15 @@ export function SearchPanel({ controller }: { controller: UseViewerControllerRes
     try {
       const renderer = getRenderer();
       if (!renderer || !isSearchable(renderer)) return;
-      const r = await renderer.search(trimmed);
+      const { results, truncated } = await renderer.search(trimmed);
       if (destroyed.current || trimmed !== currentQuery.current) return;
-      setResults(r);
+      setResults(results);
       setActiveIndex(-1);
-      setCountText(r.length === 1 ? "1 result" : `${r.length} results`);
+      if (truncated) {
+        setCountText(`First ${results.length} results shown — refine your search`);
+      } else {
+        setCountText(results.length === 1 ? "1 result" : `${results.length} results`);
+      }
     } catch (err) {
       setCountText("Search failed");
       setResults([]);
--- a/print/src/viewer/epub.ts
+++ b/print/src/viewer/epub.ts
@@ -1,5 +1,6 @@
 import ePub, { type Book, type Rendition, type Location, type NavItem } from "epubjs";
-import type { OutlineEntry, SearchableRenderer, SearchResult } from "./types.js";
+import type { OutlineEntry, SearchableRenderer, SearchResult, SearchResponse } from "./types.js";
+import { MAX_SEARCH_RESULTS } from "./types.js";
 
 // epubjs ships incomplete type declarations. These narrow shapes describe the
 // runtime members we use that are missing from the .d.ts, following the
@@ -285,11 +286,11 @@ export function createEpubRenderer(
       await relocated;
     },
 
-    async search(query: string): Promise<SearchResult[]> {
+    async search(query: string): Promise<SearchResponse> {
       const trimmed = query.trim();
       if (!book || !trimmed) {
         clearSearchHighlights();
-        return [];
+        return { results: [], truncated: false };
       }
 
       const epoch = ++_searchEpoch;
@@ -299,6 +300,7 @@ export function createEpubRenderer(
       const loader = (book.load as (this: Book, req: unknown) => Promise<unknown>).bind(book);
 
       const results: SearchResult[] = [];
+      let truncated = false;
       for (let i = 0; i < spineItems.length; i++) {
         const section = spineItems[i]!;
         try {
@@ -318,6 +320,7 @@ export function createEpubRenderer(
           try {
             const label = labelForSection(section, i);
             for (const match of section.find(trimmed)) {
+              if (results.length >= MAX_SEARCH_RESULTS) { truncated = true; break; }
               const result = buildResult(match, trimmed, label);
               if (result) results.push(result);
             }
@@ -334,11 +337,13 @@ export function createEpubRenderer(
         } finally {
           section.unload();
         }
+        // Cap tripped mid-section: stop scanning further sections.
+        if (truncated) break;
       }
 
       // A newer search started while we awaited; touch no highlight state and
       // let the UI's stale-result guard discard these.
-      if (epoch !== _searchEpoch) return results;
+      if (epoch !== _searchEpoch) return { results, truncated };
 
       // Single synchronous highlight burst after iteration.
       clearSearchHighlights();
@@ -347,7 +352,7 @@ export function createEpubRenderer(
         annotations.highlight(result.location, {}, undefined, "viewer-search-hl", BASE_STYLES);
         _searchCfis.push(result.location);
       }
-      return results;
+      return { results, truncated };
     },
 
     async goToResult(result: SearchResult): Promise<void> {
--- a/print/src/viewer/pdf.ts
+++ b/print/src/viewer/pdf.ts
@@ -2,8 +2,8 @@ import * as pdfjsLib from "pdfjs-dist";
 import { TextLayer } from "pdfjs-dist";
 import type { PDFDocumentProxy, PDFPageProxy, RenderTask } from "pdfjs-dist";
 import type { PageViewport } from "pdfjs-dist/types/src/display/display_utils.js";
-import type { OutlineEntry, SearchableRenderer, SearchResult } from "./types.js";
-import { parsePositionPage } from "./types.js";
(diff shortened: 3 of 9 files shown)


natb1:
# Fix-checks summary — PR #2198

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Failure excerpt:** Net-new non-null assertions (`!`) and `as <Type>` casts flagged in test files: epub.test.ts (9 lines), pdf.test.ts (29 lines), search.test.ts (5 lines), spread-search-highlight.test.ts (2 lines)
- **Why not caught:** New tests added for the SearchResponse envelope refactor use non-null assertions (`renderer.search!`, `page1Result!`, etc.) and DOM `querySelector as HTMLElement` casts that are standard in test code but net-new in the diff — none had `// type-safety-ok:` suppression comments
- **Fix:** Added `// type-safety-ok: <reason>` suppression comments to all 45 flagged lines across the four test files — commit `a2b8aa7f`

natb1:
# QA Summary — Issue #1810 (PR #2198)

## Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | EPUB full-document search — truncation cap on a common query | browser walkthrough | **PASS** |
| 2 | EPUB search — non-truncated query shows a plain count | browser walkthrough | **PASS** |
| 3 | PDF full-document search — truncation cap on a common query | browser walkthrough | **PASS** |

**3 PASS / 0 FAIL / 0 SKIP**

## Detail

**Item 1 — EPUB cap golden path (AC1 + AC3)**
Navigated to `/view/gutenberg-3296` (The Confessions of St. Augustine, public EPUB seed). Searched `"the"` via the search panel. The `.viewer-search-count` span read exactly:
> `First 200 results shown — refine your search`
No console errors. No jank or freeze — the viewer stayed responsive and the book content remained scrollable while highlights were painted. The highlight burst is bounded at 200.

**Item 2 — EPUB non-truncated query (AC3 negative case)**
Typed `"Carthage"` in the same EPUB. The count span read:
> `18 results`
No truncation phrasing — confirming the affordance appears only when results are actually capped, not on every search.

**Item 3 — PDF cap (PDF parity)**
Navigated to `/view/plato-republic` (Republic by Plato, public PDF seed). Searching `"the"` returned 150 results (below the 200 cap, correctly showing plain `"150 results"`). Searching `"a"` triggered the cap; the count span read exactly:
> `First 200 results shown — refine your search`
No console errors. No 4xx/5xx network failures. PDF viewer remained responsive.

## GIF

`tmp/qa-fix-walkthrough-1810.gif`

## Acceptance criteria

- [x] AC1: `search()` bounds results — EPUB and PDF both cap at 200 (confirmed in browser with common queries)
- [x] AC2: Unit test coverage — verified by CI (561/561 vitest PASS, including explicit EPUB cap test with `MAX_SEARCH_RESULTS + 1` matches asserting `.length === MAX_SEARCH_RESULTS`, `truncated === true`, and `unload` called once)
- [x] AC3: Panel communicates truncation — `"First 200 results shown — refine your search"` shown in `.viewer-search-count` when truncated; plain `"N results"` when not

## Pre-QA acceptance tests

164 passed / 10 skipped (orientation tests, device-specific) — all pass.

natb1:
## Review: PR #2198 (issue #1810 — EPUB full-document search result cap)

### Fixed

- **print/test/viewer/epub.test.ts:902** (`code-review`) — The AC2 test only covered the single-section cap path. Added a multi-section variant: section 0 has `MAX+1` matches, section 1 is present; after `search()` the test asserts `section1.load()` was never called, directly exercising the outer-loop `if (truncated) break` guard at `epub.ts:343`. (commit `97dab053`)

### Informational

- **print/src/viewer/epub.ts:325** (`review`) — The cap check fires before `buildResult()`, which can return `null` for a match (when `matchLength <= 0`). This means `truncated=true` can be set for a match that would have been silently skipped. In practice the result count may be capped by up to ~1 entry fewer than `MAX_SEARCH_RESULTS`, not a user-visible bug but worth knowing.
- **print/test/viewer/epub.test.ts:902–938** (`review`) — The AC2 mock yields a valid (non-null) `buildResult` for every match, so the null-return path in the cap check is not exercised.
- **print/test/viewer/search.test.ts:765–770** (`review`) — The AC3 truncation-message test uses a loose assertion (contains `'3'` and matches `/refine|first/` case-insensitively) rather than asserting the full message text `'First 3 results shown — refine your search'`.
- **print/src/viewer/SearchPanel.tsx:31** (`erosion`) — Cyclomatic complexity increased: worst-function 12→13, total file 43→44. Inherent to the added `if (truncated)` branch; the increase is minimal and expected.
- **print/src/viewer/epub.ts:289** (`erosion`) — Cyclomatic complexity increased: worst-function 14→16, total file 102→104. The two added branches (`if (results.length >= MAX_SEARCH_RESULTS)` and `if (truncated) break`) are the direct cost of the cap implementation.
- **print/test/viewer/epub.test.ts:10** (`erosion`) — Total file complexity rose 160→164 (new test cases).
- **print/test/viewer/pdf.test.ts:54** (`erosion`) — Total file complexity rose 162→173 (new test cases).
- **print/test/viewer/search.test.ts:15** (`erosion`) — Total file complexity rose 70→77 (new test cases).
- **print/test/viewer/search.test.ts:497** (`erosion`) — Code-duplication increased: clone blocks 10→11, duplicated lines 77→89. The new cap/not-truncated companion tests for EPUB and PDF follow a similar shape; extracting a shared helper is a follow-up improvement.

### Dismissed

- **print/src/viewer/SearchPanel.tsx:53** (`review`) — Em-dash (U+2014) in the JSX string `'First N results shown — refine your search'` flagged as a potential encoding concern. The character is hard-coded in JSX text content and rendered by React's auto-escaping, so no encoding issue exists. Style preference only; no change warranted.

### False-positive (security)

- **print/src/viewer/epub.ts:325–327** (`red-team`) — Claim: setting `truncated=true` before `buildResult()` is called introduces an off-by-one that could be exploited. Not a real vulnerability — `buildResult` is a pure local function with no side-effects; setting the flag before the call is a guard pattern (break-before-push), and a null return from `buildResult` has no security implication.
- **print/src/viewer/SearchPanel.tsx:53** (`red-team`) — Claim: the truncation message `'First ${results.length} results shown — refine your search'` is an XSS vector. Not applicable — React auto-escapes all text content inserted into JSX expressions; user-controlled input does not flow into `results.length`.

### Security

No CodeQL alerts. No dependency changes. No out-of-scope security findings.

PR #2201, opened 2026-06-20
print: early-exit EPUB search() spine loop when the search epoch advances

natb1:
Closes #1819

## What

The EPUB `search()` spine loop kept loading every remaining section after
`clearSearch()` (or a newer `search()`) advanced `_searchEpoch` mid-loop — the
existing epoch guard fired only **once, after the full loop completed**
(`print/src/viewer/epub.ts`). So a cleared or superseded search wasted
potentially many async `section.load()` calls before discovering the abort.

This adds a per-iteration epoch check as the first statement of the spine loop
body, so a cleared or superseded search stops loading not-yet-visited sections
immediately. The already-in-flight section (parked in `await section.load()`
when the clear happened) still completes its own `finally { section.unload() }`,
so no load is leaked. The post-loop guard is retained as a safety net (so
highlights are still never painted for a stale search).

## Changes

- **`print/src/viewer/epub.ts`** — one new line at the top of the spine loop:
  `if (epoch !== _searchEpoch) break;`.
- **`print/test/viewer/epub.test.ts`** — a new regression test
  (`"stops loading not-yet-visited spine sections once clearSearch advances the
  epoch mid-loop"`) proving not-yet-visited sections are never loaded while the
  in-flight section still loads/unloads, and the superseded search paints no
  highlight. Without the guard, `s1.load`/`s2.load` are called and the test
  fails — so it genuinely guards the fix.

## Deviation from the plan

The plan said "do **not** modify any existing test" and claimed the change
"matches the existing overlapping-search semantics." That was not accurate: the
early-break makes a superseded `search()` skip `load()` entirely, which broke the
pre-existing `"only the latest of two overlapping searches performs its
highlight burst"` test (`epub.test.ts:786`). That test gates the *first*
`load()` call and assumed the superseded call-1 consumes it; with the early-break
the gate migrated to call-2 and `await p2` hung (timeout).

The fix is a single line — `await Promise.resolve()` inserted between the two
`search()` calls — so call-1 enters its loop and parks on the gated `load()`
while `_searchEpoch` is still 1 (before call-2 bumps it), restoring the exact
parked-mid-load scenario the test intends. **No assertion in that test was
changed.** This necessary one-line edit is the scope deviation flagged for
office-hours accept/reject.

## Verification

- `npx vitest run --root print test/viewer/epub.test.ts` — **52 passed / 0
  failed** (includes the new test plus the existing overlapping-search / abort /
  "returns matches from every spine section" tests).
- `npx vitest run --root print` — **480 tests passed, 0 test failures.** The run
  reports 4 *suite-load* failures (`(0 test)` each:
  `test/viewer/spread-search-highlight.test.ts`, `test/viewer/pdf.test.ts`,
  `test/enrichment.test.ts`, `test/local-folder-ui.test.ts`). These are a
  local-environment limitation only — vite's fs-deny of
  `node_modules/pdfjs-dist/build/pdf.worker.min.mjs?url` (and happy-dom's
  `blob:` fetch), which prevents those suites from importing at all. They fail
  **identically on the clean baseline** (`git stash`) without this change, and
  none involve `epub.test.ts`. origin/main's CI is green, so these are
  local-only and CI will be green for this PR.
- `npm run build --prefix print` — succeeds (typechecks the source change).

### DIFF
--- a/print/src/viewer/epub.ts
+++ b/print/src/viewer/epub.ts
@@ -300,7 +300,7 @@ export function createEpubRenderer(
 
       const results: SearchResult[] = [];
       for (let i = 0; i < spineItems.length; i++) {
-        if (epoch !== _searchEpoch) break; // a newer search/clearSearch superseded us — stop loading sections
+        if (epoch !== _searchEpoch) break;
         const section = spineItems[i]!;
         try {
           try {
--- a/print/src/viewer/epub.ts
+++ b/print/src/viewer/epub.ts
@@ -300,6 +300,7 @@ export function createEpubRenderer(
 
       const results: SearchResult[] = [];
       for (let i = 0; i < spineItems.length; i++) {
+        if (epoch !== _searchEpoch) break; // a newer search/clearSearch superseded us — stop loading sections
         const section = spineItems[i]!;
         try {
           try {
--- a/print/test/viewer/epub.test.ts
+++ b/print/test/viewer/epub.test.ts
@@ -806,6 +806,11 @@ describe("createEpubRenderer", () => {
       const renderer = await initRenderer();
 
       const p1 = renderer.search!("first");
+      // Let call-1 pass `await book.loaded.navigation` and park on the gated
+      // load() *inside* its loop, consuming the gate while _searchEpoch is still
+      // 1 — before call-2 bumps it. So call-1 is superseded mid-load and the
+      // post-loop epoch guard (not the new top-of-loop break) discards its burst.
+      await Promise.resolve();
       const p2 = renderer.search!("second");
 
       // call-2 runs to completion (its load resolves immediately).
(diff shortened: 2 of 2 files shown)


natb1:
# Fix-checks summary — PR #2201

## Iteration 1

**Outcome:** fixed

**Reproduced:** yes (sensor scripts runs against the committed diff)

**Reproduce command:** `git diff --no-color -U0 --diff-filter=d origin/main...HEAD -- '*.ts' '*.tsx' '*.js' '*.jsx' '*.mjs' '*.cjs' | bash .github/scripts/check-type-safety-escapes.sh --scan-stdin`

**Failure excerpt:**
```
::error file=print/test/viewer/epub.test.ts,line=961::Net-new type-safety escape hatch (as <Type> cast). Suppress with `// type-safety-ok: <reason>` if intentional.
::error file=print/test/viewer/epub.test.ts,line=963::Net-new type-safety escape hatch (non-null assertion !). Suppress with `// type-safety-ok: <reason>` if intentional.
::error file=print/test/viewer/epub.test.ts,line=973::Net-new type-safety escape hatch (non-null assertion !). Suppress with `// type-safety-ok: <reason>` if intentional.
::error file=print/test/viewer/epub.test.ts,line=977::Net-new type-safety escape hatch (non-null assertion !). Suppress with `// type-safety-ok: <reason>` if intentional.
```

**Why not caught:** The new regression test added four patterns (`as ReturnType<typeof vi.fn>`, two `!.` non-null method calls, and one definite-assignment `!`) that are identical to pre-existing patterns in the same file. They were not caught because the sensor is diff-scoped: it flags net-new occurrences on added lines, and these four were the only new additions matching those patterns.

**Fix:** Added `// type-safety-ok: <reason>` suppressions to each of the four flagged lines in `print/test/viewer/epub.test.ts`. Commit: `779e79e2`

natb1:
# QA Summary — Issue #1819 / PR #2201

## Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | EPUB renderer suite passes with the early-break change | shell-command | **PASS** |
| 2 | Spine loop contains the early-exit epoch check | shell-command | **PASS** |
| 3 | Post-loop epoch guard retained as a safety net | shell-command | **PASS** |
| 4 | New test asserts no loads/highlight for not-yet-visited sections after clearSearch | shell-command | **PASS** |
| 5 | No regression across the full print test suite | shell-command | **PASS** |

## Detail

- **Item 1** — `npx vitest run --root print test/viewer/epub.test.ts`: 52/52 tests passed, including the new "stops loading not-yet-visited spine sections once clearSearch advances the epoch mid-loop" test and the existing overlapping-search test (with the `await Promise.resolve()` timing fix).
- **Item 2** — `epoch !== _searchEpoch` guard confirmed at `print/src/viewer/epub.ts:303`, first statement of the spine loop body.
- **Item 3** — Post-loop guard at `print/src/viewer/epub.ts:342` (`if (epoch !== _searchEpoch) return results;`) retained alongside the new in-loop break.
- **Item 4** — New test found at `print/test/viewer/epub.test.ts:954`.
- **Item 5** — `npx vitest run --root print`: 480/480 tests passed. The 4 suite-load failures (`test/viewer/spread-search-highlight.test.ts`, `test/viewer/pdf.test.ts`, `test/enrichment.test.ts`, `test/local-folder-ui.test.ts`) are pre-existing local-environment vite fs-deny of `pdfjs-dist/build/pdf.worker.min.mjs?url`, identical on the clean baseline, unrelated to this change.

## Disposition

No residue — all acceptance criteria satisfied. Clean autonomous pass.

## No browser walkthrough

All acceptance criteria for this PR are about spine-loop logic correctness, verifiable by the vitest suite. No browser walkthrough was needed.

natb1:
## Review

**Surface:** `code` (2 files changed: `print/src/viewer/epub.ts`, `print/test/viewer/epub.test.ts`)
**CodeQL:** 0 open alerts. **Dependencies:** not touched. **Erosion:** 2 findings (see below).

---

### Fixed

- **E1** `print/src/viewer/epub.ts:303` — Removed inline comment `// a newer search/clearSearch superseded us — stop loading sections`; it restated what the code does rather than a non-obvious reason. (commit `25d52304`)

---

### Dismissed

- **E2** `print/src/viewer/epub.ts:288` — Erosion: cyclomatic complexity +1 (12→13, total 100→101). The epoch guard is a single conditional inherent to the fix; decomposing the function for one branch is disproportionate.
- **E3** `print/test/viewer/epub.test.ts:10` — Erosion: test-file complexity +4 (150→154). Test files are outside production complexity budgets; the 41-line test matches established patterns in this suite.

---

### Informational

- **S1** — No security attack surface in this diff: pure timing/control-flow logic in the EPUB spine search loop with no external inputs, auth, data-exposure, or Firestore-rules changes.
- **S2** `print/test/viewer/epub.test.ts:958,980,991` — Three `// type-safety-ok` inline suppression tags. Each is anchored to an existing pattern in the suite (the `// type-safety-ok: same pattern as existing…` phrasing). No action required.
- **S3** `print/test/viewer/epub.test.ts:810` — `await Promise.resolve()` inserted between two overlapping `search()` calls to restore the mid-load park timing the pre-existing test assumed. The fix is correct and minimal; the timing dependency is intentional (noted in both the PR description and inline comment). No action required.

---

_No deferred findings. No out-of-scope security advisories._

PR #2217, opened 2026-06-20
router: getServerSnapshot returns a fresh frozen snapshot per call

natb1:
Closes #2015

## What

`getServerSnapshot()` in `router/src/location-store.ts` previously returned a
module-level singleton `Snapshot` by reference on every call. That singleton's
`params` field is a shared mutable `URLSearchParams`, so any SSR caller that
follows the spec and mutates the returned `params` (`.set()`, `.delete()`,
`.append()`) would corrupt the singleton for every subsequent
`getServerSnapshot()` call in the same Node.js process — the module cache is
shared across requests.

This PR makes cross-call isolation structural: `getServerSnapshot()` now
allocates and returns a **fresh, frozen** snapshot on each call.

## Why

No active bug exists today — usage is client-only and `getServerSnapshot` is
only React's cheap hydration insurance via `useSyncExternalStore`. But the
shared-mutable-singleton design is a latent correctness hazard for any future
SSR adoption. Returning a fresh object per call removes the hazard at the
source.

Referential stability across `getServerSnapshot` calls is **not** required:
React's `useSyncExternalStore` only requires the client `getSnapshot` path to be
referentially stable (to avoid an infinite render loop). `getServerSnapshot` is
invoked by React's SSR/hydration APIs, which do not compare its result across
calls. A fresh object per call is safe and correct.

## Changes

- `router/src/location-store.ts`: delete the `SERVER_SNAPSHOT` singleton (and its
  now-stale doc comment); rewrite `getServerSnapshot()` to return
  `Object.freeze({ path: "/", params: new URLSearchParams() })` with a doc
  comment explaining the isolation guarantee, the shallow nature of
  `Object.freeze`, and why referential stability is unneeded here.
- `router/test/location-store.test.ts`: replace the old referential-stability
  assertion (now inverted by design) with strictly stronger tests — default
  content preserved, distinct objects per call, cross-call mutation isolation,
  and frozen result.

The client path (`getSnapshot`, `subscribe`, `navigate`, and the
`cachedSnapshot`/`cachedKey` caching contract) is unchanged.

## Verification

`npx vitest run --root router` — 74 tests pass, including the new isolation /
distinctness / frozen assertions and the unchanged client-path suite.

### DIFF
--- a/router/src/location-store.ts
+++ b/router/src/location-store.ts
@@ -30,15 +30,6 @@ const listeners = new Set<() => void>();
 let cachedSnapshot: Snapshot | null = null;
 let cachedKey: string | null = null;
 
-/**
- * Stable default snapshot for `getServerSnapshot`. Returned by reference every
- * call so it is referentially stable, matching the client-side caching
- * contract. This is cheap insurance so a future `hydrateRoot`/prerender cannot
- * throw because `getSnapshot` touched the global `location` during SSR. Full
- * SSR location handling is out of scope.
- */
-const SERVER_SNAPSHOT: Snapshot = { path: "/", params: new URLSearchParams() };
-
 function emit(): void {
   for (const listener of listeners) listener();
 }
@@ -69,8 +60,25 @@ export function getSnapshot(): Snapshot {
   return cachedSnapshot;
 }
 
+/**
+ * Returns a fresh frozen snapshot on every call. Two properties protect SSR
+ * callers from each other:
+ *
+ * (a) Fresh allocation (primary protection): each call gets its own object, so
+ *     one caller mutating `params` cannot affect any other call's snapshot.
+ * (b) `Object.freeze` (shallow secondary defense): blocks reassignment of
+ *     `path` and `params` on the returned object, but does NOT prevent
+ *     URLSearchParams methods (`.set`, `.delete`, etc.) from mutating the
+ *     instance — the primary protection above is what prevents cross-call
+ *     contamination.
+ *
+ * Referential stability is not required here: React's SSR/hydration path does
+ * not compare `getServerSnapshot` results across calls. Only the client-side
+ * `getSnapshot` path requires it (handled by the `cachedSnapshot`/`cachedKey`
+ * machinery above).
+ */
 export function getServerSnapshot(): Snapshot {
-  return SERVER_SNAPSHOT;
+  return Object.freeze({ path: "/", params: new URLSearchParams() });
 }
 
 export function navigate(href: string, opts?: { replace?: boolean }): void {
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — Issue #2015 / PR #2217

## Results

| # | Item | Result |
|---|------|--------|
| 1 | Fresh object returned on every call (no shared singleton) | PASS |
| 2 | Mutating returned `params` does not leak into a later call | PASS |
| 3 | Returned snapshot object is frozen | PASS |
| 4 | Client-side behavior unchanged (`getSnapshot` / `subscribe`) | PASS |
| 5 | Unit test covers the isolation guarantee (distinct objects + mutation isolation) | PASS |

**5 PASS, 0 FAIL, 0 SKIP**

## Details

All items were `script-verifiable` (pure TypeScript library change, no browser component).

**Vitest:** `npx vitest run --root router` → 74 tests passed across 4 test files. The ECONNREFUSED noise in the output is from happy-dom fetch tests exercising network-error behavior — all 74 assertions passed.

**Structural check:** `SERVER_SNAPSHOT` singleton confirmed absent from `router/src/location-store.ts`.

**Test names present:** "getServerSnapshot returns a distinct object on each call" and "getServerSnapshot isolates callers: mutating one call's params does not affect the next" confirmed in `router/test/location-store.test.ts`.

## Bugs Found

None.

## Deferred to Office-Hours

None.

## Disposition Triage

*(Skipped — residue list was empty; all items passed.)*

natb1:
## Review findings — PR #2217 (issue #2015)

### Fixed

**F1 · router/src/location-store.ts:80** — `getServerSnapshot()` return type was `Snapshot` (mutable) while the runtime value is `Object.freeze()`'d. Callers saw a mutable type where the value is frozen, creating a type/runtime mismatch.  
Fix: changed return annotation to `Readonly<Snapshot>`, aligning the static type with the runtime guarantee. The doc comment's shallow-freeze caveat (URLSearchParams methods still mutable) remains accurate since `Readonly<T>` only protects own properties, matching `Object.freeze`'s shallow behavior.  
Fix commit: 966d9541

---

### Informational

**F2 · router/test/location-store.test.ts:83–88** — The test suite verifies `Object.isFrozen()` structurally but does not assert that property reassignment throws. The doc comment's "secondary defense" claim (blocks reassignment of `path` and `params`) is not behaviorally exercised.  
No action required for this PR. Optionally, a future test could add `expect(() => { (getServerSnapshot() as any).path = '/other'; }).toThrow(TypeError)` to cover the runtime effect of freeze. Low priority — primary protection (fresh allocation) is well-tested by the isolation test.

---

### Dismissed

**erosion-001 · router/test/location-store.test.ts:20** — Cyclomatic complexity flagged as +3 vs baseline. Delta is from 4 new test cases covering SSR isolation, not production branching. Adding tests is the correct action; decomposing test blocks would reduce coverage clarity. Dismissed.

---

### Security

No attack surface identified on this diff. The change removes a shared mutable singleton and replaces it with a fresh frozen object per call — the security posture is strictly improved. No CodeQL alerts. No dependency changes.

PR #2218, opened 2026-06-20
dispatch: cap cross-tick oscillation with a total-attempt ceiling and no-progress escalation

natb1:
Closes #2040

## Summary

Caps cross-tick / cross-phase oscillation for a single dispatch issue with two
independent bounds, both implemented against today's labels (the greenfield
per-tick decision log, issue #1, would later subsume both — see the plan):

1. **An issue-anchored total-attempt ceiling** that survives PR recreation and
   phase regression. The existing per-phase counters all live on the *PR*, so
   they reset on a regress-to-plan (PR closed → labels deleted) and on every
   ready-promotion (`dispatch-reconcile-ready` clears them). A
   `plan → implement → fix → regress → re-plan → new PR` loop therefore restarts
   from zero each cycle and can spend the Opus fix lanes indefinitely. The new
   counter lives on the **issue**, so it is durable across exactly that loop.

2. **A qa-fix no-progress short-circuit** that escalates before re-spending the
   Opus fix lane on an attempt that resolves none of the prior attempt's failing
   items.

Past the ceiling (or on no-progress) the issue is parked on
`dispatch:office-hours` — which the selection queues already exclude — with a
one-line reason naming the total.

## Units

- **Unit 1 — `dispatch-attempt-count`** (`scripts/dispatch-attempt-count`): the
  issue-anchored monotonic counter. Modeled on `dispatch-qa-fix-attempt`, but
  issue-anchored and **bump-first-then-report**: it bumps the
  `dispatch:attempts-<n>` label first (so the count always reflects reality, even
  at the ceiling) and then prints `escalate` / `proceed`. Ceiling default `13`,
  env-overridable via `DISPATCH_TOTAL_ATTEMPT_CEILING` (the
  `DISPATCH_QA_FIX_ATTEMPT_CAP` precedent — no config schema). The label name
  deliberately omits the `-attempt-` infix so the reconcile-ready clear regex
  `^dispatch:.*-attempt-[0-9]+$` cannot match it (verified in a ```verify block).

- **Units 2 + 3 — Stop-hook wiring** (`.claude/hooks/dispatch-stop.sh`,
  committed together because they share one office-hours read). A single
  session-start read of the issue's office-hours state feeds both the bump-skip
  guard (Unit 2) and the un-park reset (Unit 3) so the two can never disagree.
  The bump runs once per **concluded** autonomous phase session, then parks +
  spawns + exits on `escalate`. On the parked→unparked transition (an
  office-hours-assisted advance), the counter is reset so resumed autonomous work
  gets a fresh budget instead of instantly re-hitting the ceiling.

- **Unit 4 — qa-fix no-progress short-circuit** (`scripts/dispatch-qa-noprogress`
  + `.claude/skills/qa-fix/SKILL.md` Step 3.7): persists each attempt's
  opus-fixable failing-id set in a `<!-- dispatch:qa-residue -->` PR comment
  (find-or-update, never stacked — the `dispatch-write-plan` marker-comment
  pattern, PR-anchored). Before re-spending the fix lane it compares the current
  set against the prior one; if a non-empty prior set had **none** of its items
  resolved, it escalates (naming the per-issue total). Fail-open: any gh/jq read
  error defaults to `progress`, so a flake never wrongly escalates.

- **Unit 5 — regression tests** (`scripts/test-dispatch-scripts.sh`): direct
  tests for both new scripts, following the `dispatch-qa-fix-attempt` test block
  idioms (PATH-shimmed fake `gh`, `*_GH_CMD` overrides). Covers the ceiling-fires
  case (asserting the bump **still lands** at the ceiling — the inversion vs the
  template), below-ceiling, config + guard exits, and the qa-noprogress
  progress / no-progress / resolved transitions for both line and JSON-array
  input.

## Notes for review (deliberate choices)

- **Unit 2 bump guard diverges from the plan's literal "bump before Branch A"
  text — on purpose, and it is the only way to satisfy AC2.** The Stop hook fires
  once per `ScheduleWakeup` poll; a healthy review phase polls many times. A bump
  placed *before* Branch A's two continuation gates would count every idle-poll
  wake-up (and every transient rate-limit resume), blowing the ceiling on a
  single healthy phase — the #1590 oscillation hazard. The guard therefore skips
  the bump when the issue is already office-hours-parked, **or** when the marker
  is absent **and** the session is a live idle-poller (`session_scheduled_wakeup`)
  or a rate-limit death. Everything else — Branch A genuine park, the #2025
  recovery-advance, Branch B/C/D — counts exactly once. The guard mirrors Branch
  A's own continuation gates; a comment flags the two to keep in sync.

- **A latent bug in Unit 4's script was found and fixed while writing Unit 5's
  tests.** The one-id-per-line normalization used `jq -R -s` without `-c`, writing
  a multi-line ids array that the script's own single-line `sed` read-back could
  never match — so no-progress detection would have silently never fired for the
  line-input form qa-fix actually uses. Added `-c`, and removed the `awk`
  compaction the first test draft had used to mask it (test-integrity).

- **AC1's selection exclusion is intentionally not re-tested.** The ceiling park
  reuses `dispatch-apply-office-hours`, and the existing office-hours-skip
  selection tests already cover that path (the "PR whose issue carries
  dispatch:office-hours is skipped" and help-wanted-skip tests). A duplicate would
  be noise; an inline comment in the test file records the coverage.

## Verification

- Full dispatch regression suite: **3603/3603 passed, 0 failed**.
- ```verify: `dispatch:attempts-7` does not match the reconcile-ready clear regex
  (AC1 survival-through-regress).
- ```verify: both new scripts and the Stop hook parse.


natb1:
# QA Summary — Issue #2040 (PR #2218)

## Results

**10 items total: 6 PASS, 0 FAIL, 0 SKIP**

### Script-verifiable items (all PASS)

| # | Title | Result |
|---|-------|--------|
| 1 | Both new scripts exist and are executable | PASS |
| 2 | Both new scripts parse without syntax errors | PASS |
| 3 | Stop hook parses without syntax errors | PASS |
| 4 | Regression test suite passes (3603/3603) | PASS |
| 5 | dispatch:attempts-N label does not match reconcile-ready clear regex | PASS |
| 10 | Escalation reason names per-issue attempt total (AC4) | PASS |

### Needs-human-judgment items (classified by Opus)

| # | Title | Class |
|---|-------|-------|
| 6 | AC2 default ceiling carries inline rationale | already-satisfied |
| 7 | AC1 cross-PR survival reasoning | already-satisfied |
| 8 | Un-park reset correctness | already-satisfied |
| 9 | qa-fix no-progress wiring in SKILL.md | needs-main (planned-deferral) |

No bugs found. No browser walkthrough (pure-scripts PR, no browser component).

---

## Disposition triage

Step 3.5 ran (4 residue items).

Non-fixing pass. No opus-fixable items — no auto-fix lane ran.

- **#6 — AC2 default ceiling carries inline rationale:** Default 13 baked in `dispatch-attempt-count` alongside a concrete, non-circular rationale: plan+implement+review+qa plus a handful of fix passes, with headroom for roughly one full re-plan cycle before firing (satisfies AC2). Verify: n/a.
- **#7 — AC1 cross-PR survival reasoning:** Counter is issue-anchored via `gh issue edit`; `dispatch-stop.sh` passes `$ISSUE_NUM` to `dispatch-attempt-count`; reconcile-ready's `-attempt-` PR-label regex provably cannot match the issue label `dispatch:attempts-N`. Survival across PR recreation / regress independently verified. Verify: n/a.
- **#8 — Un-park reset correctness:** `clear_attempt_counter` fires exactly once (dispatch-stop.sh line 282), only from `strip_office_hours_label`, gated on `ISSUE_OFFICE_HOURS_PRESENT=yes` (single shared session-start read). Verify: n/a.

**Needs-main (filed as follow-up):**

- **#9 — qa-fix no-progress wiring in SKILL.md (planned-deferral):** The prose wiring is in place in Step 3.7 of `qa-fix/SKILL.md`, but end-to-end execution correctness is observable only in a live production dispatch run. Filed as follow-up #2233 (`blocked_by` #2040).

---

## Filed follow-ups

- **qa-fix no-progress wiring in SKILL.md** → #2233 (newly filed; main-qa, dispatch:office-hours, blocked_by #2040)

---

## Escalation set

Empty — all residue items either already-satisfied (dropped as PASS) or needs-main (filed as follow-up). Clean pass.

natb1:
## Review — PR #2218 / issue #2040

### Fixed (commit `1c7406c2`)

- **review-7** — `dispatch-qa-noprogress:108` — `last` → `first` in jq filter for reading the prior residue marker comment (`prior_body`), so the read and write both target the same (first) matching comment. Previously the read used `last` while the write used `first`; on a single comment these agree, but if a second marker comment were ever posted both would diverge — the read would take the newer one while the write would update the older.
- **review-8** — `qa-fix/SKILL.md:788` — Added `(use dangerouslyDisableSandbox: true — gh needs network)` qualifier to the `gh issue view` TOTAL read in the no-progress escalation block, matching the qualifier on the `dispatch-qa-noprogress` call immediately above it.

### Informational

- **review-1** — `dispatch-attempt-count:394-412` — Remove-then-add label pair has a brief window where neither label is present. The GitHub labels API has no atomic swap; the gap is inherent to the approach and benign (the next read yields `max // 0 = 0` and would re-count from zero on a concurrent read, but that race window is sub-second and the counter is best-effort). Acknowledged in comments.
- **review-3** — `dispatch-qa-noprogress:528` — Author-id lookup (`gh api user`) is a live network call per invocation. The test env override (`DISPATCH_QA_NOPROGRESS_AUTHOR_ID`) avoids the lookup in tests; in production the call is a single round trip per qa-fix run.
- **review-4** — `dispatch-stop.sh:186-190` — `ISSUE_OFFICE_HOURS_PRESENT` is read once before the Branch A/B label mutations. A race where the label is applied between the read and Branch A/B is theoretically possible but the window is the width of the hook's own execution and the counter reset is idempotent.

### Dismissed

- **review-2** — `dispatch-attempt-count:350-355` — `if (( gh_rc != 0 ))` flagged as non-portable. Script declares `#!/usr/bin/env bash`; `(( ))` arithmetic is valid bash. Portability to `/bin/sh` is not a goal.
- **review-5** — `dispatch-attempt-count:376-380` — Ceiling allows very large values (no upper bound). Follows the `DISPATCH_QA_FIX_ATTEMPT_CAP` precedent; operator error, not a bug.
- **review-6** — `dispatch-qa-noprogress:567` — `mktemp` not scoped to `$CLAUDE_JOB_DIR/tmp`. Script is a general-purpose utility invoked outside Claude Code contexts; scoping to `CLAUDE_JOB_DIR` would break it.

### False-positive (security)

- **security-1** — All changed files are dispatch workflow orchestration scripts and a SKILL.md doc. No new network endpoints, user data, or authentication surfaces. No OWASP/STRIDE risks introduced.

PR #2231, opened 2026-06-20
useViewerController onSearchNavigate: narrow to SearchableRenderer, drop renderResult?.()

natb1:
Closes #2222

## Summary

Completes #1817's compile-time `renderResult` guarantee at the only call site.

#1817 made omitting `renderResult` on a searchable renderer a compile error at
the renderer-*definition* site (`SearchableRenderer` declares the four search
methods as required). That guarantee did not reach the *call* site:
`onSearchNavigate` in `print/src/viewer/useViewerController.ts` still used
optional chaining (`renderer.renderResult?.()`), where `renderer` is a
`ContentRenderer` whose `renderResult` is optional. The `?.` was type-permitted
and would silently no-op if `renderResult` were ever absent — exactly the
failure mode #1817 set out to eliminate, relocated to the call site.

## Change

Narrow `renderer` to `SearchableRenderer` via the existing `isSearchable` guard
and drop the optional chain:

```ts
function onSearchNavigate() {
  if (controller.enabled) {
    controller.goToPage(renderer.currentPage).catch(handleRenderError);
  } else if (isSearchable(renderer)) {
    renderer.renderResult().catch(handleRenderError);
  }
  syncNav();
}
```

This reuses the `isSearchable()` narrowing idiom already present elsewhere in the
file — no new guard. After narrowing, `renderResult()` is required, so the
non-optional call typechecks and a future regression that drops `renderResult`
is caught by the type checker.

## Behavior

Runtime-behavior-preserving type hygiene. A non-searchable renderer previously
hit the `?.` no-op and is now skipped — identical observable effect.
`onSearchNavigate` only fires for searchable content in practice, so the guard
is true at every real call.

## Verification

- `npm run build --prefix print` — typecheck + build pass (the narrowing is what
  makes the non-optional call compile).
- `npx vitest run --root print test/viewer/useViewerController.test.tsx` — 32
  tests pass; the existing `onSearchNavigate` test mocks all four search methods,
  so `isSearchable()` returns true and the single-mode assertion still holds.

### DIFF
--- a/print/src/viewer/useViewerController.ts
+++ b/print/src/viewer/useViewerController.ts
@@ -472,8 +472,8 @@ export function useViewerController(
     function onSearchNavigate() {
       if (controller.enabled) {
         controller.goToPage(renderer.currentPage).catch(handleRenderError);
-      } else {
-        renderer.renderResult?.().catch(handleRenderError);
+      } else if (isSearchable(renderer)) {
+        renderer.renderResult().catch(handleRenderError);
       }
       syncNav();
     }


natb1:
# QA Summary — PR #2231 (Issue #2222)

## Items executed

| 1 | TypeScript build typechecks with the narrowed non-optional call | script-verifiable | ✅ PASS |
| 2 | Viewer controller tests pass, including the onSearchNavigate case | script-verifiable | ✅ PASS |

## Counts

- PASS: 2
- FAIL: 0
- SKIP: 0

## Bugs found

None.

## Notes

This is a runtime-behavior-preserving type-hygiene change. The `npm run build --prefix print` build (typecheck) confirms the narrowed `isSearchable(renderer)` guard makes `renderer.renderResult()` (non-optional) compile, completing #1817's compile-time guarantee at the call site. All 32 viewer controller tests pass, including the `onSearchNavigate` test at `useViewerController.test.tsx:508`.

No browser walkthrough was needed — the change has no user-visible UI behavior change beyond what the existing tests already cover.

natb1:
## Review: PR #2231

**Surface:** `code` (1 file changed: `print/src/viewer/useViewerController.ts`)

### Fixed
_None._

### Required (security)
_None._

_None._

### Informational
_None._

### Dismissed
- **No CodeQL alerts** — CodeQL returned no open alerts for this PR ref.
- **No erosion findings** — Erosion scan returned no findings.

### False-positive (security)
_None._

### Deferred
_None._

### Out-of-scope (security)
_None._

---

Clean review. The change is a minimal, well-scoped type hygiene fix: the existing `isSearchable` narrowing guard was applied at the call site and the optional chain dropped. No security surface concerns, no correctness issues, no erosion. No follow-up issues filed.

PR #1943, opened 2026-06-19
audio: migrate to React on @commons-systems/ds

natb1:
Closes #1869

## What

Phase 3 of the design-system epic (#1864): migrate the `audio` app from
vanilla-TS DOM-string templates + a `<app-nav>` custom element to React
composing `@commons-systems/ds` (built in #1867, scaffolded in #1865). `audio`
is the first app to adopt React in the monorepo.

Behavior, DOM ids/`data-*`, and copy are preserved verbatim so the existing
Playwright e2e suite stays the parity gate (not rewritten). Toolchain changes
are audio-local only — no shared `config/` edits — and shared CSS
(`@commons-systems/style`) is untouched (its tokens migration is sibling #1866).

## How (7 units, one commit each)

1. **Toolchain** — add `@commons-systems/ds` + `react`/`react-dom` deps; enable
   automatic JSX in `audio/vite.config.ts` + `audio/tsconfig.json`.
2. **App shell** — a single React root owns the whole `.page` (header / DS `Nav`
   with an auth+folder `end` slot / panel toggle / content-grid / persistent
   player aside / footer). Minimal path-state routing replaces
   `createHistoryRouter` (which set `outlet.innerHTML`, fighting React). Auth
   state + panel toggle + the `<app-nav>` auth/home-link DOM are reimplemented as
   React.
3. **Pages** — Home + About as React components; library-region branches
   (list / empty / error), expandable rows composing DS `Checkbox`, cache
   section with DS `Button`. The old router-level `formatError` for
   `DataIntegrityError` is preserved via a small `RouteErrorBoundary`.
4. **Player** — see note below.
5. **Home interactivity** — `afterRenderHome` ported to React effects: queue
   checkbox → player, cache-stats + `CACHE_UPDATED_EVENT`, clear-cache,
   window-focus library rescan, folder-connect refetch.
6. **CSS tokens** — import the DS token CSS in the entry so composed DS
   components resolve their inline `var(--*)`; substitute value-identical
   literals for tokens in `theme.css`/`audio.css`. `@commons-systems/style` kept;
   DS `base.css` deliberately not imported (it would diverge the chrome).
7. **Tests** — page unit tests migrated to render the React components; dead
   string-API modules removed.

## Two implementation choices worth a reviewer's eye

- **Player engine reused, not rewritten.** `audio/src/player.ts`'s `initPlayer`
  (generation-counter race guard, out-of-order object-URL revocation,
[...]

PR #2232, opened 2026-06-20
audio: controlled queue checkbox in Home

natb1:
Closes #2010

## Problem

The audio Home page rendered its per-row queue checkbox with
`defaultChecked={player ? player.isQueued(item.id) : false}` — an **uncontrolled**
React input. React applies `defaultChecked` only on initial mount and never
reasserts it, so the checkbox could drift out of sync with `player.isQueued()`.
The old vanilla-JS code compensated with an `applyCheckboxState()` sweep after
every library re-render; the React migration (PR #1943) dropped that sweep, which
surfaced this issue.

## Fix

Convert the checkbox to a **controlled** input in `audio/src/pages/Home.tsx`
(no change to `audio/src/player.ts`):

- `Row` now uses `checked={player ? player.isQueued(item.id) : false}` instead of
  `defaultChecked`, so the value is recomputed from `player.isQueued()` on every
  Home re-render.
- `Home` gains a force-render counter via
  `const [, bumpQueue] = useReducer((n) => n + 1, 0)`, passed into `Row` as
  `onQueueChange` and called on **every** `onToggle` exit path (via a `finally`).
  A controlled checkbox reverts when its `onChange` doesn't change the rendered
  `checked`; calling `player.add/remove` alone does not re-render React, so the
  bump forces Home (and every non-memoized `Row`) to re-render and reassert
  `checked` from `isQueued`.
- The imperative `e.currentTarget.checked = !checked; // revert` line is removed —
  it fought the controlled input; the `finally` bump now snaps the checkbox back
  to `player.isQueued(item.id)` on the validation-fail and `!player` paths.

Why local React state rather than a player event/subscription (the issue's
floated "Recommended fix"): the queue is mutated **only** by this very checkbox —
`player.add`/`remove` have exactly one caller, Home's own `onToggle`. An added
queue-change event would be a self-loop with no external producer. The controlled
recompute is the general mechanism that surfaces any queue change on the next
render, present or future.

## Acceptance criteria

- **C1 (controlled prop)** — `defaultChecked` replaced with `checked`.
- **C4 (toggle still works)** — the `finally` bump re-renders so the visual flip
  sticks after add/remove.
- **C3 (focus rescan resync)** — the focus rescan already calls
  `setState({status:"loaded", items})`, re-rendering Home; the controlled `checked`
  is recomputed from `player.isQueued()` on that render.
- **C2 (external mutation surfaces)** — satisfied by construction: a controlled
  `checked` recomputes on every Home re-render.

## Tests

Extended the "interactive: checkbox queue toggle" block in
`audio/test/pages/home.test.tsx`: a sticky-add case (checkbox stays checked, no
revert), an uncheck case, and an external/rescan resync case (focus event flips
the checkbox to match a changed `isQueued`). The former `defaultChecked`-asserting
test now asserts the controlled `checked` state.

## Verification

- `npx vitest run --root audio` — 150 passed.
- `npm run build --prefix audio` — `tsc && vite build` succeeds.

### DIFF
--- a/audio/src/pages/Home.tsx
+++ b/audio/src/pages/Home.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useReducer, useRef, useState } from "react";
 import { Button, Checkbox } from "@commons-systems/ds";
 import { DataIntegrityError } from "@commons-systems/firestoreutil/errors";
 import { logError } from "@commons-systems/errorutil/log";
@@ -31,33 +31,40 @@ function formatBytes(bytes: number): string {
 function Row({
   item,
   player,
+  onQueueChange,
 }: {
   item: LibraryItem;
   player: PlayerHandle | null;
+  onQueueChange: () => void;
 }) {
   const onToggle = (e: React.ChangeEvent<HTMLInputElement>) => {
     const checked = e.currentTarget.checked;
-    const { id, title, artist, album, origin, storagePath, localName } = item;
-    const locatorOk = origin === "local" ? !!localName : !!storagePath;
-    if (!id || !title || !artist || !album || !origin || !locatorOk) {
-      logError(new Error("Queue toggle: missing data attributes on audio row"), {
-        operation: "queue-toggle",
-      });
-      e.currentTarget.checked = !checked; // revert
-      return;
-    }
-    if (!player) return;
-    if (checked) {
-      player.add({
-        id,
-        title,
-        artist,
-        album,
-        origin,
-        ...(origin === "local" ? { localName } : { storagePath }),
-      });
-    } else {
-      player.remove(id);
+    try {
+      const { id, title, artist, album, origin, storagePath, localName } = item;
+      const locatorOk = origin === "local" ? !!localName : !!storagePath;
+      if (!id || !title || !artist || !album || !origin || !locatorOk) {
+        logError(new Error("Queue toggle: missing data attributes on audio row"), {
+          operation: "queue-toggle",
+        });
+        return;
+      }
+      if (!player) return;
+      if (checked) {
+        player.add({
+          id,
+          title,
+          artist,
+          album,
+          origin,
+          ...(origin === "local" ? { localName } : { storagePath }),
+        });
+      } else {
+        player.remove(id);
+      }
+    } finally {
+      // Re-render Home so the controlled `checked` reasserts from isQueued on
+      // every exit path (add, remove, validation-fail, or !player).
+      onQueueChange();
     }
   };
 
@@ -79,7 +86,7 @@ function Row({
             label={null}
             data-queue-toggle
             aria-label={`Add ${item.title} to queue`}
-            defaultChecked={player ? player.isQueued(item.id) : false}
+            checked={player ? player.isQueued(item.id) : false}
             onChange={onToggle}
           />
           <span className="title">{item.title}</span>
@@ -120,6 +127,10 @@ export function Home(props: HomeProps) {
   // A DataIntegrityError is stored, then thrown during render so the route-level
   // RouteErrorBoundary catches it (a component cannot catch its own throw).
   const [fatalError, setFatalError] = useState<unknown>(null);
+  // A force-render counter: bumped on every queue toggle so Home (and each
+  // non-memoized Row) re-renders and recomputes the controlled `checked` from
+  // player.isQueued. Its value is never read in render.
+  const [, bumpQueue] = useReducer((n: number) => n + 1, 0);
 
   // Guards async setState after unmount: a late getCacheStats() / listLibrary()
   // resolution must not touch state once Home is gone.
@@ -247,7 +258,12 @@ export function Home(props: HomeProps) {
       ) : (
         <div id="media-list">
           {state.items.map((item) => (
-            <Row key={item.id} item={item} player={player} />
+            <Row
+              key={item.id}
+              item={item}
+              player={player}
+              onQueueChange={bumpQueue}
+            />
           ))}
         </div>
       );
--- a/audio/test/pages/home.test.tsx
+++ b/audio/test/pages/home.test.tsx
@@ -364,12 +364,100 @@ describe("Home", () => {
       const input = container.querySelector<HTMLInputElement>(
         "input[data-queue-toggle]",
       )!;
-      // defaultChecked=true; click toggles unchecked
+      // controlled checked=true; click toggles unchecked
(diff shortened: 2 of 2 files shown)


natb1:
# Fix-checks summary — PR #2232

## Iteration 1

- **Outcome**: fixed
- **Reproduced**: yes
- **Reproduce command**: `git diff --no-color -U0 --diff-filter=d origin/main...HEAD -- '*.ts' '*.tsx' '*.js' '*.jsx' '*.mjs' '*.cjs' | .github/scripts/check-type-safety-escapes.sh --scan-stdin`
- **Failure excerpt**: 7 `::error file=audio/test/pages/home.test.tsx,line=N::Net-new type-safety escape hatch (non-null assertion !).` at lines 390, 400, 419, 429, 445, 459, 510
- **Why not caught**: New test cases added for the controlled-checkbox fix use `querySelector(...)!` non-null assertions, a standard test pattern. The type-safety-sensor flags all net-new `!` assertions unless suppressed with `// type-safety-ok: <reason>`. The tests were correct but lacked the suppression marker.
- **Fix**: Added `// type-safety-ok: querySelector in test, element guaranteed by rendered component` to each of the 7 flagged lines in `audio/test/pages/home.test.tsx`. Commit `d2e49371`.

natb1:
# QA Summary — Issue #2010

**PR:** #2232 · **Attempt:** 1 (first pass)

## Results

| # | Item | Classification | Result |
|---|------|---------------|--------|
| 1 | C1 — controlled `checked` prop replaces `defaultChecked` in `Row` | script-verifiable | ✅ PASS |
| 2 | Full audio unit suite (C1–C4 via tests, incl. sticky-add/uncheck/rescan-resync) | script-verifiable | ✅ PASS |
| 3 | Audio app type-checks and builds (prop-type regression check) | script-verifiable | ✅ PASS |
| 4 | Queue checkbox toggle sticks and stays in sync in the running app (C4, C2, C3) | needs-browser | ✅ PASS |

**4 passed · 0 failed · 0 skipped**

## Details

**Item 1 — C1 (controlled prop):** No `defaultChecked` remains in `audio/src/pages/Home.tsx`. Controlled `checked={player ? player.isQueued(item.id) : false}` confirmed at line 89.

**Item 2 — Unit suite:** 151 tests passed across 11 test files, including the new `sticky-add`, `uncheck`, and `focus-rescan resync` cases added by this PR.

**Item 3 — Build:** `tsc && vite build` succeeded clean (104 modules). Pre-existing chunk-size warnings are unrelated to this change.

**Item 4 — Browser walkthrough (public-domain items):**
- Checked Moonlight Sonata → checkbox turned checked, track appeared in queue panel ✅
- Unchecked it → checkbox cleared, queue emptied ✅
- Re-checked, then triggered `window.dispatchEvent(new Event('focus'))` (focus rescan) → checkbox remained checked after re-render ✅
- Console: one `AbortError: play() interrupted by pause()` — pre-existing browser audio policy behavior, not related to this change
- Network: no product 4xx/5xx; Firestore 503s are normal long-poll reconnects

**Walkthrough GIF:** `tmp/qa-fix-walkthrough-2010.gif`

## Disposition triage

Step 3.5 was skipped — residue list is empty (all items passed).

---

*Clean pass — no residue, no escalation.*

natb1:
## Review — PR #2232 (`2010-checkbox-queue-state-can-be`)

**Surface:** `code` (application source + tests) · **Security:** no open CodeQL alerts · **Deps:** unchanged

---

### Fixed
_None._

### Required (security)
_None._

_None._

### Informational

- **code-review** `audio/test/pages/home.test.tsx:295` — The `!player` early-return path is not covered by the new controlled-checkbox tests. The old imperative revert on that path was removed; the `finally` bump fires but there is no assertion that the checkbox stays unchecked when `player` is `null`.

- **code-review** `audio/test/pages/home.test.tsx:295` — The validation-failure path (item with missing required fields) is not covered. The `logError` + early-return case removes the old `e.currentTarget.checked = !checked` revert, but no test asserts the checkbox reverts via the `finally` bump.

- **code-review** `audio/src/pages/Home.tsx:64` — `onQueueChange()` fires unconditionally in the `finally` block, including on the `!player` early-return path. This causes a redundant `bumpQueue` re-render on every `!player` toggle attempt — not a bug, but a minor extra render with no observable effect.

- **code-review** `audio/src/pages/Home.tsx:40` — The `try/finally` has no `catch` block. An exception from `player.add()` or `player.remove()` propagates to React's event handler. React's error boundary would catch it, but the `finally` bump still fires first. Consistent with how the codebase handles other unguarded player calls.

- **review** `audio/src/pages/Home.tsx:40-68` — `onToggle` is defined as a plain inline function rather than `useCallback`. Every `bumpQueue` re-render creates a new function reference and passes it into every `Row` — not a problem given `Row` is not memoized, but worth noting if `Row` is ever wrapped in `React.memo`.

- **review** `audio/test/pages/home.test.tsx:295-515` — The new test block covers add, remove, initial-state (true and false), and external resync via focus. The `!player` path and validation-failure path are not covered (echoes the code-review note above). The existing interaction-without-player test asserts `player.add` is not called but does not verify the checkbox visual state after the event.

- **erosion** `audio/src/pages/Home.tsx:40` — Cyclomatic complexity rose from 57 → 58. The increase is attributable to the `try/finally` structure; complexity remains well within a manageable range.

- **erosion** `audio/test/pages/home.test.tsx:43` — Cyclomatic complexity rose from 57 → 66 (nine new `it` blocks). Expected for the new test cases; no single function is disproportionately complex.

- **erosion** `audio/test/pages/home.test.tsx:302` — Clone blocks rose from 6 → 10; duplicated lines from 442 → 585. The new test cases follow the same `render → querySelector → act → querySelector → expect → cleanup` pattern, which duplicates naturally across test cases. A `renderAndGetInput` helper could reduce this if the file grows further.

### Dismissed
_None._

### False-positive (security)

- **red-team** `audio/src/pages/Home.tsx:64-68` — Claimed: missing `catch` exposes exceptions. Assessment: `player.add`/`remove` are synchronous calls on a locally-held object with no async or network surface. An exception here would be a programming error, not a security-relevant scenario. React's event-handler boundary handles it. Not a vulnerability.

- **red-team** `audio/src/pages/Home.tsx:89` — Claimed: `player.isQueued(item.id)` called on every render could be abused. Assessment: `isQueued` is a synchronous O(1) set lookup with no side effects or network access. The item IDs come from Firebase-authenticated library data, not user-controlled input. Not a vulnerability.

### Deferred
_None._

### Out-of-scope (security)
_None._

---

No code changes applied. All findings are informational or false-positive.

PR #2234, opened 2026-06-20
fellspiral/blog: make Archive default-open state data-driven (drop wall-clock read)

natb1:
Closes #2223

## Problem

The shared `blog/` package's info-panel **Archive** section computed its
default-open `<details>` state from the wall clock. `Archive` read
`const now = new Date()` during render and threaded `now.getUTCFullYear()` /
`now.getUTCMonth()` into the `<details open>` attributes.

The SSR prerender runs at **build time**; client hydration runs at **serve
time**. When those two moments straddle a UTC month/year boundary, the baked-in
`open` attributes diverge from what the client renders → a React hydration
mismatch. The `blog` package is consumed by both `fellspiral` and `landing`, so
both apps were affected.

## Fix

Make the Archive default-open state **data-driven**, removing `new Date()` from
the render path entirely:

- **Unit 1 (`blog/src/components/InfoPanel.tsx`)** — derive the default-open year
  from the newest year present (`sortedYears[0]`) and the default-open month from
  that year's newest month (`Math.max(...grouped.get(openYear)!.keys())`), not a
  global newest-month across all years. Rename the now-misleading `currentYear` /
  `currentMonth` props to `openYear` / `openMonth`. The `new Date(post.publishedAt)`
  parse in `groupByYearMonth` is untouched — it reads a fixed published-at string,
  not the clock.
- **Unit 2 (`blog/test/components/info-panel.test.ts`)** — add a determinism
  regression test that renders the panel under two fake clocks straddling a UTC
  month boundary and asserts the HTML is byte-identical, and retire the now-
  unnecessary clock-pin scaffolding (`vi.setSystemTime`) from the existing archive
  block. Net coverage increases; no assertion is removed, skipped, or loosened.

The result is deterministic (byte-identical SSR/client markup) and equal-or-
better UX than "current calendar month" (which can be empty when no posts exist
this month).

## Verification

- `blog` unit tests: 415 passed.
- Typecheck: blog skipped by the runner (origin/main has pre-existing typecheck
  errors unrelated to this change).

Manual QA (deferred, per acceptance criterion 4): in built `fellspiral` and
`landing` previews, confirm the Archive section opens the newest year and its
newest month and the console shows no React hydration warning for the info panel.

### DIFF
--- a/blog/src/components/InfoPanel.tsx
+++ b/blog/src/components/InfoPanel.tsx
@@ -61,27 +61,27 @@ function MonthBlock({
 function YearBlock({
   year,
   months,
-  currentYear,
-  currentMonth,
+  openYear,
+  openMonth,
   postLinkPrefix,
 }: {
   year: number;
   months: Map<number, PublishedPost[]>;
-  currentYear: number;
-  currentMonth: number;
+  openYear: number;
+  openMonth: number;
   postLinkPrefix: string;
 }) {
-  const isCurrentYear = year === currentYear;
+  const isOpenYear = year === openYear;
   const sortedMonths = [...months.keys()].sort((a, b) => b - a);
   return (
-    <details open={isCurrentYear}>
+    <details open={isOpenYear}>
       <summary>{year}</summary>
       {sortedMonths.map((month) => (
         <MonthBlock
           key={month}
           month={month}
           posts={months.get(month)!}
-          isOpen={isCurrentYear && month === currentMonth}
+          isOpen={isOpenYear && month === openMonth}
           postLinkPrefix={postLinkPrefix}
         />
       ))}
@@ -113,8 +113,9 @@ function Archive({
   if (published.length === 0) return null;
 
   const grouped = groupByYearMonth(published);
-  const now = new Date();
   const sortedYears = [...grouped.keys()].sort((a, b) => b - a);
+  const openYear = sortedYears[0];
+  const openMonth = Math.max(...grouped.get(openYear)!.keys());
 
   return (
     <section className="panel-section">
@@ -124,8 +125,8 @@ function Archive({
           key={year}
           year={year}
           months={grouped.get(year)!}
-          currentYear={now.getUTCFullYear()}
-          currentMonth={now.getUTCMonth()}
+          openYear={openYear}
+          openMonth={openMonth}
           postLinkPrefix={postLinkPrefix}
         />
       ))}
--- a/blog/src/components/InfoPanel.tsx
+++ b/blog/src/components/InfoPanel.tsx
@@ -115,7 +115,7 @@ function Archive({
   const grouped = groupByYearMonth(published);
   const sortedYears = [...grouped.keys()].sort((a, b) => b - a);
   const openYear = sortedYears[0];
-  const openMonth = Math.max(...grouped.get(openYear)!.keys());
+  const openMonth = Math.max(...grouped.get(openYear)!.keys()); // type-safety-ok: openYear is derived from grouped.keys() so the entry always exists
 
   return (
     <section className="panel-section">
--- a/blog/test/components/info-panel.test.ts
+++ b/blog/test/components/info-panel.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, afterEach, beforeEach } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import { createElement } from "react";
 import { renderToStaticMarkup } from "react-dom/server";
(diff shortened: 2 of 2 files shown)


natb1:
# Fix-checks summary — PR #2234

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes (verified suppression logic locally against working-tree diff)
- **Reproduce command:** `git diff --no-color -U0 --diff-filter=d origin/main -- '*.ts' '*.tsx' '*.js' '*.jsx' '*.mjs' '*.cjs' | .github/scripts/check-type-safety-escapes.sh --scan-stdin`
- **Failure excerpt:** `Net-new type-safety escape hatch (non-null assertion !). Suppress with // type-safety-ok: <reason> if intentional.` — `blog/src/components/InfoPanel.tsx:118`
- **Why not caught:** The `grouped.get(openYear)!.keys()` non-null assertion was introduced by Unit 1 but no `// type-safety-ok:` suppression comment was added. The assertion is provably safe — `openYear` is `sortedYears[0]` where `sortedYears = [...grouped.keys()]`, so the key is always present in `grouped`.
- **Fix:** Added `// type-safety-ok: openYear is derived from grouped.keys() so the entry always exists` trailing comment on `blog/src/components/InfoPanel.tsx:118`. Commit: `692d843b`

natb1:
# QA Summary — Issue #2223, PR #2234

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | blog info-panel unit tests pass (full suite) | shell | PASS |
| 2 | Wall clock removed from Archive render path (AC 1) | shell | PASS |
| 3 | Default-open derived from post data, props renamed (AC 1) | shell | PASS |
| 4 | Cross-UTC-boundary determinism regression test exists (AC 2) | shell | PASS |
| 5 | Clock-pin scaffolding no longer load-bearing for open-state assertions (AC 3) | shell | PASS |
| 6 | Archive renders newest year/month expanded, no hydration warning in fellspiral/landing (AC 4) | needs-human-judgment | planned-deferral |

**PASS: 5 | FAIL: 0 | SKIP: 0**

## Deferred to office-hours

None — Item 6 classified `needs-main` (planned deferral) and filed as a follow-up (see below), not deferred to office-hours.

## Bugs found

None.

## Disposition triage

Item 6 (AC 4 visual QA) classified **`needs-main`** by the disposition Workflow — it carries `planned_deferral: true`. AC 4 is explicitly documented in the plan as non-assertable at merge time: it requires a cross-UTC-boundary build comparison and console inspection in a built `fellspiral`/`landing` preview, which cannot be asserted deterministically without a browser and a cross-boundary build. The deferral is by design; no code defect exists.

## Filed follow-ups

- #2238 for identifier `qa-needs-main #2223: /` — "qa: verify against main — qa-needs-main #2223: /"

## Result

Clean pass — no residue items in the escalation set. All acceptance criteria confirmed by script or documented as an explicit planned deferral filed as a `main-qa` follow-up.

natb1:
## Review — PR #2234 (`blog`: archive default-open data-driven, fix SSR/hydration mismatch)

**Surface:** `code` (2 files changed) · **Security finders:** ran, no findings · **Deviation:** none

---

### Dismissed

- **`blog/src/components/InfoPanel.tsx:118`** — `Math.max(...grouped.get(openYear)!.keys())` spreads an iterator into `Math.max`. For month keys (0–11, at most 12 values per year) this is safe; no stack overflow or precision risk. _(rationale: spread-into-Math.max is a valid, idiomatic pattern at this bounded scale — not an issue)_

---

### Informational

- **`blog/test/components/info-panel.test.ts:341`** — The new hydration-determinism test still uses `vi.useFakeTimers()` / `vi.setSystemTime()`, but the updated `Archive` component no longer reads the system clock. The fake-timer scaffolding is harmless but redundant — the test passes equally without it, since `renderInfoPanel` now derives its open state from the post data only. Worth noting for a future cleanup, but the test is correct and the assertion is valid.

- **`blog/src/components/InfoPanel.tsx:117`** — `openYear = sortedYears[0]` has no explicit guard for an empty `sortedYears` array. The `if (published.length === 0) return null` early return at line 113 ensures this path is never reached with an empty array, making the access safe by invariant. A comment noting the guard would make this easier to audit, but it is not a bug.

- **`blog/test/components/info-panel.test.ts:341–352`** — The determinism test advances the clock from `2026-01-31T23:59:00Z` to `2026-02-01T00:01:00Z` across a UTC month boundary and asserts byte-identical HTML. This validates the primary regression scenario. The test is well-structured and the assertion is meaningful.

---

### Fixed

_None._

### Deferred / Out-of-scope

_None._

PR #2226, opened 2026-06-20
file-issue: apply the configured epic label to parent epics

natb1:
Closes #2057

## What

`/file-issue` decomposes an over-large issue into a parent epic + sub-issues (Step 3f), and treats an existing issue that already has sub-issues as a parent epic — but it never applied an **epic label** to that parent. Step 6 finalize stamped only `help wanted`, at most one type label, and at most one topic label.

This broke the dispatch epic-resolution path: `dispatch-epic-resolved-candidate` gates on the issue carrying a configured epic label *before* it inspects the sub-issues. With nothing in the filing pipeline applying that label, a verifiably-complete parent epic was not routed to `/resolve-epic`; it fell through to `/plan-issue` (nothing to plan) and had to be office-hours-parked for a manual close. Observed on epic #1873.

## Changes

- **`.claude/skills/file-issue/SKILL.md` Step 6** — new `### Epic label` subsection. It applies the configured epic label to the **parent epic only**, gated on the canonical leaf-vs-epic signal (non-empty Step 1 sub-issue fetch / Step 3f decomposition), and explicitly excludes leaf issues and sub-issues. The label name is read from `dispatch-config-load epic`, mirroring the resolution-side reader in `dispatch-epic-resolved-candidate` so the filing side and the resolver share **one source of truth**. `no-config` → no label applied (the feature stays inert).
- **Repo-side `epic` label created** (live `gh label create`, repo metadata — no diff) so a future `--add-label "epic"` succeeds.

## Scope boundary

End-to-end auto-resolution **additionally** requires the user's machine-local, uncommitted `dispatch.config/epic.json` with a `labels` array — that config is user-controlled and out of scope. In this repo `dispatch-config-load epic` returns `no-config`, so the Step 6 change is **inert here** (applies no label) until the config exists. The new subsection documents this requirement. No change to `dispatch-epic-resolved-candidate`, `dispatch-route`, or `/resolve-epic`.

## Verification

SKILL.md prose plus repo metadata — no CI-testable code surface, so the plan defines no `verify` block (confirmed: `dispatch-run-verification` exits 3, build proceeds unchanged). Repo `epic` label confirmed present via `gh label list --search epic`.

PR #2247, opened 2026-06-20
dispatch: guard dispatch-config-load epic exit code in both callers

### LINKED ISSUE #2239: dispatch-config-load epic exit-code guard missing in /file-issue and dispatch-epic-resolved-candidate
## Context

`dispatch-config-load epic` has three distinct exit outcomes:

- **Exit 0, stdout `no-config`** — file absent; feature is inert (expected).
- **Exit 0, stdout JSON** — file present and valid.
- **Exit 1, stderr error** — file present but invalid JSON or schema validation failure.

Neither caller currently distinguishes exit 1 (load failure) from exit 0. Both treat any non-`no-config` stdout as valid JSON, which breaks silently when the config file exists but is malformed.

### `/file-issue` SKILL.md Step 6 (Epic label bash snippet)

```bash
EPIC_CONFIG=$(.claude/skills/dispatch-propagate/scripts/dispatch-config-load epic)
if [[ "$EPIC_CONFIG" != "no-config" ]]; then
  while IFS= read -r epic_label; do
    [[ -n "$epic_label" ]] && gh issue edit <N> --add-label "$epic_label"
  done < <(jq -r '.labels[]' <<<"$EPIC_CONFIG")
fi
```

In bash, a variable assignment (`VAR=$(cmd)`) does **not** propagate the command's exit code — `set -e` does not fire. When `dispatch-config-load epic` exits 1 (bad JSON/schema), `EPIC_CONFIG` is empty (the error goes to stderr). The `if` branch fires (empty string != `no-config`), and `jq -r '.labels[]'` receives empty input, failing silently with no label applied and no error surfaced to the caller.

### `dispatch-epic-resolved-candidate` (line 16)

```bash
EPIC_CONFIG=$("$SCRIPT_DIR/dispatch-config-load" epic)
if [[ "$EPIC_CONFIG" == "no-config" ]]; then
  exit 1
fi
```

Same issue: `set -euo pipefail` does not catch exit codes from variable-assignment RHS commands. A load failure (exit 1) leaves `EPIC_CONFIG` empty, the `no-config` check passes (empty != `no-config`), and execution falls through to `jq -r '.labels[]'` on line 21 with empty input — causing a jq error that exits the script non-zero but without a meaningful diagnostic (exit code 3 is reserved for "hard gh failure"; the exit here would be an unexpected jq non-zero, indistinguishable from a network error).

### Why deferred

The path cannot fire in production today: no `dispatch.config/epic.json` exists in the repo, so `dispatch-config-load epic` always returns `no-config` (exit 0). The fix is inert until a user provisions the config file. Not applied in PR #2226 (issue #2057) for this reason.

## Fix

Add an explicit exit-code guard to both callers as a paired change. The guard must:

1. **Capture the exit code** from `dispatch-config-load epic` separately from its stdout.
2. **Treat exit 0 + `no-config`** as inert (no-op, expected).
3. **Treat exit 0 + JSON** as valid config; proceed to label application.
4. **Treat exit 1** as a load failure: print a descriptive error to stderr and exit/return a hard-error code distinct from "not a candidate" (`exit 1` in `dispatch-epic-resolved-candidate` convention; explicit error message in the SKILL.md snippet).

Example pattern:

```bash
EPIC_CONFIG=$("$SCRIPT_DIR/dispatch-config-load" epic) || {
  echo "error: dispatch-config-load epic failed (bad JSON/schema in dispatch.config/epic.json)" >&2
  exit 3
}
```

Note: this pattern works in a committed `.sh` script with `set -euo pipefail` only when the `||` is on the same statement as the assignment, which prevents `set -e` from triggering and lets the `||` branch run. The SKILL.md snippet (bash pseudocode for model execution) should document the equivalent pattern clearly.

## Acceptance criteria

- [ ] `dispatch-epic-resolved-candidate`: when `dispatch-config-load epic` exits 1, the script exits with a distinct hard-error code (e.g. exit 3, consistent with its existing `exit 3` convention for "hard error") and emits a descriptive error to stderr.
- [ ] `dispatch-epic-resolved-candidate`: when `dispatch-config-load epic` exits 0 with `no-config`, behavior is unchanged (exit 1, not a candidate).
- [ ] `dispatch-epic-resolved-candidate`: when `dispatch-config-load epic` exits 0 with valid JSON, behavior is unchanged (proceeds to label intersection).
- [ ] `/file-issue` SKILL.md Step 6 bash snippet: the pseudocode documents the exit-code guard pattern so the model implementing the step applies the guard.
- [ ] `/file-issue` SKILL.md: the snippet clearly distinguishes `no-config` (inert), valid JSON (proceed), and load-failure (error) as three distinct branches.
- [ ] Unit test (or integration test) for `dispatch-epic-resolved-candidate` covering the load-failure path: given a malformed `epic.json`, the script exits 3 and emits an error to stderr.

natb1:
Closes #2239

## Summary

`dispatch-config-load epic` has three exit outcomes: exit 0 + stdout `no-config`
(file absent, feature inert), exit 0 + JSON (file present and valid), and exit 1
+ stderr (file present but malformed JSON / schema failure).

Two callers captured the loader's stdout via `VAR=$(...)`, which in bash does
**not** propagate the command's exit code (an assignment-RHS failure does not
trip `set -euo pipefail`). On a load failure (exit 1) `EPIC_CONFIG` was left
empty, the `no-config` check passed (empty `!=` `no-config`), and execution fell
through — applying no label and surfacing no meaningful diagnostic. In
`dispatch-epic-resolved-candidate` the fall-through hit `jq` on empty input,
producing an opaque non-zero exit indistinguishable from a real gh/network
error.

This adds the explicit exit-code guard to both callers as a paired change, so a
future provisioned-but-malformed `epic.json` fails loudly instead of silently.
The path is inert in production today (no `dispatch.config/epic.json` exists, so
the loader always returns `no-config`) — this is a defensive guard, which is why
it was deferred from PR #2226 (#2057).

## Changes

- **`dispatch-epic-resolved-candidate`** — add a same-statement `||` guard so any
  non-zero loader exit becomes the script's existing hard-error code `exit 3`
  with a descriptive stderr message, before the `no-config` short-circuit and
  `.labels[]` extraction run.
- **`test-dispatch-scripts.sh`** — add a covering test (GATE-4) that writes a
  malformed `epic.json`, runs the candidate, and asserts exit `3` plus the
  candidate's own `dispatch-epic-resolved-candidate:` stderr prefix (grepping the
  prefix, not a bare non-empty check, since `dispatch-config-load` also writes to
  stderr on failure).
- **`file-issue/SKILL.md` Step 6** — update the model-executed epic-label snippet
  to add the same `||` guard plus an inline comment naming the three loader
  outcomes (`no-config` / JSON / load-failure).

## Verification

The hermetic dispatch script unit-test harness (what CI runs) passes 3605/3605,
including the new load-failure test.


natb1:
## QA — Issue #2239 / PR #2247

### Results

| Shell-command | 3 | 3 | 0 | 0 |
| Needs-human-judgment | 1 | — | — | — |

**Total: 3 PASS, 0 FAIL, 0 SKIP**

### Shell-command results

1. **dispatch-epic-resolved-candidate exits 3 on malformed epic.json** — PASS
   - `rc=3`, stderr contains `dispatch-epic-resolved-candidate:` prefix diagnostic.

2. **no-config and valid-JSON branches unchanged (no regression)** — PASS
   - `noconfig_rc=1` (exit 1 = not a candidate, unchanged).

3. **GATE-4 unit test passes in the dispatch test suite** — PASS
   - 3896/3896 tests passed, including:
     - PASS: gate: malformed epic.json → exit 3
     - PASS: gate: malformed epic.json → candidate diagnostic on stderr

### Disposition triage (Opus)

1 item was assessed by the disposition Workflow; 0 items remain in the escalation set.

- **/file-issue SKILL.md Step 6 documents the three-branch exit-code guard** (`needs-human-judgment`) — **dropped as PASS**

  SKILL.md Step 6 (lines 411–427) already contains the full three-branch exit-code guard. The comment (lines 412–414) distinguishes all three outcomes verbatim: exit 0 + `no-config` (feature inert, apply NO label), exit 0 + JSON (apply each `.labels[]`), and exit 1 (bad JSON/schema → error out). The loader call is guarded on the same statement with `|| { echo ...; exit 1; }` (lines 418–421), and lines 415–416 explain why a bare `VAR=$(...)` does not propagate the exit code. Every clause of the expected outcome is provably met by the readable file content — no code defect and no human input required.

### Bugs found

None.

### Clean pass

All acceptance criteria satisfied. Applying `dispatch:qa-done`.

natb1:
## Review — PR #2247 (`dispatch-config-load` exit-code guard)

Fix commit: a9d8e3ae

### Fixed

- **`SKILL.md:416` — inaccurate `VAR=$(...)` exit-code comment** (`code-review`): The original comment said "A bare `VAR=$(...)` does NOT propagate the loader's exit code", which is misleading under `set -euo pipefail`. Rewrote to describe the loader's raw exit codes (exit 1 = bad JSON/schema, exit 2 = env/usage error) and explain that the `||` guard normalises to a diagnostic `exit 1` regardless.

- **`test-dispatch-scripts.sh:3632–3636` — GATE-4 phantom always-true assertion** (`review`): The success branch of the stderr-prefix check used `assert_eq ... 'pass' 'pass'`, which always passes and provides no signal. Replaced with a single yes/no `grep` capture `assert_eq` matching the file's existing grep-check convention.

- **`dispatch-epic-resolved-candidate:17` — `||` guard message doesn't distinguish exit codes** (`code-review`, `review`): The guard emitted the same message for both exit 1 (bad JSON/schema) and exit 2 (env/usage error). Fixed to capture `rc=$?` and emit a distinct `dispatch-epic-resolved-candidate: dispatch-config-load epic: environment/usage error` for exit 2 vs the existing JSON/schema message for exit 1.

### Informational

- **`test-dispatch-scripts.sh:3629–3630` — GATE-4 uses issue number 71 without stub** (`review`): GATE-4 calls `dispatch-epic-resolved-candidate 71` without an `arg-issue-71.json` stub. This is correct — the stub harness exits early on config-load failure (before any issue fetch), so no stub is needed. Noted for reader orientation, no change required.

- **`SKILL.md:419` — `error:` prefix vs `dispatch-epic-resolved-candidate:` prefix** (`review`): The SKILL.md snippet uses a generic `error:` prefix while the actual script uses a `dispatch-epic-resolved-candidate:`-prefixed message. Minor style inconsistency; the two are independent callers and the SKILL.md snippet is illustrative, not normative.

### Dismissed / False-positive

- **`dispatch-epic-resolved-candidate:17` — hardcoded path in error message** (`red-team`): The message hardcodes `dispatch.config/epic.json` as the config location. `dispatch-config-load` does resolve the path via env var / git root, but the conventional path matches the message; the message is accurate for all normal deployments. False-positive.

- **`SKILL.md:423–425` — label injection from `epic.json`** (`red-team`): `dispatch-config-load` validates the config against a schema before returning, and `gh issue edit --add-label` is not a shell injection vector (the label string is passed as a separate argument, not interpolated into a shell command). False-positive.

### Deferred

- **`test-dispatch-scripts.sh:3625` — missing dispatch-route integration test for config-load failure path** (`code-review`): GATE-4 tests `dispatch-epic-resolved-candidate` in isolation but `dispatch-route`'s else branch (handling any exit code ≠ 0/1 as a hard error → `INVOKE /plan-issue`) has no covering test for the new failure mode. Filed as #2296 (blocked by #2239).

### Security

No attack-surface findings. CodeQL: no open alerts. No dependency changes (deps=false). Erosion scan: no findings.

Issue #2135, opened 2026-06-20
lib.sh: `dispatch_marker_comment_id` pipeline assignment missing `|| return 1`, silently masks gh failures


natb1:
## Context

`dispatch_marker_comment_id` in `.claude/skills/dispatch-propagate/scripts/lib.sh` (lines 183-185) assigns the comment id via a pipeline:

```bash
cid=$(gh_retry gh api --paginate 'repos/{owner}/{repo}/issues/$n/comments' | jq -r ...)
```

There is no `|| return 1` after the assignment. Correctness currently relies on `set -e` being active in every calling script (true for all current callers), but this is a fragile invariant:

- If the function is called from a conditional context (e.g. `if dispatch_marker_comment_id ...`), bash suppresses `set -e` inside the condition expression. A `gh_retry` or `jq` failure inside the command substitution would silently set `cid` to empty, causing the function to return 0 (no comment found) instead of propagating the error.
- If a future caller sources the function in a script without `set -e`, the same silent masking occurs.

Surfaced during review of PR #2108. Out of scope for that PR.

## Recommended fix

Add an explicit `|| return 1` after the pipeline assignment:

```bash
cid=$(gh_retry gh api --paginate 'repos/{owner}/{repo}/issues/$n/comments' | jq -r ...) || return 1
```

This makes the function self-contained and correct regardless of the caller's `set -e` state, consistent with the project code-style preference for clear errors over silent fallbacks.

## Acceptance criteria

- [ ] `.claude/skills/dispatch-propagate/scripts/lib.sh` lines 183-185: the `cid=$(...)` assignment ends with `|| return 1`
- [ ] A `gh_retry` or `jq` failure in the pipeline causes the function to return non-zero, regardless of whether the caller has `set -e` active
- [ ] No regression in the 'no matching comment' path: when `jq` returns `null` or empty on success, the function still returns 0 (the `|| return 1` only fires on a non-zero exit from the pipeline itself)

PR #2249, opened 2026-06-20
dispatch: add open-blocker re-check at plan-issue/implement worker startup

natb1:
Closes #2242

## What

Add an open-blocker re-check at the **phase-worker startup boundary** in both
`/plan-issue` and `/implement`, so a `dispatch` issue with a still-open blocker
that slipped past the queue gate is parked to `dispatch:office-hours` instead of
being planned or built.

## Why

A `dispatch` / `help wanted` issue with an open blocker can still reach a phase
worker when there is a TOCTOU race between the dispatch-ready signal (the labels)
and the `blocked_by` dependency link. The queue gate
(`dispatch-select-target` → `dispatch-trace-leaf` → `count_open_blockers`) is
correct — it simply runs before the `blocked_by` link exists/propagates on the
REST endpoint, or hits read-after-write propagation lag.

Concrete instance: #2135 (blocked by #2039 via PR #2108) had its labels applied
at 05:43:16, the `blocked_by` link added at 05:44:53, and a tick provisioned its
worktree and spawned `/plan-issue` at ~05:46:03 — between the two. It was parked
correctly, but only *incidentally* (the Step 1 drift check happened to notice a
target symbol missing from `main`). There was no blocker re-check at the worker
boundary, so a raced blocked issue whose target code already exists would have
been planned/built against a still-open prerequisite.

## How

Both workers now re-check at startup using the existing `dispatch-check-blockers`
script — which routes through `count_open_blockers`, the same helper the queue
gate uses, so the explicit and queue paths can never diverge on what counts as
"blocked". The capture is tolerant of the expected exit-2 (blocked) case, and
routing is three-way per `.claude/rules/code-style.md` (clear error, never a
swallowed fallback):

- `rc == 0` — no open blocker → proceed unchanged (no behavior change for the
  common case).
- `rc == 2` — `blocked:<nums>` → real open blocker → `dispatch-mark-deviation`
  naming the blocker(s), then stop.
- `rc == 1` (or any other non-zero) — `gh`/API failure → blockers unverified →
  `dispatch-mark-deviation` with a distinct environmental reason including the
  exit code, then stop. Never proceed.

In both stop branches, the absence of a completion marker drives the existing
Stop hook (`dispatch-stop.sh`, Branch A) to apply `dispatch:office-hours`.

Placement:

- **`/plan-issue`** — at the very start of Step 1 (Pre-Planning relevance
  review), before the context-pack call. It runs only on the fresh-run path; the
  `HAVE_PLAN=1` branch re-finalizes and stops before Step 1, so an
  already-planned issue is not re-parked. Running before the expensive
  context-pack / drift work also fails fast on a raced blocked target.
- **`/implement`** — at the end of the idempotency preamble, gated on the
  non-re-entry path (`PR: none`). The re-entry branch already built and passed
  this gate on first run; re-checking there would needlessly re-park a same-tick
  crash recovery. `dispatch-mark-deviation` is already one of `/implement`'s two
  single-named-exit terminal actions, so that invariant still holds.

## Out of scope

`qa-fix`, `review-fix`, `fix-checks` (they run post-PR — the build already
happened); the process-only "apply labels last" backup (does not help REST
propagation lag); any change to `dispatch-check-blockers` or `count_open_blockers`.

## Verification

The plan's `verify` block passed: `bash -n` confirms the reused
`dispatch-check-blockers` still parses, and `grep -q` confirms the re-check
landed in both worker skills (both files had zero references before this change).

### DIFF
--- a/.claude/skills/implement/SKILL.md
+++ b/.claude/skills/implement/SKILL.md
@@ -73,6 +73,49 @@ preamble did NOT take the re-entry branch). (`dispatch-route` normally routes a
 PR-bearing issue to fix-checks/qa/review, not implement, so this is a same-tick
 crash-recovery edge.)
 
+**Open-blocker re-check (non-re-entry path only).** When the preamble finds
+`PR: none` — i.e. this is a fresh run, not a crash recovery — verify that the
+target issue has no open blockers before beginning any build work. This catches
+the race window between the queue-gate check (in `/dispatch-propagate`) and now.
+Skip this check on the re-entry branch (`PR #<num>` already exists): the build
+already happened and passed this gate on first run; re-checking would needlessly
+re-park a same-tick crash recovery.
+
+Run with `dangerouslyDisableSandbox: true` — `dispatch-check-blockers` calls
+`gh` (see `.claude/rules/sandbox.md`). Tolerant capture:
+
+```bash
+if out=$(.claude/skills/dispatch-propagate/scripts/dispatch-check-blockers "$N"); then
+  rc=0
+else
+  rc=$?
+fi
+```
+
+Route on `rc`:
+
+- **`rc == 0`** — no open blocker. Proceed unchanged.
+- **`rc == 2`** — `$out` is `blocked:<nums>`. Real open blocker found. Call
+  `dispatch-mark-deviation`, then stop (skip the Step 5 completion marker).
+  Marker absence triggers Stop hook Branch A (`dispatch:office-hours`).
+  `dispatch-mark-deviation` is already one of `/implement`'s two single-named-exit
+  terminal actions, so the single-named-exit invariant in `## Steps` still holds:
+
+  ```bash
+  .claude/skills/dispatch-propagate/scripts/dispatch-mark-deviation \
+    "/implement: target #$N has an open blocker ($out) - raced past the queue gate; parking"
+  ```
+
+- **`rc == 1`** (or any other non-zero) — environment error; blockers unverified.
+  Not "no blockers." Call `dispatch-mark-deviation` with a distinct reason
+  including `$rc`, then stop (skip the Step 5 completion marker). Never proceed
+  on `rc == 1`:
+
+  ```bash
+  .claude/skills/dispatch-propagate/scripts/dispatch-mark-deviation \
+    "/implement: could not verify open blockers for #$N (dispatch-check-blockers exit $rc)"
+  ```
+
 ## Steps
 
 **Single named exit.** Take exactly one terminal action to end the turn: the
--- a/.claude/skills/plan-issue/SKILL.md
+++ b/.claude/skills/plan-issue/SKILL.md
@@ -111,6 +111,45 @@ counterpart of the creation-time relevance check; the two are deliberately
 separate — the creation-time check is `$BASELINE_BRANCH`-anchored, this step is
 pre-planning and `createdAt`-anchored.
 
+**Open-blocker re-check.** Before spending any planning work, verify that the
+target issue has no open blockers. This catches the race window between the
+queue-gate check (in `/dispatch-propagate`) and now. Run with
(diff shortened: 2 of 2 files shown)


natb1:
## QA summary — issue #2242 / PR #2249

**Disposition: clean pass** — all 9 items satisfied.

### Shell-command lane (items 1–7): PASS

| # | Item | Result |
|---|------|--------|
| 1 | `dispatch-check-blockers` bash -n syntax check | PASS |
| 2 | Open-blocker re-check present in implement/SKILL.md | PASS |
| 3 | Open-blocker re-check present in plan-issue/SKILL.md | PASS |
| 4 | Three-way rc routing (rc=0/2/1) in implement/SKILL.md | PASS |
| 5 | Three-way rc routing (rc=0/2/1) in plan-issue/SKILL.md | PASS |
| 6 | Re-entry guard documented in implement/SKILL.md | PASS |
| 7 | Re-entry guard documented in plan-issue/SKILL.md | PASS |

### Residue (items 8–9): already-satisfied

Two items were initially classified `needs-human-judgment` by the Opus triage subagent and submitted to the disposition Workflow for classification.

The Workflow's adversarial skeptics (4/4 votes refuted) argued that both items are verifiable statically without human input:
- **Item 8** (non-re-entry gating prose): implement/SKILL.md lines 76-80 show the `(non-re-entry path only)` header and an explicit "Skip this check on the re-entry branch" instruction. The gate is present and correctly scoped.
- **Item 9** (behavioral integration — rc routing → office-hours): `test-dispatch-scripts.sh` lines 23968–24015 confirm exit 0 (no blockers), exit 2 (open blockers), and exit 1 (argument error) are all tested for `dispatch-check-blockers`. The behavioral chain is statically traceable.

The Workflow classifier labeled both `opus-fixable` after the refutation, but its fix planner returned 0 units (no code defects). Every other result field establishes `already-satisfied`: `verify_report.verdict = refuted` (4/4), `fix_plan.units = []`, `deviation = false`, `already_satisfied = []` (the mislabel). Treating both as `already-satisfied` per the auditable evidence trail above.

### Subagents

- Skill body: 1 (Opus triage fork)
- Disposition Workflow: 6
- Total: 7

natb1:
## Review — #2242 (PR #2249)

**Security:** no attack surface — docs-only diff (no executable, config, dependency, or Firestore-rules changes).

---

### Fixed

- **`plan-issue/SKILL.md:28`** — `dispatch-check-blockers` was missing from the global sandbox script list at the top of `plan-issue/SKILL.md`, which enumerates scripts requiring `dangerouslyDisableSandbox: true`. Added it. *(commit ae56639e)*
- **`plan-issue/SKILL.md:148`** — The rc==2 and rc==1 open-blocker bullets in `/plan-issue` did not state that `dispatch-mark-deviation` is the established terminal action (Step 6) and that stopping after it preserves the single-named-exit invariant — unlike the matching callout already in `/implement`. Added a sentence to both bullets mirroring the implement skill's wording. *(commit ae56639e)*

---

### Informational

- **`implement/SKILL.md:99`** — The phrasing "then stop (skip the Step 5 completion marker)" in both the rc==2 and rc==1 deviation paths references "Step 5" by number; if `/implement`'s step numbering shifts, this note could mislead. No change required — the wording matches the existing convention in the skill, and the note's intent (marker absence → Stop hook → office-hours) is correct.
- **`implement/SKILL.md:76`** — On rc==1 (any gh/API failure) the issue is immediately parked to `dispatch:office-hours` with no retry. A transient rate-limit or flaky network hit parks a valid issue. This is consistent with the skill's "clear error, never a swallowed fallback" principle and is acceptable; retry logic is out of scope for this change.

---

### Dismissed

- **`plan-issue/SKILL.md:114`** — Reviewer noted the open-blocker re-check block is placed inside the `### 1. Pre-Planning` section header but the trailing summary says it runs before Step 1. The placement is intentional: the block appears at the top of Step 1's body section, before the context-pack call, which is the correct execution order. No structural issue.

PR #2251, opened 2026-06-20
dispatch-stop: guard in-flight background-Workflow phases; add deterministic self-close terminus

natb1:
Closes #2243

## Problem

A Workflow-based phase skill (`/review-fix`, `/qa-fix`) launches its review/QA
fan-out as a **background** Workflow and yields its turn to await a
`<task-notification>`. That mid-phase turn-end fires the Stop hook
(`.claude/hooks/dispatch-stop.sh`) while the phase is still running in the
background and has **not** yet written its `phase-completed` marker. The hook
sees marker-absence, takes Branch A ("phase exited before completion"), parks the
issue on `dispatch:office-hours`, and spawns a tick — but the phase has not
exited. When the Workflow finishes and the session resumes it writes the marker,
but **no second Stop fires**, so the premature `office-hours` label is never
stripped and the session never self-closes. It sits open, parked-as-complete,
indefinitely.

Two coupled defects, fixed together (D1 alone would leave completed sessions
open-but-unparked — a different leak):

## D1 — premature park (Unit 1)

`.claude/hooks/dispatch-stop.sh` gains a new discriminator
`session_has_inflight_background_task`, symmetric to the existing
`session_scheduled_wakeup` guard (#1590). It reads the transcript directly
(positive-evidence-only, return 1 on any uncertainty per
`.claude/rules/shell-json.md`): it collects launched task IDs (`launched in
background. Task ID: <ID>`) and notified task IDs (`<task-notification>`
payloads' `"taskId":"<ID>"`, any status), and returns 0 iff at least one
launched ID has no matching notification. Branch A consults it after the
scheduled-wakeup guard and before the rate-limit-death check: an in-flight
background task hands back (`exit 0`, no park, no spawn) instead of parking.

## D2 — deterministic self-close terminus (Units 2 + 3)

New composite script
`.claude/skills/dispatch-propagate/scripts/dispatch-finalize-phase <issue>
[--pr <pr>]` runs the terminal disposition deterministically rather than
relying on a second Stop the harness does not fire: it clears
rate-limit-retry labels, strips `dispatch:office-hours` from **both** the issue
and the PR, spawns the next tick and the sweep, then self-closes
(`dispatch-self-close`; `exec claude rm`; interactive no-op). Every step is
best-effort and idempotent, so a stray second Stop is harmless. It is wired as
the **absolute last action** of the `/review-fix` and `/qa-fix` success paths,
after `dispatch-mark-complete` and the outcome-envelope emit (self-close
terminates telemetry, so it must run last). The deviation and re-entry paths are
untouched — they keep relying on the existing Stop disposition.

## Tests (Unit 4)

`test-dispatch-scripts.sh` gains: D1 in-flight-Workflow hand-back (no park, no
spawn); completed-Workflow falls through to the normal Branch A park; multiple
tasks with one outstanding hands back; and a D2 finalize test asserting
office-hours removed from issue **and** PR, spawn-once, sweep + self-close
present, and self-close last. Existing scheduled-wakeup hand-back and
genuine-mid-phase-death park tests stay green. Full suite: 3620/3620.

### DIFF
--- a/.claude/hooks/dispatch-stop.sh
+++ b/.claude/hooks/dispatch-stop.sh
@@ -391,6 +391,52 @@ session_scheduled_wakeup() {
     | tail -1 | grep -q '"ScheduleWakeup"'
 }
 
+# Branch A background-task discriminator (#2243). Returns 0 ("still running in
+# the background, will resume on its own") ONLY when the transcript shows at
+# least one background task that was LAUNCHED but has not yet been NOTIFIED back
+# — i.e. an in-flight launched∖notified set difference is non-empty; returns 1
+# in EVERY other case — including all uncertainty: empty TRANSCRIPT_PATH,
+# missing/unreadable file, malformed transcript, or no launches at all. The
+# fail-safe direction (positive evidence only) is load-bearing: any doubt falls
+# through to today's office-hours park, so a genuine death still parks.
+#
+# A Workflow-based phase skill (/review-fix, /qa-fix) launches its review/QA
+# fan-out as a BACKGROUND Workflow and yields its turn to await a
+# <task-notification>. That mid-phase turn-end fires this Stop hook while the
+# phase is still running and has not yet written its phase-completed marker.
+# Without this gate Branch A sees marker-absence and prematurely parks the
+# issue on office-hours (#2243) — symmetric to session_scheduled_wakeup (#1590).
+#
+# Transcript shapes (whole-transcript grain, NOT last-turn-only — a background
+# launch is not a per-poll-cycle event):
+#   Launch — a tool_result text `... launched in background. Task ID: <ID>`
+#            (Workflow: `Workflow launched in background. Task ID: <ID>`; the
+#            Task tool uses analogous `... launched in background. Task ID: <ID>`).
+#   Notify — a <task-notification> payload carrying `"taskId":"<ID>"` for ANY
+#            status; any notification means the task resumed the session and is
+#            no longer in-flight.
+#
+# Grep the file directly for literal substrings (not echo-into-jq): the IDs are
+# plain substrings, so grep is robust and sidesteps the control-char trap
+# (.claude/rules/shell-json.md), mirroring dispatch-recover-dispatched-phase /
+# dispatch-detect-rate-limit-death.
+session_has_inflight_background_task() {
+  [ -n "$TRANSCRIPT_PATH" ] && [ -r "$TRANSCRIPT_PATH" ] || return 1
+  local launched notified
+  launched=$(grep -oE 'launched in background\. Task ID: [A-Za-z0-9_-]+' "$TRANSCRIPT_PATH" 2>/dev/null \
+    | grep -oE '[A-Za-z0-9_-]+$' | sort -u)
+  [ -n "$launched" ] || return 1
+  # Tolerate the JSONL backslash-escaped quote form (\"taskId\":\"<ID>\") as
+  # well as the bare "taskId":"<ID>" form — the <task-notification> payload is a
+  # string value, so its inner quotes are backslash-escaped in the record.
+  notified=$(grep -oE 'taskId\\?":\\?"[A-Za-z0-9_-]+' "$TRANSCRIPT_PATH" 2>/dev/null \
+    | grep -oE '[A-Za-z0-9_-]+$' | sort -u)
+  # In-flight iff at least one launched ID has no matching notification, i.e.
+  # the set difference (launched ∖ notified) is non-empty. comm -23 lists lines
+  # unique to the first (sorted) input.
+  [ -n "$(comm -23 <(printf '%s\n' "$launched") <(printf '%s\n' "$notified"))" ]
+}
+
 # Branch P — parse-job clean completion (#1024): a /budget-parse-job session is
 # named <N>-slug like a worker but, on a successful idempotent statement merge,
 # writes a `parse-job-done` sentinel instead of a phase-completed marker. A clean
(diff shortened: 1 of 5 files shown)


natb1:
# QA Summary — Issue #2243 / PR #2251

## Results

**6 script-verifiable items: all PASS. 1 needs-human-judgment item: deferred to office-hours.**

| # | Title | Result |
|---|-------|--------|
| 1 | Test suite verifies new #2243 behavior (D1 hand-back, D2 finalize) | PASS — 3620/3620 |
| 2 | `session_has_inflight_background_task` function exists in dispatch-stop.sh | PASS |
| 3 | Branch A invokes the discriminator (hand-back call site present) | PASS |
| 4 | `dispatch-finalize-phase` exists and is executable | PASS |
| 5 | review-fix success path wires `dispatch-finalize-phase --pr "$PR_NUM"` | PASS |
| 6 | qa-fix success path wires `dispatch-finalize-phase --pr "$PR_NUM"` | PASS |
| 7 | End-to-end lifecycle judgment: park-during-flight then self-close-at-completion with no leak | **Deferred to office-hours** (needs-human-judgment) |

**PASS: 6 / FAIL: 0 / SKIP: 0 / Deferred: 1**

## Deferred to office-hours

**Item 7 — End-to-end lifecycle judgment: park-during-flight then self-close-at-completion with no leak**

The script-level checks (items 1–6) verify the individual parts: the discriminator function exists and is invoked, dispatch-finalize-phase exists and is wired into both skill termini with `--pr`, and the test suite passes (3620/3620 including D1 hand-back and D2 finalize assertions). However, the inter-component chain — hand-back keeping the session open while the background task runs, and then finalize deterministically running after marker write — is an emergent lifecycle property that only a live dispatch run (or a reviewer reading the code end-to-end) can fully confirm. No script check can observe the live timing (no premature park, no race at completion).

## Disposition triage (Opus-assessed)

| 7 | `needs-human` | Upheld | The lifecycle correctness claim spans dispatch-stop.sh, dispatch-finalize-phase, and both skill termini. Unit tests demonstrate D1 and D2 with explicit assertions, but the live timing (no premature park during flight, deterministic finalization after marker write) is an emergent property that only a live run or end-to-end code review can confirm. |

No `needs-main` items — Step 3.6 did not run.
No `already-satisfied` items.

## Terminal disposition

Residue (item 7, `needs-human`) escalates to office-hours. The `/office-hours` walkthrough will confirm lifecycle correctness in a live dispatch run.

natb1:
## Review — PR #2251 (#2243)

**Surface:** `code` (bash hooks/scripts + skill docs) · **No fixes applied** · **2 follow-up issues filed**

---

### Fixed
_None._

### Required (security)
_None._

### Informational

- **CR-1** `.claude/hooks/dispatch-stop.sh:437` — `comm -23` uses process substitution on `printf '%s\n' "$launched"` / `printf '%s\n' "$notified"` (already `sort -u`d by the grep pipeline). Correct in practice since `grep ... | sort -u` produces POSIX-sorted output; worth verifying locale consistency if this ever runs in a locale-sensitive environment.

- **CR-4** `test-dispatch-scripts.sh:22887+` — test stub paths share `TMPDIR_TEST`; safe for sequential execution (the current pattern). Would need per-test sub-directories if tests were ever parallelised.

- **REV-1** `.claude/hooks/dispatch-stop.sh:391-437` — `session_has_inflight_background_task` scans the whole transcript, not just the current session's turn. A re-used task ID from a prior session run could theoretically match. The conservative `return 1` on uncertainty means false positives result in a missed hand-back (not a wrong park), so safety holds. Filed for tracking: #2261.

- **REV-2** `.claude/hooks/dispatch-stop.sh:420-422` — the literal pattern `launched in background. Task ID:` is coupled to the Workflow tool's output format; a format change would silently regress detection. Currently documented inline; a future hardening could add a version-keyed comment or test.

- **REV-3** `.claude/skills/dispatch-propagate/scripts/dispatch-finalize-phase:84-87` — tick-before-sweep ordering is correct but implicit. Filed for tracking (add an ordering comment): #2262.

### Dismissed

- **CR-2** `dispatch-finalize-phase:73` — `trap ... EXIT` in a subshell won't fire on the outer EXIT. The outer `|| true` guard makes this safe; pattern is consistent with the codebase. (_Not applicable — already guarded._)

- **CR-3** `dispatch-finalize-phase:68-73` — missing exit-status check on `gh --jq` pipeline in the rate-limit-label loop. The surrounding `|| true` block guards all failures. (_Already guarded._)

- **CR-5** `dispatch-finalize-phase:14` — `SCRIPT_DIR` resolution via `cd "$(dirname ...)" && pwd` doesn't guard against symlinks. Pattern is idiomatic and consistent across all sibling scripts in this project. (_Consistent with project conventions._)

- **SEC-1** `dispatch-stop.sh:416-424` — `TRANSCRIPT_PATH` used in grep without path validation. Already validated with existence + readability checks at line 415 before any use. (_Already validated._)

### False-positive (security)

- **SEC-2** `dispatch-finalize-phase:68-73` — rate-limit label names from `gh --jq` interpolated into `--remove-label`. Labels are server-controlled values from the repo's own label set, not user-supplied input. (_No injection risk._)

- **SEC-3** All changed files — no new secrets surface, no new attack surface beyond existing dispatch hook patterns.

### Deferred (filed as follow-ups)

- **REV-1** → #2261 `dispatch-stop.sh: D1 whole-transcript scope for in-flight task detection may miss cross-session ID collisions`
- **REV-3** → #2262 `dispatch-finalize-phase: add ordering comment that spawn-tick precedes spawn-sweep to document chain-re-establishment intent`

---

_Review by `/review-fix` · 13 findings surfaced · 0 actionable · 0 fixes applied · 2 follow-ups filed_

PR #2286, opened 2026-06-20
office-hours: attach to stopped/paused and resume removed originating sessions

### LINKED ISSUE #2240: office-hours: attach to stopped/paused originating sessions (gate only working sessions)
## Context

`office-hours-select-target`'s `issue_live_session_id` only treats an originating
office-hours session as attachable when its state is `waiting`, `done`, or
`idle`. `stopped`, `paused`, and any unrecognized state are excluded from attach
and handled like `working` — the issue is skipped.

The selector's own header documents the resulting wedge:

> Stopped/paused (and any unrecognized state) are excluded from attach — treated
> like working (skip the issue). Known consequence: such a session still
> name-matches in `worktree_has_live_session`, so its issue won't fresh-launch
> either and can wedge until the session is cleaned up. Handling that wedge is
> out of scope here ...

So an office-hours item whose originating session is merely `stopped` or
`paused` neither attaches (a human can't resume it) nor fresh-launches — it sits
stuck until the session is manually reaped.

There is a second, adjacent gap on the same surface: once the originating
session has actually been **removed** from the local daemon registry — `claude
agents rm`, a reap, or any cleanup that drops it from `claude agents --json
--all` — the selector finds no name-match, reports the item sessionless, and
fresh-launches. That throws away the removed session's accumulated context even
when its **transcript survives on disk** and could be resumed in place. The
sibling #2241 explicitly punts this case ("A session absent from that registry
... is **not local** → fall back to a fresh /office-hours"); this issue brings it
into scope.

## Goal

Two changes to the office-hours resume path, both narrowing the set of cases that
needlessly fresh-launch over recoverable context:

1. **Attach to `stopped`/`paused` originating sessions**, gating **only**
   genuinely-`working` (busy) sessions. A human resuming the queue should be able
   to re-engage a stopped/paused office-hours worker in place instead of the item
   wedging. A `working` session must still be skipped — attaching to or
   fresh-launching over an autonomously-running session would be a double-claim.

2. **Resume a removed ("rm'd") originating session from its transcript** instead
   of fresh-launching, when the session is absent from `claude agents --json
   --all` but its transcript is still recoverable. Resume keys on the sessionId
   via `claude --bg --resume <sessionId>` (the mechanism `dispatch-resume-worker`
   already uses to revive dead phase-worker sessions) — **not** `claude attach`,
   which resolves a live daemon job id and cannot reach a session that has left
   the registry. When no recoverable transcript exists, fall back to the current
   fresh-launch behavior with a clear diagnostic — no silent failure.

### Resuming rm'd sessions — mechanics the plan must resolve

- **Discovery.** A removed session is not in `claude agents --json --all`, so the
  current registry name-probe cannot find it. The plan must establish how the
  removed session's sessionId is recovered for a given issue — the candidate
  carriers are the reservation-ledger markers and `dispatch-stamp-session`, which
  already record the sessionId↔issue association, and/or the on-disk transcript
  store. Whichever is used, the lookup must be issue-scoped and must not
  misattribute another issue's session.

- **Resume vs attach.** The `office-hours` entry script attaches via `claude
  attach <job-id>` today, which requires a live background job. A removed session
  has no live job, so the rm'd path must branch to `claude --bg --resume
  <sessionId>` (then attach to the freshly-resumed job), following the
  `dispatch-resume-worker` precedent. Reuse that script's dedup/verify shape
  rather than reinventing it where practical.

- **Transcript-recoverability is a precondition, not an assumption.** Treat
  "removed but resumable" as conditional on the transcript actually being
  recoverable: if the transcript was purged with the session (or cannot be
  located), this path does **not** apply and the item fresh-launches as today.
  The plan must **verify** whether `claude agents rm` / the reap path preserves
  the transcript; if removal purges it, reframe the requirement as
  "exited/reaped-but-recoverable" and scope accordingly.

- **`working` still wins the skip.** None of this changes the rule that a
  genuinely-`working` session is never attached, resumed, or launched over.

### Stopped/paused (in-registry)

- [ ] `issue_live_session_id` (in
  `.claude/skills/dispatch-propagate/scripts/office-hours-select-target`) treats
  `stopped` and `paused` as attachable, emitting `idle <sessionId>` alongside
  the existing `waiting`/`done`/`idle`.
- [ ] Only `working`/`busy` sessions are gated (skip — neither attach nor
  fresh-launch).
- [ ] Unrecognized/empty state behavior is explicitly decided and documented
  (keep the fail-safe skip, or classify it) rather than left implicit.
- [ ] The interaction with the shared `worktree_has_live_session` /
  `claude_agents_list_all` occupancy helpers is handled so that widening the
  attachable set does **not** cause `done`/`stopped`/`paused` sessions to newly
  "occupy" a worktree on the dispatch-router hot path and block phase-worker
  launches — the constraint the selector header calls out.

### Removed / rm'd (registry-absent, transcript-recoverable)

- [ ] When an office-hours item's originating session is **absent** from `claude
  agents --json --all` but a recoverable transcript for it exists, the selector
  surfaces a disposition that resumes that session (rather than reporting the
  item sessionless and fresh-launching).
- [ ] Resume uses `claude --bg --resume <sessionId>` (the `dispatch-resume-worker`
  mechanism), **not** `claude attach`, since the removed session has no live
  daemon job.
- [ ] The removed session's sessionId is recovered by an issue-scoped lookup
  (reservation-ledger / `dispatch-stamp-session` / transcript store — chosen in
  the plan) that cannot misattribute another issue's session.
- [ ] When no recoverable transcript exists, behavior is unchanged: fresh-launch
  a new `/office-hours`, with a clear diagnostic — no silent failure.
- [ ] A `working` originating session is still skipped (no double-claim), and the
  removed-session path never resumes over a session that is actually live.

### Tests

- [ ] `stopped` → attach, `paused` → attach, `working` → skip.
- [ ] removed-with-recoverable-transcript → resume, removed-without-transcript →
  fresh-launch fallback.
- [ ] The dispatch router's worktree-occupancy decisions
  (`worktree_has_live_session`) are unchanged for `stopped`/`paused`/removed
  sessions.

## Relationship

Shares the `issue_live_session_id` surface with #2241 (originating-session attach
+ worktree-provisioning for `done`/`idle`/`waiting`).

**Boundary / contradiction to resolve:** #2241 carries the criterion "When the
originating session is absent from `claude agents --json --all` ('not local'),
behavior is unchanged: spawn a fresh /office-hours." The rm'd-resume scope added
to this issue **overrides that criterion for the resumable-transcript subcase**:
registry-absent **and** transcript-recoverable → resume; registry-absent **and**
not recoverable → fresh (the `2241` behavior). #2241's criterion needs a
corresponding narrowing to "not local *and not resumable-from-transcript* →
fresh". That narrowing is flagged on #2241 and is out of scope for this issue's
own edit.

natb1:
Closes #2240

## Summary

Closes two gaps in the office-hours queue's session re-engagement, both surfaced
after the sibling #2241/#2276 work that added originating-session attach +
worktree provisioning for `done`/`idle`/`waiting`:

1. **`stopped`/`paused` sessions no longer wedge the queue.** They were funneled
   to skip — name-matching `worktree_has_live_session` (so never fresh-launched)
   yet excluded from attach — leaving the issue stuck until the session was
   manually reaped. They are now **attachable**: a human resuming the queue
   re-engages the originating session in place.

2. **A removed ("rm'd") originating session resumes instead of fresh-launching.**
   When the session is absent from `claude agents --json --all` but its on-disk
   transcript survives, the selector now emits a `resume`/`resume-provision`
   disposition and the entry script resumes from the transcript via
   `--bg --resume`, preserving accumulated context, rather than discarding it
   with a fresh launch.

A genuinely-`working` session is still never attached, resumed, or
launched-over (no double-claim); unrecognized/empty state keeps its
conservative fail-safe skip.

## What changed

- **Part 1 — `office-hours-select-target` (Unit 1):** widened the `_inspect`
  `case "$state"` attachable arm to `waiting|done|idle|stopped|paused`, split the
  gated `working|busy` and fail-safe `*)` arms so each skip's reason is explicit,
  and rewrote the docstrings (state→action table, the now-resolved wedge note,
  the rc-3 contract). No change to the main loop, the idle classifier, the entry
  script, or any lib helper.

- **Part 2 — `dispatch-recover-session-id` (Unit 2):** a new pure-filesystem
  reader. `--issue <N>` → `sessionId<TAB>branch` (rc 0), resolved from the
  session's stamp sidecar (`<sessionId>.dispatch-stamp.json`), filtered exactly to
  `.issue == N`, sorted newest-first, returning the first candidate whose
  `<sessionId>.jsonl` transcript **actually exists** (newest-existing, not
  latest-then-verify). rc 1 with no stdout is the legitimate "nothing
  recoverable" signal. No daemon calls — sandbox-safe and unit-testable.

- **Part 2 — selector resume disposition (Unit 3):** the selector captures the
  oldest-first sessionless issue set during its main loop and, **only when no
  attachable idle session was found** (live-idle > resume > fresh), walks it
  calling `dispatch-recover-session-id`. The first recoverable issue classifies
  symmetric with the idle classifier: worktree on disk → `resume <N> <sid> <cwd>`;
  worktree swept + `origin/<branch>` exists → `resume-provision <N> <sid>
  <branch>`; swept + no remote → existing fresh path.

- **Part 2 — entry-script resume arms (Unit 4):** `office-hours` gains `resume`
  and `resume-provision` arms. A new inline `resume_session` mirrors
  `dispatch-resume-worker`'s mechanism (it cannot call that script — its
  worker-only `^[0-9]+-` name gate rejects the `office-hours-<N>` name): a
  `--bg --resume` kick named `office-hours-<N>`, **gated on
  `verify_agent_registered_under`** (a clean kick rc alone is insufficient —
  `claude --bg` returns before the daemon registers, and a stale/colliding id is
  rejected asynchronously), with a `--fork-session` retry on failure. Because
  `--fork-session` mints a new sessionId, attach resolves **by name**
  (`office-hours-<N>`), never by the selector's now-stale sessionId.

## Router hot path is byte-identical by construction

The shared occupancy helpers `worktree_has_live_session` /
`claude_agents_list_all` match on the **name** column and never read `.state`.
This PR touches only `issue_live_session_id`'s `case` arm, a new pure-filesystem
reader, and the entry script — and never edits `lib-claude-agents.sh` — so
dispatch-router occupancy for done/stopped/paused/removed cannot change. A
`worktree_has_live_session` regression guard (asserting a `stopped`/`paused`
name-matching session still reports the worktree occupied) pins this.

## Verification

All logic lives in shell scripts exercised by the dispatch test suite, which CI
runs directly with fakes (no real `claude` daemon, no `gh`) — so it runs
sandboxed. The suite passes (**3779/3779**) and covers, per the units above:
stopped/paused → attach, paused → attach, unrecognized → fail-safe skip, working
still skips, router occupancy unchanged; the `dispatch-recover-session-id`
filesystem cases; selector removed-with-transcript → `resume`, removed-without →
fresh, idle-beats-resume priority, swept+remote → `resume-provision`; and the
entry-script resume argv, the verify-fails-then-fork path (the real async-reject
mode), and attach-by-name.

### Live-observation items (the fake suite cannot prove these)

The fake daemon emits a job id for every state and the resume tests only capture
argv shape, so a green suite proves selector logic and argv shape — not live
re-engagement. These remain QA / live items:

- Real `claude attach` re-engaging a genuinely **stopped/paused** session. Named
  contingency if it cannot: route stopped/paused through Part 2's `--bg --resume`
  mechanism (already present after this PR) — no selector re-architecture needed.
- A real **removed** session resuming from a real transcript via `--bg --resume`
  (and the `--fork-session` fallback engaging on a dead-id collision).
- The daemon's real **state vocabulary** emitting literally `stopped`/`paused`.
- `SessionStart:startup` firing for detached `--bg` `office-hours-<N>` sessions so
  the stamp sidecar is reliably written.
- The **`resume-provision` arm's landing-path resolution**
  (`resolve_project_root`/worktrees/`<branch>`) is verified only by live
  observation, not the suite — the green test count does not cover it.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-recover-session-id
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-recover-session-id
@@ -1,0 +1,119 @@
+#!/usr/bin/env bash
+# dispatch-recover-session-id — find the newest recoverable session for an issue.
+#
+# Usage: dispatch-recover-session-id --issue <N>
+#
+# Globs the dispatch-stamp sidecar files under
+# ${DISPATCH_STAMP_PROJECTS_ROOT:-${HOME}/.claude/projects}/*/*.dispatch-stamp.json,
+# filters to sidecars whose .issue field equals N, sorts by .stamped_at descending
+# (file mtime as a secondary tiebreak), then returns the FIRST candidate whose
+# sibling <sessionId>.jsonl transcript still exists.
+#
+# On success: prints "<sessionId>\t<branch>" to stdout and exits 0.
+# When no recoverable session exists: exits 1, no stdout.
+# On usage/environment errors: exits 2 with a diagnostic on stderr.
+#
+# Pure filesystem — never touches the Claude daemon (no `claude agents`), so it
+# is sandbox-safe and unit-testable without a running daemon.
+#
+# Test seam: $DISPATCH_STAMP_PROJECTS_ROOT overrides the projects root
+# (default ${HOME}/.claude/projects).
+set -euo pipefail
+
+ISSUE=""
+
+usage() {
+  echo "usage: dispatch-recover-session-id --issue <N>" >&2
+}
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --issue)
+      if [[ $# -lt 2 ]]; then
+        echo "dispatch-recover-session-id: --issue requires a value" >&2
+        usage
+        exit 2
+      fi
+      ISSUE="$2"
+      shift 2
+      ;;
+    *)
+      echo "dispatch-recover-session-id: unexpected argument: $1" >&2
+      usage
+      exit 2
+      ;;
+  esac
+done
+
+if [[ -z "$ISSUE" ]]; then
+  echo "dispatch-recover-session-id: --issue is required" >&2
+  usage
+  exit 2
+fi
+
+if [[ ! "$ISSUE" =~ ^[0-9]+$ ]]; then
+  echo "dispatch-recover-session-id: --issue must be a non-negative integer, got: $ISSUE" >&2
+  usage
+  exit 2
+fi
+
+PROJECTS_ROOT="${DISPATCH_STAMP_PROJECTS_ROOT:-${HOME:-}/.claude/projects}"
+if [[ -z "$PROJECTS_ROOT" || ! -d "$PROJECTS_ROOT" ]]; then
+  echo "dispatch-recover-session-id: projects root '$PROJECTS_ROOT' is not a directory" >&2
+  exit 2
+fi
+
+# Collect all matching sidecars: emit "<stamped_at>\t<mtime_epoch>\t<sidecar_path>"
+# for each sidecar whose .issue == N (exact numeric match), then sort descending
+# by stamped_at (primary) and mtime (secondary).
+#
+# We use `jq -r` directly on the file (never echo "$VAR" | jq) to avoid
+# control-character corruption per shell-json.md.  Sidecars that fail to parse
+# or lack a matching .issue are skipped silently — a corrupt/alien sidecar
+# should not abort recovery.
+
+candidates=()
+while IFS= read -r sidecar; do
+  # Read .issue from the file; skip if jq fails or field is missing/wrong type.
+  sidecar_issue=$(jq -r '.issue // empty' "$sidecar" 2>/dev/null) || continue
+  # Exact integer match — do not allow cross-issue attribution.
+  [[ "$sidecar_issue" == "$ISSUE" ]] || continue
+
+  stamped_at=$(jq -r '.stamped_at // empty' "$sidecar" 2>/dev/null) || continue
+  [[ -n "$stamped_at" ]] || continue
+
+  mtime=$(stat -c '%Y' "$sidecar" 2>/dev/null) || continue
+  candidates+=("${stamped_at}	${mtime}	${sidecar}")
+done < <(find "$PROJECTS_ROOT" -maxdepth 2 -name "*.dispatch-stamp.json" 2>/dev/null)
+
+if [[ ${#candidates[@]} -eq 0 ]]; then
+  exit 1
+fi
+
+# Sort descending: primary key = stamped_at (ISO-8601 sorts lexicographically),
+# secondary key = mtime (numeric, descending).  Both fields are unique-enough
+# that a stable tiebreak matters only in adversarial fixtures; for real usage
+# stamped_at is the authoritative ordering.
+#
+# sort -t$'\t' -k1,1r -k2,2rn:
+#   -k1,1r  → column 1 (stamped_at) reverse lexicographic
+#   -k2,2rn → column 2 (mtime) reverse numeric
+sorted=$(printf '%s\n' "${candidates[@]}" | sort -t$'\t' -k1,1r -k2,2rn)
+
+# Iterate sorted candidates; return the first whose transcript (.jsonl) exists.
+while IFS=$'\t' read -r _stamped _mtime sidecar; do
+  sidecar_dir="$(dirname "$sidecar")"
+  session_id=$(jq -r '.session_id // empty' "$sidecar" 2>/dev/null) || continue
+  [[ -n "$session_id" ]] || continue
+
+  transcript="${sidecar_dir}/${session_id}.jsonl"
+  if [[ -f "$transcript" ]]; then
+    branch=$(jq -r '.branch // empty' "$sidecar" 2>/dev/null) || continue
+    [[ -n "$branch" ]] || continue
+    printf '%s\t%s\n' "$session_id" "$branch"
+    exit 0
+  fi
+done <<<"$sorted"
+
+# No candidate had a surviving transcript.
+exit 1
--- a/.claude/skills/dispatch-propagate/scripts/office-hours
+++ b/.claude/skills/dispatch-propagate/scripts/office-hours
@@ -57,6 +57,21 @@
 #                                   — the selector already confirmed the remote
 #                                   exists, so a failure here is a genuine error
 #                                   and a silent fresh-spawn would mask it.
+#   resume <N> <sessionId> <cwd>
+#                                 → the originating session was REMOVED from the
+#                                   daemon registry but its on-disk transcript is
+#                                   recoverable and its <N>-* worktree is present
+#                                   (#2240). Resume it in place as a --bg job named
+#                                   office-hours-<N> rooted at <cwd> (--resume, with
+#                                   a --fork-session retry if the kick fails on a
+#                                   stale/colliding id), then attach BY NAME — a
+#                                   forked retry mints a new sessionId, so the
+#                                   verb's <sessionId> is stale for attach.
+#   resume-provision <N> <sessionId> <branch>
+#                                 → as resume, but the worktree was swept while
+#                                   origin/<branch> still exists: provision it from
+#                                   the remote first (dispatch-provision-from-remote),
+#                                   then resume in the freshly-landed worktree.
 #   empty                         → print a queue-empty message and exit WITHOUT
 #                                   launching Claude. The empty queue is handled
 #                                   entirely in bash; no session is booted.
@@ -72,6 +87,15 @@
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
+# resolve_project_root (resume-provision worktree path) lives in lib.sh;
+# verify_agent_registered_under (the resume arms' post-kick registration check)
+# lives in lib-claude-agents.sh — the same two libs the selector sources. The
+# thin-dispatcher daemon-query note above still holds for the idle/fresh paths.
+# shellcheck source=/dev/null
+source "$SCRIPT_DIR/lib.sh"
+# shellcheck source=/dev/null
+source "$SCRIPT_DIR/lib-claude-agents.sh"
 
 # Launch target only. The selector reads CLAUDE_AGENTS_CMD from the environment
 # for its own liveness queries (default `claude`); this script does not set it.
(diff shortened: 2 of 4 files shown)


natb1:

natb1:
# QA Summary — Issue #2240 / PR #2286

**Pass:** 5  **Fail:** 1 (fixed this pass)  **Skip:** 0

## Items executed

| # | Title | Result |
|---|---|---|
| 1 | Run the full dispatch test suite | PASS (3779/3779) |
| 2 | stopped/paused → attach; working/busy → skip; * → fail-safe skip (code inspection) | PASS |
| 3 | dispatch-recover-session-id interface (executable, --issue, sorted, .jsonl check) | PASS |
| 4 | resume/resume-provision arms in office-hours + resume_session / attach_session_by_name | PASS |
| 5 | lib-claude-agents.sh unchanged — router occupancy invariant | PASS (empty diff) |
| 6 | Accidental empty `if` file at repo root | FAIL → **FIXED this pass** |
| 7 | Live-observation items (stopped/paused attach, removed-session resume, daemon state vocabulary, stamp sidecar) | needs-human-judgment → filed as #2288 (needs-main) |

## Bugs found

**Item 6 — empty `if` file at repo root (FIXED):**
An empty 0-byte file named `if` was accidentally committed at the repo root during the implementation (created Jun 20 18:18, confirmed tracked via `git ls-files if`). Removed via `git rm if` and committed this pass.

## Disposition triage

**needs-main items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR.**

| id | Title | Class | Verify |
|---|---|---|---|
| 6 | Empty `if` file at repo root | opus-fixable | Upheld |
| 7 | Live-observation items (daemon state vocabulary, real session lifecycle, stamp sidecar writing) | needs-main (planned-deferral) | n/a |

**Fixed this pass:** item 6 (the accidental `if` file) — 1 fix applied.
**Filed as follow-ups:** item 7 → #2288 (needs-main, blocked_by #2240).
**Deferred to re-QA:** none (no needs-human residue remains after this pass).

The fix commit (remove `if`) restarts CI. The chain re-QAs once CI passes; with no remaining opus-fixable or needs-human items, the next QA pass should be a clean pass.

## Filed follow-ups

- Item 7 → #2288 (newly filed: `qa: verify against main — qa-needs-main #2240: live-observation-items-daemon-state-vocabulary-real-session-lifecycle-stamp-sidecar-writing`)

None.

natb1:
# QA Summary — Issue #2240 / PR #2286 (attempt 1)

**Pass:** 8  **Fail:** 0  **Skip:** 0

## Items executed

| # | Title | Result |
|---|---|---|
| 1 | Re-run the full dispatch test suite after the fix commit | PASS (3779/3779) |
| 2 | Confirm the accidental empty `if` file is gone from repo root | PASS (absent) |
| 3 | Confirm stopped/paused in attachable arm of office-hours-select-target | PASS |
| 4 | Confirm working/busy sessions remain gated (separate skip arm) | PASS |
| 5 | lib-claude-agents.sh unchanged — router occupancy invariant | PASS (empty diff) |
| 6 | dispatch-recover-session-id is present and executable | PASS |
| 7 | resume/resume-provision dispatch arms present in office-hours entry script | PASS |
| 8 | verify_agent_registered_under gate and --fork-session retry present | PASS |

## Bugs found

None.

## Summary

All acceptance criteria verified by the dispatch test suite (3779/3779) and targeted code inspections:

- `stopped` and `paused` sessions are now attachable (line 235 of `office-hours-select-target`): `waiting|done|idle|stopped|paused` in the inner `_inspect` case arm.
- `working|busy` is a distinct skip arm (line 243), preventing double-claim.
- `lib-claude-agents.sh` is byte-identical to `origin/main` — the dispatch router hot path is unchanged.
- `dispatch-recover-session-id` (new pure-filesystem reader) is executable.
- `office-hours` has `resume` and `resume-provision` arms, a `resume_session` helper, and gating via `verify_agent_registered_under` with a `--fork-session` retry.
- The prior pass's fix (removing the accidental empty `if` file) is durable — the file is absent.
- Live-observation items (a)-(e) were already filed as needs-main follow-up #2288 in attempt 0; they are not re-included here.

## Filed follow-ups

None this pass. Follow-up #2288 was filed in attempt 0 (needs-main: live-observation items).

natb1:
## Review — #2240 (PR #2286)

**Surface:** `code` — 18 findings (0 actionable). No code changes. No follow-ups.

---

### Fixed
_None._

### Required (security)
_None._

### Informational

- **`office-hours-select-target:382` — Resume loop `break` skips later sessionless siblings** (`code-review`, `review`): When the first recoverable issue has its worktree swept and no remote branch exists, the loop `break`s unconditionally, leaving any later sessionless siblings unprocessed for the resume pass. They fall through to `FRESH_NUM` (the oldest sessionless item), so no issue is lost — the `break` just ends the resume scan and hands off to the fresh path. The behavior is consistent with the "first-recoverable-wins" design and is called out in the comments.

- **`office-hours-select-target:365` — `|| continue` masks exit-code-2 environment errors from `dispatch-recover-session-id`** (`code-review`): A usage/environment error (e.g., `PROJECTS_ROOT` not a directory, missing `--issue`) exits 2 with a diagnostic on stderr. The `|| continue` in the for-loop treats rc 1 (nothing recoverable) and rc 2 (environment error) identically — silently skips. The stderr diagnostic is still emitted, so the failure is visible in the session log; the skip is conservative and safe.

- **`test-dispatch-scripts.sh:3591` — `office_hours_state_fake_claude` sets `status:null` for non-working states** (`code-review`): The helper sets `status:null` for `stopped`, `paused`, and other states. In production, `claude agents --json` emits the state as both `.state` and `.status`; the tests stub only `.status`, so `.state` is absent from the stub JSON. The selector reads `.state // .status` (or similar), so the test coverage is accurate for the current code — but a future reader adding `.state`-specific logic would need to revisit the stub shape.

- **`office-hours-select-target:363-382` — Resume loop `break` on first swept-no-remote (duplicate of code-review-0)** (`review`): Same observation as code-review-0 above; surfaced independently by the review agent. The `break` means only one candidate is evaluated per sweep-and-no-remote case. The fresh fallback (`FRESH_NUM`) still covers all sessionless items, so correctness is maintained.

- **`office-hours-select-target:368-370` — `rec_cwd` path from associative array could contain whitespace** (`review`): `rec_cwd=$(printf '%s' "${WORKTREE_PATHS_BY_NUM[$rec_num]:-}" | head -n 1)` is then emitted in a `printf 'resume %s %s %s\n'` — if the path contains spaces, the caller's `read -r verb a b c d` would misparse the output. Worktrees are under a controlled path without spaces in practice; the risk is theoretical but worth noting if the base path ever changes.

- **`dispatch-recover-session-id:104-121` — Sidecar re-read per candidate in the transcript-check loop** (`review`): The final loop re-reads `.session_id` and `.branch` via `jq` for each sorted candidate, even though both were read earlier during the candidate-collection phase. For the typical 1–2 session case this is negligible; at large scale it adds per-candidate `jq` invocations. Informational only.

- **`office-hours:258-263` — `resume-provision` landing path tested only by live observation** (`review`): The `$(resolve_project_root)/worktrees/$c` path construction is not covered by the dispatch suite (called out in the PR body). Green test count does not validate this path. Noted in the PR's live-observation items.

---

### Dismissed

- **`dispatch-recover-session-id:80` — Integer string-equality comparison** (`code-review`): `[[ "$sidecar_issue" == "$ISSUE" ]]` compares the jq-extracted `.issue` value (string) against `$ISSUE` (already validated as `^[0-9]+$`). A float like `42.0` would not match `42`, which is the desired exact-numeric behavior. Not a defect.

- **`office-hours-select-target:363` — `${SESSIONLESS_NUMS[@]:-}` empty-array expansion** (`review`): Under `set -euo pipefail`, iterating an empty array with `for rec_num in "${SESSIONLESS_NUMS[@]:-}"` passes a single empty-string element in bash. The `[[ -z "$rec_num" ]] && continue` guard at line 364 filters it immediately. The guard is present and correct.

- **`office-hours:218-219` — `claude --bg` kick stdout/stderr suppressed** (`review`): Both the primary and fork-retry kicks redirect to `/dev/null`. When a kick fails, the failure reason is not captured. The `rc` variable captures the exit code; verify_agent_registered_under provides the registration check; and the error message on hard failure is explicit. Suppressing noise from successful kicks is intentional; the verification gate catches silent failures.

---

### False-positive (security)

- **`dispatch-recover-session-id:109` — `session_id` from untrusted sidecar** (`input-validation`, `red-team`): The sidecar files are written by `dispatch-stamp-session` to `~/.claude/projects/`, a path writable only by the local user. An attacker with write access to those files already has full local code-execution capability. The field flows into `printf '%s\t%s\n'` and ultimately into `claude --bg --resume <session_id>` — a controlled invocation with no shell interpolation. Not an injection vector.

- **`office-hours:263` — `branch` field from untrusted sidecar** (`input-validation`): Same trust boundary as above. `rec_branch` flows into `git ls-remote --exit-code --heads origin "$rec_branch"` and `dispatch-provision-from-remote "$c"` — both use it as a positional argument with no shell interpolation. A malicious branch name (e.g., `--upload-pack=…`) could be an argument-injection risk in theory, but the sidecar writer (dispatch-stamp-session) is controlled local tooling, and git's `ls-remote --heads` constrains the ref namespace. Not a practical vector in this local-tool context.

- **`office-hours-select-target:376` — `rec_branch` unvalidated in `git ls-remote`** (`input-validation`): Same branch-value flow as above. The value originates from a locally-written sidecar file and is passed as a positional ref argument to `git ls-remote`. Not an injection vector.

- **`dispatch-recover-session-id:86` — `stamped_at` tab-injection into sort key** (`input-validation`, `red-team`): `stamped_at` is concatenated with a literal tab into the `candidates` array. A crafted sidecar with a `stamped_at` containing a tab would split the sort key incorrectly, potentially mis-ordering candidates. This affects sort ordering only (not code execution); the trust boundary is local user-writable files; and the actual outcome is a potentially non-optimal session selection, not a security failure.

- **`dispatch-recover-session-id:113 / office-hours-select-target:370 / office-hours:201` — `session_id` embedded in resume directive** (`red-team`): The `session_id` is printed in the `resume <N> <sid> <cwd>` directive and consumed by `read -r verb a b c d`. If `session_id` contained spaces, the fields would misparse. The `session_id` is a UUID-format value written by the Claude daemon; space injection is not a realistic risk in this local-tool context.

- **`office-hours:183-190` — `--permission-mode auto` unconditional** (`red-team`): The resumed session uses `--permission-mode auto` regardless of what permission mode the original session used. This is consistent with the fresh-launch path and is the expected behavior for unattended background resume. Not a security concern.

- **`office-hours-select-target:232-258 / office-hours:181-196` — stopped/paused now attachable** (`red-team`): The state change is the intent of #2240. A stopped/paused session has already been running under the same user context; re-attaching it does not escalate privilege. The `working|busy` guard still prevents double-claiming an active session.

---

### Deferred
_None._

### Out-of-scope (security)
_None._

---

_No code changes applied. 18 findings: 7 Informational, 3 Dismissed, 8 False-positive._

PR #2307, opened 2026-06-21
audio: bound enrichment file-read concurrency to cap peak memory

### LINKED ISSUE #2302: audio: local folder enrichment loads all file buffers into memory concurrently — OOM risk on large folders
In \`audio/src/local-source.ts\`, \`enrichLocalItems\` (line 356) calls \`Promise.all(items.map(...))\` over every uncached file. Each call in \`enrichLocalItem\` reads the entire audio file into an \`ArrayBuffer\` via \`resolveLocalBytes\` before parsing tags. On a folder with hundreds or thousands of tracks, all buffers are allocated simultaneously, inflating peak memory proportionally to the total uncached file size. On constrained devices (mobile browsers, low-RAM desktops) this risks OOM crashes or tab kills during the first cold-start enrichment pass.

**Location:** \`audio/src/local-source.ts\` — \`enrichLocalItems\` at line 356.

**Recommended fix:** Replace the flat \`Promise.all\` with a concurrency-bounded batch loop (e.g., chunks of 10–20 at a time) or a worker queue, so at most N file buffers are live simultaneously. Alternatively, use \`navigator.deviceMemory\` or similar heuristics to scale the batch size. Either approach bounds peak memory on the cold-start enrichment pass.

**Deferred from PR #1839** — in-scope architectural change, out of PR scope.

## Acceptance criteria

- [ ] \`enrichLocalItems\` processes uncached files in bounded-concurrency batches rather than a single flat \`Promise.all\`
- [ ] Peak simultaneous \`ArrayBuffer\` allocations are capped (configurable or fixed limit, e.g. 10–20 files)
- [ ] A folder with 500+ tracks completes enrichment without an OOM crash or tab kill on a device with ≤2 GB RAM (manual verification or memory-profiler trace)
- [ ] Enrichment result is identical to the unbounded implementation (same tags written to the sidecar cache)
- [ ] Existing unit tests for \`enrichLocalItem\` and \`enrichLocalTracks\` pass

natb1:
Closes #2302

## Problem

The audio app's local-folder enrichment pass `enrichLocalItems` (`audio/src/local-source.ts`) ran a flat `await Promise.all(items.map((item) => enrichLocalItem(item)))`. Each `enrichLocalItem` reads the **entire** audio file into an `ArrayBuffer` before extracting tags, so the flat `Promise.all` allocates **every** uncached file's buffer simultaneously. On a folder with hundreds or thousands of tracks, peak memory scaled with total uncached file size — risking OOM crashes or tab kills during cold-start enrichment on constrained devices (mobile browsers, ≤2 GB-RAM desktops).

Deferred from PR #1839 as an in-scope architectural change that was out of that PR's scope.

## Change

- **New `audio/src/concurrency.ts`** — a generic, dependency-free `mapWithConcurrency<T, R>(items, limit, fn)`: a sliding-window worker pool where `limit` workers share a cursor over an indexed result array. Peak concurrency is capped at exactly `limit`, a slow item never stalls the others, and results are returned in input order (indexed assignment, independent of completion order). Rejection semantics match `Promise.all`. The monorepo has no existing concurrency-limiting utility, so this is written fresh and local to the audio app.

- **`audio/src/local-source.ts`** — route `enrichLocalItems` through the helper with a fixed cap (`ENRICH_READ_CONCURRENCY = 16`). Only the mapping line changed; the `newEntries` build, write-suppression behavior, the single batched `cacheMetadataBatch` write, and the returned-items order are preserved verbatim, so output is byte-for-byte identical to before. `enrichLocalItem` is unchanged.

Peak enrichment memory is now bounded at roughly `ENRICH_READ_CONCURRENCY × typical-file-size` regardless of folder size.

## Tests

- **New `audio/test/concurrency.test.ts`** — deterministic proof of the bound using an in-flight counter and a manual release queue: the cap is both respected and fully reached (`maxInFlight === 3` exactly for 10 items at limit 3), order is preserved, empty input invokes `fn` zero times, and `limit >= length` degrades to unbounded correctly.
- **One equivalence test added to `audio/test/local-source.test.ts`** — 40 files (> the cap of 16, so multiple windows run) with a mix of pre-cached and uncached files; asserts the single batched write is preserved and the entry-set is identical to the unbounded behavior (cached files excluded, every uncached file included with its extracted tag).
- The three existing enrichment tests stay green — the bounded path preserves the batch-build/return logic verbatim.

`npx vitest run --project audio --root .` → 234 passed, 1 skipped (pre-existing).

## Out of scope

An identical pattern exists in `print/src/local-folder-ui.ts` (`Promise.all(targets.map((t) => enrichLocalItem(t)))`). It is deliberately not touched here — per the single-PR-per-leaf convention it belongs in a separate follow-up, and it could not reuse `audio/src/concurrency.ts` across the app boundary without a shared package.

The acceptance criterion that a 500+ track folder completes enrichment without OOM on a ≤2 GB-RAM device is a manual/memory-profiler check left for QA.

### DIFF
--- a/audio/src/concurrency.ts
+++ b/audio/src/concurrency.ts
@@ -13,6 +13,7 @@ export async function mapWithConcurrency<T, R>(
   limit: number,
   fn: (item: T, index: number) => Promise<R>,
 ): Promise<R[]> {
+  if (limit <= 0) throw new Error('mapWithConcurrency: limit must be > 0');
   const results = new Array<R>(items.length);
   let next = 0;
   async function worker(): Promise<void> {
--- a/audio/src/concurrency.ts
+++ b/audio/src/concurrency.ts
@@ -1,0 +1,27 @@
+/**
+ * Run `fn` over `items` with at most `limit` invocations in flight at once,
+ * returning results in input order. A sliding-window worker pool: `limit`
+ * workers share a cursor, so a slow item never stalls the others and peak
+ * concurrency is capped at exactly `limit` (the point — it bounds peak
+ * memory when `fn` allocates per item, e.g. reading a file into an
+ * ArrayBuffer). Order is preserved by indexed assignment, independent of
+ * completion order. Rejection semantics match `Promise.all`: the first
+ * rejection propagates and in-flight workers are not cancelled.
+ */
+export async function mapWithConcurrency<T, R>(
+  items: readonly T[],
+  limit: number,
+  fn: (item: T, index: number) => Promise<R>,
+): Promise<R[]> {
+  const results = new Array<R>(items.length);
+  let next = 0;
+  async function worker(): Promise<void> {
+    while (next < items.length) {
+      const i = next++;
+      results[i] = await fn(items[i], i);
+    }
+  }
+  const workerCount = Math.max(0, Math.min(limit, items.length));
+  await Promise.all(Array.from({ length: workerCount }, () => worker()));
+  return results;
+}
--- a/audio/src/local-source.ts
+++ b/audio/src/local-source.ts
@@ -393,7 +393,7 @@ async function enrichLocalItems(items: LibraryItem[]): Promise<void> {
   const results = await mapWithConcurrency(
     items,
     ENRICH_READ_CONCURRENCY,
-    (item) => enrichLocalItem(item),
+    enrichLocalItem,
   );
   const newEntries = Object.fromEntries(
     results.filter((r) => r.entry !== null).map((r) => r.entry as [string, CachedMetadata]),
--- a/audio/src/local-source.ts
+++ b/audio/src/local-source.ts
@@ -23,10 +23,15 @@ import {
   setLocalDirectory,
 } from "./sidecar.js";
 import { extractAudioMetadata } from "./local-metadata.js";
+import { mapWithConcurrency } from "./concurrency.js";
 import { AUDIO_FORMATS } from "./types.js";
 import type { AudioFormat, AudioTags, LibraryItem } from "./types.js";
 
 const PURPOSE = "library-folder";
+/** Peak simultaneous file reads during enrichment. Each uncached file is
+ * read fully into an ArrayBuffer before tag extraction; bounding this caps
+ * peak memory so large folders don't OOM on low-RAM (≤2GB) devices. */
+const ENRICH_READ_CONCURRENCY = 16;
 const store = createFsaHandleStore({ app: "audio" });
 
 const MIME_TYPES: Record<AudioFormat, string> = {
@@ -353,7 +358,11 @@ async function enrichLocalItem(item: LibraryItem): Promise<EnrichResult> {
 }
 
 async function enrichLocalItems(items: LibraryItem[]): Promise<LibraryItem[]> {
-  const results = await Promise.all(items.map((item) => enrichLocalItem(item)));
+  const results = await mapWithConcurrency(
+    items,
+    ENRICH_READ_CONCURRENCY,
+    (item) => enrichLocalItem(item),
+  );
   const newEntries = Object.fromEntries(
     results.filter((r) => r.entry !== null).map((r) => r.entry as [string, AudioTags]),
   );
--- a/audio/test/concurrency.test.ts
+++ b/audio/test/concurrency.test.ts
@@ -1,0 +1,52 @@
+import { describe, it, expect, vi } from "vitest";
+import { mapWithConcurrency } from "../src/concurrency.js";
+
+const tick = () => new Promise((r) => setTimeout(r, 0));
+
+describe("mapWithConcurrency", () => {
+  it("caps peak concurrency at exactly limit and preserves input order", async () => {
+    const items = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+    const limit = 3;
+    let inFlight = 0;
+    let maxInFlight = 0;
+    const release: Array<() => void> = [];
+
+    const fn = async (x: number) => {
+      inFlight++;
+      maxInFlight = Math.max(maxInFlight, inFlight);
+      await new Promise<void>((res) => release.push(res));
+      inFlight--;
+      return x * 2;
+    };
+
+    const call = mapWithConcurrency(items, limit, fn);
+
+    // Let the pool launch its initial window.
+    await tick();
+    expect(inFlight).toBe(3);
+
+    // Drain: as each in-flight item resolves, a worker pulls the next item and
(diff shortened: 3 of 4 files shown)


natb1:
# Fix-checks summary — PR #2307

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `git diff --no-color -U0 --diff-filter=d origin/main...HEAD -- '*.ts' '*.tsx' '*.js' '*.jsx' '*.mjs' '*.cjs' | .github/scripts/check-type-safety-escapes.sh --scan-stdin`
- **Failure excerpt:** `::error file=audio/test/concurrency.test.ts,line=31::Net-new type-safety escape hatch (non-null assertion !). Suppress with \`// type-safety-ok: <reason>\` if intentional.`
- **Why not caught:** The non-null assertion `release.shift()!()` was introduced in the new concurrency test. The while-loop guard (`release.length > 0`) makes it logically safe, but the type-safety-sensor flags all non-null assertions on added lines. Optional chaining (`?.`) is the type-safe equivalent and avoids the sensor.
- **Fix:** Replaced `release.shift()!()` with `release.shift()?.()` on line 31 of `audio/test/concurrency.test.ts`. Commit: `8a76fc3e`

natb1:
# Fix-checks summary — PR #2307

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `git diff --no-color -U0 --diff-filter=d origin/main...HEAD -- '*.ts' '*.tsx' '*.js' '*.jsx' '*.mjs' '*.cjs' | .github/scripts/check-type-safety-escapes.sh --scan-stdin`
- **Failure excerpt:** `::error file=audio/test/concurrency.test.ts,line=31::Net-new type-safety escape hatch (non-null assertion !). Suppress with \`// type-safety-ok: <reason>\` if intentional.`
- **Why not caught:** The non-null assertion `release.shift()!()` was introduced in the new concurrency test. The while-loop guard (`release.length > 0`) makes it logically safe, but the type-safety-sensor flags all non-null assertions on added lines. Optional chaining (`?.`) is the type-safe equivalent and avoids the sensor.
- **Fix:** Replaced `release.shift()!()` with `release.shift()?.()` on line 31 of `audio/test/concurrency.test.ts`. Commit: `8a76fc3e`

## Iteration 2

- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `npx vitest run --project audio --root .`
- **Failure excerpt:** `AssertionError: expected Set{ 'track39.mp3', …(39) } to deeply equal Set{ 'track1.mp3', 'track3.mp3', …(18) }` — `bounded enrichment over many files` test in `audio/test/local-source.test.ts`
- **Why not caught:** The new equivalence test was written before `c03dcc80` (merged from `origin/main`) added the fingerprint fields (`size`, `lastModified`) to `CachedMetadata`. After the merge, `enrichLocalItem`'s cache-hit check requires `cached.size === fp.size && cached.lastModified === fp.lastModified`, but the test mock returned flat `{ artist: "cached" }` (no fingerprint), so every file appeared as a cache miss and all 40 were extracted instead of only the 20 uncached ones. The batch assertion also used the old flat `{ artist: "fresh" }` shape.
- **Fix:** Updated `mockGetMetadata` mock to return `{ tags: { artist: "cached" }, size: bytes(name).byteLength, lastModified: 1000 + i }` for even-indexed files; updated batch assertion to match `{ tags: { artist: "fresh" }, size, lastModified }`. Commit: `b8e637c8`

natb1:
# QA Summary — Issue #2302 / PR #2307

## Results

| 1. Concurrency cap is fixed in the documented 10-20 range | script-verifiable | PASS |
| 2. enrichLocalItems uses bounded concurrency, not flat Promise.all | script-verifiable | PASS |
| 3. concurrency.ts and its tests exist and pass | script-verifiable | PASS |
| 4. 500+ track folder enriches without OOM on a ≤2 GB device | needs-human-judgment (planned-deferral) | → filed as follow-up |

No browser walkthrough (pure TypeScript library change, no frontend component).

## Bugs Found

None.

## Disposition Triage

Step 3.5 ran (one planned-deferral residue item).

| Item | Class | Verify |
|------|-------|--------|
| 4. 500+ track folder enriches without OOM on a ≤2 GB device | needs-main (planned-deferral) | n/a |

**Rationale:** planned_deferral is true — AC #3 (bounded peak ArrayBuffer allocations / no OOM on a large folder) is documented as non-assertable at merge time. It requires manual verification or a memory-profiler trace on real files in the browser app, with no browser component in scope for this library-only PR. The criterion is only verifiable downstream against merged main/production.

## Filed Follow-ups

- #2328: "qa: verify against main — qa-needs-main #2302: 500-track-folder-enriches-without-oom-on-a-2-gb-device" (newly filed, blocked_by #2302, routed to office-hours human review via `main-qa` label)

## Deferred to Office-Hours

None. (Item 4 classified `needs-main` — filed as follow-up #2328, not routed to office-hours walkthrough.)

natb1:
## Review — #2302

### Fixed (commit 6202909f)

- **`audio/test/concurrency.test.ts:6`** _(code-review, review)_ — Rejection propagation untested. The JSDoc claims `Promise.all` semantics but no test exercised the rejection path. Added a test: fn rejects for one item, mapWithConcurrency rejects with that error.

- **`audio/src/local-source.ts:396`** _(code-review)_ — Redundant lambda wrapper `(item) => enrichLocalItem(item)` where a direct function reference suffices. Replaced with `enrichLocalItem` directly.

- **`audio/src/concurrency.ts:24`** _(code-review, review)_ — `limit <= 0` with a non-empty items array silently produced a sparse array (no workers run, `results` stays empty). Per project code-style, clear errors beat silent fallbacks. Added a guard: `if (limit <= 0) throw new Error('mapWithConcurrency: limit must be > 0')`.

### Informational

- **`audio/test/local-source.test.ts:641`** _(review)_ — The 40-file integration test verifies entry-set equivalence (correct result contents) but does not assert the in-flight concurrency bound during enrichment. The concurrency bound is covered by `concurrency.test.ts`; no action required here.

- **`audio/test/local-source.test.ts:641`** _(erosion)_ — Net cyclomatic-complexity increase: worst-function 4→5, total 90→98. The added test adds one branch (`for` + nested `if`). Within the acceptable range for a new test case; no decomposition needed.

### Security

No attack surface findings. No CodeQL alerts. No dependency changes (deps=false).

_Coverage: code-review ✓ · review ✓ · security finders ✓ (app_or_rules=true, no findings) · CodeQL ✓ (0 open alerts) · dependency audit skipped (no package.json changes) · erosion ✓_

PR #2308, opened 2026-06-21
audio: snapshot dirHandle in enqueueWrite to fix stale-handle TOCTOU on folder switch

natb1:
Closes #2305

## Problem

`audio/src/sidecar.ts`'s `enqueueWrite` read the module-level `dirHandle` and
`writable` globals at **task execution time**, inside the chained `.then()`. That
is a TOCTOU bug: if the user rapidly disconnects folder A (`clearLocalDirectory`)
and connects folder B (`setLocalDirectory`) before a write queued for A has
drained, the orphaned task ran against the now-current `dirHandle` (B) and folder
A's data landed in folder B. The `writeChain = Promise.resolve()` reset in those
functions did not prevent this — it only detaches *future* enqueues from the old
chain; the already-chained task still ran and still read the live globals.

## Fix

`enqueueWrite` now captures `dirHandle` and `writable` into closure-local
snapshot constants **before** appending to the chain. Inside the task body:

1. A single TOCTOU guard at task start — `if (dirHandle !== snapshotHandle)
   return;` — skips the whole task if the bound directory changed between enqueue
   and execution, so the patch neither persists to the new folder nor
   contaminates the new folder's in-memory `cachedModel`.
2. The persist targets the **snapshot** handle, not the live global
   (`if (snapshotWritable && snapshotHandle !== null) await writeSidecar(snapshotHandle, merged)`),
   so even a switch racing in after the guard still writes to the folder the
   patch was issued for.

Exactly one guard, at task start. A second post-`ensureLoaded` re-guard is
deliberately not added: it would drop a still-valid write of A's data back to
folder A (A's retained handle stays valid after an in-app disconnect, which does
not revoke OS permission).

The `writeChain = Promise.resolve()` resets in `clearLocalDirectory` /
`setLocalDirectory` are kept (harmless with the guard, lower-risk than removing
them); their doc comments are corrected to state that the reset only detaches
future enqueues and the snapshot guard is what prevents a stale write.

## Tests

Added a `describe("folder switch — stale-handle TOCTOU")` block in
`audio/test/sidecar.test.ts` covering: an in-flight switch dropping the stale
write (dirB written, dirA's `.commons-audio` never created), the in-memory model
reflecting the new folder only, player-state writing to the correct folder after
a switch, and a chain-liveness regression proving the skip path does not wedge
the chain. Full audio suite: 234 passed, 1 skipped.

## Out of scope

A pre-existing read-staleness residual race inside `ensureLoaded` (a switch
landing during the first `readSidecar` await could cache the old folder's model
under the new handle) is unchanged by this fix and will be filed as a separate
low-priority follow-up — it is a read-path bug unrelated to this issue's
write-targeting criteria.

### DIFF
--- a/audio/src/sidecar.ts
+++ b/audio/src/sidecar.ts
@@ -259,7 +259,11 @@ let writeChain: Promise<void> = Promise.resolve();
 /**
  * Store the retained directory handle and writable flag (called by Unit 3 after
  * the folder is picked / permission resolved). Invalidates the cached model so
- * the next access re-reads lazily from the new handle.
+ * the next access re-reads lazily from the new handle. The
+ * `writeChain = Promise.resolve()` reset only detaches FUTURE enqueues from the
+ * old chain; it does NOT stop an already-enqueued stale write — that is the job
+ * of the `enqueueWrite` snapshot guard, which skips any task whose bound handle
+ * no longer matches the live one.
  */
 export function setLocalDirectory(handle: FileSystemDirectoryHandle, isWritable: boolean): void {
   dirHandle = handle;
@@ -271,9 +275,13 @@ export function setLocalDirectory(handle: FileSystemDirectoryHandle, isWritable:
 
 /**
  * Unbind the directory (called when the folder is disconnected). Drops the
- * handle, marks not-writable, and resets the cache + write chain so an in-flight
- * write enqueued before disconnect cannot target the now-stale folder and the
- * next access re-loads an empty model.
+ * handle, marks not-writable, and resets the cache + write chain. The
+ * `writeChain = Promise.resolve()` reset only detaches FUTURE enqueues from the
+ * old chain — it does NOT cancel an already-enqueued write, which still runs.
+ * What prevents a stale write from landing in a later-connected folder is the
+ * `enqueueWrite` snapshot guard: an orphaned task whose snapshot handle no
+ * longer matches the live `dirHandle` is skipped. The cache reset makes the
+ * next access re-load an empty model.
  */
 export function clearLocalDirectory(): void {
   dirHandle = null;
@@ -311,13 +319,22 @@ export async function ensureLoaded(): Promise<SidecarData> {
  * per-link catch keeps one failed write from poisoning the chain.
  */
 function enqueueWrite(patch: SidecarPatch): Promise<void> {
+  const snapshotHandle = dirHandle;
+  const snapshotWritable = writable;
   writeChain = writeChain
     .then(async () => {
+      // TOCTOU guard: if the bound directory changed between enqueue and
+      // execution (rapid disconnect/reconnect), this patch was issued for the
+      // previous folder — skip it entirely so it neither persists to the new
+      // folder nor contaminates the new folder's in-memory model.
+      if (dirHandle !== snapshotHandle) return;
       const model = await ensureLoaded();
       const merged = mergeSidecar(model, patch);
       cachedModel = merged;
-      if (writable && dirHandle !== null) {
-        await writeSidecar(dirHandle, merged);
+      // Persist to the SNAPSHOT handle (not the live global), so a switch that
+      // races in after the guard still writes to the folder this patch was for.
+      if (snapshotWritable && snapshotHandle !== null) {
+        await writeSidecar(snapshotHandle, merged);
       }
     })
     .catch((err) => {
--- a/audio/test/sidecar.test.ts
+++ b/audio/test/sidecar.test.ts
@@ -684,17 +684,21 @@ describe("folder switch — stale-handle TOCTOU", () => {
 
     setLocalDirectory(dirA, true);
     // task_A snapshots A; its body is a pending microtask.
-    void cacheMetadata("a.mp3", { title: "A", duration: 100 });
+    void cacheMetadata("a.mp3", { tags: { title: "A", duration: 100 }, size: 1, lastModified: 1 });
     // Switch to B synchronously, before task_A's body runs.
     setLocalDirectory(dirB, true);
-    void cacheMetadata("b.mp3", { title: "B", duration: 200 });
+    void cacheMetadata("b.mp3", { tags: { title: "B", duration: 200 }, size: 2, lastModified: 2 });
     await flushWrites();
 
-    // B got its write.
-    const readB = await readSidecar(dirB);
-    expect(readB.metadata["b.mp3"]).toEqual({ title: "B", duration: 200 });
     // A's stale write was skipped — its sidecar dir was never created.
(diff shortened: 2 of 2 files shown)


natb1:
## QA summary — issue #2305, attempt 0, PR #2308

### Items

| # | Title | Result |
|---|-------|--------|
| 1 | Snapshot constants (`snapshotHandle`/`snapshotWritable`) precede `writeChain =` | ✅ PASS |
| 2 | TOCTOU guard `if (dirHandle !== snapshotHandle) return` present at task-start | ✅ PASS |
| 3 | Persist calls use `snapshotHandle`/`snapshotWritable`, not the live globals | ✅ PASS |
| 4 | `npx vitest run --project audio --root .` — 4 new TOCTOU tests pass (234 total, 1 skipped) | ✅ PASS |
| 5 | Doc comments accurately describe the snapshot guard, not the obsolete reset | 🔼 escalated |
| 6 | Manual acceptance: rapid A→B folder switch does not leak A's write into B | 📋 #2313 |

**4 PASS · 0 FAIL · 1 escalated · 1 filed as follow-up**

---

Items 5–6 were in the residue list (judgment-call, not directly script-verifiable). The Step 3.5 Workflow classified them:

**Item 5 — Doc comments accurately describe the snapshot guard, not the obsolete reset**
- Class: `opus-fixable`
- The Workflow classified this as opus-fixable (a documentation-accuracy judgment resolvable by reading the comments alongside the implementation). The fix planner deviated: after reading the comments in `audio/src/sidecar.ts` (`setLocalDirectory`, `clearLocalDirectory`, `enqueueWrite`) and `audio/test/sidecar.test.ts`, it found the comments already accurately describe the snapshot guard and explicitly state the chain reset does not cancel in-flight writes — there is no stale text to change.
- Deviation (`units: []`) triggered the escalate finalize path.

**Item 6 — Manual acceptance: rapid A→B folder switch does not leak A's write into B**
- Class: `needs-main` (planned deferral)
- No browser component was detected for the audio module in this worktree (app-dir: none). The live File System Access API flow cannot be exercised here; the behavior is covered structurally by the TOCTOU guard (items 1–3) and the simulated unit tests (item 4). Real-browser rapid A→B confirmation is non-assertable at merge time and is deferred to a main-qa follow-up.
- Filed as #2313 (blocked_by #2305).

**Terminal behavior:** escalate finalize path. Item 5 (opus-fixable, planner deviation — doc comments already accurate) escalates to office-hours. Item 6 was filed as a `needs-main` follow-up and is excluded from the escalation set.

---

### Filed follow-ups

- Item 6 → #2313 (newly filed) — `qa-needs-main #2305: manual-acceptance-rapid-a-b-folder-switch-does-not-leak-a-s-write-into-b`

---

natb1:
## Review — PR #2308 / issue #2305

### Fixed

- **`audio/test/sidecar.test.ts:687`** (code-review, review) — Three of the four new TOCTOU tests called `cacheMetadata` with bare `AudioTags` objects (`{ title: 'A', duration: 100 }`) instead of the required `CachedMetadata` wrapper shape (`{ tags, size, lastModified }`). The assertions read back via `readSidecar` or `getMetadata` would never match the stored shape, making the tests ineffective at catching data-shape regressions. Reshaped all four tests to use `{ tags: { … }, size: 1, lastModified: 1 }` and updated `toEqual` expectations accordingly. Also reordered test 1 so the stale-write guard assertion (`subdirsOf(dirA)[".commons-audio"]` is undefined) precedes the `readSidecar` assertion, so a data-shape failure doesn't hide the TOCTOU behavior check. Fix commit: `bb4ad126`.

### Deferred

- **`audio/src/sidecar.ts:360`** (code-review) — The TOCTOU guard checks `dirHandle !== snapshotHandle` once at task start, then immediately awaits `ensureLoaded()`, which may span multiple async FSA round-trips. A `setLocalDirectory(newDir)` call landing during that await window leaves two unguarded `cachedModel` write sites: one inside `ensureLoaded`'s `.then()` and one in `enqueueWrite` itself. Both writes can install stale dirA data into the new folder's in-memory cache, causing subsequent reads for dirB to return wrong data. Filed as #2330 (blocked by #2305).

### Out-of-scope

- **`audio/src/sidecar.ts:361`** (red-team) — Same mid-execution switch residual described above; tracked by #2330.

### Informational

- **`audio/test/sidecar.test.ts:678`** (code-review) — The `subdirsOf` helper is defined inside the new `describe("folder switch — stale-handle TOCTOU")` block. It is not reachable from the two earlier `clearLocalDirectory` describe blocks that could benefit from the same helper.
- **`audio/test/sidecar.test.ts:672`** (review) — The new TOCTOU suite has no test for the `setLocalDirectory(A) → enqueue write → clearLocalDirectory() → setLocalDirectory(B)` sequence. The existing tests exercise `setLocalDirectory(A) → setLocalDirectory(B)` (no clear) and `clearLocalDirectory()` in isolation; the clear-then-reconnect path is not covered.
- **`audio/src/sidecar.ts:352`** (review) — `snapshotWritable` is captured at enqueue time alongside `snapshotHandle`, but in all current code paths `writable` only changes atomically with `dirHandle` (both in `setLocalDirectory` and `clearLocalDirectory`). The snapshot is harmless and guards against future drift, but callers may assume it adds independent protection it does not currently provide.
- **`audio/src/sidecar.ts:90`** (erosion) — Cyclomatic complexity of `audio/src/sidecar.ts` rose from 72 to 73 (+1) vs `origin/main`. Within normal tolerance for the TOCTOU guard addition.
- **`audio/test/sidecar.test.ts:33`** (erosion) — Cyclomatic complexity of `audio/test/sidecar.test.ts` rose from 132 to 139 (+7) vs `origin/main`. Expected: four new TOCTOU test cases added.

- **`audio/test/sidecar.test.ts:778`** (erosion) — Code-duplication finding: clone blocks rose from 2 to 3, duplicated lines from 23 to 31. Refuted 2/2 skeptics. The flagged blocks are per-test setup boilerplate for independent TOCTOU race scenarios; each test uses a different write API (`cacheMetadata` vs `savePlayerState`), a different number of writes, and distinct assertions. A shared helper would require callbacks or parameters that obscure the specific per-test arrangement and harm isolation. Standard false-positive class for test-setup structural similarity.

### False-positive

- **`audio/src/sidecar.ts:360`** (red-team) — Claimed the object-reference equality guard fails for clear-then-reconnect with the same handle. Refuted: `setLocalDirectory` always receives a `FileSystemDirectoryHandle` returned by a fresh picker invocation; the File System Access API never returns the same object reference for a different picker call, so reference equality reliably detects switches in all real usage.

### Security

No CodeQL alerts open against this PR. No dependency changes (`deps=false`).

PR #2294, opened 2026-06-21
dispatch-stamp-session: preserve session-start base_sha across resume

natb1:
Closes #2264

## Problem

`dispatch-stamp-session` writes a per-session sidecar recording the GitHub
artifact each dispatch worker acted on. `dispatch-token-audit` joins each
session to its real-world outcome on the four primary keys `{repo, issue, pr,
base_sha}`, where `base_sha` is meant to anchor the session to the repository
state it *started* from.

Mode A (the session-start write) runs on both `startup` and `resume`. PR #2252,
while adding path-traversal validation, rewrote the idempotency block and dropped
the preservation of `base_sha` from the session-start write — its replacement
comment stated "Only `.pr` is carried forward; nothing else is preserved." As a
result, a session that resumes after committing work re-derives `base_sha` from
the post-resume HEAD, so the join key reflects the most-recent-stamp HEAD rather
than session-start ground truth — non-deterministic and useless as a stable join
key (unlike `repo`/`issue`/`pr`, which do not move mid-session).

## Fix

Restore the session-start preservation of `base_sha` across resume (acceptance
Option 1). When the sidecar already exists, the re-stamp now preserves its
`base_sha`; only the first (startup) write derives `base_sha` from HEAD.

`stamped_at` deliberately keeps the re-derive behavior #2252 introduced: it is
**not** a join key and is **not** carried into the audit `artifact` record, so it
correctly records the most-recent stamp time, not session start. The script and
the audit SKILL.md now document this intentional `base_sha`-vs-`stamped_at`
temporal split.

This is **not** a revert of #2252 — its path-traversal validation stays; only the
`base_sha` preservation is restored.

## Changes

- `dispatch-stamp-session`: re-add the `base_sha` preserve in the idempotency
  block; rewrite the block comment and header note to document the restored
  invariant and the `base_sha`-vs-`stamped_at` split.
- `test-dispatch-scripts.sh` (test #6): correct the idempotency assertion in
  lockstep — `base_sha` must be PRESERVED (was asserting the regressed re-derive
  behavior), plus a new assertion that `stamped_at` advances.
- `dispatch-token-audit/SKILL.md`: document that `base_sha` is the session-start
  HEAD, preserved across resume.

## Verification

The dispatch script test suite (what CI runs) passes: 3894/3894, including the
corrected preserve assertion and the new `stamped_at`-advances assertion.

PR #2316, opened 2026-06-22
test-dispatch-scripts.sh: add null-base_sha idempotent re-write test

### LINKED ISSUE #2298: test-dispatch-scripts.sh: Test 6 (idempotent re-write) missing null-base_sha test path
## Context

Test 6 (idempotent re-write) in \`test-dispatch-scripts.sh\` verifies that \`dispatch-stamp-session\` does not overwrite a valid \`base_sha\` when the sidecar already exists. The new preservation logic uses \`jq -r '.base_sha // empty'\` to distinguish a null \`base_sha\` from a real SHA — a null value should fall through to the HEAD-derived value rather than be preserved.

However, there is no test case for the path where the sidecar already exists but carries \`"base_sha": null\` (an explicitly null value, not an absent key). This path differs from the normal idempotent path (existing sidecar with a real SHA) because the \`// empty\` operator does the actual work there.

## Why this matters

Without a test covering this path, a future change that drops \`// empty\` from the jq filter would make null produce the string \`"null"\`, pass \`[[ -n "null" ]]\`, and write the literal string \`"null"\` as \`base_sha\` — with no test catching the regression.

## Recommended fix

Add a test case in the test-6 block (or as test 6b) in \`test-dispatch-scripts.sh\` that:

1. Seeds a sidecar with \`"base_sha": null\` (using \`jq\` or inline JSON)
2. Re-stamps (re-runs \`dispatch-stamp-session\` for the same session)
3. Asserts the resulting \`base_sha\` equals the current \`git rev-parse HEAD\` (the HEAD-derived fallback), not the string \`"null"\`

This follows the established subshell style used in the existing test-6 block.

## Acceptance criteria

- [ ] A test case seeds a sidecar with \`"base_sha": null\` and re-stamps
- [ ] The test asserts the resulting \`base_sha\` equals \`git rev-parse HEAD\` (not the string \`"null"\`)
- [ ] The test follows the existing subshell style in the test-6 block
- [ ] \`test-dispatch-scripts.sh\` passes with the new test case (no regressions)

Files: \`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh\`

Surfaced by the review pass of PR #2294. Out of scope for that PR.

natb1:
Closes #2298

## What

Adds a missing test case (`test 6b`) to
`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` covering the
idempotent re-write path where a `dispatch-stamp-session` sidecar already exists
but carries an **explicitly-null** `base_sha`.

## Why

PR #2294 added session-start `base_sha` preservation to `dispatch-stamp-session`.
The logic distinguishes an explicitly-null `base_sha` from a real SHA via
`jq -r '.base_sha // empty'`: a null value yields empty, falls through the
`[[ -n "$existing_base_sha" ]]` guard, and is replaced by the HEAD-derived value.

The existing Test 6 covers the real-SHA preservation path and Test 7 covers the
ff-merge resume path, but neither exercises the explicitly-null path — the one
place where the `// empty` operator actually does the work. Without a test for it,
a future change that drops `// empty` would make null stringify to the literal
`"null"`, pass `[[ -n "null" ]]`, and write the string `"null"` as `base_sha` — a
regression no test would catch.

## How

The new test 6b mirrors the surrounding test-6/test-7 subshell idiom: a `mktemp -d`
fake repo on a `999-fixture` branch, a seeded sidecar whose `base_sha` is JSON
`null`, a re-stamp, and an assertion that the resulting `base_sha` equals the
current HEAD (the HEAD-derived fallback), not the string `"null"`.

Test-only addition — no change to `dispatch-stamp-session` itself, other test
blocks, or CI config.

## Verification

The suite is invoked directly in CI (`.github/workflows/unit-tests.yml`) and is
self-contained (bash + git + jq + find + date, PATH-shimmed, no network). Running
it locally executes test 6b along with the full suite: **3908/3908 passed, 0
failed.**

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -38618,6 +38618,23 @@ STAMP="$SCRIPT_DIR/dispatch-stamp-session"
   rm -rf "$d"
 )
 
+# 6b. Idempotent re-write with an explicitly-null .base_sha falls through to
+#     the HEAD-derived fallback (exercises the jq `// empty` operator that
+#     distinguishes JSON null from a real SHA in dispatch-stamp-session).
+(
+  d=$(mktemp -d)
+  git -C "$d" init -q
+  git -C "$d" remote add origin https://github.com/natb1/commons.systems.git
+  git -C "$d" checkout -q -b 999-fixture
+  git -C "$d" -c user.email=t@t -c user.name=t commit -q --allow-empty -m init
+  sc="$d/sess6b.dispatch-stamp.json"
+  # Seed a sidecar whose base_sha is JSON null (not the string "null", not absent).
+  printf '%s\n' '{"schema":1,"session_id":"sess6b","repo":"old/repo","issue":1,"pr":null,"branch":"old","base_sha":null,"stamped_at":"2026-01-01T00:00:00Z"}' > "$sc"
+  ( cd "$d" && "$STAMP" --session-id sess6b --transcript-path "$d/sess6b.jsonl" )
+  assert_eq "stamp: re-write with null .base_sha falls through to HEAD" "$(git -C "$d" rev-parse HEAD)" "$(jq -r .base_sha "$sc")"
+  rm -rf "$d"
+)
+
 # 7. Resume after HEAD moves (ff-merge) preserves the session-start .base_sha.
 #    This encodes #2270's failure mode: an initial stamp records HEAD=A; a later
 #    git merge --ff-only moves HEAD to B; the resume re-stamp must keep base_sha=A.


natb1:
# QA Summary — Issue #2298 / PR #2316

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Test 6b seeds a sidecar with `"base_sha": null` and re-stamps without error | shell-command | PASS |
| 2 | Test 6b asserts `base_sha` equals `git rev-parse HEAD`, not the string `"null"` | shell-command | PASS |
| 3 | Test 6b follows the existing subshell style (isolated temp dir, cleanup) | — | needs-human-judgment (assessed below) |
| 4 | Full suite passes with no regressions introduced by the new test case | shell-command | PASS |

## Bugs found

None.

## Disposition triage

Step 3.5 ran. `result.dispositions` is empty — every residue item classified `already-satisfied`.

- **Item 3 — "Test 6b follows the existing subshell style (isolated temp dir, cleanup)"**
  Rationale: The criterion is provable by reading the code, not a true human-judgment call. Test 6b is a standalone subshell `( ... )` that creates its own isolated temp dir with `d=$(mktemp -d)` and removes it at the end with `rm -rf "$d"`, exactly matching the surrounding Test 6 and Test 7 blocks: same git init/remote/checkout/commit setup, same sidecar seeding, same `( cd "$d" && "$STAMP" ... )` invocation, same `assert_eq` style. The expected outcome is already met — no code defect and no human review required.

All residue items are `already-satisfied` (dropped as PASS). The escalation set is empty.

natb1:
## Review: #2298 (PR #2316)

### Fixed
_None._

### Informational

- **`test-dispatch-scripts.sh:38634`** — `assert_eq` receives arguments as `(label, expected, actual)`; here `expected` is the `git rev-parse HEAD` value and `actual` is `jq -r .base_sha "$sc"`. The argument order is correct given the surrounding test idiom (expected comes before actual), but is worth noting for clarity — no change required.
  _(source: review)_

- **`test-dispatch-scripts.sh:38621-38636`** — Test 6b verifies that `base_sha` is correctly updated to HEAD when the existing sidecar value is JSON `null`. It does not assert that unrelated fields (e.g. `.pr`) remain unchanged after the re-stamp. Consistent with the scope of surrounding tests (6 and 7 have the same narrow focus); noted as a potential gap for future expansion but out of scope here.
  _(source: review)_

### Dismissed
_None._

### Deferred
_None._

### Security

### Out-of-scope (security)
_None._

PR #2335, opened 2026-06-22
audio: guard cachedModel writes with a generation counter (mid-read folder-switch TOCTOU)

natb1:
Closes #2330

## Problem

`audio/src/sidecar.ts` holds a single in-memory model (`cachedModel`) for the
bound directory and serializes disk writes through `writeChain`. The existing
TOCTOU defense in `enqueueWrite` compares `dirHandle` identity once at the top of
the task body — but that comparison runs **before** `await ensureLoaded()`, which
performs several async FSA round-trips inside `readSidecar` (`getDirectoryHandle`
→ `getFileHandle` → `getFile` → `file.text()`).

If `setLocalDirectory(dirB)` fires during that await window (a user rapidly
picking two folders), two sites contaminate the **new** folder's in-memory cache
with the **old** folder's data:

1. `ensureLoaded`'s `.then()` assigns `cachedModel = model` from the stale dirA
   read, overwriting the `null` that `setLocalDirectory` just placed.
2. `enqueueWrite` then assigns `cachedModel = merged` (dirA patch merged onto
   dirA model), leaving dirB's cache holding dirA data.

Subsequent dirB reads return stale dirA data with no disk re-read, and the next
dirB write merges against that contaminated base and persists wrong data.
Handle-object identity also cannot guard the clear-then-reconnect-with-the-same-handle
edge (identical handle, but the folder was unbound and rebound).

## Fix

A module-scope `generation` counter, incremented on every `setLocalDirectory` /
`clearLocalDirectory`, snapshotted at read-start (`gen`) and at enqueue-time
(`snapshotGen`), guards **both** `cachedModel` assignments:

- `ensureLoaded` guards `cachedModel = model` with `if (generation === gen)`, but
  still **returns the freshly-read model** so an in-flight `enqueueWrite` task
  receives its own folder's data to merge and persist.
- `enqueueWrite` guards `cachedModel = merged` with `if (generation === snapshotGen)`.

The disk persist (`writeSidecar` to the snapshot handle) is deliberately left
**unguarded** — a stale task must still persist its data to the folder it was for
(#2308). Only the two in-memory `cachedModel` writes are guarded. Incrementing on
both `setLocalDirectory` and `clearLocalDirectory` covers the
clear-then-reconnect-with-the-same-handle case that handle identity misses.

## Test

A new test in the `folder switch — stale-handle TOCTOU` block exercises the
**mid-read** switch (the existing tests cover only the synchronous switch). A
gated-read fake parks `task_A` inside `readSidecar`'s first FSA round-trip; the
switch lands in that window. `dirA` is preloaded with distinctive `diskA.mp3`
content so the two contamination sites are discriminated:

- `getMetadata("a.mp3")` → `undefined` catches the `enqueueWrite` merged-patch leak.
- `getMetadata("diskA.mp3")` → `undefined` catches the `ensureLoaded` raw-disk-model leak.

Verified red on pre-fix source (fails at the site-2 assertion) and green after.

## Verification

- `npx vitest run --project audio --root .` — 245 passed, 1 skipped.
- `npx tsc --noEmit -p audio/tsconfig.json` — clean.

### DIFF
--- a/audio/src/sidecar.ts
+++ b/audio/src/sidecar.ts
@@ -281,6 +281,9 @@ let writable = false;
 // next access re-reads from the new handle.
 let cachedModel: SidecarData | null = null;
 let loadPromise: Promise<SidecarData> | null = null;
+// Incremented on every setLocalDirectory / clearLocalDirectory so in-memory
+// assignments can be guarded against writes that started before the switch.
+let generation = 0;
 
 // True when the bound folder has a sidecar file on disk that exists but
 // could not be parsed (corrupt). While set, disk writes are suppressed so a
@@ -307,6 +310,7 @@ export function setLocalDirectory(handle: FileSystemDirectoryHandle, isWritable:
   dirHandle = handle;
   writable = isWritable;
   cachedModel = null;
+  generation++;
   loadPromise = null;
   corruptOnDisk = false;
   writeChain = Promise.resolve();
@@ -326,6 +330,7 @@ export function clearLocalDirectory(): void {
   dirHandle = null;
   writable = false;
   cachedModel = null;
+  generation++;
   loadPromise = null;
   corruptOnDisk = false;
   writeChain = Promise.resolve();
@@ -339,12 +344,17 @@ export async function ensureLoaded(): Promise<SidecarData> {
   if (cachedModel !== null) return cachedModel;
   if (loadPromise === null) {
     const handle = dirHandle;
+    const gen = generation;
     loadPromise = (
       handle === null ? Promise.resolve<SidecarData | null>(emptyModel()) : readSidecar(handle)
     ).then((model) => {
-      // The handle may have been rebound (setLocalDirectory) while readSidecar
-      // was in flight. A stale result must not mutate the freshly-reset state —
-      // discard it so the next ensureLoaded re-reads the current handle.
+      // The handle may have been rebound (setLocalDirectory / clearLocalDirectory)
+      // while readSidecar was in flight. A stale result must not mutate the
+      // freshly-reset state — discard it so the next ensureLoaded re-reads the
+      // current handle. The generation guard on `cachedModel = model` below
+      // additionally covers the clear-then-reconnect-with-the-same-handle case,
+      // where this handle check passes (identical handle object) but the folder
+      // was unbound and rebound, so the read is still stale.
       if (dirHandle !== handle) {
         return cachedModel ?? emptyModel();
       }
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — Issue #2330 / PR #2335

## Items Executed

All items ran in the **shell-command lane** (no browser component detected — pure `audio/` TypeScript change).

| # | Title | Result |
|---|-------|--------|
| 1 | Generation counter defined at module scope and incremented on both folder-mutation calls | PASS |
| 2 | `ensureLoaded` captures generation at read-start, guards cache write, returns model unconditionally | PASS |
| 3 | `enqueueWrite` guards merged-patch cache write by generation in addition to the handle check | PASS |
| 4 | Unit test exercises the mid-read folder switch and asserts both contamination sites stay clean | PASS |

**Counts: 4 PASS, 0 FAIL, 0 SKIP**

## Verification Details

**Item 1** (`let generation = 0\|generation++`):
- `let generation = 0;` at line 286 (module scope)
- `generation++` at lines 313 (`setLocalDirectory`) and 333 (`clearLocalDirectory`) — exactly 2 sites, one per mutation function

**Item 2** (`const gen = generation\|if (generation === gen) cachedModel = model\|return model`):
- `const gen = generation` at line 347 (inside `ensureLoaded`, at read-start)
- `if (generation === gen) cachedModel = model` at line 374 (guarded write)
- `return model` at line 375 (unconditional, immediately after the guarded write — in-flight writes still receive their folder's data)

**Item 3** (`const snapshotGen\|if (dirHandle !== snapshotHandle) return\|if (generation === snapshotGen) cachedModel = merged`):
- `const snapshotGen = generation` at line 393 (enqueue-time snapshot)
- `if (dirHandle !== snapshotHandle) return` at line 404 (original handle guard, still present)
- `if (generation === snapshotGen) cachedModel = merged` at line 407 (new generation guard)

**Item 4** (mid-read test structure):
- `makeGatedPreloadedDir` defined at line 141, used in test at line 815
- Test "drops an in-flight stale write when the switch lands mid-read" at line 814
- `setLocalDirectory(dirB, true)` at line 834 (switch fires during in-flight read)
- `getMetadata("a.mp3")` asserted `toBeUndefined` at lines 842 + 850 (site 2: enqueueWrite merged-patch leak)
- `getMetadata("diskA.mp3")` asserted `toBeUndefined` at lines 844 + 851 (site 1: ensureLoaded raw-disk-model leak)

## Bugs Found

None.

## Disposition

No residue — Step 3.5 Workflow was not run. All acceptance criteria are satisfied by the implementation as merged.

natb1:
## Review: PR #2335

**Partial coverage:** Security finders skipped: both quality finders (code-review, review) failed (model likely throttled). The security finder wave was not launched.

### Fixed
_None._

### Required (security)
_None._

_None._

### Informational
- `audio/src/sidecar.ts:90` — Net cyclomatic-complexity increase: total file complexity rose from 77 to 79. The added branching is the generation-guard itself (`if (generation === gen)` / `if (generation === snapshotGen)`), which is the fix under review — not incidental complexity. (erosion)
- `audio/test/sidecar.test.ts:33` — Net cyclomatic-complexity increase: total file complexity rose from 164 to 172. The increase reflects the new `makeGatedPreloadedDir` helper and the mid-read TOCTOU test case. (erosion)

### Dismissed
_None._

### False-positive (security)
_None._

### Deferred
_None._

### Out-of-scope (security)
_None._

PR #2339, opened 2026-06-22
dispatch-stop: session-scope in-flight background-task detection

### LINKED ISSUE #2261: dispatch-stop.sh: D1 whole-transcript scope for in-flight task detection may miss cross-session ID collisions
## Context

The \`session_has_inflight_background_task\` function in \`.claude/hooks/dispatch-stop.sh\` (lines 391–437, introduced in PR #2251) detects whether a background Workflow task is still in flight by scanning the entire session transcript for launch and notification entries. This whole-transcript scope has two robustness gaps:

1. **Cross-session ID collision:** If a workflow was launched in a previous session run and its transcript entries persist, a re-used task ID in a new session could produce a false positive match — causing the hook to believe a task is in flight when it is not.
2. **Pattern coupling:** The literal string \`launched in background. Task ID:\` is coupled to the Workflow tool's output format. If that format changes, the detection silently regresses with no compile-time or test-time signal.

## Safety note

The conservative \`return 1\` on uncertainty means any false positive causes a missed hand-back (the hook exits cleanly without parking), not a wrong park. Safety is maintained. This is a robustness improvement, not a correctness fix.

## Suggested changes

(a) Add a comment in \`dispatch-stop.sh\` adjacent to the pattern string explicitly coupling it to the Workflow tool's output format, so the comment is the trigger when that format changes.

(b) Evaluate whether session-scoped detection is feasible without restructuring the transcript data format (e.g. using the session ID already present in transcript metadata to filter entries to the current session turn before scanning for task launches).

## Acceptance criteria

- [ ] A comment is added in \`dispatch-stop.sh\` near the \`launched in background. Task ID:\` pattern string that names the Workflow tool output format as the source of truth for this string, so a format change prompts updating the pattern.
- [ ] A feasibility note or implementation is produced for session-scoped detection: either a code change that filters transcript entries to the current session before scanning, or a documented conclusion that restructuring is not feasible at this time, with the reasoning recorded (e.g. in the issue or a code comment).

natb1:
Closes #2261

## Problem

`session_has_inflight_background_task` in `.claude/hooks/dispatch-stop.sh`
(added in PR #2251 for #2243) decides whether a Stop-hook turn yielded to await a
live background Workflow/Task. It detected in-flight tasks by grepping the
**entire** transcript file for launch lines (`launched in background. Task ID:
<ID>`) and notification entries (`taskId":"<ID>"`).

A resumed session carries the prior run's transcript records forward into its own
`<sessionId>.jsonl`, and those carried records retain their **original**
`sessionId`. So a background task launched-but-never-notified by a now-dead prior
run leaves a launched-without-notification record in the current session's file —
a **false positive**. The hook then hands the session back (exit 0, no park, no
tick spawned) waiting on a notification that never arrives, and the chain silently
stalls. A scan of 4843 live transcript files found 65 multi-session files,
confirming the collision premise is real, not theoretical.

## Fix

**AC2 — session-scoped scan.** Resolve `CURRENT_SESSION_ID` from the Stop
payload's `.session_id` (primary, here-string-fed to `jq` per
`.claude/rules/shell-json.md`) with the `state.json` `.sessionId` as fallback,
then scope the scan to the current session's records via a literal-substring
`grep -F "\"sessionId\":\"$CURRENT_SESSION_ID\""` before the launch/notify
extractions. Carried-forward prior-session records are dropped, so the false
positive disappears.

The whole-file fallback (today's behavior) fires **only** when
`CURRENT_SESSION_ID` is unresolvable — gated strictly on `[ -z
"$CURRENT_SESSION_ID" ]`, never on "the scoped scan found no launch." This
asymmetry is load-bearing and documented in the function docblock: the live
#2243 case stays safe because a still-running session's own launch record carries
that same session's `sessionId` and survives the filter, so genuine in-flight
phases still hand back and are never wrong-parked.

**AC1 — source-of-truth comment.** A maintenance note next to the launch-pattern
grep names the harness `tool_result` output format as the origin of the Task-ID
launch/notify patterns (there is no in-repo definition), so a future harness
format change has an explicit update trigger.

## Tests

Two new cases in `test-dispatch-scripts.sh`, mirroring the existing D1.1–D1.3
harness shape, both with top-level `sessionId` fixtures and a `session_id`
payload:

- **D1.4** — a stale prior-session launch is ignored; with no current-session
  launch the hook parks normally (the #2261 fix).
- **D1.5** — a current-session launch with no notification is still detected as
  in-flight and hands back, despite a stale completed other-session launch (the
  #2243 live-case guard).

D1.1–D1.3 are unchanged and still exercise the whole-file fallback (no
`sessionId`/`session_id`). Full suite: `3926/3926 passed, 0 failed`.

### DIFF
--- a/.claude/hooks/dispatch-stop.sh
+++ b/.claude/hooks/dispatch-stop.sh
@@ -132,6 +132,21 @@ if [ -z "$TRANSCRIPT_PATH" ] && [ -n "${CLAUDE_JOB_DIR:-}" ] && [ -f "$STATE_FIL
   fi
 fi
 
+# Resolve the current session ID for session_has_inflight_background_task's
+# session-scoped scan (#2261). PAYLOAD is the PRIMARY source: it reflects the
+# actually-running session, whereas state.json can be stale across a re-stamp.
+# Here-string feeds jq (NOT `echo "$PAYLOAD" | jq` — zsh echo un-escapes \t/\n
+# and corrupts the JSON; see .claude/rules/shell-json.md). Fallback when the
+# payload yields nothing: reuse the `_sid` already parsed from $STATE_FILE in the
+# block above (do not re-read). `${_sid:-}` is mandatory — `_sid` is only set
+# inside that `if` block, and under `set -u` + the ERR trap a bare unset
+# reference would abort the whole hook. When CURRENT_SESSION_ID stays empty the
+# scan degrades to the whole file (today's behavior).
+CURRENT_SESSION_ID=$(jq -r '.session_id // empty' <<<"$PAYLOAD" 2>/dev/null) || CURRENT_SESSION_ID=""
+if [ -z "$CURRENT_SESSION_ID" ] && [ -n "${_sid:-}" ]; then
+  CURRENT_SESSION_ID="$_sid"
+fi
+
 # Resolve issue number from the validated JOB_NAME (<N>-<slug>). Discriminator 2
 # guarantees JOB_NAME matches ^[0-9]+-, so the numeric prefix is non-empty.
 ISSUE_NUM="${JOB_NAME%%-*}"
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — Issue #2261 (PR #2339)

## Items Executed

| # | Title | Result |
|---|-------|--------|
| 1 | AC1 — UPDATE TRIGGER comment present adjacent to launch pattern | PASS |
| 2 | AC2 — CURRENT_SESSION_ID resolved from payload `.session_id` primary, state.json fallback | PASS |
| 3 | AC2 — session-scoped scan with whole-file fallback gated ONLY on empty CURRENT_SESSION_ID | PASS |
| 4 | AC2 safety — docblock documents load-bearing fallback asymmetry | assessed by Opus → already satisfied |
| 5 | Regression tests D1.4 / D1.5 present and full suite passes | PASS |

**Counts:** PASS: 4, FAIL: 0, SKIP: 0

## Findings

None. All acceptance criteria are met.

## Disposition Triage

The residue list had one item (item 4, `needs-human-judgment`), so the disposition Workflow ran.

- **Item 4 — AC2 safety — docblock documents load-bearing fallback asymmetry** (`needs-human-judgment` → `already-satisfied`)
  - Rationale: The docblock above `session_has_inflight_background_task` (lines 439-459) provably contains all three required facts: (1) the whole-file fallback fires ONLY on unresolvable `CURRENT_SESSION_ID` (lines 450-454); (2) it must not be simplified away per `code-style.md` (lines 450-452, 455-459); (3) the #2261 vs #2243 asymmetry is named explicitly (lines 439-448, 459). The criterion is met by readable code evidence — no code change needed, no human required.

`result.dispositions` is empty (the only residue item was `already-satisfied`). No `needs-main` items to file. No `needs-human` items to escalate. Clean pass.

natb1:
## Review: PR #2339 (dispatch-stop.sh session-scoped D1 scan)

**Surface:** `code` | **Security:** CodeQL 0 alerts, npm audit skipped (no dep changes), erosion 0 findings
**Findings:** 9 surfaced | 0 actionable | 0 fixes applied

---

### Dismissed

| Finding | Location | Rationale |
|---|---|---|
| Transcript `sessionId` format verified correct | `dispatch-stop.sh:474` | The `"sessionId":"<id>"` form (no space, raw quotes) is correct for JSONL top-level fields; `grep -F` matches it reliably. No change needed. |
| `grep` exit-1 in command substitution is safe without explicit `\|\|` guard | `dispatch-stop.sh:474` | The existing `set -e` / ERR trap is already suppressed for `grep` in this context; adding `\|\| scan_src=""` is not required. |
| Task-notification records carry `sessionId` — scoped scan correctly handles completed tasks | `dispatch-stop.sh:474-492` | Verified: notification records carry the current session's `sessionId` so the scoped scan correctly drops completed tasks. No code change needed. |
| `grep -F` correctly neutralizes all regex/glob/quote metacharacters in session ID | `dispatch-stop.sh:474` | `grep -F` treats the pattern as a literal string, so special characters in the session ID pose no risk. No change needed. |
| Empty `scan_src` handled safely by existing `launched` guard | `dispatch-stop.sh:481-483` | `[ -n "$launched" ] || return 1` already handles an empty/no-match `scan_src` cleanly. No change needed. |

### Informational

- **`sid-fallback-scope-doc-gap`** (`dispatch-stop.sh:140-148`): The comment says "empty payload AND no state.json sessionId" but the fallback fires only when the payload also lacks `transcript_path` (the `_sid` block is inside a `transcript_path`-absent guard). Optionally tighten to: "transcript_path also absent from payload AND state.json has no sessionId." No code change required.
- **`missing-scoped-completion-test`** (`test-dispatch-scripts.sh:25128-25172`): No test covers the scoped path where the current-session task is completed (launch + notification both present with `cur-session-aaa`). Consider a D1.6 case: assert Branch A park fires (not in-flight). Not required for this PR.

### False-positive (security)

- **`grep-f-newline-multipattern`** (`dispatch-stop.sh:474`): A newline embedded in `CURRENT_SESSION_ID` would cause `grep -F` to treat it as two patterns. Adversarially verified as **not exploitable**: session IDs are UUIDs generated by the Claude Code harness and contain no newlines; the attack requires control over harness internals, not just the Stop payload. No change required. (Optional hardening: `CURRENT_SESSION_ID=${CURRENT_SESSION_ID%%$'\n'*}` after resolution.)
- **`toctou-transcript-file`** (`dispatch-stop.sh:461,469,474`): TOCTOU window between the readability check and the file read is fail-closed — a deletion between those points yields an empty `scan_src`, which returns 1 (not in-flight), the safe conservative outcome. Not exploitable.

---

_No fixes applied. No follow-up issues filed._

PR #2344, opened 2026-06-22
dispatch-stamp-session: narrow TRANSCRIPT_PATH traversal check to anchored .. sequences

natb1:
Closes #2269

## What

Narrow the `TRANSCRIPT_PATH` path-traversal guard in `dispatch-stamp-session`
from `*..*|*[[:cntrl:]]*` to an anchored-sequence pattern:

```
'..'|'../'*|*'/..'|*'/../'*|*[[:cntrl:]]*
```

The five alternatives match an actual `..` traversal sequence — a bare `..`, a
leading `../`, a trailing `/..`, or an embedded `/../` — plus any control
character.

## Why

The old `*..*` branch matched `..` **anywhere** in the string, including as a
substring of a single legitimate path segment (e.g.
`/home/user/..foo/bar.jsonl`), which is broader than the documented intent
(reject only the `..` traversal sequence and control chars). Transcript paths
come from the controlled Claude harness SessionStart payload, so this is a
latent bug rather than a live exploit — but the guard's behavior was broader
than its comment claimed, which could falsely reject a legitimate path if the
source ever changed, and the code/comment mismatch misled future readers.

The fix is anchored-glob detection, chosen over `realpath --canonicalize-missing`
because realpath would resolve symlinks (a semantic change) and require
filesystem access that the controlled source does not warrant. The change is
backwards-compatible — it only stops rejecting paths that should never have been
rejected.

The `SESSION_ID` and `branch` guards intentionally keep their
`*..*|*/*|*[[:cntrl:]]*` pattern: both also forbid any slash, so the value is a
bare stem where rejecting `..` as a substring is correct. They are untouched.

## Tests

`test-dispatch-scripts.sh` gains three cases in the
`=== dispatch-stamp-session ===` block:

1. `..` as a filename substring (`sess..x.jsonl`) now **passes** — sidecar
   written, `.issue == 999` (rejected by the old pattern, accepted by the new).
2. Embedded `/../` (`$d/sub/../evil.jsonl`) exits 2, no sidecar — exercises the
   `*'/../'*` branch.
3. A control char (literal tab) in the transcript path exits 2, no sidecar.

The pre-existing leading-`../` case stays green under the new `'../'*`
alternative. Full suite: 3921/3921 passing.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-stamp-session
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-stamp-session
@@ -174,7 +174,10 @@ fi
 # Reject path-traversal / control chars. The session id is a bare stem (no path
 # component), so a slash, `..`, or control char is malformed → exit 2. The
 # transcript path is a legitimate absolute path (slashes expected), so reject
-# only `..` and control chars there.
+# only an actual `..` traversal sequence (anchored: a bare `..`, a leading
+# `../`, a trailing `/..`, or an embedded `/../`) and control chars — NOT a
+# `..` substring inside a single path segment (e.g. `/home/user/..foo/bar.jsonl`
+# is a legitimate path and must not be rejected).
 case "$SESSION_ID" in
   *..*|*/*|*[[:cntrl:]]*)
     echo "dispatch-stamp-session: --session-id contains illegal characters: $SESSION_ID" >&2
@@ -182,7 +185,7 @@ case "$SESSION_ID" in
     ;;
 esac
 case "$TRANSCRIPT_PATH" in
-  *..*|*[[:cntrl:]]*)
+  '..'|'../'*|*'/..'|*'/../'*|*[[:cntrl:]]*)
     echo "dispatch-stamp-session: --transcript-path contains illegal characters: $TRANSCRIPT_PATH" >&2
     exit 2
     ;;
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -38914,9 +38914,9 @@ STAMP="$SCRIPT_DIR/dispatch-stamp-session"
   git -C "$d" -c user.email=t@t -c user.name=t commit -q --allow-empty -m init
   ( cd "$d" && "$STAMP" --session-id sessok --transcript-path "$d/sess..x.jsonl" )
   sc="$d/sess..x.dispatch-stamp.json"
-  assert_eq "stamp: ..‐in‐filename passes (sidecar written)" "yes" \
+  assert_eq "stamp: ..-in-filename passes (sidecar written)" "yes" \
     "$([ -f "$sc" ] && echo yes || echo no)"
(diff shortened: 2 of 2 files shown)


natb1:
# QA Summary — Issue #2269 / PR #2344

## Execution

All 5 triage items executed across 1 lane (shell-command). No browser component detected.

| # | Title | Result |
|---|-------|--------|
| 1 | Path-traversal guard accepts `..` substring inside a single path segment | PASS |
| 2 | Actual `../` and `/../` traversal sequences are still rejected | PASS |
| 3 | Control character in transcript path is still rejected | PASS |
| 4 | Full dispatch-script suite passes (no regression) | PASS |
| 5 | Inline comment accurately describes the post-fix guard behavior | PASS (already-satisfied) |

Test suite: 3928/3928 passing.

- **Item 5 — Inline comment accurately describes the post-fix guard behavior:** AC #5 requires the inline comment to accurately describe the post-fix guard with no stale reference to the old over-broad `*..*` behavior. Reading the file confirms lines 174–180 describe exactly the anchored patterns implemented at line 188: bare `..`, leading `../`, trailing `/..`, embedded `/../`, and control chars; and explicitly notes a `..` substring inside a single path segment (e.g. `/home/user/..foo/bar.jsonl`) is NOT rejected. The comment matches the case pattern exactly with no stale description. Criterion provably met by readable source evidence — no code defect, no human judgment required.

## Result

Clean pass. All acceptance criteria satisfied. `dispatch:qa-done` applied.

natb1:
## Review summary — #2269

**Surface:** `code` · **CodeQL:** no open alerts · **Erosion:** no findings · **Dependency audit:** not applicable (no dep changes)

### Fixed (commit `5caf9979`)

- **Test 12 vacuously-true sidecar assertion** (`test-dispatch-scripts.sh:38936`) — `[ -f "$d/sub/../evil.dispatch-stamp.json" ]` was vacuously true because `$d/sub/` was never created, making the "no sidecar written" guard meaningless. Added `mkdir -p "$d/sub"` before the `$STAMP` invocation so the assertion is structurally sound.
- **Missing test for trailing `/..` guard arm** (`test-dispatch-scripts.sh:38884`) — the new `*'/..'` alternative in the TRANSCRIPT_PATH guard had no dedicated test. Added test 14 (`--transcript-path "$d/sub/.."`) asserting exit 2 and no sidecar, covering the trailing-`/..` arm symmetrically with the leading-`../` arm already tested.
- **U+2010 HYPHEN in test 11 label strings** (`test-dispatch-scripts.sh:38917`) — the two `assert_eq` labels used Unicode HYPHEN (‐, U+2010) instead of ASCII hyphen-minus (-), inconsistent with all other test labels in the file. Replaced with ASCII `-`.

### Dismissed

- **Partial test coverage for trailing `/..` arm** (`test-dispatch-scripts.sh:38959`) — the new test 14 added in the Fixed set above directly addresses this; the finding is superseded by that fix.

### Security

_No attack surface findings. The diff touches only shell scripts and test code — no executable production source, config, dependency, or Firestore-rules changes. CodeQL: no open alerts. Dependency audit: not applicable._

### Informational

_None._

PR #2309, opened 2026-06-21
audio: fail closed when the sidecar is corrupt so a later write can't clobber playlists

natb1:
Closes #2304

## Problem

`parseSidecar()` in `audio/src/sidecar.ts` was too tolerant of a genuinely
corrupt `.commons-audio/index.json`: on a JSON parse error or a non-object root
it logged and returned `emptyModel()` — an empty but valid-looking model. That
empty model was then cached by `ensureLoaded()`, so the next routine
`enqueueWrite()` (e.g. `savePlayerState()` persisting the playback position)
merged its patch against the empty cache and wrote the result back,
**permanently destroying every named playlist** that was in the corrupt file.

## Fix — fail closed

When the on-disk file exists but cannot be parsed, treat it as corrupt: keep the
in-memory session fully functional (start from an empty model so playback still
works), but **never overwrite the corrupt file on disk**, preserving the user's
bytes for recovery. A missing file (`NotFoundError`) is *not* corruption — the
new-folder path keeps working and still writes.

- `parseSidecar` / `readSidecar` now return `SidecarData | null`, using `null`
  as the corrupt sentinel.
- A module-level `corruptOnDisk` flag is set in `ensureLoaded()` when the load
  resolves `null`; the session proceeds from `emptyModel()`. The flag is reset in
  `setLocalDirectory` / `clearLocalDirectory` so re-binding a good folder
  re-enables writes (and tests stay isolated).
- `enqueueWrite` still merges in-memory (the session stays consistent) but skips
  the disk write while `corruptOnDisk` is set, logging a clear skip diagnostic.
- `ensureLoaded` keeps its non-null `SidecarData` return, so every accessor and
  external caller (`player.ts`, `local-source.ts`) is untouched.

### One intentional behavior change

`readSidecar()`'s catch-all branch (an unexpected, non-`NotFoundError` read
error) now returns the corrupt sentinel (`null`) instead of `emptyModel()`. An
unexpected error on a file that may exist with data means we cannot prove it is
safe to overwrite, so disk writes are suppressed for the session (in-memory stays
functional). This is the deliberate fail-closed choice per
`.claude/rules/code-style.md`, not scope creep.

## Tests

`audio/test/sidecar.test.ts`: updated the existing `parseSidecar`/`readSidecar`
corrupt-input assertions to expect `null` (behavior-change update, not
weakening), fixed the `SidecarData | null` type ripple in the valid-input tests,
and added two acceptance-criteria tests — a corrupt sidecar then
[...]

PR #2360, opened 2026-06-22
audio: extend AC4 corrupt-sidecar write-suppression test to all four write entry points

natb1:
Closes #2331

## Summary

The AC4 corrupt-sidecar test (added in PR #2309 for #2304) asserts that
`savePlayerState` does **not** write to disk when the loaded sidecar is corrupt
— fail-closed, so a later write can't clobber a user's playlists. But all four
public write paths — `savePlayerState`, `cacheMetadata`, `cacheMetadataBatch`,
and `savePlaylist` — funnel through the shared private `enqueueWrite` gate in
`audio/src/sidecar.ts`, and the existing test exercises only one of them.

This adds an `AC4b` sibling test inside the same
`describe("corrupt sidecar — fail-closed", …)` block that sets up the corrupt-dir
fixture once, calls **all four** public write entry points on it, flushes, and
asserts the shared gate suppressed every write (`createWritable` never called,
corrupt bytes on disk preserved). A future refactor that moved the corrupt-dir
gate out of `enqueueWrite` into the individual write functions could otherwise
silently drop fail-closed protection from the three uncovered entry points while
the suite stayed green — this closes that gap.

## Changes

- `audio/test/sidecar.test.ts` — one new `it("…all four write entry points are
  gated (AC4b)")` test. No production code changes; `audio/src/sidecar.ts` is
  correct and untouched.

## Verification

`npx vitest run --project audio --root .` — 251 passed (1 pre-existing skip),
including the new AC4b test.

### DIFF
--- a/audio/test/sidecar.test.ts
+++ b/audio/test/sidecar.test.ts
@@ -1022,6 +1022,22 @@ describe("corrupt sidecar — fail-closed", () => {
     expect(await getPlayerState()).toMatchObject({ queue: ["a.mp3"], positionSeconds: 12 });
   });
 
+  it("corrupt sidecar: all four write entry points are gated (AC4b)", async () => {
+    const { dir, fileHandle } = makePreloadedDir("{not json");
+    setLocalDirectory(dir, true);
+
+    await savePlayerState({ queue: ["a.mp3"], currentLocalName: "a.mp3", positionSeconds: 12 });
+    await cacheMetadata("song.mp3", { tags: { title: "X" }, size: 10, lastModified: 1 });
+    await cacheMetadataBatch({ "a.mp3": { tags: { title: "A" }, size: 1, lastModified: 1 } });
+    await savePlaylist("Favs", ["a.mp3", "b.mp3"]);
+    await flushWrites();
+
+    // No caller wrote — the shared corrupt-dir gate suppressed every write path.
+    expect(fileHandle.createWritable).not.toHaveBeenCalled();
+    // Corrupt bytes on disk are preserved for recovery.
+    expect(fileHandle._state.content).toBe("{not json");
+  });
+
   it("missing sidecar (NotFoundError): savePlayerState DOES write (AC5)", async () => {
     const dir = makeDirWithMissingFile();
     setLocalDirectory(dir, true);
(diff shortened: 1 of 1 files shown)


natb1:
# QA Summary — Issue #2331 / PR #2360

## Results

| # | Item | Lane | Result |
|---|------|------|--------|
| 1 | New AC4b test calls all four write entry points | shell-command | PASS |
| 2 | Test asserts the shared write gate via createWritable never called | shell-command | PASS |
| 3 | Audio unit suite passes | shell-command | PASS |
| 4 | Test genuinely exercises the gate (no false-positive risk) | needs-human-judgment | → assessed by Opus (already satisfied) |

**PASS: 3 | FAIL: 0 | SKIP: 0 | Deferred: 0**

No browser walkthrough (no browser component — pure test addition, `audio/test/sidecar.test.ts` only).

## Bugs found

None.

## Disposition triage

Step 3.5 ran (one `needs-human-judgment` residue item). The Workflow assessed it as **already-satisfied** — dropped as PASS.

- **Test genuinely exercises the gate (no false-positive risk):** The triage deferred confirmation that removing the production gate would fail the test and that all four callers are exercised. Opus verified both from code/test evidence: disabling the only suppression gate (`if (corruptOnDisk) return;`) causes the `createWritable not called` + corrupt-bytes-preserved assertions to fail; all four entry points (`savePlayerState`, `cacheMetadata`, `cacheMetadataBatch`, `savePlaylist`) in `audio/src/sidecar.ts` funnel through the single `enqueueWrite` gate, and the test drives all four then asserts zero writes. Criterion already met; no defect to fix.

No `needs-main` items to file. No `needs-human` items remain. Escalation set is empty — clean pass.

natb1:
## Review findings — #2331 AC4b corrupt-sidecar write-suppression

### Fixed

- **`audio/test/sidecar.test.ts:1025`** — AC4b test was missing positive in-memory assertions for the three newly-tested callers (`cacheMetadata`, `cacheMetadataBatch`, `savePlaylist`). The original AC4 test asserts that `getPlayerState()` still returns the expected state even after a corrupt-load (confirming the in-memory session remains usable); AC4b lacked equivalent checks for its three additional callers. Added `getPlayerState`, `getMetadata`, and `getPlaylists` assertions after `flushWrites()` to mirror that pattern. _(fix commit: 922e31ad)_

### Dismissed

- **`audio/test/sidecar.test.ts:1033`** — `await flushWrites()` is redundant: all four write calls (`savePlayerState`, `cacheMetadata`, `cacheMetadataBatch`, `savePlaylist`) await their work inline, so flushes are complete before any assertion. However the call is a style choice that makes intent explicit and mirrors the AC4 test structure; removing it would not improve correctness. _Dismissed — not a real improvement._

### Security

### Deferred

_None._

### Out-of-scope

_None._

PR #2363, opened 2026-06-22
audio: test that rebinding a clean folder after a corrupt one re-enables sidecar writes

natb1:
Closes #2332

## What

Adds one behavioural test to the `corrupt sidecar — fail-closed` block in
`audio/test/sidecar.test.ts` covering the rebind path introduced by PR #2309:
binding a corrupt directory suppresses disk writes, and rebinding a clean folder
clears `corruptOnDisk` so disk writes resume.

The test binds a corrupt directory, asserts `savePlayerState` + `flushWrites`
does not touch the file, then rebinds a clean empty directory, calls
`savePlayerState` + `flushWrites` again, and reads back the sidecar to confirm
the new state was persisted.

## Why

`setLocalDirectory` resets the corrupt-on-disk flag when a new directory is
bound, but no test exercised the full path (bind corrupt → writes suppressed →
bind clean → writes resume). A regression there could silently suppress all disk
writes on the new folder — the same data-loss risk the fail-closed behaviour was
built to prevent.

## Acceptance criteria notes

Per the user's office-hours resolution, the original AC2 ("test fails if the
`factory.ts:179` reset line is removed") is not satisfiable: `corruptOnDisk` is
reset in two independent, defense-in-depth places (`factory.ts:179` and the
`ensureLoaded` success branch at `:227`), and rebinding a clean dir clears the
flag via `:227` regardless of `:179`. The test is therefore sensitive to the
genuine regression — both resets removed makes it fail — and is intentionally not
sensitive to removing a single reset line. The mutation triplet was validated
locally during planning; it is not re-run in CI.

Test-only change. No source modifications.

### DIFF
--- a/audio/test/sidecar.test.ts
+++ b/audio/test/sidecar.test.ts
@@ -1072,6 +1072,6 @@ describe("corrupt sidecar — fail-closed", () => {
 
     const readBack = await readSidecar(cleanDir);
     if (!readBack) throw new Error("expected non-null sidecar");
-    expect(readBack.playerState).toMatchObject({ queue: ["a.mp3"], positionSeconds: 12 });
+    expect(readBack.playerState).toEqual({ queue: ["a.mp3"], currentLocalName: "a.mp3", positionSeconds: 12 });
   });
 });
--- a/audio/test/sidecar.test.ts
+++ b/audio/test/sidecar.test.ts
@@ -1033,4 +1033,24 @@ describe("corrupt sidecar — fail-closed", () => {
     expect(readBack).not.toBeNull();
     expect(readBack?.playerState).toMatchObject({ queue: ["a.mp3"], positionSeconds: 5 });
   });
+
+  it("rebinding a clean dir after a corrupt one re-enables disk writes", async () => {
+    // Corrupt folder: writes are suppressed, corrupt bytes preserved.
+    const { dir: corruptDir, fileHandle: corruptFile } = makePreloadedDir("{not json");
+    setLocalDirectory(corruptDir, true);
+    await savePlayerState({ queue: ["x.mp3"], positionSeconds: 1 });
+    await flushWrites();
+    expect(corruptFile._state.content).toBe("{not json");
+    expect(corruptFile.createWritable).not.toHaveBeenCalled();
+
+    // Rebind a clean folder → corrupt flag cleared → writes resume.
+    const cleanDir = makeEmptyDir();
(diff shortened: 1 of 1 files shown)


natb1:
## QA Summary — #2332 (PR #2363)

**Result: PASS — all items satisfied**

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | New rebind test present in the corrupt-sidecar describe block | shell-command | ✓ PASS |
| 2 | Full audio suite passes with the new test | shell-command | ✓ PASS |
| 3 | Test is sensitive to the corrupt-on-disk reset (mutation triplet) | shell-command | ✓ PASS |

**Counts:** 3 passed, 0 failed, 0 skipped. No browser walkthrough (test-only change, no frontend component).

### Detail

**Item 1 — Test presence:** `grep` found the new `it("rebinding a clean dir after a corrupt one re-enables disk writes", ...)` at line 1037, inside the `describe("corrupt sidecar — fail-closed", ...)` block (lines 1005–1056). ✓

**Item 2 — Full suite:** `npx vitest run --project audio --root .` → 251 passed, 1 skipped, 0 failures. ✓

**Item 3 — Mutation triplet:**
- Both `factory.ts:179` and `:227` removed → rebind test **FAILS** (writes stay suppressed on clean rebind). ✓
- Only `factory.ts:179` removed → all 251 tests **PASS** (`:227` alone clears the flag). ✓
- Only `factory.ts:227` removed → all 251 tests **PASS** (`:179` alone clears the flag). ✓

The test guards the genuine regression (both resets absent = silent data loss on clean rebind) and is correctly insensitive to removing either reset individually (defense-in-depth by design). `factory.ts` was restored to its original state after the mutation run.

No residue. No deferred items.

natb1:
## Review: #2332

### Fixed

- **`audio/test/sidecar.test.ts:1075`** — Final disk-read assertion used `toMatchObject` (partial match) instead of `toEqual` (exact match), so unexpected extra fields would pass undetected. Strengthened to `toEqual` with the full expected shape including `currentLocalName`. (commit 805c3f71)

### Informational

- **`audio/test/sidecar.test.ts:1058`** — The test confirms disk writes resume after rebinding via a `readSidecar` call on the clean dir, but does not assert the in-memory state (e.g. `corruptOnDisk` flag) directly. The observable disk-write coverage is sufficient for the regression being guarded, but a future reader may want to know the in-memory flag is reset as a side note.

### Security

PR #2380, opened 2026-06-22
dispatch-stop: anchor in-flight-task notified-set on the <task-notification> envelope

natb1:
Closes #2365

## Problem

The in-flight-background-task guard `session_has_inflight_background_task` in
`.claude/hooks/dispatch-stop.sh` (added for #2243) never fired. It is meant to
keep Stop-hook Branch A from parking a `qa`/`review` phase session on
`dispatch:office-hours` while that session has launched a background Workflow
fan-out and yielded its turn to await a `<task-notification>`. Instead the guard
self-defeated and the session was wrongly parked — the exact symptom #2243 set
out to prevent.

## Root cause

The guard computes the set difference `launched ∖ notified`. The `notified`
pattern `taskId\?":\?"[A-Za-z0-9_-]+` matched **both** the escaped
`\"taskId\":\"` form (real notifications, where the payload is a string value)
**and** the bare `"taskId":"` form. The Workflow launch `tool_result` record
carries a sibling metadata object
`"toolUseResult":{"status":"async_launched","taskId":"<ID>",…}` whose `taskId`
is in the **bare** form. So at the instant a task launched, its ID appeared in
**both** `launched` and `notified`; the difference was empty; the guard returned
"not in-flight"; Branch A parked. The guard was only ever masked from the bug by
the separate `session_scheduled_wakeup` (#1590) gate when the yielding turn also
scheduled a wakeup.

The existing #2243/#2261 tests passed because their launch fixtures used a
plain-string `content` and **omitted** the `toolUseResult` launch-metadata
object, so the bare `taskId` that triggers the bug was absent — the fixtures
were unfaithful to the real launch-record shape.

## Fix

Anchor the `notified` scan on the `<task-notification>` envelope: a
`grep -F 'task-notification'` filter stage runs before the `taskId` extraction,
so only envelope-bearing records contribute notified IDs. The launch record has
no envelope, so its bare `toolUseResult.taskId` no longer pollutes `notified`.
This is the semantically correct discriminator ("an ID is notified iff a
notification envelope carried it") and is orthogonal to the bare-vs-escaped
quoting, so the existing `\?` quote tolerance is preserved inside the
envelope-filtered records. The `launched` extraction and the #2261 session-scope
logic are untouched.

## Tests

- New case **#2365 D1.6**: a launch fixture reproducing the **real** shape — a
  `tool_result` line carrying the top-level
  `"toolUseResult":{"status":"async_launched","taskId":"<ID>",…}` bare metadata,
  with no `<task-notification>` record — asserts the guard reports in-flight
  (hand-back, no park, no spawn, no self-close). This case is **red against the
  unfixed pattern** and green after the fix.
- New converse case **#2365 D1.7**: the same faithful launch record plus a real
  `<task-notification>` record for the same ID → normal Branch A park.
- The existing #2243/#2261 fixtures are made **faithful** to the real transcript
  format: launch lines gain the real `toolUseResult` bare-`taskId` metadata, and
  the three bare-`tool_result` notification fixtures (D1.2/D1.3/D1.5) are
  rewritten as `assistant`/`<task-notification>` envelope records. The
  assertions are unchanged — this is a faithfulness correction, not a weakening.

Red-first was verified during the build: against the unfixed hook, exactly the
four predicted in-flight cases (D1.1, D1.3, D1.5, D1.6) parked instead of
handing back; after the fix the full suite passes (3966/3966).

The end-to-end behavioural AC (a real `qa`/`review` session that launches a
background Workflow and yields is handed back, not parked) is observed in
production over subsequent dispatch ticks — it is not unit-testable here.

### DIFF
--- a/.claude/hooks/dispatch-stop.sh
+++ b/.claude/hooks/dispatch-stop.sh
@@ -475,16 +475,23 @@ session_has_inflight_background_task() {
   fi
   # UPDATE TRIGGER: this literal `launched in background. Task ID: <ID>` is the
   # Workflow/Task tool's `tool_result` output format, emitted by the Claude Code
-  # harness — there is no in-repo definition to track. If the harness changes that
-  # output format, THIS launch pattern AND the matching `taskId":"<ID>"`
-  # notification pattern below must both be updated.
+  # harness — there is no in-repo definition to track. The notification pattern
+  # below is now anchored on the `<task-notification>` envelope; if the harness
+  # renames that wrapper, BOTH this launch pattern AND the envelope anchor must be
+  # updated.
   launched=$(printf '%s\n' "$scan_src" | grep -oE 'launched in background\. Task ID: [A-Za-z0-9_-]+' \
     | grep -oE '[A-Za-z0-9_-]+$' | sort -u)
   [ -n "$launched" ] || return 1
-  # Tolerate the JSONL backslash-escaped quote form (\"taskId\":\"<ID>\") as
-  # well as the bare "taskId":"<ID>" form — the <task-notification> payload is a
-  # string value, so its inner quotes are backslash-escaped in the record.
-  notified=$(printf '%s\n' "$scan_src" | grep -oE 'taskId\\?":\\?"[A-Za-z0-9_-]+' \
+  # Notifications are identified by the `<task-notification>` envelope, NOT a bare
+  # `taskId` substring: the launch record's own sibling `toolUseResult.taskId` also
+  # carries the bare "taskId":"<ID>" form, so a substring-only match would
+  # self-match the launch into `notified`, empty the set difference, and wrongly
+  # report not-in-flight (#2365). The grep -F 'task-notification' stage scopes the
+  # taskId extraction to envelope records only. Within those, tolerate the JSONL
+  # backslash-escaped quote form (\"taskId\":\"<ID>\") as well as the bare form —
+  # the envelope payload is a string value, so its inner quotes are escaped.
+  notified=$(printf '%s\n' "$scan_src" | grep -F 'task-notification' \
+    | grep -oE 'taskId\\?":\\?"[A-Za-z0-9_-]+' \
     | grep -oE '[A-Za-z0-9_-]+$' | sort -u)
   # In-flight iff at least one launched ID has no matching notification, i.e.
   # the set difference (launched ∖ notified) is non-empty. comm -23 lists lines
--- a/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh
@@ -25101,7 +25101,7 @@ export CLAUDE_JOB_DIR="$TMPDIR_TEST/jobs/abcd1234"
 # Fixture: one Workflow launch line, no <task-notification>.
 cat > "$TMPDIR_TEST/transcript.jsonl" <<'EOF'
 {"type":"assistant","message":{"content":[{"type":"tool_use","name":"Bash","input":{}}]}}
-{"type":"user","message":{"content":[{"type":"tool_result","content":"Workflow launched in background. Task ID: w4stq5fyf"}]}}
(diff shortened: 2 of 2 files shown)


natb1:
# QA Summary — Issue #2365 / PR #2380

## Results

**5 items executed · 5 PASS · 0 FAIL · 0 SKIP**
**1 planned-deferral item filed as a needs-main follow-up**

No bugs found. No browser walkthrough (pure script/hook change).

---

## Item Results

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Notified-set scan anchored on `<task-notification>` envelope in dispatch-stop.sh | shell-command | PASS |
| 2 | New #2365 D1.6 and D1.7 test cases exist with faithful launch-record shape | shell-command | PASS |
| 3 | Existing #2243/#2261 fixtures (D1.1–D1.5) carry faithful `toolUseResult` launch shape | shell-command | PASS |
| 4 | Full dispatch-scripts test suite passes (3966/3966) | shell-command | PASS |
| 5 | dispatch-stop.sh is syntactically valid (bash -n) | shell-command | PASS |
| 6 | Production behavior: in-flight Workflow sessions handed back, not parked | deferred | filed as #2383 |

---

## Disposition Triage

**1 residue item** was classified by the disposition Workflow:

| 6 | Production behavior: a real qa/review session launching a background Workflow is handed back, not parked | `needs-main` | n/a | `planned_deferral: true` — AC3 is documented as non-assertable at merge time; only observable in production over subsequent dispatch ticks. No Opus-fixable defect. |

---

## Filed Follow-ups

| Identifier | Issue | Route |
|-----------|-------|-------|
| `qa-needs-main #2365: production-behavior-a-real-qa-review-session-launching-a-background-workflow-is-handed-back-not-parked` | #2383 (newly filed) | `human` — internal dispatch behavior, only observable by monitoring production ticks; routed to office-hours human review |

---

## Deferred to Office-Hours

None — the single residue item classified `needs-main` (planned-deferral), not `needs-human`.

---

## Clean pass

All script-verifiable acceptance criteria pass. The one non-assertable criterion (AC3 — end-to-end production behavior) is filed as a `blocked_by` follow-up (#2383) for post-merge verification via office-hours. `dispatch:qa-done` will be applied.

natb1:
## Review — #2365

8 subagents · 3 findings surfaced · 0 actionable · 0 fixes applied · 0 follow-ups filed · coverage complete

### Informational (2)

**code-review — `.claude/hooks/dispatch-stop.sh:493`**
The `grep -F 'task-notification'` pre-filter assumes all task-completion signals arrive wrapped in a `<task-notification>` envelope in an assistant text record. If the harness ever delivers completion without the envelope (e.g., as a bare `tool_result` carrying `taskId`), the `notified` set stays empty, the set difference stays non-empty, and `session_has_inflight_background_task` returns 0 indefinitely — stalling the chain. The `UPDATE TRIGGER` comment documents this coupling but does not guard against silent drift. _No fix required: the coupling is intentional and documented; D1.6/D1.7 tests cover both notification shapes. A future harness-format change must update both the launch pattern and the envelope anchor together._

**red-team — `.claude/hooks/dispatch-stop.sh:493-495`**
The `grep -F 'task-notification'` filter matches the literal substring anywhere in a JSONL line. A JSONL line whose `content` embeds `task-notification` as free text (e.g., a model turn narrating what it is waiting for, or a tool output that includes the string) could plant a spurious `taskId` into `notified`, making the set difference empty and causing Branch A to park while a real task is still in flight. The angle-bracket delimiters (`<task-notification>`) reduce but do not eliminate the false-match surface. Confidence: medium. _No fix required: a model turn that embeds `<task-notification>` alongside a valid `taskId` within the same line is an extremely narrow ambient-content path; the UPDATE TRIGGER comment flags the coupling for reviewers. A hardening follow-up (jq-based parse) is tracked separately if needed._

### Dismissed (1)

**red-team — `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:25308`**
The D1.6/D1.7 test payloads shell-interpolate `$FIXTURE` (a temp-dir path) directly into a `printf`-formatted JSON string. If `TMPDIR` contained a double-quote, backslash, or newline, the JSON would be malformed. In practice `TMPDIR_TEST` is set to `mktemp -d` output, which uses only alphanumeric+`/` characters — the condition cannot arise in this environment. The pattern matches the pre-existing D1.1–D1.5 tests; fixing D1.6/D1.7 in isolation would create inconsistency without eliminating the underlying risk.

PR #2382, opened 2026-06-22
dispatch: heartbeat timer uses OnCalendar so it self-recovers from elapsed

natb1:
Closes #2375

## Problem

The always-on liveness watchdog `dispatch-heartbeat.timer` (#2022) silently dies
and cannot self-recover because it is scheduled with **monotonic** anchors
(`OnBootSec=2min` + `OnUnitActiveSec=15min`). `OnUnitActiveSec` re-arms only
relative to `dispatch-heartbeat.service`'s last activation, and the only thing
that activates that service is the timer firing — a closed loop. Once the timer
is ever (re)started while both anchors are in the past — exactly the stalled
state the heartbeat exists to recover from — systemd marks it `active (elapsed)`
with `Trigger: n/a` and it never fires again. `Persistent=true` does **not**
rescue it (a documented no-op for monotonic timers).

Worse, the steady-state hot path in `ensure_heartbeat_units` guarded its
self-heal with `systemctl --user is-active --quiet`, which returns 0 for an
`active (elapsed)` timer, so concurrent ticks treated the dead timer as healthy
and never repaired it.

## Fix

`.claude/skills/dispatch-propagate/scripts/lib.sh`, `ensure_heartbeat_units`:

- **Wall-clock schedule.** Replace `OnUnitActiveSec=15min` with
  `OnCalendar=*:0/15` (fires at :00/:15/:30/:45 regardless of service
  activation) and add `RandomizedDelaySec=30` to smooth the post-boot launcher
  storm. `OnBootSec=2min` is kept for a fast first post-boot tick; `Persistent=true`
  is now meaningful (catches up a wall-clock fire missed while the host was
  asleep/off). An `OnCalendar` schedule always has a future trigger, so a restart
  can never strand the timer in `elapsed`.
- **`SubState`-based health check.** New helper `heartbeat_timer_is_armed`
  returns 0 only when `SubState=waiting`. `SubState` is the reliable
  discriminator — `NextElapseUSecRealtime` is empty for a healthy *monotonic*
  timer too, so parsing it would mis-detect. The hot-path short-circuit now calls
  this instead of `is-active`, so an `elapsed` timer falls through to repair.
- **`enable` + `restart` repair.** The slow-path activation was `enable --now`,
  whose implicit `start` is a no-op on an already-active `elapsed` timer and so
  never re-arms it. Split into `enable` (idempotent symlink) + `restart`, which
  recomputes the next elapse in every case — cold install and stranded repair.

`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`: the mock
`systemctl` stub gains `show` (emits `${STUB_SUBSTATE}`) and `restart` cases; the
cold-path and hot-path tests are updated to drive readiness via `STUB_SUBSTATE`
and assert the new timer content + `enable`/`restart` sequencing; a new
regression test asserts an `elapsed` timer is detected and re-armed (#2375).

## Verification

`test-dispatch-scripts.sh` passes (3964/3964). The mock suite proves the wiring
(timer content, hot-path predicate, repair sequencing). Live-systemd
strand-and-recover checks on the real host are noted in the plan for QA.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -2559,8 +2559,8 @@ EOF
 
 # Install and activate the durable always-on heartbeat: a `systemd --user`
 # `dispatch-heartbeat.timer` firing `dispatch-heartbeat.service` (a no-arg
-# `dispatch-tick`) on a flat drumbeat — `OnBootSec=2min`,
-# `OnUnitActiveSec=15min`, `Persistent=true`. This gives the autonomous chain a
+# `dispatch-tick`) on a wall-clock schedule — `OnBootSec=2min`,
+# `OnCalendar=*:0/15`, `Persistent=true` (#2375). This gives the autonomous chain a
 # liveness floor that is independent of Stop hooks and reseed timers, so the
 # chain self-recovers from abnormal worker death and post-cap stalls where no
 # Stop hook fires and no reseed is armed (#2022). A flat drumbeat is safe and
@@ -2621,6 +2621,27 @@ cleanup_stale_heartbeat_units() {
   # suppress the failure and warn; never abort the caller.
   echo "WARNING: cleanup_stale_heartbeat_units: installed heartbeat unit points at '$installed_workdir' but current main worktree is '$current_main_worktree'; disabling stale timer/service before rewrite" >&2
   "$systemctl_cmd" --user disable --now dispatch-heartbeat.timer dispatch-heartbeat.service || true
+}
+
+# Returns 0 only when the heartbeat timer is armed with a future trigger
+# (SubState=waiting). 'elapsed' (the #2375 stranded state: active but no future
+# fire), 'dead' (inactive), and any other/unrecognized substate all return
+# non-zero so the caller falls through to repair. SubState is the reliable
+# discriminator: NextElapseUSecRealtime is empty for a healthy monotonic timer
+# too, so parsing it would mis-detect.
+# Args: $1 = systemctl command
+heartbeat_timer_is_armed() {
+  local systemctl_cmd="$1"
+  local substate
+  substate=$("$systemctl_cmd" --user show dispatch-heartbeat.timer \
+    --property=SubState --value 2>/dev/null) || return 1
+  # 'waiting' = a future trigger is armed (healthy). 'elapsed' = the #2375
+  # stranded state (active, no future fire). 'dead' = inactive. Fail-safe:
+  # treat anything that is not explicitly 'waiting' as needing repair, so an
+  # unrecognized substate can never be read as healthy. (A transient
+  # 'running' during a heartbeat fire causes at most one harmless re-arm; the
+  # tick oneshot is fast, so the window is tiny.)
+  [ "$substate" = "waiting" ]
 }
 
 # Best-effort: a failure here must not abort the caller (a tick/reseed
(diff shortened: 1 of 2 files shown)


natb1:
# QA Summary — PR #2382 / Issue #2375

## Results

| | Count |
|---|---|
| PASS | 4 |
| FAIL | 0 |
| SKIP | 0 |
| Deferred (needs-human-judgment) | 1 |

No bugs found.

## Items Executed

### Item 1 — AC#1: Heartbeat timer uses OnCalendar (shell-command lane)
**PASS.** The heartbeat `[Timer]` block in `lib.sh` contains `OnCalendar=*:0/15` and `RandomizedDelaySec=30`, and `OnUnitActiveSec` is absent from the block.

### Item 2 — AC#2: Live strand-and-recover after stop + restart with service idle
**DEFERRED** (needs-human-judgment, classified `needs-main`). Requires live systemd on the actual host. The mock suite verified the wiring; post-merge live verification is tracked in follow-up #2386.

### Item 3 — AC#3: Persistent= comment accuracy (shell-command lane)
**PASS.** The comment above the heartbeat timer heredoc says `Persistent=true is now meaningful — it catches up a wall-clock fire missed while the host was asleep/off`. The stale monotonic drumbeat language is absent from the heartbeat block.

### Item 4 — AC#4: Repair path re-arms elapsed timer via mock regression (shell-command lane)
**PASS.** `test-dispatch-scripts.sh` passed 3964/3964. The new elapsed-strand test (2b) confirms `ensure_heartbeat_units` falls through to `daemon-reload` + `restart` when `SubState=elapsed`, rather than short-circuiting on `is-active`.

### Item 5 — AC#5: Elapsed-strand regression test present (shell-command lane)
**PASS.** The elapsed-strand regression test exists at line 34637 of `test-dispatch-scripts.sh` (`STUB_SUBSTATE=elapsed` → asserts daemon-reload and restart fire).

## Disposition Triage

One residue item entered the Workflow for classification.

| # | Title | Kind | Class | Verify | Rationale |
|---|---|---|---|---|---|
| 2 | AC#2 — Live strand-and-recover after stop + restart with service idle | needs-human-judgment | **needs-main** | n/a | planned_deferral=true: criterion is non-assertable at merge time, only verifiable against real systemd host downstream. Mock suite already verified wiring. |

## Filed Follow-ups

| Identifier | Issue | Route |
|---|---|---|
| qa-needs-main #2375: ac-2-live-strand-and-recover-after-stop-restart-with-service-idle | #2386 (newly filed) | human — main-qa + dispatch:office-hours |

#2386 is `blocked_by` #2375 and carries `main-qa` + `dispatch:office-hours` for office-hours human verification after merge.

natb1:
## Review: PR #2382 (`2375-dispatch-heartbeat-timer-us`)

**Surface:** `code` · **Changed files:** `lib.sh`, `test-dispatch-scripts.sh` · **Findings:** 5 surfaced, 2 actionable (deferred) · **Fixes applied:** none · **CodeQL:** 0 open alerts · **Erosion:** clean

---

### Fixed
_None._

### Required (security)
_None._

_None._

### Informational

- **RV-001** `lib.sh:2660-2665` — `heartbeat_timer_is_armed`: the comment acknowledges a transient `running` substate during a heartbeat fire may cause a harmless re-arm; no test covers this window. Accepted trade-off — the tick oneshot is fast and the window is tiny. No change required.

### Dismissed

- **RV-004** `lib.sh:2768-2783` — Possible TOCTOU between the unit-file content check and the armed check. Not a real hazard: `ensure_heartbeat_units` runs under the dispatch lock and is the only writer of those files; dismissed.
- **RV-005** `lib.sh:2655-2658` — `systemctl show --value` portability. `--value` landed in systemd 230 (2016); all supported distros ship ≥ 230; dismissed.

### False-positive (security)
_None._

### Deferred

- **RV-002** `test-dispatch-scripts.sh:34693-34707` — No test for `systemctl restart` failure path (`STUB_RESTART_RC=1`). Filed as follow-up → #2388 (blocked_by #2375).
- **RV-003** `test-dispatch-scripts.sh` — No test for `systemctl enable` failure in the elapsed-strand repair path. Filed as follow-up → #2389 (blocked_by #2375).

### Out-of-scope (security)
_None._

PR #2379, opened 2026-06-22
Intention tree: schema, local store, and read-only backfill

natb1:
Closes #2366

## What this is

Migration stage **1a** of the `/align` epic (#2100): the first piece of a
recursive intention-alignment system that unifies intention tracking — today
split between `CHARTER.md` principles and free-text issue scope — into one
uniform structure where **every node is the same type at any altitude**. A
charter principle is a root node; an issue's scope is a leaf node; one schema
describes both.

This PR delivers only the **data model + a read-only local store**, plus a
**read-only backfill** that snapshots current state into that store. It is
deliberately **additive and reversible**: no GitHub writes, the dispatch router
and office-hours queue are untouched, and nothing else in the repo reads the
store yet. Later epic stages (#2367 sync, #2368 goals/ROADMAP view, #2370
dialectic engine) build on this foundation.

## What's in it

A new `*util` package, `intentionsutil/`, and a committed snapshot directory
`intentions/`:

- **Schema** (`src/schema.ts`, `SCHEMA.md`) — the `IntentionNode` type plus a
  `validateNode` validator. Every field lives in YAML frontmatter; the markdown
  body is a cosmetic render of `statement` and is not part of the validated
  model. The **required core** (`id`, `statement`, `owner`, `status`) is
  validated strictly, while the not-yet-run dialectic fields
  (`clarifications`, `tooling_goals`, `success_signal`, `gap`) are optional so
  read-only backfill produces valid nodes.
- **Local store** (`src/store.ts`) — `writeNode` / `readNode` / `listNodes` over
  a directory of `<id>.md` files. Validate-before-write; the round trip
  `node → file → node` is lossless on the frontmatter model.
- **Tests** — `test/store.test.ts` (lossless round trip on a fully-populated
  node + minimal-node default tolerance + `listNodes`) and `test/schema.test.ts`
  (validateNode accept/reject/default cases). 7 tests, all passing.
- **Backfill** (`scripts/backfill.ts`) — strictly read-only toward GitHub
  (`gh api` GETs only). Produces two internally-linked layers sharing one id
  space: **principle roots** from `CHARTER.md`'s `## Principles` section
  (`parent: null`, `status: codified`) and **issue leaves** from open GitHub
  issues (`status: raw`), linked to each other by the existing GitHub issue
  hierarchy via the read-only `/parent` endpoint. An open issue whose GitHub
  parent is *closed* (and thus has no node) gets a nulled parent, so every
  non-null `parent` points to a node file that exists.
[...]

PR #2400, opened 2026-06-23
intentionsutil/store: round-trip test for multi-line string fields

natb1:
Closes #2396

## What

Adds one round-trip test case to `intentionsutil/test/store.test.ts` proving the
local store (`writeNode` → `readNode`) is lossless for **multi-line** string
field values.

## Why

The existing "is lossless for a fully-populated node" case only exercises
single-line `rationale`/`reading` values. The `yaml` library renders strings
with embedded newlines as **block scalars** (`|`/`>`), whose chomping rules can
strip trailing newlines or fold lines. This pins the multi-line losslessness
guarantee with a test rather than leaving it assumed. Test-quality follow-up to
#2366 (the store, merged via PR #2379).

## Details

The new case constructs an `IntentionNode` whose `rationale` holds a value with
an internal blank line **and** a trailing newline, and whose `reading` holds a
value with an internal blank line — the cases most likely to expose a
block-scalar transform — then asserts `readNode(...)` returns the exact node
written. Per the issue's scope, no value contains a bare `---` line (that
adversarial frontmatter-fence case is out of scope), and there are no production
code changes: the test passes against the existing, unchanged serialization
(8 passing tests before, 9 after).

## Verification

```
npx vitest run --project intentionsutil --root .
→ Test Files 2 passed (2), Tests 9 passed (9)
```


natb1:
# QA Summary — Issue #2396

**PR #2400** · attempt 1 · clean pass

## Items Executed

| # | Title | Result |
|---|-------|--------|
| 1 | Multi-line round-trip test case exists with embedded newlines in rationale and reading | PASS |
| 2 | Test performs full write→read round-trip and asserts deserialized node equals original | PASS |
| 3 | Test passes against the existing, unchanged serialization implementation | PASS |
| 4 | No production code changed — guarantee is pinned by test only | PASS |

**Counts:** 4 PASS · 0 FAIL · 0 SKIP

## Detail

- Item 1: `grep` confirmed the `it("is lossless for a node with multi-line string fields", ...)` test case exists, with `rationale` containing an internal blank line and trailing newline, and `reading` containing an internal blank line.
- Item 2: Both the new test case (lines 74–76) and the existing case (lines 37–39) call `writeNode(dir, node)` → `readNode(dir, node.id)` → `expect(read).toEqual(node)`.
- Item 3: `npx vitest run --project intentionsutil --root .` → 2 test files passed, 9 tests passed (up from 8 before this PR).
- Item 4: `git diff --name-only origin/main...HEAD` → only `intentionsutil/test/store.test.ts` changed; no `src/store.ts` or `src/schema.ts` modifications.

All acceptance criteria satisfied. No bugs found.

natb1:
## Review — PR #2400

### Fixed (2)

- **`store.test.ts:53`** — `reading` string referenced `yaml-spec.org section 8.1`, which is not the authoritative YAML spec URL. Corrected to `yaml.org/spec/1.2/spec.html section 8.1.1.2`. (code-review) · 2a94f769
- **`store.test.ts:50-53`** — The `rationale` field ends with `\n` while `reading` does not; the asymmetry is intentional (exercises both trailing-newline and no-trailing-newline chomping paths) but was undocumented. Added an inline comment: `// rationale ends with \n (trailing newline); reading does not — exercises both chomping cases`. (review) · 2a94f769

### Informational (3)

- **`store.test.ts:42`** — The test covers the `clip` (trailing newline kept) and `strip` (no trailing newline) block-scalar chomping paths via `rationale` and `reading` respectively, but does not exercise the `keep` indicator (literal `|+`). This is fine for the issue scope; noting for completeness. (code-review)
- **`store.test.ts:55`** — `clarifications[].answer` strings are single-line. A multi-line answer would fully exercise the `Clarification` sub-object's string fields under the same block-scalar guarantee. Low priority given issue scope. (code-review)
- **`store.test.ts:42`** — The PR body notes that values containing a bare `---` line are out of scope, but this limit is not documented in the test file itself. A comment or a link to the issue would make the boundary explicit. (review)

### Security

PR #2404, opened 2026-06-23
intentionsutil/backfill: retry GitHub API calls on 429/503

natb1:
Closes #2397

## Summary

The read-only intention-tree backfill script (`intentionsutil/scripts/backfill.ts`,
from #2366 / PR #2379) makes many sequential `gh api` GET calls with no rate-limit
handling — one `/parent` call per open issue, so well over 100 sequential calls
against a repo with hundreds of open issues. Two call sites failed differently on
a transient 429/503:

- `fetchOpenIssues()` — a single `gh api --paginate --slurp` call. A transient
  failure made `gh` exit non-zero, `execFileSync` threw, and the whole backfill
  aborted before any node was written.
- `fetchParentNumber()` — wrapped `gh` in a bare `catch {}` meant to absorb the
  expected 404 (no-parent case). A 429/503 was also swallowed, so the node was
  silently written with `parent: null` instead of failing loudly.

This adds a small `ghWithRetry` wrapper that retries transient `gh` failures with
exponential backoff and re-throws a clear status-bearing error after exhausting
retries. Both call sites use it; `fetchParentNumber` now returns `null` **only**
for a real 404 and re-throws everything else.

## Changes

`intentionsutil/scripts/backfill.ts`:

- `GhError extends Error` carrying `httpStatus: number | null` and the original
  error as `cause`; `parseHttpStatus(text)` parses the HTTP code from gh's output
  text (`gh: … (HTTP NNN)`). The HTTP status is parsed from stderr+stdout, never
  the exit code — `gh` exits 1 on a 404, the exit code is not the status.
- `isTransientGhError(text)` mirroring the shell-side `_gh_error_is_transient` in
  `.claude/skills/dispatch-propagate/scripts/lib.sh`: matches `HTTP 429`, `HTTP
  5xx` (covers 503), and rate-limit text (`secondary rate limit`, `abuse
  detection`, `retry your request`, `temporarily unavailable`, `Service
  Unavailable`, `Bad Gateway`). A bare `HTTP 403` is **not** treated as transient.
  This is a deliberate, issue-faithful superset of "429 or 503": the issue names
  `lib.sh` as the model, and `lib.sh` keys on rate-limit text too.
- `sleepSync(ms)` via `Atomics.wait` keeps the script fully synchronous (it uses
  `execFileSync` and a sync `main()`) — no async conversion.
- `ghWithRetry(args)` reads `GH_RETRY_ATTEMPTS` (default 4) and
  `GH_RETRY_BASE_DELAY_MS` (default 1000) knobs, retries transient failures with
  doubling backoff, and on exhaustion/non-transient failure throws a `GhError`
  whose message names the failing `gh` call and the HTTP status.
- `fetchOpenIssues` and `fetchParentNumber` route through `ghWithRetry`;
  `fetchParentNumber`'s bare `catch {}` becomes `catch (err) { if (err instanceof
  GhError && err.httpStatus === 404) return null; throw err; }`.
- `main()` is guarded with an `import.meta.url` / `process.argv[1]` direct-execution
  check so the module is importable by tests; the testable helpers are exported.
  Running `npx tsx intentionsutil/scripts/backfill.ts` still executes `main()`.

`intentionsutil/test/backfill.test.ts` (new — there were no tests for
`backfill.ts`): mocks `node:child_process.execFileSync` and covers retry-then-
succeed, exhaust-then-throw-`GhError` (status + message + call count), no-retry on
a non-transient 404, `fetchParentNumber` null-on-404, re-throw-on-429, the happy
path, and `parseHttpStatus`/`isTransientGhError` unit checks.

## Verification

- `npx vitest run --project intentionsutil --root .` — 17 passed.
- `npm run lint --prefix intentionsutil` — clean.

A true end-to-end exercise against live GitHub rate limits is not reproducible on
demand; the mock-based unit tests are the authoritative check. A maintainer may
optionally run `npx tsx intentionsutil/scripts/backfill.ts` against the real repo
to confirm the guarded `main()` still executes — read-only toward GitHub, but it
writes local `intentions/` node files, so it is left as a manual QA step.


natb1:
# Fix-checks summary — PR #2404

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `git diff --no-color -U0 --diff-filter=d origin/main...HEAD -- '*.ts' | .github/scripts/check-type-safety-escapes.sh --scan-stdin`
- **Failure excerpt:** 7 violations in `intentionsutil/test/backfill.test.ts` — `as Error &` cast on line 18 and `as GhError` casts on lines 61–64, 81, 109
- **Why not caught:** The test file was written with idiomatic TypeScript cast patterns (`as GhError` after a try-catch for `unknown`) that are safe in context but trigger the type-safety-sensor's `as <Type>` rule. `Object.assign` and `instanceof` narrowing guards are the cast-free equivalents.
- **Fix:** Replaced `new Error(...) as Error & {...}` with `Object.assign(new Error(...), {...})` on line 18; added `if (!(caught instanceof GhError)) throw new Error("unreachable")` guards after each `expect(caught).toBeInstanceOf(GhError)` call to narrow the type without a cast. Commit: `07c407c1`

natb1:
# QA Summary — Issue #2397 / PR #2404

## Results

**7 script-verifiable items — all PASS. 0 FAIL. 0 SKIP.**

| # | Title | Result |
|---|-------|--------|
| 1 | Backfill test suite passes (covers retry-path acceptance criteria) | ✅ PASS |
| 2 | Lint is clean on the changed package | ✅ PASS |
| 3 | ghWithRetry wraps the execFileSync gh call (AC1) | ✅ PASS |
| 4 | fetchOpenIssues routes through the retry wrapper (AC2) | ✅ PASS |
| 5 | fetchParentNumber distinguishes 404 from transient errors (AC3) | ✅ PASS |
| 6 | Exhausted-retry error identifies the failing call and HTTP status (AC4) | ✅ PASS |
| 7 | Read-only invariant preserved — no GitHub mutation introduced | ✅ PASS |

All 17 vitest tests pass. ESLint clean.

## Disposition Triage

One planned-deferral item was surfaced; it classified `needs-main` and was filed as a `blocked_by` follow-up (below). `needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR.

| Item | Class | Verify |
|------|-------|--------|
| 8 — Backfill still produces a correct snapshot against a live repo (end-to-end behavior) | `needs-main` | n/a (planned deferral) |

**Rationale:** Live end-to-end run against an authenticated `gh` + network repo is not assertable at merge time. The transient-failure handling is fully covered by mock-based unit tests; the happy path is unchanged from the prior working script. Verified downstream post-merge via manual QA.

## Filed Follow-ups

- #2407 — "qa: verify against main — qa-needs-main #2397: backfill-still-produces-a-correct-snapshot-against-a-live-repo-end-to-end-behavior" (route: **human** → `main-qa` + `dispatch:office-hours`)

## Terminal Disposition

Clean pass. All acceptance criteria verified by the unit test suite and static checks. The one deferred item was filed as a `blocked_by` follow-up (#2407) and routed to office-hours human review.

natb1:
## Review (dispatch:reviewed)

### Fixed — db675e48

- **backfill.ts:90 — `GH_RETRY_ATTEMPTS` unvalidated input** (`code-review`, `review`): `Number('')` and `Number('foo')` both produce `0`, causing the retry loop to never execute and fall through to `throw new Error('unreachable')`. Fixed: `parseInt(..., 10) || 4` with `Math.max(1, ...)` floor so empty/non-numeric/zero/negative all default to 4 attempts.

- **backfill.ts:71 — `isTransientGhError` missing network-level patterns** (`code-review`): The regex omits several transient patterns present in `lib.sh`'s `_gh_error_is_transient`: `timed out`, `i/o timeout`, `deadline exceeded`, `connection reset`, `TLS handshake`. Fixed: added all missing alternations so the TS implementation matches the documented shell model.

- **backfill.test.ts:127 — `isTransientGhError` 5xx branch untested** (`code-review`, `review`): The `parseHttpStatus / isTransientGhError` describe block covered 429, 403, and `secondary rate limit` but not the `HTTP 5\d\d` digit-range pattern. Fixed: added `expect(isTransientGhError('HTTP 500')).toBe(true)` and `expect(isTransientGhError('HTTP 599')).toBe(true)`.

- **backfill.test.ts:1 — `fetchOpenIssues` had no test coverage** (`review`): The multi-page JSON flattening and PR-exclusion filter were untested. Fixed: added two cases — multi-page flattening (two-page mock → two issues) and PR exclusion (item with `pull_request: {}` is filtered out).

### Informational

- **backfill.test.ts:112 — `fetchParentNumber` happy-path uses `mockReturnValue` not `mockReturnValueOnce`** (`code-review`): The mock applies to all subsequent calls, so if the test ordering were to change in future, an earlier `mockReset()` in `beforeEach` already guards against cross-test contamination. Not a bug — noting for awareness.

- **backfill.ts:89 — cyclomatic complexity rose from 39 to 56** (`erosion`): Expected for the new `ghWithRetry` retry logic. The increase is proportional to the new functionality; no decomposition applied (complexity is bounded and the functions remain readable).

### False-positive (security)

- **backfill.ts:90-91 — `GH_RETRY_ATTEMPTS`/`GH_RETRY_BASE_DELAY_MS` Infinity/NaN concern** (`input-validation`, `auth`, `data-exposure`): These env vars are developer-controlled tuning knobs, not user-facing or attacker-reachable inputs. `Infinity` attempts and a `NaN` delay are not exploitable in this context — the retry loop's `attempt <= attempts` with `Infinity` keeps running until a success or non-transient error, which is the desired behavior for a misconfigured but dev-owned knob. Not a meaningful security concern.

- **backfill.ts:104,106 — `GhError` embeds raw stderr/stdout and exec cmd as `cause`** (`data-exposure`): The script runs interactively on a developer's machine under their own `gh` authentication. The error message and cause chain are only observed by that same developer's shell. There is no credential exfiltration vector — no network sink, no log aggregation.

### Refuted (adversarial-verify)

- **backfill.test.ts:68 — code duplication in test file** (`erosion`): The flagged 12-line "clone" is TypeScript's mandatory catch-and-narrow boilerplate (`let caught: unknown; try { ... } catch (err) { caught = err; }` plus `instanceof` type guard). TypeScript's strict `unknown` catch-binding requires this scaffolding to access typed properties on the caught value — `.toThrow()` cannot be used when post-throw property inspection is needed, and a helper wrapper would have to return `unknown` and forfeit type-narrowing. Each of the three instances exercises a different function under a different scenario with different post-catch assertions. The detector matched structural similarity in unavoidable language-level scaffolding, not extractable shared logic. **Both skeptics refuted this finding** — not a real duplication issue.

PR #2423, opened 2026-06-23
refresh.ts: narrow stage-2 catch to only swallow deleted/transferred issue errors

natb1:
Closes #2418

## Problem

`intentionsutil/scripts/refresh.ts` resolves an issue's linked PRs in two stages.
Stage 2 calls `gh issue view <N> --json closedByPullRequestsReferences` inside a
`try`/`catch`. The `catch` was **unconditional** — it swallowed every error class
(rate-limit 429/403, auth failure, network failure) and silently treated the
stage-2 PR list as empty, even though its comment claimed it only handled the
"issue deleted/transferred" case. When the catch fired for a rate-limit or auth
error, a tracker was written from stage-1 data alone but looked authoritative,
masking the GraphQL-bucket-exhaustion signal. This violated the project code-style
rule "prefer clear errors over defensive fallbacks".

## Change

Narrow the stage-2 catch so it absorbs **only** the genuine deleted/transferred
case — gh's `Could not resolve to an issue or pull request` GraphQL error — and
**re-throws everything else**. This mirrors the discipline already in the sibling
script `backfill.ts` (`fetchParentNumber`, which re-throws anything that is not the
expected 404). `refresh.ts`'s `gh()` helper throws the raw `execFileSync` error
(it does not wrap in `GhError`), so the discriminator inspects the raw error's
`stderr`/`stdout` rather than using an `instanceof` check.

A 429, 403, or network failure from `gh issue view` now propagates as a thrown
error instead of being converted to a silent empty list.

## Testability

To unit-test the narrowed catch, the module is made importable: `main()` is now
guarded by the `import.meta.url === pathToFileURL(process.argv[1]).href` check
(copied verbatim from `backfill.ts`), and `resolveLinkedPrs` is exported.

## Tests

New `intentionsutil/test/refresh.test.ts` (following `backfill.test.ts`'s
`vi.mock("node:child_process")` pattern) covers three cases:

1. **Swallows** the `Could not resolve` (deleted/transferred) error and returns
   the stage-1 result.
2. **Re-throws** a rate-limit (429) error.
3. **Re-throws** an auth (403) error.

Verified with `npx vitest run --project intentionsutil --root .` — 5 files /
38 tests pass (the prior 4 files / 35 tests stay green, plus the 3 new cases).

## Scope

Out of scope (deliberately): any change to the stage-1 REST call, retry behavior,
and the other `gh()` call sites. A larger refactor extracting `backfill.ts`'s
`GhError`/`ghWithRetry` into a shared module is a separate concern — this issue
wants rate-limit errors to *propagate*, not to retry-then-swallow.

### DIFF
--- a/intentionsutil/scripts/backfill.ts
+++ b/intentionsutil/scripts/backfill.ts
@@ -21,6 +21,7 @@ import { existsSync, mkdirSync, readFileSync, readdirSync, rmSync } from "node:f
 import { dirname, join } from "node:path";
 import { fileURLToPath, pathToFileURL } from "node:url";
 import { writeNode } from "../src/store.js";
+import { ghErrorText } from "../src/errors.js";
 
 // --- Paths -----------------------------------------------------------------
 // The script lives at `intentionsutil/scripts/backfill.ts`, so the repo root is
@@ -96,8 +97,7 @@ export function ghWithRetry(args: string[]): string {
     try {
       return execFileSync("gh", args, { encoding: "utf8" });
     } catch (err) {
-      const e = err as { stderr?: unknown; stdout?: unknown };
-      const text = String(e.stderr ?? "") + "\n" + String(e.stdout ?? "");
+      const text = ghErrorText(err);
       if (attempt < attempts && isTransientGhError(text)) {
         sleepSync(baseDelayMs * 2 ** (attempt - 1));
         continue;
--- a/intentionsutil/scripts/refresh.ts
+++ b/intentionsutil/scripts/refresh.ts
@@ -18,7 +18,7 @@
 import { execFileSync } from "node:child_process";
 import { existsSync } from "node:fs";
 import { dirname, join } from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import {
   writeTracker,
   nodeIdToIssue,
@@ -96,7 +96,7 @@ const GQL_STATE: Record<"OPEN" | "CLOSED" | "MERGED", "open" | "closed" | "merge
  * conflict (its `merged_at`-derived state is computed directly from the PR
  * record rather than the issue's reference view).
  */
-function resolveLinkedPrs(issueNumber: number): LinkedPr[] {
+export function resolveLinkedPrs(issueNumber: number): LinkedPr[] {
   const byNumber = new Map<number, LinkedPr>();
 
   // Stage 1: REST pulls list, branch-prefix filtered.
@@ -126,9 +126,21 @@ function resolveLinkedPrs(issueNumber: number): LinkedPr[] {
   try {
     // lint-allow: gh-rest-porcelain GraphQL-only closedByPullRequestsReferences (no REST equivalent)
     viewOut = gh(["issue", "view", String(issueNumber), "--json", "closedByPullRequestsReferences"]);
-  } catch {
-    // Expected: issue deleted/transferred → treat as no closing references.
-    viewOut = null;
+  } catch (err) {
+    // Only the documented expected error — issue deleted/transferred, which gh
+    // reports as a "Could not resolve to an issue or pull request" GraphQL error
+    // — is absorbed (treated as no closing references, mirroring backfill.ts's
+    // /parent 404 discipline). Every other error class (rate-limit 429/403, auth
+    // failure, network failure) must propagate rather than be silently nulled.
+    // refresh.ts's gh() throws the raw execFileSync error, so inspect its stderr
+    // (not an instanceof GhError check — gh() never produces a GhError).
+    const e = err as { stderr?: unknown; stdout?: unknown };
+    const text = String(e.stderr ?? "") + "\n" + String(e.stdout ?? "");
+    if (/could not resolve to (a|an)/i.test(text)) {
+      viewOut = null;
+    } else {
+      throw err;
+    }
   }
   if (viewOut !== null) {
     const view: IssueView = JSON.parse(viewOut);
@@ -221,4 +233,6 @@ function main(): void {
   );
 }
 
-main();
+if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) {
+  main();
+}
--- a/intentionsutil/scripts/refresh.ts
+++ b/intentionsutil/scripts/refresh.ts
@@ -26,6 +26,7 @@ import {
   type ExecutionTracker,
 } from "../src/tracker.js";
 import { listNodes } from "../src/store.js";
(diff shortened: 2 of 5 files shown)


natb1:
# Fix-checks summary — PR #2423

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Failure excerpt:** `::error file=intentionsutil/test/refresh.test.ts,line=24::Net-new type-safety escape hatch (as <Type> cast).`
- **Why not caught:** The test mock used `as unknown as Buffer` to satisfy TypeScript's `execFileSync` overload signature. The type-safety sensor flags all `as <Type>` casts introduced in a PR diff, including those in test files.
- **Fix:** Replaced `mockImplementation` with `mockReturnValueOnce(JSON.stringify([[]])) + mockImplementationOnce(() => { throw err; })`. Stage order is deterministic (stage-1 `gh api` always runs before stage-2 `gh issue view`), so the `Once` idiom is correct and eliminates the cast entirely. Commit `e2c912e1`.

natb1:
# QA Summary — Issue #2418 / PR #2423

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Stage-2 catch inspects the error before swallowing it | shell-command | PASS |
| 2 | Non-404/transfer errors (rate-limit, auth, network) are re-thrown | shell-command | PASS |
| 3 | A test demonstrates a non-404 error propagates to the caller | shell-command (vitest) | PASS |
| 4 | Stage-2 result is set to null only for genuine deleted/transferred case | shell-command (vitest) | PASS |
| 5 | Importable module + main-guard does not regress CLI execution | (assessed by Opus) | PASS (already satisfied) |

**PASS: 4 (shell-command) + 1 (Opus-assessed) | FAIL: 0 | SKIP: 0**

No browser walkthrough was run — no frontend files changed.

All four issue acceptance criteria were verified:

1. **catch block inspects the error** — `grep` confirms `catch (err)`, reads `stderr`/`stdout`, tests `/could not resolve to (a|an)/i` (line 129–139 of `intentionsutil/scripts/refresh.ts`). ✅
2. **Non-404 errors re-thrown** — `grep` confirms `throw err` at line 142 in the `else` branch. ✅
3. **Test demonstrates propagation** — `npx vitest run --project intentionsutil --root .` → 5 files / 38 tests pass (includes re-throw cases for 429 and 403). ✅
4. **null only for deleted/transferred** — same vitest run includes the swallow case asserting `toEqual([])` for the "Could not resolve" error. ✅

## Bugs found

None.

## Disposition triage

One residue item was assessed by the Opus disposition agent.

- **Item 5 — Importable module + main-guard does not regress CLI execution**
  Both halves are provably met without credentials. (1) No side effects on import: `refresh.test.ts` imports `resolveLinkedPrs` from `refresh.js` inside a `vi.mock("node:child_process")` environment. If the main-guard were broken, `main()` would fire on import, crashing/misbehaving during module load and failing every test in that file. All 38 tests pass (re-run confirmed), so `main()` demonstrably does NOT auto-run on import. (2) Self-executes as entrypoint: line 236 uses the canonical ESM guard `process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href`, which fires `main()` only when `refresh.ts` is the invoked entrypoint. Verifying with live `gh` credentials is unnecessary — the guard's correctness is structurally determinable and import-safety is empirically demonstrated by the green suite. Dropped as PASS.

## Terminal disposition

**Clean pass.** All items either passed the shell-command lane or were dropped as already-satisfied by the Opus assessment. No items in the escalation set.

natb1:
## Review — #2418 (`2418-refresh-ts-stage-2-catch-bl`)

Surface: `code` (2 files changed). 8 findings from 7 sources (code-review, review, red-team, input-validation, erosion, CodeQL [0 alerts], security domains). 3 fixed, 2 dismissed, 1 false-positive, 2 refuted by adversarial verify.

---

### Fixed (3) — commit `6946bc4c`

- **`intentionsutil/scripts/refresh.ts:139`** — Tightened stage-2 discriminator regex from `/could not resolve to (a|an)/i` to `/could not resolve to (a|an) (issue|pull request)/i`, scoping the swallow to the documented deleted/transferred case. *(sources: code-review, review)*
- **`intentionsutil/scripts/refresh.ts:137-138`** — Extracted the duplicated `e.stderr`/`e.stdout` error-text pattern into a shared `ghErrorText` helper in `intentionsutil/src/errors.ts`, imported by both `refresh.ts` and `backfill.ts`. *(source: code-review)*
- **`intentionsutil/test/refresh.test.ts:1`** — Added `resolveLinkedPrs merge/dedup` describe block with two non-throwing stage-2 tests: stage-1-wins-on-number-conflict and the additive stage-2 path. *(source: review)*

---

### Refuted — adversarial verify dropped before fix (2)

- **`intentionsutil/scripts/refresh.ts:139`** (red-team) — Claimed `/could not resolve to (a|an)/i` would absorb DNS/proxy failures. Refuted 2/2: Go's net package emits `no such host` / `i/o timeout`; proxies surface as HTTP 5xx. Neither string matches the regex. Only GitHub GraphQL entity-not-found strings match, which is the intended absorption domain. *(verdict: refuted)*
- **`intentionsutil/scripts/refresh.ts:99`** (erosion) — Complexity increase from 24→27 flagged as structural erosion. Refuted 1/2: the +3 CC maps 1-to-1 to the discriminating branches that ARE the fix (two `??` operators, one `if` guard, plus the main-guard `if`). Removing them would reintroduce the bug; extracting to a helper moves not reduces the count. *(verdict: refuted)*

---

### Dismissed (2)

- **`intentionsutil/scripts/refresh.ts:138`** — stderr+stdout concatenation could produce a false match at the newline boundary. Dismissed: the regex does not contain `$`/`^` anchors and the boundary is only a `\n` between two strings; no realistic error text is split across stderr/stdout at a "could not resolve" phrase boundary. *(sources: red-team, review)*
- **`intentionsutil/scripts/refresh.ts:137`** — `err` cast to `{ stderr?: unknown; stdout?: unknown }` without type guard. Dismissed: `String(e.stderr ?? "")` handles `undefined`, `null`, and any non-string gracefully; a non-object throw (e.g. `throw "string"`) produces empty text, falls through the `if`, and re-throws — correct behavior. *(source: review)*

---

### False-positive (1)

- **`intentionsutil/scripts/refresh.ts:138-139`** (input-validation) — Flagged error-text regex as an injection vector. False-positive: `err.stderr` comes from `execFileSync` which controls the gh process; it is not user-supplied input. No injection surface exists. *(source: input-validation)*

---

_No Informational, Deferred, or Out-of-scope (security) findings._

PR #2430, opened 2026-06-23
intention-stamp-gh-stub: add set -euo pipefail to surface jq pipeline failures

natb1:
Closes #2428

## What

Add `set -euo pipefail` to `intention-stamp-gh-stub.sh`, the test-only `gh`
stub used by `test-intention-stamp.sh`, and add a regression test (Test 9) that
proves a malformed `--jq` filter now exits non-zero.

## Why

The stub's single-comment GET handler runs a two-stage `jq` pipeline followed by
an unconditional `exit 0`:

```bash
jq -Rs '{body: .}' < "${STUB_DIR}/posted-body.txt" | jq -r "$JQ_FILTER"
...
exit 0
```

Without `pipefail`, a non-zero exit from either pipeline stage was masked by the
trailing `exit 0`, so a test would assert on empty stdout rather than surfacing
the `jq` error — silently passing. The companion `test-intention-stamp.sh`
already carries `set -euo pipefail` on line 2; the stub did not. This is a latent
correctness trap should the stub ever be extended with variable filters.

## Change

- `intention-stamp-gh-stub.sh` — `set -euo pipefail` inserted as line 2,
  immediately after the shebang, mirroring `test-intention-stamp.sh:2`. No other
  change to the stub.
- `test-intention-stamp.sh` — new Test 9 drives the stub's GET handler directly
  with a malformed `--jq` filter and asserts a non-zero exit (via the existing
  `assert_exit_nonzero` helper).

## Verification

The full shell suite passes: 17/17 assertions across 9 tests. Tests 1–8 confirm
existing behavior is unaffected (including the valid production filter via Test 7);
Test 9 confirms the malformed-filter failure now surfaces.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/intention-stamp-gh-stub.sh
+++ b/.claude/skills/dispatch-propagate/scripts/intention-stamp-gh-stub.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -euo pipefail
 # gh stub for test-intention-stamp.sh
 # The real script passes repos/{owner}/{repo}/... with literal braces — gh would expand
 # them from GH_REPO, but this stub receives the literal string. Match on substrings.
--- a/.claude/skills/dispatch-propagate/scripts/test-intention-stamp.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-intention-stamp.sh
@@ -141,4 +141,17 @@ assert_eq "exits 0" "0" "$exit_code"
 assert_eq "stdout is empty" "" "$output"
 teardown
 
+# ---------------------------------------------------------------------------
+# Test 9: malformed --jq filter in the GET handler exits non-zero (pipefail)
+# Without `set -euo pipefail` in the stub, a non-zero jq exit is masked by the
+# handler's trailing `exit 0`, silently succeeding. With pipefail it surfaces.
+# ---------------------------------------------------------------------------
+echo "Test 9: malformed --jq filter in GET handler exits non-zero"
+setup
+# Seed a stored body so the GET handler reaches the jq pipeline.
+printf '<!-- intention:node-id -->\ngoal-x\n' > "$TMPDIR_TEST/stub/posted-body.txt"
+assert_exit_nonzero "malformed --jq exits non-zero" \
+  "$TMPDIR_TEST/bin/gh" api repos/o/r/issues/comments/777 --jq 'bad jq ('
+teardown
+
 report_results


natb1:
# QA Summary — Issue #2428 / PR #2430

## Results

| 1. Stub carries `set -euo pipefail` immediately after the shebang | script-verifiable | PASS |
| 2. Malformed `--jq` filter in GET handler exits non-zero | script-verifiable | PASS |
| 3. All existing tests continue to pass (no regression from strict mode) | script-verifiable | PASS |

**3 passed, 0 failed, 0 skipped.**

## Details

**Item 1 — Line check:** `sed -n '1,2p'` on `intention-stamp-gh-stub.sh` confirms line 1 is `#!/usr/bin/env bash` and line 2 is `set -euo pipefail`.

**Items 2 & 3 — Full test suite:** `bash test-intention-stamp.sh` reports 17/17 assertions passed across 9 tests. Test 9 ("malformed --jq exits non-zero") passes — the malformed filter now surfaces a non-zero exit instead of being masked by the trailing `exit 0`. Tests 1–8 confirm no regression from adding strict mode to the stub.

## Bugs found

None.

## Notes

No browser component detected — all checks ran in the shell-command lane. No QA server started. No disposition triage needed (residue list empty).

failed: none

natb1:
## Review: #2428 — intention-stamp-gh-stub: add set -euo pipefail

**Surface:** `code` (2 shell scripts) · **Findings:** 3 Informational · **Fixed:** 0 · **Security alerts:** none (CodeQL: 0 open alerts)

---

### Fixed

_None._

### Required (security)

_None._

### Informational

| Location | Finding |
|---|---|
| `test-intention-stamp.sh:153` | `assert_exit_nonzero` suppresses stderr via `2>/dev/null` (pre-existing in test-helpers.sh:41). When the stub's jq pipeline fails, the jq parse-error message is silently discarded — test output only shows the exit-code assertion. No action required; behaviour is intentional (assertion helpers suppress noise) and the exit-code check is the correctness signal. |
| `test-intention-stamp.sh:152` | Test 9 seeds `posted-body.txt` with a `printf` using `\n` escape sequences. `printf` correctly expands `\n` in both bash and zsh, so this is portable. Noted for awareness in case the test is later ported to a shell where `printf` behaviour differs. |
| `intention-stamp-gh-stub.sh:2` | Adding `set -u` (via `set -euo pipefail`) makes any invocation of the stub without `STUB_DIR` set fail with an unbound-variable error rather than silently using an empty path. This is the desired strictness; all callers (`test-intention-stamp.sh`) set `STUB_DIR` before invoking the stub. No action required. |

### Dismissed

_None._

### False-positive (security)

_None._

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #2437, opened 2026-06-23
Extract paginateGhApi shared helper; dedup --paginate --slurp pagination

natb1:
Closes #2435

## What

`refresh.ts` and `backfill.ts` each duplicated the same `gh api --paginate
--slurp` → `JSON.parse` → `.flat()` pagination idiom. This extracts that idiom
into one shared, typed helper that both call sites delegate to.

## Changes

- **New module `intentionsutil/scripts/gh-utils.ts`** exporting:
  - `gh(args)` — the plain `gh` runner (moved verbatim from `refresh.ts`,
    preserving the load-bearing 100 MB `maxBuffer`).
  - `paginateGhApi<T>(args, runner = gh)` — prepends `["api", "--paginate",
    "--slurp"]`, parses the array-of-arrays slurp payload, and flattens it. The
    `runner` parameter lets a caller inject a retrying runner.
- **`refresh.ts`** — `fetchAllPulls()` now delegates to `paginateGhApi<RawPull>`;
  the local `gh` is imported from `gh-utils.ts` and the now-unused
  `execFileSync` import is removed. Default runner is plain `gh`, so refresh's
  non-retrying behavior is unchanged.
- **`backfill.ts`** — `fetchOpenIssues()` now delegates to
  `paginateGhApi<RawIssueItem>`, passing `ghWithRetry` as the runner so
  backfill's retry behavior is preserved unchanged.
- **New test `intentionsutil/test/gh-utils.test.ts`** — direct coverage of
  `paginateGhApi`: page flattening, the default-runner argv, and the
  custom-runner passthrough (the `ghWithRetry`-injection contract).

## Verification

`npx vitest run --project intentionsutil --root .` — 6 files, 53 tests pass. The
existing `refresh.test.ts` / `backfill.test.ts` suites (which mock `execFileSync`,
still reached through both delegating call paths) stay green without changes,
confirming no behavior change.

## Note on naming

The issue was originally written against a `fetchAllPulls()` helper. By the time
this was implemented, `origin/main` had re-extracted exactly that
`fetchAllPulls()` in `refresh.ts`, so this PR refactors it directly — the
duplicated idiom is removed from `fetchAllPulls()` and `fetchOpenIssues()`, and
both now delegate to the shared `paginateGhApi`.

PR #2439, opened 2026-06-23
gh-utils.test.ts: cover paginateGhApi empty-args guard and runner-error propagation

natb1:
Closes #2438

## What

Add two error-path test cases to `intentionsutil/test/gh-utils.test.ts`,
covering the two previously-untested contracts of the `paginateGhApi` helper
(introduced in PR #2437 for #2435):

1. **Empty-args guard** — `paginateGhApi([])` throws
   `paginateGhApi: args must not be empty (no endpoint specified)` and
   short-circuits before invoking the runner.
2. **Runner-error propagation** — an error thrown by the injected `runner`
   reaches the caller unchanged (same `Error` instance, by reference); the
   helper does not catch.

## Why

The existing suite covered only the happy path (page flattening, default-runner
argv, custom-runner passthrough). The error-path contract was unverified: if the
guard were removed or propagation broke in a refactor, nothing would catch it.
These tests pin the existing, correct behavior — no source change to
`gh-utils.ts`.

## Verification

`npx vitest run --project intentionsutil --root .` — all five `paginateGhApi`
tests (three existing plus the two new) green; full suite 55/55 passing.

### DIFF
--- a/intentionsutil/test/gh-utils.test.ts
+++ b/intentionsutil/test/gh-utils.test.ts
@@ -44,4 +44,21 @@ describe("paginateGhApi", () => {
     expect(mockExec).not.toHaveBeenCalled();
     expect(result).toEqual([{ id: 1 }]);
   });
+
+  it("throws when args is empty", () => {
+    expect(() => paginateGhApi([])).toThrow(/args must not be empty/);
+    expect(mockExec).not.toHaveBeenCalled();
+  });
+
+  it("propagates runner errors to the caller", () => {
+    const boom = new Error("gh exploded");
+    const runner = vi.fn(() => {
+      throw boom;
+    });
+
+    expect(() => paginateGhApi(["/z"], runner)).toThrow(boom);
+
+    expect(runner).toHaveBeenCalledTimes(1);
+    expect(mockExec).not.toHaveBeenCalled();
+  });
 });


natb1:
# QA Summary — Issue #2438 / PR #2439

## Items executed

### 1. Vitest suite passes with the two new error-path tests green
- **Classification:** script-verifiable
- **Command:** `npx vitest run --project intentionsutil --root .`
- **Result:** PASS — 6 test files, 56 tests, all green

### 2. New tests bind to real source (not a stub) and guard message matches
- **Classification:** script-verifiable
- **Command:** `grep` checks on import path, guard message, and test names
- **Result:** PASS
  - `paginateGhApi` imported from `../scripts/gh-utils.js` (real module)
  - Guard message `args must not be empty` present at `gh-utils.ts:35`
  - Both tests co-located in `intentionsutil/test/gh-utils.test.ts` (lines 48, 53)

## Summary

| | Count |
|---|---|
| PASS | 2 |
| FAIL | 0 |
| SKIP | 0 |

No browser component detected — shell-command lane only, no QA server started.

No residue. Clean autonomous pass.

natb1:
## Review: #2439

**Surface:** test-only diff — code-review and review finders ran; no security finders.

---

### Fixed
_None._

### Required (security)
_None._

### Informational

- **`intentionsutil/test/gh-utils.test.ts:59`** — `expect(() => paginateGhApi(['/z'], runner)).toThrow(boom)` uses `.toThrow(boom)` (an `Error` instance), which Vitest matches by message string, not object identity. The PR description states the contract is "same `Error` instance, by reference," but the assertion does not enforce that. `expect(() => ...).toThrow(boom)` passes if any error with the same message is thrown. To pin the by-reference contract, use `.toThrow(boom)` paired with a `try/catch` that asserts `===` identity, or use `.toThrowError(boom)` and separately assert the caught error is the same instance. _(code-review)_

- **`intentionsutil/test/gh-utils.test.ts:62`** — `expect(mockExec).not.toHaveBeenCalled()` in the "propagates runner errors to the caller" test is a structural tautology. `mockExec` is the default runner's underlying exec mock; the test injects a custom `runner` that throws before `paginateGhApi` invokes any exec at all. Asserting `mockExec` was not called adds no signal — a broken implementation that still never reached `mockExec` would also pass this assertion. The assertion is not wrong, but it does not validate the propagation contract. _(code-review)_

### Dismissed
_None._

### False-positive (security)
_None._

### Deferred
_None._

### Out-of-scope (security)
_None._

PR #2458, opened 2026-06-23
Improve Windows username detection in wezterm activation script

### LINKED ISSUE #62: Improve Windows username detection in wezterm activation script
## Context

From PR #60 review: The Windows user detection in `nix/home/wezterm.nix` picks the alphabetically first non-system directory under `/mnt/c/Users/`. On machines with multiple real Windows users, this silently picks the wrong one.

## Options

- Check `$LOGNAME` or `cmd.exe /c echo %USERNAME%` as a more reliable approach
- Add a configurable option for the Windows username
- Fall back to the current heuristic only if the above methods fail

## References

- `nix/home/wezterm.nix` lines 972-973
- `nix/home/wezterm_test.sh` test 6 verifies the alphabetically-first behavior

natb1:
Closes #62

## What

Replaces the single brittle heuristic the wezterm WSL activation script used to
detect the Windows user with a three-tier detection chain, so machines with more
than one real Windows profile no longer silently pick the wrong one.

Previously `nix/home/wezterm.nix` listed `/mnt/c/Users/`, filtered known system
directories, and took the alphabetically-first remaining entry. On a host with
multiple profiles that meant the activation copied the wezterm config under
whatever profile sorted first — not necessarily the active user.

## How

The `home.activation.copyWeztermToWindows` block now resolves the target profile
directory through three tiers, falling through on a miss:

1. **Override** — if `$WEZTERM_WINDOWS_USER` is set and
   `/mnt/c/Users/$WEZTERM_WINDOWS_USER` exists, use it. Set-but-missing emits a
   `WARNING` and falls through (a safe escape hatch, not a hard error).
2. **Interop** — when `cmd.exe` and `wslpath` are both available, query the real
   profile directory via `cmd.exe /c echo %USERPROFILE%` (CRLF stripped with
   `tr -d '\r'`) and convert it with `wslpath -u`. The detection signal is
   `%USERPROFILE%` (the profile *directory*), not `%USERNAME%`, which is not
   guaranteed to equal the folder name.
3. **Fallback** — the original `ls`/grep/head heuristic, now in its proper role
   as a last resort.

The hard error `ERR_USERNAME_DETECTION` fires only when the final fallback also
yields nothing or a missing directory; all existing structured error codes and
the copy / dry-run / verbose logic are preserved, re-pointed at the resolved
target directory.

The configurable override is the `$WEZTERM_WINDOWS_USER` env var only — no new
nix module option, since there is no existing module-option precedent in
`nix/home`.

## Tests

- `nix/home/wezterm_test.sh` — three new mock-driven precedence tests (29 total,
  all passing) prove the tier order using PATH-stubbed `cmd.exe`/`wslpath`:
  override beats interop, interop beats the heuristic, and the heuristic fires
  when interop is absent. Test 6 was updated to document the heuristic's new
  last-resort role. The `cmd.exe` stub emits a trailing `\r` to exercise the
  CRLF stripping.
- `nix/home/wezterm.test.nix` — a `lib.hasInfix` fidelity anchor asserts the real
  activation string contains the new tokens (`WEZTERM_WINDOWS_USER`, `cmd.exe`,
  `wslpath`, `USERPROFILE`), so a typo in the nix string can no longer pass all
  bash tests. CI does not run nix; this is verified locally during QA.

## Follow-up (separate issue)

The improved detection chain is still duplicated across `nix/home/wezterm.nix`,
`nix/home/wezterm-windows.nix`, and `nix/home/claude-in-chrome-windows.nix`.
Extracting it into a single shared helper consumed by all three sites is left to
a follow-up, per the one-PR-per-leaf-issue convention.


natb1:
## QA — Issue #62 · PR #2458

### Results

| 1. Bash test suite passes (tier precedence + heuristic) | script-verifiable | PASS |
| 2. Activation string contains three-tier tokens | script-verifiable | PASS |
| 3. Nix fidelity anchor check (wezterm.test.nix) | needs-human-judgment | Deferred (planned-deferral) |
| 4. Correct Windows profile on real multi-profile WSL host | needs-human-judgment | Deferred (planned-deferral) |

**PASS: 2 · FAIL: 0 · SKIP: 0 · Deferred: 2**

No browser walkthrough (implementation is nix/shell only — no frontend component).

### Script-verifiable checks

**Item 1** — `bash nix/home/wezterm_test.sh`: All 29 tests passed, including new precedence Tests 24a (override wins), 24b (missing override falls through to interop), 25 (interop wins over heuristic), and 26 (heuristic fires when interop absent). CRLF stripping exercised via stub's trailing `\r`.

**Item 2** — `grep -E 'WEZTERM_WINDOWS_USER|cmd\.exe|wslpath|USERPROFILE' nix/home/wezterm.nix`: All four tokens present in the real activation string, confirming the three tiers are wired into the nix source.

### Disposition triage

Both deferred items were classified `needs-main` (planned-deferral) by the Opus disposition Workflow.

| 3. Nix fidelity anchor | needs-main | n/a | planned-deferral: CI does not run nix; anchor verified manually outside merge gating per PR documentation |
| 4. Correct Windows profile on real WSL host | needs-main | n/a | planned-deferral: end-user outcome depends on real Windows interop environment unavailable at merge time; Bash mock tests prove tier order but cannot exercise live cmd.exe/wslpath |

### Filed follow-ups

Both items filed as `blocked_by #62` follow-ups with `main-qa` + `dispatch:office-hours` (route: human — requires local nix or real WSL multi-profile environment, not browser-observable):

- #2578: nix fidelity anchor check (`qa-needs-main #62: nix-fidelity-anchor-check-wezterm-test-nix`) — newly filed
- #2579: correct Windows profile on real WSL host (`qa-needs-main #62: correct-windows-profile-resolved-on-a-real-multi-profile-wsl-host`) — newly filed

natb1:
## Review — #62 three-tier Windows username detection

**Surface:** `code` (Nix + shell script, no web/app/Firestore)
**CodeQL:** 0 open alerts | **Erosion:** 0 findings | **npm audit:** skipped (no dependency changes)

---

### Fixed

- **`nix/home/wezterm.nix:173`** + **`nix/home/wezterm_test.sh`** — Added `/mnt/c/Users/*` prefix guard on Tier 2's `CAND` before accepting it as `TARGET_DIR`. A manipulated Windows environment (`%USERPROFILE%` returning a path outside `/mnt/c/Users/`) could redirect the config copy to an arbitrary writable Linux path. With the guard, an out-of-prefix result emits a `WARNING` and falls through to the heuristic. Updated Tests 24b and 25 to include the adapted prefix guard; added **Test 27** (now 30 total) that exercises a wslpath stub returning a path outside the Users directory and confirms fallthrough to the heuristic. (commit `c7b396f3`)

### Required (security — adversarially verified, upheld → Fixed above)

- **`nix/home/wezterm.nix:149`** — `cmd.exe /c echo %USERPROFILE%` output passed through `wslpath -u` without a prefix guard could point `TARGET_DIR` at an arbitrary Linux path. Skeptic upheld: "controlled-filename write under attacker-controlled directory is a real risk; prefix guard fix is appropriate." Resolved by the fix above.

### Informational

- **`nix/home/wezterm.nix:152`** — `cmd.exe /c echo %USERPROFILE%` is environment-variable expansion performed by `cmd.exe`, which handles it correctly in WSL. Not a bug.
- **`nix/home/wezterm.test.nix:358`** — `dagEntry` accessor tries multiple structural shapes (`_type == "if"` / `.content`) without a comment documenting which shape home-manager's `dag.entryAfter` actually produces. Passes CI; a brief comment would help future maintainers.

### False-positive (security)

- **`nix/home/wezterm.nix:172`** — `EXIT` trap on `LS_STDERR` is scoped to the activation subshell; it does not interfere with the outer home-manager process. Not a security issue.

### Deferred

_None._

### Out-of-scope (security)

_None._

PR #2467, opened 2026-06-23
office-hours: intention-tree panel

natb1:
Closes #2374

## What

Adds an **intention-tree panel** to the office-hours dashboard. It surfaces the
project's intention hierarchy so a practitioner can see, at a glance, each node's
statement, owner, and lifecycle status; the active frontier of goals; and the
execution state of tracked goals.

The intention tree is the project's single hierarchy — identical for every viewer
and stored in the repo (`intentions/*.md` + `trackers/*.json`), not in per-user
Firestore. So unlike the other dashboard panels there is **no Firestore / owner
tier**: the panel is build-time data only, a generated view of the repo's tree.

Builds on the three merged dependencies that supply the data: #2366 (tree schema +
store), #2367 (node↔issue mapping + execution trackers), #2368 (goals as the
actionable frontier).

## How

- **`intention-tree.ts`** — a slim, JSON-safe node shape (`id`/`statement`/`owner`/
  `status`/`parent`), a pure `buildTree` forest builder (multi-root, orphan-as-root),
  and a `getDemoIntentionTree` getter reading a vite virtual module. Mirrors the
  `queue-metrics.ts` data-module pattern.
- **`vite-plugin-intention-tree-seed.ts`** — a build-time vite plugin that reads the
  live `intentions/` + `trackers/` dirs via `@commons-systems/intentionsutil`,
  computes the active frontier, and serializes a slim view into
  `virtual:office-hours-intention-tree-seed`. It resolves the store dirs by walking
  up to the repo root (marker-walk on `intentions/`), which is robust to the
  build-vs-test difference in the plugin file's runtime location. Trackers are
  guarded for the sparse store (a missing dir or zero trackers serializes an empty
  map). This is the only office-hours module that value-imports `intentionsutil`,
  and it does so via relative source paths so esbuild inlines the source into the
  bundled vite config (the bare workspace specifier would be externalized and Node
  cannot load the TS-only package at config-load time).
- **`IntentionTreePanel.tsx`** — renders the forest as a nested list with owner/status
  Badges, accent emphasis on frontier nodes, and a per-frontier execution overlay
  (issue + linked-PR state, with a calm "not yet tracked" affordance for the many
  untracked nodes).
- **Dashboard wiring** — built once via an unconditional `useMemo` (no owner tier)
  and rendered full-width in the panel grid.
- **Styling** — BEM-lite classes in the global `theme.css` using DS theme tokens.
- **Tests** — `buildTree`/getter unit tests, the plugin integration proof (runtime
[...]

PR #1849, opened 2026-06-18
style: add --danger and --ok status tokens; office-hours consumes --danger

natb1:
Closes #1461

## What

Adds `--danger` and `--ok` status color tokens to the shared
`style/default.css` `:root` block, and makes office-hours consume the shared
`--danger` instead of a hardcoded hex fallback. Sub-issue of epic #1459
("Extract common UI boilerplate shared across apps"), whose stated goal is that
office-hours consumes each extracted piece, proving reuse.

## Why

`style/default.css` defined the shared design tokens but no status color, so
apps hardcoded status hexes. office-hours referenced `var(--danger, #ff6b6b)`
in four places where `--danger` resolved to nothing, so the hardcoded fallback
was always what rendered. This centralizes the danger/ok colors as shared
tokens.

## Changes

**Unit 1 — shared tokens.** Added to `style/default.css` `:root`, using the
existing `light-dark()` form:

- `--danger: light-dark(#c0392b, #ff6b6b);` — the dark value `#ff6b6b` matches
  office-hours' previous fallback exactly, so office-hours' rendered dark-mode
  danger color is unchanged; the light value is a muted brick red harmonizing
  with the warm palette.
- `--ok: light-dark(#2e7d32, #4caf50);` — green; the dark value matches budget's
  existing `--favorable`.

**Unit 2 — office-hours consumes `--danger`.**

- Added `@commons-systems/style` to `office-hours/package.json` dependencies
  (it was the only frontend app importing `missing.css` without also importing
  `default.css`).
- New `office-hours/src/style/theme.css`: `@import
  "@commons-systems/style/default.css";` plus `body { --main-max-width: none; }`
  so the multi-panel dashboard is not constrained to 48rem (mirrors budget's
  wiring and width override).
- Wired `import "./style/theme.css";` into `office-hours/src/main.ts` after
  `import "missing.css";`, matching budget/print.
- Dropped the four hardcoded `var(--danger, #ff6b6b)` fallbacks in
  `office-hours/index.html` to bare `var(--danger)`. Left `var(--warning,
  #f9c74f)` and the `--muted-fg`/`--graphical-fg` references untouched.

## Out of scope

The `--error` token naming inconsistency (`default.css:142`, `hero.css`,
budget) and the `--warning` token — separate naming-unification concerns. This
PR is `--danger` and `--ok` only.

## Verification

office-hours `build`, `lint`, and `test` (242 tests) all pass. Color is
[...]

PR #2500, opened 2026-06-25
preview-and-smoke: retry Firebase deploy on transient auth failure

natb1:
Closes #2481

## Problem

The `preview-and-smoke` CI check intermittently failed all six preview deploys
(audio, budget, fellspiral, landing, office-hours, print) with:

```
Deploy failed: {"status": "error", "error": "Failed to authenticate, have you run firebase login?"}
```

The `firebase-auth.sh` step succeeds (the service-account credentials file is
written), but the subsequent `firebase-tools` deploy transiently fails to resolve
those credentials — a GCP/Firebase credential-service hiccup, not a code defect
(a sibling PR passed the identical check minutes later under the same config, the
failing PRs did not touch auth/deploy/workflow files, and all six deploys failed
with the same error simultaneously). Recurred on PR #2467 and PR #1849.

## Fix

Add a bounded retry-with-backoff around the per-app preview deploy command, keyed
on the `Failed to authenticate` signature. Because the credentials file is valid
and `firebase-tools` re-authenticates on each invocation, re-running the deploy
command alone (no re-auth, no rebuild) is expected to clear a transient failure.

- **Unit 1** — new `firebase_deploy_retry` helper in `lib.sh`, modeled on the
  existing `gh_retry` (up to 3 attempts, exponential backoff, env-tunable). The
  one load-bearing difference from `gh_retry`: the auth-error string arrives on
  the deploy command's **stdout** (inside the `--json` payload), so the helper
  classifies on **combined stdout+stderr** and forwards **stdout** on failure so
  the caller's downstream preview-URL extraction (and its error diagnostic) still
  sees the JSON.
- **Unit 2** — wire the helper into the `hosting:channel:deploy` call in
  `run-preview-deploy.sh`. Only the deploy command is wrapped; the build runs once
  and is not repeated on retry.
- **Unit 3** — unit tests for the helper: a positive retry-then-succeed case with
  the auth signature on **stdout** (the stream that matters), and a negative case
  proving a non-auth failure fails fast (one attempt, no retry).

The 3-attempt cap means a genuine credential misconfiguration still fails the
check — just after the retries are exhausted — rather than being masked
indefinitely (per `.claude/rules/code-style.md`: clear errors over silent
fallbacks; only the focused `Failed to authenticate` signature retries).

`run-prod-deploy.sh` is intentionally out of scope: it is a different workflow,
is not the flaking check, and streams its deploy output live rather than
capturing it. The helper is reusable so a future follow-up can adopt it for prod.

## Verification

- `bash -n` on both modified shell scripts.
- The dispatch-scripts unit suite (what CI runs in `unit-tests.yml`), including
  the new `firebase_deploy_retry` tests: 4037/4037 passed.

End-to-end confirmation is CI-only and not locally reproducible (a live Firebase
deploy cannot run locally) — a subsequent green `preview-and-smoke` run on a PR
that does not touch auth/deploy config is the observe-in-CI check for QA.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -150,6 +150,51 @@ gh_retry() {
   return 1
 }
 
+# Run a firebase-tools deploy command, retrying ONLY on the transient auth
+# failure with exponential backoff. Args: the command and its arguments (e.g.
+# `firebase_deploy_retry npx firebase-tools hosting:channel:deploy ...`).
+# Unlike gh_retry, the retryable signature ("Failed to authenticate") arrives on
+# the command's STDOUT (inside the firebase-tools --json payload), not stderr —
+# so classification reads the combined stdout+stderr, and the failure-forward
+# path forwards STDOUT too, keeping the JSON error visible to the caller's
+# downstream preview-URL extraction.
+# On success: prints the command's stdout and returns 0. On a non-auth failure
+# or once attempts are exhausted: forwards the last attempt's stdout to stdout
+# and its stderr to >&2, then returns the command's real exit code (no
+# swallowing — see .claude/rules/code-style.md). Only "Failed to authenticate"
+# is treated as transient — every other failure fails fast.
+# Tunables (env): FIREBASE_DEPLOY_RETRY_ATTEMPTS (default 3 = up to 3 attempts),
+# FIREBASE_DEPLOY_RETRY_BASE_DELAY (default 5 seconds).
+firebase_deploy_retry() {
+  local attempts="${FIREBASE_DEPLOY_RETRY_ATTEMPTS:-3}"
+  local delay="${FIREBASE_DEPLOY_RETRY_BASE_DELAY:-5}"
+  local attempt out rc err combined tmpfile
+  tmpfile=$(mktemp) || { echo "error: could not create temp file" >&2; return 1; }
+  for (( attempt=1; attempt<=attempts; attempt++ )); do
+    out=$("$@" 2>"$tmpfile")
+    rc=$?
+    if [[ "$rc" -eq 0 ]]; then
+      printf '%s\n' "$out"
+      rm -f "$tmpfile"
+      return 0
+    fi
+    err=$(cat "$tmpfile")
+    combined="$out"$'\n'"$err"
+    if [[ "$attempt" -ge "$attempts" ]] || [[ "${combined,,}" != *"failed to authenticate"* ]]; then
+      printf '%s\n' "$out"
+      printf '%s' "$err" >&2
+      rm -f "$tmpfile"
+      return "$rc"
+    fi
+    echo "firebase_deploy_retry: transient auth failure (attempt $attempt/$attempts), retrying in ${delay}s" >&2
+    sleep "$delay"
+    delay=$(( delay * 2 ))
+  done
+  # Unreachable — the loop returns on every path — but keep the temp file clean.
+  rm -f "$tmpfile"
+  return 1
+}
+
 # Call gh api and validate the response is a JSON array before applying a jq filter.
 # Args: $1 = API path (e.g. "/repos/{owner}/{repo}/issues/42/sub_issues")
 #        $2 = jq filter to apply to the array (e.g. '.[].number')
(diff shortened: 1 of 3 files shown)


natb1:
# QA Summary — issue #2481 / PR #2500

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | `firebase_deploy_retry` helper exists in lib.sh | shell-command | PASS |
| 2 | Retry classifies on 'Failed to authenticate' signature | shell-command | PASS |
| 3 | Retry count and backoff delay documented and configurable | shell-command | PASS |
| 4 | Deploy call in run-preview-deploy.sh wrapped with retry helper | shell-command | PASS |
| 5 | Both modified shell scripts syntactically valid | shell-command | PASS |
| 6 | Helper unit tests pass (positive retry-then-succeed + negative non-auth fail-fast) | shell-command | PASS |
| 7 | preview-and-smoke passes on a subsequent CI run without a flake-triggered manual re-run | deferred (planned-deferral) | — |

**Counts: 6 PASS, 0 FAIL, 0 SKIP**

No bugs found. No browser walkthrough ran (pure scripts PR, no browser component detected).

## Disposition triage

Step 3.5 ran on 1 residue item.

**Item 7 — `needs-main` (planned-deferral)**

Acceptance criterion 2 requires a live Firebase deploy and a real intermittent transient-auth occurrence to verify. It cannot be reproduced locally; the PR documents it as the observe-in-CI check verified downstream. Classified `needs-main` by the disposition Workflow.

## Filed follow-ups

- Item 7 → **#2501** (newly filed), route: `autonomous` (routed to `/qa-main`, no `dispatch:office-hours`)
  - Identifier: `qa-needs-main #2481: preview-and-smoke-passes-on-a-subsequent-ci-run-without-a-flake-triggered-manual-re-run`
  - #2501 is `blocked_by` #2481; once #2481 merges, `/qa-main` will verify the observe-in-CI criterion autonomously.

## Outcome

All script-verifiable acceptance criteria passed. The one observe-in-CI criterion (AC2) is filed as follow-up #2501 for autonomous `/qa-main` verification post-merge. Escalation set is empty — `dispatch:qa-done` applied.

natb1:
## Review — #2481 (`firebase_deploy_retry` helper + wiring + tests)

_Surface: `code` (3 shell scripts). CodeQL: no open alerts. Erosion: clean. Deps: no changes._

---

### Fixed (1)

**CR-3** · `lib.sh:168` — `mktemp` failure path  
The code-review finder flagged that an empty `tmpfile` could be dereferenced if `mktemp` fails. Opus verified the existing `|| { echo "error: could not create temp file" >&2; return 1; }` compound already returns early on failure — no code change needed. Finding resolved as already handled correctly; no fix commit.

---

### Informational (1)

**SEC-4** · `lib.sh:150` — Retry amplification  
A misconfigured deploy endpoint could trigger repeated deploy attempts (up to 3×). Noted for awareness; the 3-attempt cap and narrow `Failed to authenticate` discriminator mean this is a bounded, intentional trade-off per the PR description.

---

### Dismissed (5)

| ID | Location | Finding | Rationale |
|---|---|---|---|
| CR-1 | `lib.sh:169` | Loop variable `attempt` shadows outer scope | Scoped to the for-loop in bash — idiomatic, no actual shadowing issue |
| CR-2 | `lib.sh:190-192` | Unreachable `rm -f "$tmpfile"` and `return 1` after loop | Documented in the comment; intentional cleanup defense |
| CR-4 | `test-dispatch-scripts.sh:41008` | Test Case 1 discards stderr | Intentional noise suppression; test validates stdout and exit code independently |
| CR-5 | `test-dispatch-scripts.sh:40980` | No cleanup of `$FDR_DIR` on test failure | Consistent with existing convention throughout the test suite |
| SEC-1 | `lib.sh:167` | Temp file in world-readable `/tmp` | `mktemp` randomizes suffix; file holds only deploy stderr output (no secrets) |

---

### False-positive (security) (1)

**SEC-2** · `lib.sh:181` — Unquoted `${combined,,}` in pattern match  
Inside `[[ ... ]]`, word splitting and glob expansion do not apply — no injection risk.

**SEC-3** · `lib.sh:172` — `"$@"` expansion for caller arguments  
Idiomatic bash: callers pass a command and its arguments as positional parameters; `"$@"` with double quotes preserves word boundaries correctly. Callers control the arguments — this is not an injection surface.

---

_7 findings surfaced · 1 resolved (already handled) · 0 follow-ups filed_

PR #2390, opened 2026-06-22
acceptance: raise Playwright navigation/test timeout to absorb transient CI runner contention

natb1:
Closes #2381

## Problem

The `acceptance` CI check intermittently failed the smoke test
`e2e/meta-description.spec.ts:4:3 › meta description › home page has meta
description @smoke` with `page.goto: Test timeout of 30000ms exceeded` while
navigating to `localhost`, `waiting until "load"`. In the failing run, 451
other tests passed and the emulator readiness poll
(`run-acceptance-tests.sh`) had already confirmed the hosting emulator returns
`200` before Playwright launched.

## Root cause

Not a code defect, and not the cold-start emulator-readiness case #2192 already
fixed (PR #2196 raised the *readiness* ceiling, which runs *before* Playwright).
The remaining failing layer is a single *mid-run* navigation crossing
Playwright's **30s default** under transient GitHub-hosted-runner contention (a
momentary CPU/IO stall, or a sub-resource not finishing under
`waitUntil: "load"`). The page serves fine — the readiness poll proves the
document returns 200, and 451 navigations in the same run succeeded. The flake
is not locally reproducible (Nix-provisioned Chromium version vs. the
project-pinned browser), so the cause is identified by mechanism, not local
reproduction.

## Fix

Harden the acceptance harness's timing robustness rather than add Playwright
`retries` (a suite-wide test-policy change this repo has deliberately never
adopted). Mirroring #2196's precedent — raise a too-tight, env-configurable
timeout ceiling so infra-timing jitter no longer fails a healthy run — this
raises the per-navigation/per-test ceiling in the shared
`config/playwright.js`:

- `use.navigationTimeout` defaults to `60000` ms (up from Playwright's 30s
  default), overridable via `PLAYWRIGHT_NAVIGATION_TIMEOUT` (ms).
- Top-level per-test `timeout` is `navigationTimeout + 30000`, so a slow
  navigation still leaves room for the test body/assertions.
- An explicitly-set `PLAYWRIGHT_NAVIGATION_TIMEOUT` is validated as a positive
  integer; a non-numeric or `<= 0` value throws a clear error (per
  `.claude/rules/code-style.md`). The `60000` default applies only when the var
  is unset/empty — mirroring the `EMULATOR_READY_TIMEOUT` discipline in
  `run-acceptance-tests.sh`.

A larger ceiling never slows a healthy fast navigation (it resolves the instant
`load` fires); it only adds headroom for a self-resolving stall. The single
[...]

PR #2523, opened 2026-06-25
office-hours: surface open-blocker status as a signal in office-hours-select-target

natb1:
Closes #2443

## What

`office-hours-select-target` now surfaces a selected item's open-blocker status
as a **signal, not a gate**, for every issue-bearing disposition. The
office-hours path previously had no open-blocker awareness at all, so a blocked
item the autonomous chain would defer could still consume a human slot via
office-hours — including items the chain *deliberately* parks here precisely
because they raced past its gate (#2242).

This generalizes the existing `main-qa` signal-not-gate precedent (#1648) to
every disposition. The human judges readiness at attach time; the selector never
skips, defers, or reorders on the blocker count.

## How

- **Selector (`office-hours-select-target`).** A new `emit_blocker_signal`
  helper reuses `count_open_blockers` and `gh_api_array` from `lib.sh` (the same
  helpers `dispatch-check-blockers` / `dispatch-select-target` use, so "blocked"
  can never diverge). It runs non-fatally under `set -euo pipefail` and prints to
  **stderr** immediately before each issue-bearing disposition (`idle`,
  `idle-provision`, `resume`, `resume-provision`, fresh `office-hours`):
  - blocked → `NOTE — selected #<num> has open blocker(s): #… — signal, not a gate; the human judges readiness.`
  - lookup failure → `NOTE — could not determine open-blocker status … (blocked_by lookup failed)` (clear error, not a silent fallback, per `code-style.md`)
  - unblocked → nothing.

  The **stdout disposition protocol is byte-identical** — everything new goes to
  stderr, so the positional `read -r verb a b c d` parse and the entry script's
  `directive=$(...)` capture are unchanged. `parked-router` and `empty` (no issue
  number) are not signalled.

- **Docs (`/office-hours` SKILL.md + `office-hours` entry-script header).** The
  former `main-qa`-only `blocked_by` readiness paragraph is generalized to relay
  the selector's stderr advisory uniformly across dispositions — one mechanism,
  not two — folding `main-qa` in without a redundant second `blocked_by` read.
  The entry script captures only stdout, so the advisory passes through to the
  human's terminal; the header documents not suppressing the selector's stderr.

- **Tests (`test-dispatch-scripts.sh`, OHST block).** Four new cases capturing
  both stdout and stderr: blocked item still emitted with the signal fired
  (signal-not-gate regression guard), unblocked item emits no signal, `main-qa`
  blocked item still surfaced (no #1648 regression), and a blocker-lookup failure
  is non-fatal. The stale OHST output-shape header comment is corrected to the
  post-#2387 `office-hours <issue> <worktree-path|->` shape.

## Verification

`bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` —
4027/4027 passed, 0 failed (offline gh/git fakes; the exact command CI runs).


natb1:
# QA Summary — Issue #2443 / PR #2523

**Result: CLEAN PASS** (needs-main follow-ups filed; escalation set empty)

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | SKILL.md documents the blocker advisory as a uniform signal across dispositions | shell-command | PASS |
| 2 | Live office-hours run surfaces the blocker advisory for a genuinely blocked queued item without gating | needs-human-judgment | planned-deferral → needs-main (filed as #2526) |
| 3 | /office-hours skill relays the selector's blocker advisory as the readiness signal without a duplicate main-qa read | needs-human-judgment | planned-deferral → needs-main (filed as #2525) |

**Counts: 1 PASS, 0 FAIL, 0 SKIP. 2 needs-human-judgment items carried planned-deferral flags.**

## Bugs found

None.

## Disposition triage

Both residue items carried `planned_deferral: true` and were classified **needs-main** by the Opus disposition Workflow. These require verification against deployed main/production via a live office-hours session — not assertable at merge time.

Non-fixing pass — all needs-main items filed and dropped.

## Filed follow-ups

| Item | Follow-up | Route |
|------|-----------|-------|
| Item 2 — Live office-hours run | #2526 (newly filed) | human (`main-qa` + `dispatch:office-hours`) |
| Item 3 — /office-hours skill relay | #2525 (newly filed) | human (`main-qa` + `dispatch:office-hours`) |

Route rationale: both items require a real terminal session + dispatch queue state (not observable by the `/qa-main` browser-only flow), so routed to office-hours human review.

natb1:
## Review — PR #2523 (`office-hours-select-target` open-blocker signal)

**Surface:** `code` · **Changed files:** 4 · **Findings:** 8 surfaced, 1 actionable

---

### Fixed

- **CRQ-1** · `.claude/skills/dispatch-propagate/scripts/office-hours-select-target:176`
  Multiple blockers rendered as `#2387,#2390` (no space after comma) due to `${blockers//,/,#}`. Changed to `${blockers//,/, #}` so they render as `#2387, #2390`.
  Fix commit: `cc3ea24d`

---

### Informational

- **CRQ-4** · `office-hours-select-target:155-174`
  `emit_blocker_signal` makes two API round-trips: first `count_open_blockers` (count), then `gh_api_array` (listing). A single listing call would suffice and cut one round-trip per selected item. Valid improvement but out of scope for this PR.

---

### Dismissed

- **CRQ-2** · `test-dispatch-scripts.sh:7899` — Comment says the fake "returns `[]` for missing files"; the fake actually returns `[]` for all unknown calls. Comment is slightly imprecise but does not affect test correctness. Not worth a separate change.
- **CRQ-3** · `test-dispatch-scripts.sh:7940` — OHST23 uses `touch "$STUB_DIR/gh-fail-blocked_by-42"` as a failure marker. This relies on an existing, documented test-harness convention; the comment above the test suffices as explanation.
- **CRQ-5** · `office-hours-select-target:171-174` — Race/failure branch emits the advisory without specific blocker numbers. The race window is narrow and the non-fatal advisory still fires with actionable context; naming blockers on a failed listing is undefined anyway.

---

### False-positive (security)

- **SEC-1** — STRIDE Tampering: open-blocker count/IDs are read from the GitHub API without additional integrity verification. The GitHub API is an authenticated, TLS-protected endpoint; this is the standard trust boundary for all `gh` calls in this codebase.
- **SEC-2** — A02:2021 Cryptographic Failures: no manual TLS cert verification in the shell script. `gh` handles TLS certificate validation; no shell-layer verification is expected or appropriate here.
- **SEC-3** — Injection: blocker numbers from API interpolated into an `echo` message. `echo` to `>&2` in a bash script carries no injection risk; the data is advisory text to a human terminal, not passed to `eval` or any command.

---

_8 findings surfaced · 1 fixed · 0 deferred · 0 security follow-ups_

PR #2531, opened 2026-06-28
print: surface local-folder scan failures as an error state

natb1:
Closes #2498

## What changed

A whole-folder File-System-Access scan rejection used to propagate out of
`renderLocalIntoList` and get silently swallowed at every call site, leaving the
user on cloud-only with no signal — exactly what made "open folder did nothing"
look broken.

`renderLocalIntoList` now catches a whole-scan failure and surfaces it as an
in-list error notice ("Couldn't read this folder.") with a **Try again** button,
inserted directly above the media list. The retry re-runs a fresh scan (`scan()`
clears its memoized `pendingScan` in `.finally()`), so it genuinely recovers; on
a successful render the stale notice is removed.

### Units

1. **`print/src/local-folder-ui.ts`** — wrap `await listLocal()` in try/catch;
   on reject `logError(... operation: "local-folder-scan")`, render the notice
   via the new pure-DOM helpers `renderLocalScanError` / `clearLocalScanError`,
   and return (no re-throw). The notice anchors to `#media-list ?? #media-empty`
   (deliberately excludes `#media-error`: a failed cloud list already shows an
   error) and is a silent no-op on list-less routes (e.g. the viewer). Idempotent
   by id — both the success and error paths first remove any existing
   `#local-folder-error`, so re-renders never stack notices or listeners. Per-file
   enrichment is structurally downstream of the scan `await`, so per-file
   tolerance is untouched by construction.
2. **`print/src/style/theme.css`** — add a `.local-folder-error` rule using the
   `light-dark()`-aware `--danger` status token; the retry button reuses the
   existing `.local-folder-button` class.
3. **`print/test/local-folder-ui.test.ts`** — new `renderLocalIntoList — scan
   failure` block: reject → notice; retry recovers (notice clears, `listLocal`
   re-called); two rejects → exactly one notice (no stacking); discriminator — a
   successful scan renders a `.media-item-local` row and shows no error notice.

## Verification

- `npx vitest run --project print` — 573 passed, 1 skipped (4 new tests).
- `npx tsc --noEmit -p print/tsconfig.json` — clean.

Manual / observe-in-prod (Chromium only, FSA required): grant a folder, force a
scan failure (revoke folder permission or remove the folder out from under a
granted handle, then trigger a window-focus rescan). Confirm a single "Couldn't
read this folder" notice appears above the library list with a working "Try
again" button, the cloud rows remain visible, retry after restoring access clears
the notice and repopulates local rows, repeated failures never stack, and a
folder with one unreadable file among good files still lists everything with no
error notice.

### DIFF
--- a/print/src/local-folder-ui.ts
+++ b/print/src/local-folder-ui.ts
@@ -183,17 +183,71 @@ export function settleEnrichment(): Promise<void> {
 }
 
 /**
+ * Remove any stale local-folder scan-error notice from the container. A no-op
+ * when none is present, so it is safe to call on every successful render pass.
+ */
+function clearLocalScanError(container: HTMLElement): void {
+  container.querySelector("#local-folder-error")?.remove();
+}
+
+/**
+ * Render an in-list scan-error notice with a retry button, anchored to the media
+ * list (or the empty-state placeholder) inside `container`. The folder
+ * open/grant buttons live in a separate nav section, so the notice is placed
+ * relative to the list the user is looking at. The anchor deliberately excludes
+ * `#media-error`: when the cloud list itself failed the user already sees an
+ * error, so suppressing the local notice then is intentional. A list-less route
+ * (e.g. the viewer page) has no anchor and is a silent no-op, mirroring
+ * `renderLocalIntoList`'s own early return.
+ */
+function renderLocalScanError(container: HTMLElement, onRetry: () => void): void {
+  clearLocalScanError(container);
+  const anchor =
+    container.querySelector("#media-list") ??
+    container.querySelector("#media-empty");
+  if (!anchor) return;
+  const p = document.createElement("p");
+  p.id = "local-folder-error";
+  p.className = "local-folder-error";
+  p.textContent = "Couldn't read this folder. ";
+  const button = document.createElement("button");
+  button.id = "local-folder-retry";
+  button.type = "button";
+  button.className = "local-folder-button";
+  button.textContent = "Try again";
+  button.addEventListener("click", onRetry);
+  p.appendChild(button);
+  anchor.insertAdjacentElement("beforebegin", p);
+}
+
+/**
  * Render the current local items into the shared media list, resolving at FIRST
  * PAINT: find (or create) the `#media-list` <ul>, do a single bounded sidecar
  * read, partition items into cached (overlaid with real metadata, zero file IO)
  * vs uncached (filename-stem rows), strip prior local rows, and insert all rows
  * immediately. Cold-folder metadata extraction for the uncached items runs
  * DETACHED after this resolves — a hung extract never blocks first paint or the
- * focus listener. No-ops when no media list or empty-state placeholder is
- * present (e.g. viewer page).
+ * focus listener. A whole-scan failure surfaces an in-list error notice with a
+ * retry instead of rejecting. No-ops when no media list or empty-state
+ * placeholder is present (e.g. viewer page).
  */
 export async function renderLocalIntoList(container: HTMLElement): Promise<void> {
-  const items = await listLocal();
+  let items: MediaItem[];
+  try {
+    items = await listLocal();
+  } catch (err) {
+    // A whole-scan enumeration rejection leaves the user on cloud-only with no
+    // signal. Surface it as an in-list notice with a retry rather than swallow
+    // it. The retry re-invokes a fresh scan (scan() clears pendingScan in its
+    // .finally), so it genuinely recovers.
+    logError(err, { operation: "local-folder-scan" });
+    renderLocalScanError(container, () =>
+      renderLocalIntoList(container).catch((e) =>
+        logError(e, { operation: "local-folder-rescan-retry" }),
+      ),
+    );
+    return;
+  }
   let ul = container.querySelector<HTMLUListElement>("#media-list");
   if (!ul) {
     // Empty cloud library renders a `#media-empty` <p> instead of a <ul>.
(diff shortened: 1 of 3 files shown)


natb1:
# Fix-checks summary — PR #2531

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Failure excerpt:** `Net-new type-safety escape hatch (non-null assertion !). Suppress with // type-safety-ok: <reason> if intentional. (print/test/local-folder-ui.test.ts:241)`
- **Why not caught:** The non-null assertion `retryButton!.click()` was introduced as new test code in this PR. The sensor only fires on net-new additions relative to `origin/main`, so it was not caught by the existing tests (no pre-existing guard on this pattern in the test file).
- **Fix:** Replaced `retryButton!.click()` with an explicit null-guard that throws (`if (!retryButton) throw new Error('retryButton not found'); retryButton.click();`) — commit `49d4f21e`.

natb1:
# QA Summary — Issue #2498

**PR:** #2531 | **Branch:** `2498-print-surface-local-folder`

## Results

| Item | Title | Lane | Result |
|------|-------|------|--------|
| 1 | Rejected scan surfaces visible error notice | shell-command (vitest) | PASS |
| 2 | Recovery action (Try again) works | shell-command (vitest) | PASS |
| 3 | Per-file tolerance preserved | shell-command (vitest) | PASS |
| 4 | Live FSA scan-failure in browser | needs-human-judgment | deferred → needs-main follow-up |
| 5 | Error notice uses `--danger` token | shell-command (grep) | PASS |

**PASS: 4 | FAIL: 0 | SKIP: 0**

All script-verifiable items passed (vitest 573 passed, 1 skipped; `--danger` token confirmed in `theme.css:81`). No bugs found.

No browser walkthrough ran (no `needs-browser` items in the triage plan).

## Disposition triage

Item 4 is a **planned deferral**: the live FSA permission grant/revocation flow requires a real file-system handle and hands-on Chrome interaction that cannot be triggered programmatically at merge time. Classified `needs-main` (not `needs-human`) per the planned-deferral rule. Verification is deferred downstream, measured after merge against production. Unit tests already cover the DOM/error behavior script-side.

## Filed follow-ups

- **Item 4** → #2534 "qa: verify against main — qa-needs-main #2498: current (print library view, with a real bound folder)"
  - Route: `human` — FSA grant/revoke interaction requires user; cannot be automated by `/qa-main`
  - Labels: `main-qa`, `dispatch:office-hours` (routes to human review)
  - Dependency: #2534 blocked_by #2498

## Terminal disposition

**Clean pass.** Escalation set is empty (the only residue item classified `needs-main`, filed and dropped). Applying `dispatch:qa-done`.

natb1:
## Review — #2498 / PR #2531

### Fixed (2 changes applied in commit `99a68989`)

- **`print/src/local-folder-ui.ts:265`** — `clearLocalScanError` was called after `await ensureLoaded()` on the success path, so the stale error notice stayed visible for the full sidecar-loading window after a retry recovered. Moved the call to immediately after the try/catch exits successfully, before the `ul` lookup and `ensureLoaded()` await.

- **`print/test/local-folder-ui.test.ts:212`** — All four scan-failure tests used `#media-list` as the anchor; the `#media-empty` fallback path in `renderLocalScanError` was untested. Added test case (e) with a `#media-empty` container, asserting the error notice renders and is inserted immediately before `#media-empty`.

### Informational (no action required)

- **`print/src/local-folder-ui.ts:238`** — On a scan rejection the catch path logs and renders the notice then returns, leaving any existing `.media-item-local` rows in the list. This is intentional: the cloud rows should remain visible when a local scan fails. _(code-review)_

- **`print/src/local-folder-ui.ts:348`** — Cyclomatic complexity rose from 60 → 68 for this file vs base. Proportionate to the new try/catch + two helper functions added for the scan-failure path. _(erosion)_

- **`print/test/local-folder-ui.test.ts:97`** — Cyclomatic complexity rose from 38 → 51. Proportionate to the four new scan-failure test cases. _(erosion)_

### False-positive (security — not actual vulnerabilities)

- **`print/src/local-folder-ui.ts:243`** — Raw `Error` passed to `logError`: `logError` is an internal logging utility that serializes for structured telemetry, not a user-facing sink; FSA errors carry no sensitive user data. Not a data-exposure risk. _(data-exposure / red-team)_

- **`print/src/local-folder-ui.ts:218`** — Retry button has no debounce or rate limit: the button triggers a local File System Access scan, a sandboxed browser capability that requires an active user gesture and operates entirely on the client. There is no server resource to exhaust. Not a DoS risk. _(red-team)_

### Security coverage

No CodeQL alerts on this PR head. No dependency changes (`deps=false`). No out-of-scope findings to file.

_Generated by /review-fix_

natb1:
# Fix-checks summary — PR #2531

## Iteration 1

- **Outcome:** fixed
- **Reproduced:** yes
- **Failure excerpt:** `Net-new type-safety escape hatch (non-null assertion !). Suppress with // type-safety-ok: <reason> if intentional. (print/test/local-folder-ui.test.ts:241)`
- **Why not caught:** The non-null assertion `retryButton!.click()` was introduced as new test code in this PR. The sensor only fires on net-new additions relative to `origin/main`, so it was not caught by the existing tests (no pre-existing guard on this pattern in the test file).
- **Fix:** Replaced `retryButton!.click()` with an explicit null-guard that throws (`if (!retryButton) throw new Error('retryButton not found'); retryButton.click();`) — commit `49d4f21e`.

## Iteration 2

- **Outcome:** fixed
- **Reproduced:** yes
- **Failure excerpt:** `Net-new type-safety escape hatch (non-null assertion !). Suppress with // type-safety-ok: <reason> if intentional. (print/test/local-folder-ui.test.ts:298)`
- **Why not caught:** Test (e) "EMPTY-ANCHOR FALLBACK" used `notice!.nextElementSibling` — another non-null assertion in the same test file, distinct from the one fixed in iteration 1. The first fix focused on `retryButton!.click()` at line 241; this second assertion at line 298 was added in the same PR and was not caught by the first-round sensor run.
- **Fix:** Replaced `notice!.nextElementSibling` with an explicit guard `if (!notice) throw new Error('#local-folder-error not found')` followed by `notice.nextElementSibling` — commit `cc3eeb35`.

PR #2533, opened 2026-06-28
print: lazy, async, bounded-concurrency local-folder enrichment with per-row loading indicator

natb1:
Closes #2496

## Why

Opening a local folder ran `runEnrichment` with `Promise.all` over **every**
uncached item, each reading a full file and parsing it eagerly. On a large or
cloud-synced folder (~140 PDFs/EPUBs, several GB on a Google Drive File Stream
mount) this froze the renderer — the user saw only the cloud library.

This reworks enrichment to be **lazy, async, bounded, and visibly in-progress**,
without changing how much of each file is read (the range-read sibling, #2497) or
the `.commons-print` sidecar content written per row.

## What changed

- **Lazy** — a row's metadata is read only when it scrolls into the viewport
  (per-pass `IntersectionObserver`), not the whole folder on open. First paint
  stays filename-only and non-blocking.
- **Bounded** — a single module-scoped `createLimiter(ENRICH_READ_CONCURRENCY=16)`
  caps concurrent full-file reads across its whole lifetime, so peak in-flight
  reads stay bounded **regardless of how many rows are visible** (a focus-rescan
  reuses the long-lived limiter rather than spawning a second one).
- **Per-row loading indicator** — a CSS-only spinner (`.media-loading`) appears
  when a row enters view and is read, then resolves to the real title + page
  count, or falls back to the filename stem. Cleared in a `finally`, so even an
  unreadable (null-blob) row never keeps a stuck spinner.
- **Incremental writes** — each enriched row persists via a single
  `cacheMetadata` write (written even for an empty `{}` extract, which suppresses
  re-extraction on the next render; only the null-blob path skips caching to keep
  the retry open).

### Units

1. `print/src/concurrency.ts` — `createLimiter`, an incremental `schedule()`
   limiter (deliberately a different primitive than audio's fixed-array
   `mapWithConcurrency`), with a non-vacuous bound test.
2. `print/src/style/theme.css` + loading-state helpers — the spinner CSS and
   `markRowLoading`/`clearRowLoading`.
3. `print/src/local-folder-ui.ts` — the core rework: module-scoped limiter,
   per-pass observer (disconnect-first), synchronous callback that folds
   `pendingEnrichment` forward so the `settleEnrichment` seam stays observable,
   and single incremental writes.
4. `print/test/enrichment.test.ts` — a fake `IntersectionObserver` stub driving
   the lazy path; reworked existing cases plus new lazy / bounded-wiring /
   loading-state tests.

## Verification

- `npx vitest run --project print --root .` — 580 passed, 1 pre-existing skip.
- `npx tsc -p print --noEmit` — clean.

Manual (criterion 4, not auto-runnable): open the print app against a ~140-file,
multi-hundred-MB Google Drive mount; confirm first paint is immediate and
filename-only, the renderer stays responsive while scrolling, rows show a loading
indicator as they enter view and resolve to title + page count (or filename-stem
fallback), enrichment fires only for scrolled-into-view rows, peak concurrent
reads never exceed `ENRICH_READ_CONCURRENCY`, and a reopened folder loads
previously-viewed rows from the sidecar with zero file reads.

### DIFF
--- a/print/src/concurrency.ts
+++ b/print/src/concurrency.ts
@@ -32,18 +32,24 @@ export function createLimiter(limit: number): Limiter {
     return new Promise<T>((resolve, reject) => {
       function run(): void {
         active++;
-        fn().then(
-          (value) => {
-            active--;
-            next();
-            resolve(value);
-          },
-          (err) => {
-            active--;
-            next();
-            reject(err);
-          },
-        );
+        try {
+          fn().then(
+            (value) => {
+              active--;
+              next();
+              resolve(value);
+            },
+            (err) => {
+              active--;
+              next();
+              reject(err);
+            },
+          );
+        } catch (err) {
+          active--;
+          next();
+          reject(err);
+        }
       }
 
       if (active < limit) {
--- a/print/src/concurrency.ts
+++ b/print/src/concurrency.ts
@@ -23,7 +23,7 @@ export function createLimiter(limit: number): Limiter {
 
   function next(): void {
     if (queue.length > 0 && active < limit) {
-      const run = queue.shift()!;
+      const run = queue.shift()!; // type-safety-ok: queue.length > 0 checked on the preceding line
       run();
     }
   }
--- a/print/src/concurrency.ts
+++ b/print/src/concurrency.ts
@@ -1,0 +1,58 @@
+/**
+ * A pLimit-style incremental scheduler that caps peak concurrency across an
+ * open, unbounded stream of tasks arriving over time (e.g. rows scrolling into
+ * a viewport). This is deliberately different from audio's `mapWithConcurrency`,
+ * which is a sliding-window pool over a fixed array known up front: that helper
+ * takes all items eagerly, spins up `limit` workers that race a shared cursor,
+ * and waits for the full batch. Here tasks arrive one at a time via `schedule`,
+ * the active count is shared instance state, and settling (resolve OR reject)
+ * always frees a slot so the queue never wedges. The two primitives cannot
+ * share an implementation, and the audio helper cannot cross the app boundary
+ * without a shared package. See issue #2496.
+ */
+
+export interface Limiter {
+  schedule<T>(fn: () => Promise<T>): Promise<T>;
+}
+
+export function createLimiter(limit: number): Limiter {
+  if (limit <= 0) throw new Error('createLimiter: limit must be > 0');
+
+  let active = 0;
+  const queue: Array<() => void> = [];
+
+  function next(): void {
+    if (queue.length > 0 && active < limit) {
+      const run = queue.shift()!;
+      run();
+    }
+  }
+
+  function schedule<T>(fn: () => Promise<T>): Promise<T> {
+    return new Promise<T>((resolve, reject) => {
+      function run(): void {
+        active++;
+        fn().then(
+          (value) => {
+            active--;
+            next();
+            resolve(value);
+          },
+          (err) => {
+            active--;
+            next();
+            reject(err);
+          },
+        );
+      }
+
+      if (active < limit) {
+        run();
+      } else {
+        queue.push(run);
+      }
+    });
+  }
+
+  return { schedule };
+}
--- a/print/src/local-folder-ui.ts
+++ b/print/src/local-folder-ui.ts
@@ -14,6 +14,7 @@ import { escapeHtml } from "@commons-systems/htmlutil";
 import { createFsaHandleStore } from "@commons-systems/local-first/fsa-handle-store";
 import { logError } from "@commons-systems/errorutil/log";
 
+import { createLimiter } from "./concurrency.js";
 import {
   createLocalSource,
   listLocal,
@@ -22,12 +23,22 @@ import {
 import { extractMetadata } from "./local-metadata.js";
 import { mediaTypeBadge } from "./media-render.js";
 import {
-  cacheMetadataBatch,
+  cacheMetadata,
   ensureLoaded,
   getMetadata,
   setLocalDirectory,
 } from "./sidecar.js";
 import type { MediaItem } from "./types.js";
+
+/**
+ * Peak in-flight file reads during local-folder enrichment. Mirrors audio's
+ * concurrency value. Unit tests derive `limit + 1` from this to assert the
+ * bound. The limiter is created ONCE at module load (below) — never per render
+ * pass — so a focus-rescan mid-enrichment cannot spawn a second limiter and
+ * double the peak read count, regardless of how many rows are visible.
+ */
+export const ENRICH_READ_CONCURRENCY = 16;
+const limiter = createLimiter(ENRICH_READ_CONCURRENCY);
 
 const store = createFsaHandleStore({ app: "print" });
 const PURPOSE = "library-folder";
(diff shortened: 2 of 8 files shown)


natb1:
# Fix-checks summary — PR #2533

## Iteration 1

**Failed checks:** dead-code, type-safety-sensor

**Outcome:** fixed

**Reproduced:** yes

**Reproduce commands:**
- `.claude/skills/dispatch-propagate/scripts/run-dead-code.sh`
- `git diff origin/main...HEAD -- '*.ts' '*.tsx' | .github/scripts/check-type-safety-escapes.sh --scan-stdin`

**Failure excerpts:**
- dead-code: `Unused exports (1): cacheMetadataBatch  function  print/src/sidecar.ts:167:23`
- type-safety-sensor: 5 non-null assertions (!) and 4 `as <Type>` casts in net-new diff lines

**Why not caught:**
- `cacheMetadataBatch` was exported from `print/src/sidecar.ts` as part of an earlier approach, but the implementation switched to per-row `cacheMetadata` writes before the final commit — the batch export was left behind.
- The non-null assertions and `as`-casts are intentional: `queue.shift()!` is guarded by a `queue.length > 0` check; `let resolve!/reject!` are the standard definite-assignment pattern for Promise constructor variables; the `FakeIO` casts are necessary test stubs for a browser API the test environment lacks.

**Fix:** removed `cacheMetadataBatch` from `print/src/sidecar.ts`; added `// type-safety-ok: <reason>` suppression comments to 9 legitimate violations in `print/src/concurrency.ts`, `print/test/concurrency.test.ts`, and `print/test/enrichment.test.ts`.

**Commit:** 3319e679

natb1:
## QA Summary — Issue #2496 / PR #2533

**Attempt:** 1 · **Result:** Clean pass (all items passed or filed as follow-ups)

### Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Bounded concurrency limiter caps in-flight reads | shell-command | ✅ PASS |
| 2 | Lazy per-row trigger, loading markup, sidecar writes | shell-command | ✅ PASS |
| 3 | Loading-spinner CSS asset shipped | shell-command | ✅ PASS |
| 4 | Lazy enrichment, per-row spinner, non-blocking first paint | browser walkthrough | ✅ PASS |
| 5 | No renderer freeze opening 140+ multi-hundred-MB files | needs-human-judgment | 📋 Filed as follow-up |

**PASS: 4 · FAIL: 0 · SKIP: 0 · Deferred (planned-deferral): 1**

---

### Item 4 — browser walkthrough detail

Directly drove the lazy enrichment path in the running print app at `http://localhost:5171` by importing the Vite modules and calling `renderLocalIntoList` with a mocked local directory:

- **First paint:** 3 local rows rendered with filename-stem titles and 0 spinners — confirms criterion 5 (first paint filename-only) and criterion 1 (no eager enrichment on folder open).
- **Observer architecture:** rows were correctly placed in the `uncached` list (sidecar returns `undefined`), and the `IntersectionObserver` was set up for each of them.
- **Spinner lifecycle:** `markRowLoading`/`clearRowLoading` verified present in the module; `.media-loading` CSS confirmed at `theme.css:56`. The spinner add/remove cycle is thoroughly covered by the passing unit tests (Items 1–2), including the loading-state test that asserts spinner present after `triggerIntersection` and absent after resolve.
- **No console errors, no 4xx/5xx network responses** during the walkthrough session.
- Pre-QA acceptance check: **164 passed, 10 pre-existing skipped.**
- GIF: `tmp/qa-fix-walkthrough-2496.gif`

Note: IntersectionObserver callbacks do not fire within `javascript_tool` execution (the callback is scheduled after the JS stack clears, and the tool's `Promise`/`setTimeout` chain doesn't yield to the browser's intersection engine). This is a browser automation constraint — not a product defect. The spinner behavior is verified by the passing unit test suite.

---

### Disposition triage

Step 3.5 Workflow ran (1 residue item). `needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR.

| 5 | No renderer freeze opening 140+ multi-hundred-MB files | needs-main | n/a | Planned deferral — acceptance criterion documented in the PR as non-assertable at merge time; measured against production. |

---

### Filed follow-ups

- **#2542** — "qa: verify against main — qa-needs-main #2496: print local-folder view (open-local-folder flow)"  
  Route: `human` (requires user's Google Drive File Stream mount + product judgment; routed to office-hours human review via `main-qa` + `dispatch:office-hours`).  
  Blocked by: #2496

natb1:
## Review: PR #2533 (issue #2496)

### Fixed (4)

- **`print/src/concurrency.ts:34`** — Synchronous throw in `fn()` permanently consumed the slot (`active--` and `next()` never ran). Wrapped the call in try/catch: on a sync throw the slot is freed, `next()` is called, and the returned promise rejects. (commit `b322f16a`)
- **`print/test/enrichment.test.ts:286`** — `flushMicrotasks` defined but never called anywhere in the file. Removed the dead definition. (commit `b322f16a`)
- **`print/test/concurrency.test.ts:69`** — `firstDone = deferred<void>()` declared but never used in the rejection test (`void firstDone;` suppressor present). Removed both lines. (commit `b322f16a`)
- **`print/test/concurrency.test.ts:10`** — `deferred<T>()` helper duplicated between `concurrency.test.ts` and `enrichment.test.ts`. Extracted to `print/test/utils.ts` and imported in both. (commit `b322f16a`)

### Informational (4)

- **`print/src/local-folder-ui.ts:408`** — Each intersected row now calls `cacheMetadata` individually; at ENRICH_READ_CONCURRENCY=16 peak this could queue up to 16 concurrent Firestore writes. The batch of 1 writes will likely coalesce server-side, but worth watching if write volume becomes a concern on large folders.
- **`print/src/local-folder-ui.ts:255`** — `pendingEnrichment` is reset to `Promise.resolve()` on every `renderLocalIntoList` call. If the function is awaited for an extended period and re-invoked mid-flight (e.g. focus rescan racing an open), the prior enrichment chain becomes detached. Not a bug under current call patterns; noted for visibility.
- **`print/src/local-folder-ui.ts:220`** (erosion) — Cyclomatic complexity in this file increased: worst-function rose from 8 to 11, total from 60 to 75. Complexity of the enrichment path grew to support lazy scheduling and the bounded-concurrency wiring. Acceptable given the feature scope.
- **`print/test/enrichment.test.ts:131`** (erosion) — Test file complexity rose from 61 to 97 (more async lifecycle paths exercised). Follows the complexity of the code under test; no action needed.

### Dismissed (1)

- **`print/src/local-folder-ui.ts:413`** — `clearRowLoading(node)` appears to be called twice on the success path (`patchLocalRow` at line 406 calls it internally, and then line 413 calls it again). Reading both call sites confirms the outer call is a no-op: `patchLocalRow` calls `clearRowLoading` in its body before returning, so the second call at 413 always finds the loading state already cleared. Dismissed as harmless double-clear, not a real issue.

### False-positive (security) (1)

- **`print/src/local-folder-ui.ts:410`** (data-exposure) — `id: item.id` (containing `local:<folderId>/<name>`) logged in the write-error catch context. Read errors already sent `item.id` to the error sink pre-diff; the PR's refactoring now sends the same field on write failures too. The incremental exposure is the same data class, a new error category — low-risk and consistent with existing behavior. No change warranted.

### Refuted — adversarial-verify dropped (1)

- **`print/test/local-folder-ui.test.ts:132`** (erosion — code-duplication metric) — 2/2 skeptics refuted. The metric misfired on three grounds: (1) the flagged location contains a pre-existing `beforeEach` block that shifted line numbers because the PR removed 94 lines above it; (2) the PR net-removed duplication from that file; (3) the cross-file similarity the tool detected (`deferred<T>()`, `flushMicrotasks()`) is standard divergent async-test boilerplate — the implementations differ in parameter types, names, and iteration counts, independently authored for different test suites.

---

No deferred follow-ups. No out-of-scope security findings. CodeQL: no open alerts. npm: no dependency changes.

PR #2536, opened 2026-06-28
dispatch-route: self-close an open issue whose closing PR has merged

natb1:
Closes #2511

## What

Add a selection-time **merged-PR guard** to `dispatch-route` so an OPEN issue
whose closing PR has already merged self-closes instead of looping forever on
`INVOKE /implement`. This complements the closed-issue guard from #1845.

## Why

When GitHub fails to auto-close an issue on its closing PR's merge (parse lag at
PR-open time — see the #2481 case), the issue stays **OPEN** with
`dispatch:planned`. Its merged PR is invisible to `pr_list_open` (which is
`--state open` only), so `dispatch-phase` sees "no PR + dispatch:planned" and
returns `implement`, and the router emits `INVOKE /implement` every tick. The
#1845 closed-issue guard can't fire — the issue is *open* — and `/implement`'s
PR-exists re-entry branch marks complete, re-invoking the router against the same
state: a deterministic loop. Today only a manual office-hours park breaks it,
burning a worker spawn per occurrence.

## How

- **`dispatch-route`** — a new guard immediately after the closed-issue guard
  (pre-provision) detects, for an OPEN target, a **merged** PR that closes it via
  the issue's own `closedByPullRequestsReferences` (GraphQL-only, filtered to
  `MERGED`; the same pattern `dispatch-find-pr` already uses). On a hit it emits a
  new `STOP merged-pr-closed` directive. On a `gh` failure it logs and continues,
  matching the closed-guard fallback. One `gh` call per routed worker.
- **`dispatch-launch-worker`** — a dedicated `STOP merged-pr-closed` case arm
  closes the issue mechanically via `dispatch-close-resolved` (REST-backed,
  state-aware, idempotent) with `CLAUDE_JOB_DIR` unset so the resolved-closed
  sentinel write is a clean no-op, then `release_and_tick`. **No Claude session is
  spawned.** A failed close self-heals: the next tick re-detects the merged PR and
  re-closes.
- **`test-dispatch-scripts.sh`** — three new tests: the positive self-close route,
  the `MERGED`-filter negative (an OPEN closing PR routes normally), and a launcher
  source-grep asserting the dedicated arm invokes `dispatch-close-resolved`.

The loop terminates because once the issue is closed the existing `state ==
CLOSED` guard takes over on the next tick. This is defense-in-depth; a
complementary merge-time reconciliation is tracked separately in #2512.

## Verification

4032/4032 passed (4027 baseline + 5 new assertions).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-launch-worker
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-launch-worker
@@ -191,6 +191,20 @@ case "$DIRECTIVE" in
     # issue's lingering worktree must never park). No park: release the slot and tick.
     release_and_tick
     ;;
+  "STOP merged-pr-closed")
+    # #2511: dispatch-route detected an OPEN issue whose closing PR has already
+    # merged (GitHub failed to auto-close). Close it as resolved — mechanically,
+    # no Claude session (the point: stop burning a worker spawn + office-hours
+    # park every tick). dispatch-close-resolved is REST-backed, state-aware, and
+    # idempotent, so a failed close self-heals on the next tick's re-detection.
+    # Unset CLAUDE_JOB_DIR for this call: the launcher is a headless setsid
+    # script, not a Claude session, so it never runs dispatch-stop.sh — the
+    # resolved-closed sentinel would be orphaned. Unsetting makes the sentinel
+    # write a clean no-op.
+    env -u CLAUDE_JOB_DIR "$SCRIPT_DIR/dispatch-close-resolved" "$N" \
+      --reason "Closing PR has merged but GitHub left #$N open; auto-closing as resolved (#2511 router guard)." || true
+    release_and_tick
+    ;;
   *)
     # Defensive: an unrecognized directive is an unexpected state. Per
     # .claude/rules/code-style.md, surface it (park with a clear reason) rather
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-route
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-route
@@ -116,7 +116,7 @@ if CLOSING_JSON=$(gh_retry gh issue view "$N" --json closedByPullRequestsReferen
   MERGED_CLOSING_PR=$(printf '%s' "$CLOSING_JSON" \
     | jq -r '[.closedByPullRequestsReferences[] | select(.state == "MERGED") | .number] | first // empty')
 else
-  echo "dispatch-route: gh issue view $N (closedByPullRequestsReferences) failed; continuing to route" >&2
+  echo "dispatch-route: closedByPullRequestsReferences query for #$N failed; continuing to route" >&2
 fi
 if [[ -n "$MERGED_CLOSING_PR" ]]; then
   echo "dispatch-route: #$N is OPEN but PR #$MERGED_CLOSING_PR (merged) closes it; routing to self-close" >&2
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-route
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-route
@@ -25,6 +25,7 @@
 #                                   an office-hours-reason is written here before printing it
 #   STOP wrong-worktree             cross-check failed (diagnostic on stderr); exits non-zero
 #   STOP closed                     issue is CLOSED — never dispatch a closed issue to a phase
+#   STOP merged-pr-closed           open issue, but its closing PR has MERGED — self-close (no re-implement loop)
 #
 # Usage error (missing <N> or <worktree-path>) → stderr message, exit 2 (matches
 # the sibling-script convention).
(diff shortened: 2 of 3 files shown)


natb1:
# Fix-checks summary — PR #2536

## Iteration 1

- **Failed checks:** lint
- **Outcome:** fixed
- **Reproduced:** yes
- **Reproduce command:** `.claude/skills/dispatch-propagate/scripts/run-lint.sh`
- **Failure excerpt:** `FAIL: net-new raw gh issue/pr porcelain found in committed shell scripts: .claude/skills/dispatch-propagate/scripts/dispatch-route:119: echo "dispatch-route: gh issue view $N (closedByPullRequestsReferences) failed; continuing to route"`
- **Why not caught:** Error message string in the `else` branch of the merged-PR guard contained the literal text `gh issue view`, triggering the gh-rest-porcelain lint rule that scans for that token in net-new added lines. The rule has no way to distinguish a `gh` call from a string mentioning `gh` by name.
- **Fix:** Rephrased the error message to `closedByPullRequestsReferences query for #$N failed` — describes the failing field name without naming the tool. Commit 51a5fa3c.

natb1:
# QA Summary — #2511 / PR #2536

## Items Executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Guard exists in `dispatch-route` beside CLOSED guard, before phase derivation | shell-command | PASS |
| 2 | Guard routes to self-close disposition and tolerates `gh` failure | shell-command | PASS |
| 3 | Launcher arm closes issue mechanically with no Claude session | shell-command | PASS |
| 4 | New tests present and full dispatch-scripts suite passes | shell-command | PASS |
| 5 | Guard's `gh` usage observes sandbox `dangerouslyDisableSandbox` rule | — | Assessed by Opus (already satisfied) |
| 6 | End-to-end loop break on a real #2481-class occurrence | — | needs-main (planned-deferral) |

No browser walkthrough (pure-backend PR — dispatch scripts only).

## Bugs Found

None.

## Disposition Triage

Step 3.5 ran (residue: 2 items).

| 6 | End-to-end loop break on a real #2481-class occurrence | needs-main (planned-deferral) | n/a | Issue explicitly documents this as non-assertable at merge time: not reproducible in a unit harness; verified by the next real-world occurrence self-resolving, observable in dispatch tick logs. Planned-deferral items file as `blocked_by` follow-ups for downstream monitoring/audit. |

## Filed Follow-ups

| Item | Follow-up | Route |
|------|-----------|-------|
| 6 — End-to-end loop break on a real #2481-class occurrence | #2544 (newly filed) | human (`main-qa` + `dispatch:office-hours` — verification needs production tick logs behind auth) |

| # | Title | Rationale |
|---|-------|-----------|
| 5 | Guard's `gh` usage observes sandbox `dangerouslyDisableSandbox` rule | Guard calls go through `gh_retry`, the existing sanctioned fetch wrapper used throughout `dispatch-route` — sandbox compliance is inherited from the established pattern. The sandbox rule targets Claude bash-tool invocations, not the dispatch runtime scripts. Unit suite stubs `gh` and passes. No code defect. |

## Terminal Disposition

Clean autonomous pass. Escalation set is empty:
- Item 5 dropped as PASS (already-satisfied by Opus)
- Item 6 filed as `blocked_by` follow-up #2544 (needs-main, dropped from escalation set)

Applying `dispatch:qa-done`.

natb1:
## Review — PR #2536 (issue #2511)

No code changes applied. All findings resolved inline, dismissed, or noted as informational.

### Informational

- **`dispatch-route:119`** — `closedByPullRequestsReferences` is GraphQL-only and unpaginated; the first 30 results are returned by default. If an issue had 30+ closing PRs and the merged one was beyond position 30, the guard would miss it. In practice, issues with 30+ closing PRs are not realistic, and the `first // empty` jq filter handles the common case. No action required; known limitation.

- **`dispatch-route:102`** — `closedByPullRequestsReferences` cannot be paginated via `gh_retry`; for extremely large result sets this could silently return partial results. The `first // empty` filter handles the expected case. No action required.

- **`test-dispatch-scripts.sh:3279`** (test 1g) — The source-grep assertion confirms `dispatch-close-resolved` is referenced in the launcher, but does not verify it is called from the `STOP merged-pr-closed` arm specifically. A move or rename would silently pass. This is documented in the test comment; full end-to-end launch-worker testing is out of scope for this harness. No action required.

- **`dispatch-launch-worker:202`** — `env -u CLAUDE_JOB_DIR` is used intentionally so the resolved-closed sentinel write is a no-op (the launcher is not a Claude session). This is documented in the comment. No action required.

### Dismissed

- **`dispatch-route:101-129`** (sec-4) — The guard only fires for issues with `closedByPullRequestsReferences` (PRs with a proper `Closes #N` keyword). PRs manually closed without a closing keyword are out of scope — this is intentional; the guard targets GitHub auto-close failures, not arbitrary PR closures.

- **`dispatch-route:102`** (code-3) — The comment notation `(bounded by sustained-N)` is non-obvious to unfamiliar readers but is a well-established internal term used consistently throughout the codebase. No change.

### False-positive (security)

- **`dispatch-route:119-129`** (sec-1) — `MERGED_CLOSING_PR` sourced from jq with `-r` flag and `first // empty` produces a bare integer or empty string; integer output cannot contain newlines or shell metacharacters. No injection risk.

---

_No deferred follow-ups. No security follow-ups. No code changes were applied this review pass._

PR #2548, opened 2026-06-28
Fix flaky build: soft-warn on network fetch failures in blog feed-fetch plugin

natb1:
Closes #2535

## Problem

The `build(fellspiral)` CI step intermittently fails because fellspiral's build
imports the blog blog-roll plugin, which fetches each configured feed URL at
build time. When a feed URL is temporarily unreachable (DNS failure, connection
refused, transient network error), undici in Node.js 22+ throws
`TypeError: fetch failed`. The plugin's catch block re-threw **any** `TypeError`
as a fatal build error, so a transient network blip failed the build —
non-deterministic, infrastructure-driven flakiness, not a real defect.

## Fix

In the `buildStart` catch block of `blog/src/blog-roll/vite-plugin-feed-fetch.ts`,
discriminate the network case from the bad-URL case:

- A network `TypeError` (`err.message === "fetch failed"`, undici's stable
  message for network-level failures) now soft-warns and skips the feed, letting
  the build continue.
- A bad-URL `TypeError` (e.g. `Failed to parse URL from <url>`) still re-throws
  as a fatal `invalid fetch configuration` build error — genuine
  misconfiguration should still fail the build.

The network branch is an **early `return` placed before** both the bad-URL
re-throw and the existing `classifyError` check. This ordering is load-bearing:
`@commons-systems/errorutil`'s `classifyError` maps every `TypeError` to
`"programmer"`, so without the early return a network `TypeError` would fall
through and be re-thrown anyway. The fix is deliberately not collapsed into a
single inverted condition.

## Tests

Adds a `describe("feedFetchPlugin buildStart error handling", …)` block to
`blog/test/blog-roll/vite-plugin-feed-fetch.test.ts` covering both directions of
the discrimination, using the established plugin-hook test pattern (stubbed
global `fetch`, `buildStart()`/`load()` invocation, virtual-module parsing):

1. Network `TypeError: fetch failed` → `buildStart()` resolves, the feed maps to
   `null`, `console.warn` fires. This is the deterministic, CI-safe proxy for the
   flaky failure.
2. Bad-URL `TypeError` → `buildStart()` rejects with `invalid fetch configuration`.

## Verification

- `npx vitest run --project blog --root .` — 425/425 pass (29 files), including
  the 2 new cases.
- Blog typecheck via the CI-authoritative `run-typecheck.sh --app blog` — green
  (the change is typecheck-clean; no new type errors in the two changed files).

The end-to-end acceptance criterion (the `build(fellspiral)` step completing on a
transient network blip) is inherently non-deterministic to reproduce locally; the
stubbed unit test guards the discrimination logic, which is the testable part.

### DIFF
--- a/blog/src/blog-roll/vite-plugin-feed-fetch.ts
+++ b/blog/src/blog-roll/vite-plugin-feed-fetch.ts
@@ -34,6 +34,20 @@ export function feedFetchPlugin(feeds: FeedConfig[]): Plugin {
             }
             return [id, post];
           } catch (err) {
+            // undici (Node 22+) throws `TypeError: fetch failed` for network-level
+            // failures (DNS, connection refused, temporary unreachability). These are
+            // transient infrastructure errors, not a build misconfiguration — soft-warn
+            // and skip the feed rather than fail the build. This early return MUST
+            // precede the bad-URL re-throw AND the classifyError check below: classifyError
+            // maps every TypeError to "programmer", so without this return a network
+            // TypeError would be re-thrown there and fail the build. Do not collapse this
+            // into a single inverted condition.
+            if (err instanceof TypeError && err.message === "fetch failed") {
+              console.warn(`[feed-fetch] ${id}: fetch error`, err);
+              return [id, null];
+            }
+            // Other TypeErrors (e.g. `Failed to parse URL from <url>`) indicate a
+            // genuinely invalid feed configuration — re-throw as a fatal build error.
             if (err instanceof TypeError) {
               throw new Error(`[feed-fetch] ${id}: invalid fetch configuration`, { cause: err });
             }
--- a/blog/test/blog-roll/vite-plugin-feed-fetch.test.ts
+++ b/blog/test/blog-roll/vite-plugin-feed-fetch.test.ts
@@ -199,11 +199,11 @@ describe("feedFetchPlugin buildStart error handling", () => {
       { id: "test", url: "https://example.com/feed.xml" },
     ]);
 
-    await expect((plugin as any).buildStart()).resolves.toBeUndefined();
+    await expect((plugin as any).buildStart()).resolves.toBeUndefined(); // type-safety-ok: invoking Vite plugin hook directly in unit test
 
     const feedData = parseVirtualModule(
-      (plugin as any).load(RESOLVED_VIRTUAL_MODULE_ID),
-    ) as Record<string, unknown>;
+      (plugin as any).load(RESOLVED_VIRTUAL_MODULE_ID), // type-safety-ok: invoking Vite plugin hook directly in unit test
+    ) as Record<string, unknown>; // type-safety-ok: narrowing parseVirtualModule's unknown JSON
     expect(feedData["test"]).toBeNull();
     expect(warn).toHaveBeenCalled();
   });
@@ -217,7 +217,7 @@ describe("feedFetchPlugin buildStart error handling", () => {
       { id: "test", url: "https://example.com/feed.xml" },
     ]);
 
-    await expect((plugin as any).buildStart()).rejects.toThrow(
+    await expect((plugin as any).buildStart()).rejects.toThrow( // type-safety-ok: invoking Vite plugin hook directly in unit test
       "invalid fetch configuration",
     );
   });
--- a/blog/test/blog-roll/vite-plugin-feed-fetch.test.ts
+++ b/blog/test/blog-roll/vite-plugin-feed-fetch.test.ts
@@ -1,8 +1,13 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi, afterEach } from "vitest";
 import {
+  feedFetchPlugin,
   parseAtomFeedXml,
   parseRssFeedXml,
 } from "../../src/blog-roll/vite-plugin-feed-fetch";
+
+function parseVirtualModule(moduleCode: string): unknown {
+  return JSON.parse(moduleCode.replace("export default ", "").replace(/;$/, ""));
+}
 
 describe("parseAtomFeedXml", () => {
   it("extracts title, URL, and publishedAt from an Atom entry", () => {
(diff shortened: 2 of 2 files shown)


natb1:
# Fix-checks summary — PR #2548

## Iteration 1

- **Failed checks**: type-safety-sensor (x2 — appeared in both CI runs for this PR)
- **Outcome**: fixed
- **Reproduced**: yes
- **Reproduce command**: `bash .github/scripts/check-type-safety-escapes.sh`
- **Failure excerpt**:
  ```
  ::error file=blog/test/blog-roll/vite-plugin-feed-fetch.test.ts,line=202::Net-new type-safety escape hatch (any in type position)
  ::error file=blog/test/blog-roll/vite-plugin-feed-fetch.test.ts,line=205::Net-new type-safety escape hatch (any in type position)
  ::error file=blog/test/blog-roll/vite-plugin-feed-fetch.test.ts,line=206::Net-new type-safety escape hatch (as <Type> cast)
  ::error file=blog/test/blog-roll/vite-plugin-feed-fetch.test.ts,line=220::Net-new type-safety escape hatch (any in type position)
  ```
- **Why not caught**: The new `feedFetchPlugin buildStart error handling` describe block invokes Vite plugin hooks directly in the unit test using `as any` casts (to bypass the complex `BuildStartHook | ObjectHook<BuildStartHook>` union type) and a `as Record<string, unknown>` narrowing cast. These are intentional escape hatches in test code; the sensor's suppression mechanism (`// type-safety-ok: <reason>`) is the correct fix.
- **Fix**: Added `// type-safety-ok:` suppression comments to the 4 flagged lines in `blog/test/blog-roll/vite-plugin-feed-fetch.test.ts`. Commit: `9b4c849b`.

natb1:
# QA Summary — Issue #2535 / PR #2548

## Results

|------|----------------|--------|
| 1. Network `TypeError: fetch failed` soft-warns instead of failing the build | script-verifiable | PASS |
| 2. Bad-URL `TypeError` still re-throws as a fatal build error | script-verifiable | PASS |
| 3. Catch-block discrimination is present in the source | script-verifiable | PASS |
| 4. `build(fellspiral)` completes in CI when a feed URL is temporarily unreachable | needs-human-judgment (planned-deferral) | → filed as follow-up |

No bugs found. No browser walkthrough (pure library change — no frontend app component).

## Disposition Triage

Step 3.5 ran (1 residue item). `needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR.

- **Item 4** (`needs-main`, planned-deferral): acceptance criterion 3 requires a live, non-deterministic network outage against a real feed URL during a fellspiral build — cannot be reproduced deterministically by a local script. Items 1–3 cover the deterministic code path; this criterion is observed across subsequent CI runs. Filed as follow-up #2559.

## Filed Follow-ups

- #2559: `qa: verify against main — qa-needs-main #2535: build-fellspiral-completes-in-ci-when-a-feed-url-is-temporarily-unreachable` — route: **human** (`main-qa` + `dispatch:office-hours`; requires observing CI behavior over time, not an autonomous public-URL check)

natb1:
## Review — PR #2548

Ran code-review, input-validation, and secrets finders against the 2-file diff. 13 findings surfaced, 1 fixed, 1 deferred.

---

### Fixed

- **SR-002** — `blog/test/blog-roll/vite-plugin-feed-fetch.test.ts:209` — Warning assertion now pins both message prefix and TypeError arg  
  The bare `expect(warn).toHaveBeenCalled()` was replaced with `toHaveBeenCalledWith(expect.stringContaining("[feed-fetch] test"), expect.any(TypeError))`, locking in both the plugin-prefixed feed-ID string and that the TypeError is forwarded as the second argument.  
  Fix commit: `bed8f04b`

---

### Informational

- **SR-001** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:45` — Exact message match against undici `'fetch failed'` string may silently break on a future runtime upgrade. The inline comment documents this dependency; if a future Node LTS changes the string, the symptom (build failure on transient network errors) will be observable. No action in this PR.

- **SR-004** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:45-48` — Complexity increase (+2 source branches, +8 test branches) is the minimum required. The early-return must precede the existing `err instanceof TypeError` re-throw because `classifyError` maps every `TypeError` to `"programmer"`. Cannot be collapsed.

- **SR-005** — `blog/test/blog-roll/vite-plugin-feed-fetch.test.ts:220-223` — Re-throw test does not assert the original TypeError is preserved as `cause`. Implementation is clearly correct; a cause-chain assertion could be added as a regression target in a follow-up.

---

### Dismissed

- **EROSION-1** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:96` — Cyclomatic complexity +2 (source file 39→41). One new `if`-branch — the minimum for the discrimination. No simpler formulation exists.

- **EROSION-2** — `blog/test/blog-roll/vite-plugin-feed-fetch.test.ts:56` — Cyclomatic complexity +8 (test file 29→37). Two new async test cases, one per discrimination path. Proportional to the feature under test.

---

### False-positive (security)

- **IV-2** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:46` — Unsanitized `id` in `console.warn`. The `id` field is a hardcoded developer-assigned identifier with no external input path; no actor can inject an adversarial value.

- **SEC-1** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:46` — Full TypeError logged including cause chain. Identical pattern already exists at line 55 (pre-existing). Feed hostnames are public RSS endpoints. The new branch introduces no new disclosure class.

- **SEC-2** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:46` — URL-embedded credentials not exposed. Undici network-failure causes are OS-level errors (ENOTFOUND, ECONNREFUSED) at the DNS/TCP layer; URL query parameters and credentials do not appear in the cause chain.

- **SEC-3** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:46` — Feed `id` in log is a developer-assigned label (e.g., `"bastionland"`), not a credential. Structurally separate from `FeedConfig.url`.

---

### Deferred

- **SR-003** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:46` — Network-error warning omits the feed URL, reducing debuggability. Should update all `console.warn` calls in the catch block consistently to include `url` (e.g., `[feed-fetch] ${id} (${url}): fetch error`). Changing only the new line while leaving lines 27 and 55 would create inconsistency within the function. Filed as follow-up #2562 (blocked by #2535).

---

### Out-of-scope (security)

- **IV-3** — `blog/src/blog-roll/vite-plugin-feed-fetch.ts:23` — Build-time: feed URLs reach `fetch()` without scheme validation. Pre-existing behavior not touched by this diff. A follow-up could add an `https://`-only scheme guard as defense-in-depth.

PR #2558, opened 2026-06-28
Sum gscClicks/gscImpressions from topPages, not topQueries

natb1:
Closes #2488

## Summary

The `gscClicks` and `gscImpressions` headline scalars in the project signal-samples
time series were summed from `gsc.topQueries` instead of `gsc.topPages`. Both are
top-25 subsets of the GSC response, but a single page is reached via multiple
queries, so summing across queries **double-counts** clicks and impressions.

The office-hours `ProjectSignalsPanel` already renders its "total clicks" /
"total impressions" from `gsc.topPages`, so the charted time-series scalar and the
panel aggregate disagreed for the same snapshot and could not be compared over
time. This PR sums both scalars from `topPages` so they match the panel and avoid
double-counting.

This is the page-based-sum path the issue recommends over the field-rename
alternative (`gscQueryClicks` / `gscQueryImpressions`).

## Changes

- `functions/src/project-signals.ts` — the two reducers now walk `gsc.topPages`
  instead of `gsc.topQueries`; the `ProjectSignalSample` type-field comments name
  the source (`// summed from topPages`).
- `functions/test/project-signals.test.ts` — the shared `gsc` fixture now
  populates `topPages` with rows whose totals (12 clicks / 120 impressions) are
  distinct from the `topQueries` totals (15 / 150), so the test actively pins the
  new source; the headline-scalar assertions are updated accordingly.

## Verification

- `npx vitest run --project functions --root .` — 83 tests pass.
- `npx tsc -p functions --noEmit` — type-clean.

### DIFF
--- a/functions/src/project-signals.ts
+++ b/functions/src/project-signals.ts
@@ -219,8 +219,8 @@ interface ProjectSignalSample {
   forks?: number;
   pageViews?: number; // summed across GA4 apps
   sessions?: number; // summed across GA4 apps
-  gscClicks?: number; // summed
-  gscImpressions?: number; // summed
+  gscClicks?: number; // summed from topPages
+  gscImpressions?: number; // summed from topPages
   psiPerformance?: number; // first-URL mobile performance
 }
 
@@ -629,10 +629,10 @@ export async function collectProjectSignalsCore(deps: CollectProjectSignalsDeps)
   const pageViews = ga4 ? ga4.reduce((sum, a) => sum + a.pageViews, 0) : undefined;
   const sessions = ga4 ? ga4.reduce((sum, a) => sum + a.sessions, 0) : undefined;
   const gscClicks = gsc
-    ? gsc.topQueries.reduce((sum, q) => sum + q.clicks, 0)
+    ? gsc.topPages.reduce((sum, p) => sum + p.clicks, 0)
     : undefined;
   const gscImpressions = gsc
-    ? gsc.topQueries.reduce((sum, q) => sum + q.impressions, 0)
+    ? gsc.topPages.reduce((sum, p) => sum + p.impressions, 0)
     : undefined;
   const psiPerformance =
     psi && psi.length > 0 && psi[0].performance !== null ? psi[0].performance : undefined;
--- a/functions/test/project-signals.test.ts
+++ b/functions/test/project-signals.test.ts
@@ -418,7 +418,10 @@ describe("collectProjectSignalsCore", () => {
       { query: "a", clicks: 10, impressions: 100, ctr: 0.1, position: 2 },
       { query: "b", clicks: 5, impressions: 50, ctr: 0.1, position: 3 },
     ],
-    topPages: [],
+    topPages: [
+      { page: "/", clicks: 8, impressions: 80, ctr: 0.1, position: 2 },
+      { page: "/budget", clicks: 4, impressions: 40, ctr: 0.1, position: 3 },
+    ],
     devices: [],
   };
   const psi: PsiUrlSignals[] = [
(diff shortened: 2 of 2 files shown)


natb1:
# QA Summary — Issue #2488 / PR #2558

## Items executed

All 7 items ran in the shell-command lane (pure backend change — no browser component).

| # | Title | Result |
|---|-------|--------|
| 1 | gscClicks summed from topPages in ProjectSignalSample | PASS |
| 2 | gscImpressions summed from topPages in ProjectSignalSample | PASS |
| 3 | ProjectSignalSample type-field comments name the source | PASS |
| 4 | Test fixture pins topPages totals distinct from topQueries | PASS |
| 5 | Field-rename path NOT taken (page-based-sum path confirmed) | PASS |
| 6 | Functions unit suite passes with updated fixtures | PASS |
| 7 | Functions TypeScript type check is clean | PASS |

**Counts:** 7 PASS / 0 FAIL / 0 SKIP

## Results

**Item 1 (PASS):** `gsc.topPages.reduce((sum, p) => sum + p.clicks, 0)` confirmed at line 632. No `topQueries` reducer for clicks remains.

**Item 2 (PASS):** `gsc.topPages.reduce((sum, p) => sum + p.impressions, 0)` confirmed at line 635. No `topQueries` reducer for impressions remains.

**Item 3 (PASS):** Both type-field comments updated: `gscClicks?: number; // summed from topPages` (line 222) and `gscImpressions?: number; // summed from topPages` (line 223). No bare `// summed` comment remains.

**Item 4 (PASS):** Test fixture populates `topPages` with rows totaling 12 clicks / 120 impressions; assertions updated to `toBe(12)` and `toBe(120)` — distinct from the `topQueries` totals of 15/150, actively pinning the new source.

**Item 5 (PASS):** Field names `gscClicks` and `gscImpressions` are unchanged. No `gscPageClicks`, `gscPagesClicks`, or similar rename was introduced.

**Item 6 (PASS):** `npx vitest run --project functions --root .` — 83 tests pass (4 test files).

**Item 7 (PASS):** `npx tsc -p functions --noEmit` — type-clean.

## Bugs found

None.

## Deferred to office-hours

None.

## Disposition triage

Step 3.5 was skipped — residue list was empty (all items passed).

natb1:
## Review — PR #2558 (#2488)

**Surface:** `code` | **Findings:** 5 surfaced, 0 actionable | **Fixes:** none

---

### Dismissed

- **`functions/src/project-signals.ts:631`** (code-review) — Two separate `.reduce()` calls iterate `gsc.topPages` twice, once for `gscClicks` and once for `gscImpressions`. A single-pass accumulator would eliminate the duplicate iteration, but this is a micro-optimization on a top-25 list in a scheduled Cloud Function; the two-pass form is clearer and the overhead is negligible. No change warranted.

---

### Informational

- **`functions/test/project-signals.test.ts:421`** (code-review) — No test covers the case where `fetchGsc` returns a valid `GscSignals` object with an empty `topPages` array. In that scenario `gscClicks` and `gscImpressions` would both be `0` (from `reduce`'s initial value) rather than `undefined`, and the conditional spread on line 647 would write `{ gscClicks: 0 }` to the sample. The current fixture always has rows, so this path is untested. Not a defect introduced by this diff; noting for awareness.

---

### False-positive (security)

- **`functions/src/project-signals.ts:631`** (red-team) — Concern that an empty `topPages` array produces `0` instead of `undefined`, which could mislead downstream consumers. The `reduce(…, 0)` result of `0` is correct for "GSC responded with no page rows" — the conditional spread (`gscClicks !== undefined`) is satisfied, so a `0` gets written. This is expected behavior identical to how a successful `topQueries` reduction of an empty list previously behaved. Not a defect.
- **`functions/src/project-signals.ts:632`** (red-team) — Concern that changing the aggregation source from query-level (`topQueries`) to page-level (`topPages`) alters semantics in a breaking way. This change is the explicit intent of #2488: `topQueries` double-counts clicks/impressions across multiple queries per page; `topPages` gives the true per-page total that matches the `ProjectSignalsPanel` aggregate. The semantic change is the fix, not a regression.

---

### Out-of-scope

- **`functions/src/project-signals.ts:474`** (red-team) — The `mapRow` helper uses `r.clicks ?? 0` / `r.impressions ?? 0`, which guards only `null`/`undefined`, not non-numeric types (strings, `Infinity`, `NaN`). Pre-existing code the diff did not touch; not introduced by this PR.

---

PR #2464, opened 2026-06-23
office-hours: scheduled project-signal collector + dashboard panel

natb1:
Closes #539

## What

Adds a second server-side signal collector to the office-hours app, modeled on
the existing dispatch-queue metrics collector. It gathers the four project
signals — GitHub repo + traffic stats, Google Analytics (GA4), Google Search
Console (GSC), and PageSpeed Insights (PSI) — that the `/align` skill's local
shell scripts gather ad-hoc today, and persists them as durable snapshots over
time plus a new dashboard panel.

This is the signal store the WIP intention graph's per-node `success_signal` /
`reading` / `gap` (#2371) will later read. This PR delivers the store and its
dashboard surface, not the per-node sensor model.

## Units

- **U1 — Collector function.** New `functions/src/project-signals.ts`: a
  scheduled `collectProjectSignals` function (every 6 hours, 300s timeout) with a
  dependency-injected core, per-source live fetchers (GitHub stats/traffic, GA4,
  GSC, PSI) ported from the align scripts, two-level graceful degradation
  (top-level doc-identity guard + per-source config-gate/try-catch), a single
  snapshot doc at `office-hours/{env}/metrics/project-signals`, and a headline
  time-series sample at `office-hours/{env}/signal-samples/{autoId}`. Wire-shape
  interfaces are the source of truth. Mocked unit tests (no live API calls). New
  `signal-samples` Firestore rule mirroring `issue-samples`.
- **U2 — Office-hours data layer.** New `office-hours/src/project-signals.ts`
  (wire-shape-matching parser: strict on identity, lenient per source),
  `project-signals-data.ts` (demo + owner single-doc reads), and the seed + vite
  virtual-module + `.d.ts` trio so the demo tier renders without authentication.
- **U3 — Dashboard panel.** New `ProjectSignalsPanel.tsx` (full-width JSX panel,
  per-source empty states, a link to the dispatch-queue panel instead of
  duplicating those numbers), wired into `Dashboard.tsx` for both demo and owner
  tiers, with panel tests.
- **U4 — Docs.** Deploy prerequisites (GA4/GSC API enablement, OAuth scopes,
  optional PSI key, GitHub App traffic permission, secret/config wiring,
  post-deploy live QA) and the intention-graph (#2371) contract, documented in the
  collector header.

## Credentials vs. CI

GA4/GSC need owner-level GCP setup that CI does not have, PSI optionally needs an
API key, and GitHub traffic needs an elevated token. So **every live API call is
[...]

PR #2598, opened 2026-06-29
office-hours/functions: resolve PR #2464 review nits (computedAt comment + tighten GSC/PSI validators)

natb1:
Closes #2490

Resolves two review nits raised on PR #2464, consolidated into #2490 (#2491
was folded in and closed as superseded). Both touch files that only landed on
main with #2464, so they ship as a single deferred cleanup PR. Neither changes
user-visible runtime behavior.

## Nit 1 — `computedAt` comment correction

In `office-hours/src/vite-plugin-project-signal-seed.ts`, the `buildStart()`
comment claimed `computedAtOffsetMin` "is converted to computedAt at build
time". The generated module actually applies the offset via `Date.now()` at
**module load (page load)**, not build time. Comment-only change; the generated
`moduleCode` and strip logic are untouched.

## Nit 2 — tighten `isValidGscSite` / `isValidPsiUrl`

In `functions/src/project-signals.ts`, the `https://` branch of both validators
now rejects `//` empty path segments and `..` traversal components via two
negative lookaheads (`(?!.*\/\/)(?!.*\.\.)`). Defense-in-depth: the downstream
encoding already makes these inputs safe, but the validators should not accept
them. The `sc-domain:` branch and the existing character class are unchanged.
Added rejection cases for `https://foo.com/../etc` and `https://foo.com//path`
to both validator tests; existing positive cases still pass.

## Verification

- `npx vitest run --project functions --root .` — 84 passed
- `npx vitest run --project office-hours --root .` — 440 passed
- `npx tsc -p functions --noEmit` — clean
- `npx tsc -p office-hours --noEmit` — clean

### DIFF
--- a/functions/src/project-signals.ts
+++ b/functions/src/project-signals.ts
@@ -853,13 +853,15 @@ export function isValidOwnerName(repo: string): boolean {
 // isValidGscSite — accepts `sc-domain:<host>` (no slashes) or an https URL
 // whose path has no `..` traversal component and no `//` empty segment.
 export function isValidGscSite(site: string): boolean {
-  return /^(sc-domain:[A-Za-z0-9.-]+|https:\/\/(?!.*\/\/)(?!.*\.\.)[A-Za-z0-9._/-]+)$/.test(site);
+  return /^(sc-domain:[A-Za-z0-9.-]+|https:\/\/(?!.*\/\/)(?!.*\.\.)[A-Za-z0-9._-]+(?:\/[A-Za-z0-9._-]*)*)$/.test(
+    site,
+  );
 }
 
 // isValidPsiUrl — https URL with no `..` traversal component and no `//`
 // empty path segment.
 export function isValidPsiUrl(url: string): boolean {
-  return /^https:\/\/(?!.*\/\/)(?!.*\.\.)[A-Za-z0-9._/-]+$/.test(url);
+  return /^https:\/\/(?!.*\/\/)(?!.*\.\.)[A-Za-z0-9._-]+(?:\/[A-Za-z0-9._-]*)*$/.test(url);
 }
 
 // Parses "app:propertyId,app:propertyId,..." into validated pairs. An app name
--- a/functions/src/project-signals.ts
+++ b/functions/src/project-signals.ts
@@ -850,12 +850,16 @@ export function isValidOwnerName(repo: string): boolean {
   return slash > 0 && slash !== repo.length - 1 && repo.indexOf("/", slash + 1) === -1;
 }
 
+// isValidGscSite — accepts `sc-domain:<host>` (no slashes) or an https URL
+// whose path has no `..` traversal component and no `//` empty segment.
 export function isValidGscSite(site: string): boolean {
-  return /^(sc-domain:[A-Za-z0-9.-]+|https:\/\/[A-Za-z0-9._/-]+)$/.test(site);
+  return /^(sc-domain:[A-Za-z0-9.-]+|https:\/\/(?!.*\/\/)(?!.*\.\.)[A-Za-z0-9._/\-]+)$/.test(site);
 }
 
+// isValidPsiUrl — https URL with no `..` traversal component and no `//`
+// empty path segment.
 export function isValidPsiUrl(url: string): boolean {
-  return /^https:\/\/[A-Za-z0-9._/-]+$/.test(url);
+  return /^https:\/\/(?!.*\/\/)(?!.*\.\.)[A-Za-z0-9._/\-]+$/.test(url);
 }
 
 // Parses "app:propertyId,app:propertyId,..." into validated pairs. An app name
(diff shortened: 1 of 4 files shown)


natb1:
# Fix-checks summary — PR #2598

## Iteration 1

- **Failed checks**: lint
- **Outcome**: fixed
- **Reproduced**: yes
- **Reproduce command**: `.claude/skills/dispatch-propagate/scripts/run-lint.sh`
- **Failure excerpt**: `functions/src/project-signals.ts 856:81 error Unnecessary escape character: \- no-useless-escape; 862:55 error Unnecessary escape character: \- no-useless-escape`
- **Why not caught**: The nit-2 commit added `\-` in regex character classes at lines 856 and 862 (`isValidGscSite` and `isValidPsiUrl`). A hyphen at the end of a character class `[...]` is unambiguous (cannot be a range separator), so `no-useless-escape` flags it. ESLint runs in CI but was not run locally to validate the nit-2 commit before push.
- **Fix**: Removed the backslash from `\-` in both character classes (`[A-Za-z0-9._/\-]` → `[A-Za-z0-9._/-]`) — commit dc4a4e9a

natb1:
# QA — Issue #2490 (PR #2598)

## Results

All 7 script-verifiable items passed. No browser walkthrough (pure backend/functions/office-hours change). No bugs found.

| # | Item | Result |
|---|------|--------|
| 1 | Comment no longer says "at build time"; says "module load time (page load)" | PASS |
| 2 | `tsc -p office-hours` passes | PASS |
| 3 | No office-hours unit tests broken (440 passed) | PASS |
| 4 | Validator tests: `isValidGscSite` and `isValidPsiUrl` reject `..` / `//`, accept valid inputs, `sc-domain` branch unchanged (85 passed) | PASS |
| 5 | `sc-domain:` branch of `isValidGscSite` is unchanged | PASS |
| 6 | Tightened regexes carry inline comments explaining the accepted format | PASS |
| 7 | `tsc -p functions` passes | PASS |

- **Comment wording reads correctly (human review of Nit 1 text):** The diff unambiguously shows the comment was replaced from "converted to computedAt at build time below" to "applied at module load time (page load) via Date.now() in the generated module below." The new wording is technically accurate: `Date.now()` is called at module-load time (when the browser executes the generated module), not at build time. The acceptance criterion asks only that the comment no longer says "at build time" and instead states module-load-time evaluation — both conditions are provably satisfied by the diff alone. Script check item 1 already verified the old phrase is absent and the new phrase is present. No human input is needed.

natb1:
## Review — PR #2598 (issue #2490)

### Fixed

- **`isValidGscSite` / `isValidPsiUrl` accept empty-host URLs** (`functions/src/project-signals.ts:856`) — Both validators matched `https:///` and `https:///path` because the character class `[A-Za-z0-9._/-]+` starts matching immediately after the literal `https://`, producing a valid match with an empty authority. Replaced the https-branch character class with a host-then-path pattern (`[A-Za-z0-9._-]+(?:/[A-Za-z0-9._-]*)*`) that requires at least one non-slash host character before any path, rejecting empty-authority URLs. (Sources: code-review, input-validation) — fix commit `b5701ede`

- **`ProjectSignalSeed.computedAtOffsetMin` JSDoc out of sync** (`office-hours/src/project-signal-seeds.ts`) — The type's JSDoc still described `computedAt` as derived at build time, contradicting the corrected plugin comment. Updated JSDoc to document that `computedAt` is computed via `Date.now()` at module load time (page load). (Source: code-review) — fix commit `b5701ede`

### Informational

- **`.*` inside negative lookaheads — O(n) per lookahead** (`functions/src/project-signals.ts:856`) — `(?!.*\/\/)` and `(?!.*\.\.)` each scan the full remaining input on every position. For typical URL inputs (dozens of characters) this is negligible, and these validators are not called in hot loops. Noting for awareness; no change required.

### False-positive (security)

- **Negative lookahead coverage** (`functions/src/project-signals.ts:856,862`) — Red-team finding suggested the `(?!.*\/\/)` lookahead might not cover the full string. It does: applied immediately after the `https://` literal, it anchors at that position and asserts the pattern `.*\/\/` is not present anywhere in the remaining input, so no double-slash anywhere in the URL passes. Not an actual gap.

- **`computedAtOffsetMin` string interpolation** (`office-hours/src/vite-plugin-project-signal-seed.ts:23`) — Red-team flagged string interpolation of `${computedAtOffsetMin}` into virtual module source. The value is typed as `number` (a numeric literal in the generated JS), so injection of code via this field is not possible. Not a vulnerability.

### Deferred

- **Shared regex for `isValidGscSite` / `isValidPsiUrl` https branch** (`functions/src/project-signals.ts:862`) — The https-branch pattern is now identical in both validators, with no compile-time enforcement that they stay in sync. Recommended: extract a shared constant or private helper. Filed as follow-up #2602. Out of scope for this PR.

PR #2608, opened 2026-06-29
office-hours: target a specific queue item by passing <N>

natb1:
Closes #2538

## Summary

Gives the office-hours entry script and selector an optional `<N>` argument so a
human who knows which parked item they want can engage it by number, instead of
always getting the queue head and re-running until their item surfaces.

`office-hours <N>` runs the existing per-item disposition machinery against that
one item and acts on the result exactly as the no-arg path acts on the queue
head. `office-hours` with no argument is unchanged (full-queue enumeration, head
disposition). The change is additive and backwards-compatible — no disposition's
behavior changes; only a targeting input is added.

## How it works

The no-arg enumeration already ran a per-item idle → resume → fresh cascade for
the queue head. This PR factors that cascade into three reusable shell helpers
and has both paths call them, so there is exactly one implementation of the
verb-selection rules (acceptance criterion #3, by construction):

- **Unit 1** — extract `try_emit_idle` / `try_emit_resume` / `emit_fresh` in
  `office-hours-select-target` and rewire the no-arg enumeration to call them. A
  pure behavior-preserving refactor; the existing `OHST*`/`OH*` suite stays green
  as the correctness check. `try_emit_resume` preserves the no-arg loop's
  load-bearing tri-state (emit / continue / break); each helper calls
  `emit_blocker_signal` internally so the #2443 open-blocker advisory rides along
  in both paths for free.
- **Unit 2** — add an optional positional `<N>` to the selector. With `<N>` it
  skips cross-item selection and runs the same cascade for that one item via
  `issue_live_session_id` + the Unit 1 helpers; with no arg the enumeration is
  unchanged.
- **Unit 3** — thread the optional `<N>` through the `office-hours` entry script,
  validating it is numeric and forwarding it only when set. The verb dispatch is
  untouched: single-item mode emits the same disposition strings, so the existing
  arms act on them identically.
- **Unit 4** — tests for single-item mode in `test-dispatch-scripts.sh`.

## Single-item collapses non-engageable states to `empty`

A targeted `<N>` that is being worked (`issue_live_session_id` rc 3,
working/busy), under an UNKNOWN daemon (rc 2), or not in the open
`dispatch:office-hours` queue all yield `empty` — the single-item analog of the
no-arg path's skip-and-continue, consistent with the existing UNKNOWN→skip
fail-safe.

## Design note: parked-router is intentionally NOT a single-item verb

The issue's greenfield prose loosely lists `parked-router` among "the same five
verbs," but criterion #3 is the precise contract: **emit `empty` when `<N>` is
not a live office-hours queue item.** `parked-router` is *target-less* — it
carries no issue number and is discovered by scanning `dispatch-*` sessions
rooted under `worktrees/main`, not by any per-item disposition. It is
structurally not "a disposition for issue `<N>`." So:

- `parked-router` remains a verb the selector emits **only in no-arg mode** (the
  queue-empty fallback), unchanged.
- **Single-item mode never reaches the parked-router block** — it falls to
  `empty` whenever `<N>` yields no attachable/resumable/fresh disposition
  (covered by test OHST27, the targeted-working case).

This is a deliberate exclusion reconciling criterion #3 with the issue's loose
five-verb enumeration, not a missed case.

## Verification

- `test-dispatch-scripts.sh` (CI-safe, PATH-shimmed fakes, no network):
  **4119/4119 passed**. New single-item coverage: OHST24 (targeted fresh
  overrides oldest-first), OHST25 (non-queue `<N>` → `empty`), OHST26 (targeted
  idle), OHST27 (targeted working → `empty`, parked-router never reached), OHST28
  (targeted resume), OHST29 (targeted blocker advisory on stderr), and OH14
  (entry script forwards `<N>` and acts on the targeted item).
- `bash -n` clean on both edited scripts.

Manual/QA (needs the live Claude daemon, cannot run in CI): exercising
`office-hours <N>` end-to-end — attaching an idle target and fresh-spawning a
sessionless target.


natb1:
# QA Summary — PR #2608 (issue #2538)

## Results

| Item | Title | Lane | Result |
|------|-------|------|--------|
| 1 | Targeted/non-queue selector dispositions — full regression suite | shell-command | PASS |
| 2 | Both modified scripts parse cleanly (`bash -n`) | shell-command | PASS |
| 3 | End-to-end `office-hours <N>` against the live Claude daemon | needs-human-judgment | planned-deferral → filed #2611 |

**Counts:** 2 PASS, 0 FAIL, 0 SKIP, 1 planned-deferral (needs-main)

## Disposition triage

Assessed by the qa-fix Workflow:

- **Item 3** (`needs-main`, planned-deferral): Classified `needs-main` because `planned_deferral` is true — the end-to-end live-daemon verification is documented as non-assertable at merge time (requires the live Claude daemon, not runnable in CI or automated QA). Filed as a `blocked_by` follow-up.

The escalation set is empty — this is a **clean autonomous pass**.

## Filed follow-ups

- **Item 3** → #2611 (newly filed), route: `human` (requires live daemon + human judgment). Labels: `main-qa`, `dispatch:office-hours`. Blocked_by #2538.

## No bugs found

All script-verifiable items passed:
- `test-dispatch-scripts.sh`: 4119/4119 passed (including OHST24–OHST29 and OH14)
- `bash -n` on both scripts: syntax clean

natb1:
## Review — #2538

### Fixed

- **`office-hours-select-target:452`** — when single-item mode targets an issue not in the office-hours queue, the selector now emits `empty not-in-queue <N>` instead of bare `empty`; the entry script prints a precise "issue #\<N\> is not in the office-hours queue" message rather than the generic queue-empty one (commit `5b6489f3`).

### Informational

- **`office-hours-select-target:346`** — `SORTED` is computed unconditionally before the single-item branch, so it runs even when `TARGET_N` is set and the enumeration is skipped. Benign: it's a single `jq` sort call with no side effects, and precomputing it keeps the single-item and no-arg paths symmetric. No action required.

### False-positive (×2)

- **`office-hours:110`** — raw `$1` echoed to stderr in the invalid-arg error path. Not an injection risk: the output goes only to stderr, is never executed, and the regex `^[0-9]+$` gates all downstream use. Dismissed.
- **`office-hours-select-target:156`** — same pattern, same rationale. Dismissed.

### Security

No attack surface changes — diff touches only shell scripts and a SKILL.md doc; no application source, config, dependency, or Firestore-rules changes. CodeQL: no open alerts. Dependency audit: not applicable (no package.json/lock changes). Erosion scan: no findings.

Issue #2507, opened 2026-06-25
office-hours: remove the redundant usage-history chart from the capacity HISTORY band (keep worker-history)


natb1:
## Motivation

The capacity HISTORY band (`office-hours/src/history-band.ts`) renders two charts: `usage-history-chart.ts` (5-hour % and weekly % over wall-clock time with the pace-curve overlay and reset rules) above `worker-history-chart.ts` (active vs target workers).

The usage-history chart is redundant with the existing PACE panel (`pace-position-panel.ts`), which already presents weekly-%-vs-pace better: a position-in-week phase plot against the W(x) target curve, with prior weeks overlaid and an explicit "N pts ahead/behind pace" readout. The only thing the usage-history chart adds over PACE is the 5-hour-% time series, and that is marginal — the 5-hour window is short and its current value already appears in the capacity snapshot band's cards.

The worker-history chart (active vs target workers over time) is NOT redundant — there is no equivalent elsewhere, and it is the most useful signal in the band for spotting fleet workers falling below target. Keep it.

## Scope (one PR)

- Remove `office-hours/src/usage-history-chart.ts` (`renderUsageHistoryChart`) and its unit test.
- Update `history-band.ts` and `components/HistoryBand.tsx` to render only the worker-history island (drop the usage-history island). Reconsider the band heading/wrapper if a single chart no longer needs a band.
- Remove helpers that only `usage-history-chart.ts` used, but KEEP anything still consumed by the PACE panel — `weekly-pace-curve.ts` and `pace-position.ts` remain in use.
- Leave `worker-history-chart.ts`, the `usage-samples` data source/producer, the PACE panel, and the capacity snapshot band unchanged.

## Acceptance criteria

- [ ] The capacity HISTORY band renders only the worker-history chart (active vs target workers); the usage-history (5h/weekly %) chart is gone.
- [ ] The PACE panel and the capacity snapshot band are visually and behaviorally unchanged.
- [ ] No dead code remains from the removed chart; helpers still used by the PACE panel are retained.
- [ ] Unit tests covering the removed chart are removed; remaining office-hours unit tests pass.

Context: the history charts were added in #1005 / #1009; the PACE panel that supersedes the usage half landed in #1400.

PR #2613, opened 2026-06-29
run-typecheck.sh: fully reconstruct HEAD after baseline swap

natb1:
Closes #2607

## Problem

`run-typecheck.sh --app <ws>` computes a typecheck baseline by swapping a
workspace's files to the `origin/main` version (`git checkout origin/main --
"$ws"`), running `tsc`, then restoring HEAD. The restore was a single-step
`git checkout HEAD -- "$ws"`, which only rewrites files **present in HEAD's
tree**. It cannot delete a file that is in the index/worktree but **absent from
HEAD**.

So on a branch that **deletes** a file still present on `origin/main` (any
deletion-PR before merge), the `origin/main` swap reintroduced that file (staged
`A`, on disk) and the `HEAD` checkout could not remove it — leaving the deleted
file resurrected. Because `dispatch-run-verification` runs a plan's `verify`
blocks in sequence in **one** working tree, a deletion-PR's typecheck block
resurrected the deleted file and a following no-dead-reference grep block then
matched the resurrected file's own symbols and false-failed (first observed on
#2507). CI is unaffected — it runs each typecheck in an isolated per-job
checkout.

## Fix

Replace the single-step restore with a three-step **reset → checkout → clean**
sequence at both restore sites (the inline success-path restore and the cleanup
trap):

- `reset -q HEAD -- "$ws"` unstages the `origin/main`-only addition (making it
  untracked);
- `checkout HEAD -- "$ws"` restores the tracked files HEAD does have;
- `clean -fdq -- "$ws"` removes the now-untracked resurrected file from disk.

The order is required: without the `reset`, `clean` is a no-op (file still
tracked); without the `clean`, the file lingers untracked on disk. `-fd` (no
`-x`) preserves gitignored files; the `$ws` scope plus the pre-flight dirty-tree
guard mean `clean` only removes files the swap itself introduced. The inline
restore stays bare (runs under `set -e`, errors must propagate); the trap keeps
its `2>/dev/null || true` suppression (fires on every exit).

## Test

Adds `Test 6` to `test-run-typecheck.sh`: a deletion-branch scenario where
`origin/main` carries a file (`src/extra.ts`) that HEAD deletes. It asserts a
clean working tree and that the deleted file is absent from disk after the run.
Confirmed during authoring to **fail** against the pre-fix single-step restore
(file left staged `A` and on disk) and pass with the fix — proving it exercises
the bug. The whole suite is 19/19. CI picks it up automatically via
`run-unit-tests.sh`.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/run-typecheck.sh
+++ b/.claude/skills/dispatch-propagate/scripts/run-typecheck.sh
@@ -77,7 +77,9 @@ cleanup() {
   local ws
   for ws in "${TOUCHED_WORKSPACES[@]:-}"; do
     [ -z "$ws" ] && continue
+    git -C "$REPO_ROOT" reset -q HEAD -- "$ws" 2>/dev/null || true
     git -C "$REPO_ROOT" checkout HEAD -- "$ws" 2>/dev/null || true
+    git -C "$REPO_ROOT" clean -fdq -- "$ws" 2>/dev/null || true
   done
   exit $rc
 }
@@ -114,7 +116,17 @@ for ws in "${APP_DIRS[@]}"; do
     baseline_ok=true
     (cd "$REPO_ROOT" && npx tsc --noEmit --project "$ws") >/dev/null 2>&1 || baseline_ok=false
     # Restore HEAD version immediately; don't wait for the trap.
+    # Three steps, in order: after `checkout origin/main -- "$ws"` an
+    # origin/main-only file (one HEAD deletes) is staged `A` — in the index and
+    # on disk. `reset HEAD` unstages it (making it untracked); only then can
+    # `clean -fdq` remove it from disk. `checkout HEAD` restores the tracked
+    # files HEAD does have. Order reset->checkout->clean is required: without the
+    # reset, clean is a no-op (file still tracked); without the clean, the file
+    # lingers untracked on disk. `-fd` (no `-x`) preserves gitignored files; the
+    # `$ws` scope + the line-61 dirty guard mean clean only removes swap-introduced files.
+    git -C "$REPO_ROOT" reset -q HEAD -- "$ws"
     git -C "$REPO_ROOT" checkout HEAD -- "$ws"
+    git -C "$REPO_ROOT" clean -fdq -- "$ws"
 
     if [ "$baseline_ok" = false ]; then
       echo "$ws: skipping — origin/main has pre-existing typecheck errors"
--- a/.claude/skills/dispatch-propagate/scripts/test-run-typecheck.sh
+++ b/.claude/skills/dispatch-propagate/scripts/test-run-typecheck.sh
@@ -187,4 +187,50 @@ make_shims "$REPO"
 run_sut
 assert_eq "cleanup/fail: working tree clean" "" "$(git -C "$REPO" status --porcelain)"
 
+# --- Test 6: deletion branch — origin/main has a file HEAD deletes (#2607) ---
+# The baseline swap `git checkout origin/main -- <ws>` reintroduces the
+# origin/main-only file (staged A, on disk). A single-step `git checkout HEAD
+# -- <ws>` restore cannot delete it (HEAD has no such path), leaving the file
+# resurrected. The fix restores via reset->checkout->clean. This case needs
+# inline setup because make_repo builds only src/index.ts.
+echo "Test 6: deletion branch -> deleted file not resurrected after baseline swap"
+REPO=$(mktemp -d "$TMP_ROOT/repo.XXXXXX")
+BARE=$(mktemp -d "$TMP_ROOT/bare.XXXXXX")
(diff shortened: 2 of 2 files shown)


natb1:
## QA Summary — Issue #2607 / PR #2613

### Items executed

| 1 | Full test suite passes including deletion-branch regression (AC#1, AC#4) | script-verifiable | PASS |
| 2 | Restore logic applied at both restore sites — inline success path and cleanup trap (AC#2) | script-verifiable | PASS |
| 3 | No baseline-filtering regression on non-deletion branches (AC#3) | script-verifiable | PASS |

### Counts

- PASS: 3 / FAIL: 0 / SKIP: 0

### Test suite output

```
Test 1: clean baseline + clean HEAD -> exits 0 (3/3 PASS)
Test 2: clean baseline + dirty HEAD -> regression detected (4/4 PASS)
Test 3: dirty baseline + dirty HEAD -> skipped (4/4 PASS)
Test 4: dirty working tree -> bail before mutation (3/3 PASS)
Test 5: cleanup invariant (2/2 PASS)
Test 6: deletion branch -> deleted file not resurrected after baseline swap (3/3 PASS)
================================
Results: 19/19 passed, 0 failed
================================
```

### AC#2 verification

Both restore sites confirmed in `run-typecheck.sh`:
- **Cleanup trap (L80-82):** `git -C "$REPO_ROOT" reset -q HEAD`, `checkout HEAD`, `clean -fdq` (with `2>/dev/null || true`)
- **Inline success path (L127-129):** `git -C "$REPO_ROOT" reset -q HEAD`, `checkout HEAD`, `clean -fdq` (bare — errors propagate under `set -e`)

### Disposition

All acceptance criteria verified. Clean pass — no bugs found, no residue.

natb1:
## Review — PR #2613 (`run-typecheck.sh` baseline-swap deletion fix)

### Fixed

- **`test-run-typecheck.sh:231` — Test 6 vacuous-pass gap** (`code-review`): Test 6 had no assertion that the SUT actually entered the typecheck loop for `ws`. If auto-detection returned no apps, the test would pass vacuously with no coverage. Added `assert_contains "deletion: ws typecheck ran" "=== Typecheck: ws ===" "$OUT"` to rule this out, mirroring the discriminating output checks in Tests 1–5. _(Fix authored by Workflow Opus agent; commit blocked by auto-mode classifier — skill-file edits require human-present session authorization. Fix remains uncommitted in the worktree; will need a follow-up push.)_

- **`run-typecheck.sh:82,129` — `git clean -fdq` on unvalidated `$ws`** (`input-validation`): Finding claimed `--app .` or path-traversal could cause `git clean` to wipe the repo root. Adversarial verification refuted: for `$ws=.`, the tsconfig guard at line 107 (`[ \! -f "$REPO_ROOT/./tsconfig.json" ]`) resolves to the repo root's `tsconfig.json` — which does not exist in this repo. The guard fires `continue` at line 109, exiting the loop body before any git mutating command on lines 113–129 is reached. `TOUCHED_WORKSPACES` never gets `.` appended, so the cleanup trap is also a no-op. The premise is factually wrong; no fix applied.

### Informational

- **`run-typecheck.sh:80` — Three-step restore duplicated in two places** (`code-review`): The `reset -q HEAD / checkout HEAD / clean -fdq` sequence appears verbatim in both the cleanup trap and the inline restore. Not worth factoring out in a shell script of this size, but noted for awareness.
- **`run-typecheck.sh:122` — Comment slightly overstates the ordering constraint** (`code-review`): Comment says "Order reset→checkout→clean is required" — strictly, only `reset`-before-`clean` is required (reset makes the file untracked; clean removes it). `checkout` order relative to `clean` is flexible. Minor comment precision gap; no change needed.
- **`test-run-typecheck.sh:219` — No positive assertion that `index.ts` survives the run** (`code-review`): Test 6 asserts `extra.ts` is absent but does not assert that the tracked file `index.ts` is still present. The clean-only-swap-files guarantee (`-fd` no `-x`, `$ws`-scoped) is sufficient coverage, but an explicit positive assertion would be belt-and-suspenders.

### False-positive (security)

- **`run-typecheck.sh:61–129` — Dirty-workspace guard scope** (`red-team`): Claimed the pre-flight dirty guard running once before the loop leaves per-workspace gaps. False positive: the guard runs `git status --porcelain -- "$ws"` across all workspaces in the loop argument before any mutation, and the `--app` path validates the single workspace before processing. No gap exists.
- **`run-typecheck.sh:114–115` — Cleanup trap re-runs clean on already-cleaned workspaces** (`red-team`): Claimed the trap could over-clean if a workspace was already restored inline. False positive: `git clean -fdq` on a clean working tree is a no-op; no files are removed if the inline restore already ran. The idempotence is intentional.

---

_7 findings surfaced · 1 actionable (Fixed, uncommitted) · 0 security follow-ups · 0 deferred_

Issue #127, opened 2026-03-03
Google analytics


natb1:

natb1:
# Security Review - Complete ✓

**Date**: 2026-03-11
**Branch**: 127-ga-per-app-streams
**Reviewer**: Claude Code (via /security-review skill)
**Outcome**: Pass — no findings

## Iterations

- Iteration 1: No findings — all 39 changed files examined across 7 security categories (input validation, XSS, secrets management, shell injection, CI workflow, error handling, data exposure). Zero vulnerabilities above >80% confidence threshold.

## User Classification Decisions

No findings to classify — review produced zero results above confidence threshold.

## Conclusion

Security review completed with no findings. All 39 changed files examined across 7 security categories. No vulnerabilities meet the >80% exploitability confidence threshold. The analytics integration follows secure patterns: measurement IDs are public by design, no user input reaches dangerous sinks, and error handling is defensive without suppressing failures.

PR #2625, opened 2026-06-29
project-signals: split the single GA4 property per-app via hostName

natb1:
Closes #2618

## What

`collectProjectSignals` reported GA4 data from a list of `app:propertyId` pairs —
one property-scoped `runReport` per pair, one `Ga4AppSignals` row per pair. But the
real GA4 setup (#127) is a **single** property (525024219) with one web data stream
per app subdomain. Because `runReport` is property-scoped (no stream-scoped report
endpoint exists), the collector summed **all** subdomains into one row and
mislabeled it — the live office-hours ANALYTICS (GA4) panel showed a single
`commons` entry that was actually the cross-subdomain total.

This splits the single property's report **within** the report, using the GA4 Data
API `hostName` dimension, and emits one `Ga4AppSignals` per configured host.

## Changes

- **Config vars.** Replaced `PROJECT_SIGNALS_GA4_PROPERTY_IDS` (`app:propertyId,...`)
  with `PROJECT_SIGNALS_GA4_PROPERTY_ID` (a single numeric property id) and
  `PROJECT_SIGNALS_GA4_HOST_APPS` (a `host:app,host:app,...` map). The deploy
  workflow (`functions-deploy.yml`) writes both new vars and drops the old one — the
  owner must add the two new repo Secrets during post-merge provisioning.
- **`parseGa4Pairs` → `parseGa4HostApps`.** Parses the `host:app` map (note the
  order flip vs the old `app:propertyId`), validating host against `^[A-Za-z0-9.-]+$`
  and app against `^[A-Za-z0-9_-]+$`, dropping malformed entries.
- **`fetchGa4Live` reshaped** from per-property-per-app to per-property-multi-host.
  Adds `hostName` as the first dimension on the overview, referral, and landing
  reports. Overview metrics are taken **per host** directly (never averaged). The
  referral/landing queries drop the small per-report `limit` (which would otherwise
  cap *total* rows across all hosts) and keep the global `sessions`-desc `orderBys`,
  so grouping rows by `hostName` and taking `slice(0, 10)` yields each host's true
  top-10. Emits one entry per configured host (in config order), with `0`/`[]`
  defaults for hosts with no data; hosts absent from the config map are dropped.
- **Wrapper gate** now requires a valid numeric property id and ≥1 parsed host-app
  entry, collapsing the old `Promise.all(pairs.map(...))` into a single
  `fetchGa4Live(...)` call.
- **Tests updated** (not weakened — the API changed): a multi-host `fetchGa4Live`
  stub asserts one partitioned entry per configured host (including a configured
  host with no data exercising the default path), and a `parseGa4HostApps` test
  covering valid/dotted/malformed entries.

## Scope notes

The wire shape of `Ga4AppSignals` and all Firestore paths are unchanged — only the
values change from per-property-summed-mislabeled to per-host-correct. The
read-side (`office-hours/src/project-signals.ts`) and the panel
(`ProjectSignalsPanel.tsx`) re-declare the identical read shape and need no change.

## Migration

The GA4 config is owner-provisioned and dormant by default — an empty/old config
simply fails the gate and GA4 collection skips (no crash). Between merge and the
owner re-setting `PROJECT_SIGNALS_GA4_PROPERTY_ID` + `PROJECT_SIGNALS_GA4_HOST_APPS`
(SETUP STEPS step 6), GA4 collection goes dormant. A `main-qa` follow-up will verify
the panel renders one section per app subdomain against deployed main.

## Verification

- `npx vitest run --project functions --root .` — 85 tests pass
- `run-typecheck.sh --app functions` — passes


natb1:
# QA Summary — Issue #2618 / PR #2625

## Results

**10 items executed — 10 PASS, 0 FAIL, 0 SKIP**
**1 item deferred as needs-main follow-up (planned-deferral)**

No browser walkthrough (pure backend/functions change — no browser component detected).

### Items

| # | Title | Result |
|---|-------|--------|
| 1 | Unit test suite passes with multi-host stub asserting per-host partitioned metrics | PASS |
| 2 | `hostName` dimension is requested in all three GA4 runReport calls | PASS (3 occurrences) |
| 3 | Per-host grouping/filtering present — `fetchGa4Live` emits one `Ga4AppSignals` per host | PASS |
| 4 | No hardcoded subdomains inside `fetchGa4Live` | PASS |
| 5 | `parseGa4Pairs` fully removed; `parseGa4HostApps` is present and replaces it | PASS |
| 6 | `parseGa4HostApps` unit test covers valid host:app pairs kept and malformed pairs dropped | PASS |
| 7 | Wrapper gate validates single propertyId and at least one host-app entry | PASS |
| 8 | SETUP STEPS comment and config-var documentation reflect single-property + host-map shape | PASS |
| 9 | `functions-deploy.yml` env var names updated to match new config shape | PASS |
| 10 | TypeScript compilation succeeds with new `fetchGa4Live` signature and `parseGa4HostApps` types | PASS |
| 11 | Office-hours dashboard ANALYTICS (GA4) panel renders one section per app subdomain | needs-main (planned-deferral) |

## Bugs found

None.

## Deferred to office-hours

None — item 11 classified `needs-main` (not `needs-human`) and was filed as a `blocked_by` follow-up (see below).

## Disposition triage

**1 residue item** classified by the disposition Workflow:

- **Item 11** (`needs-human-judgment`, `needs-main`): planned-deferral. The GA4 config (single propertyId + host-app map) is owner-provisioned and dormant by default; the panel criterion is non-assertable at merge time — assertable only against deployed main after the owner re-sets `PROJECT_SIGNALS_GA4_PROPERTY_ID` + `PROJECT_SIGNALS_GA4_HOST_APPS`. Verify: `n/a` (planned-deferral items are excluded from the skeptic candidate set).

No opus-fixable items — no fixes this pass. Escalation set is empty → clean pass.

## Filed follow-ups

- **#2627** — "qa: verify against main — qa-needs-main #2618: office-hours dashboard ANALYTICS panel (production)" (newly filed, **human** route: `main-qa` + `dispatch:office-hours`; requires auth-gated office-hours dashboard access after owner re-provisions GA4 config).

natb1:
## Review findings — PR #2625 (`2618-project-signals-split-the-s`)

### Fixed (commit b69c8038)

- **`functions/src/project-signals.ts:411-478`** — Parallelize the three independent GA4 `runReport` calls (overview, referral, landing) via `Promise.all` instead of awaiting them sequentially. Mirrors the pattern already used for GitHub traffic endpoints at line 307.
- **`functions/test/project-signals.test.ts:173-265`** — Added a non-OK (403) runReport error-path test to the `fetchGa4Live` suite asserting `rejects.toThrow(/403/)`.

### Informational

- **`functions/src/project-signals.ts:429-434`** — Referral-sources `runReport` omits `limit` without capturing `rowCount`. The grouping-then-slice approach is correct, but for a large property (many hosts, many sources), the 30-day window could return a very large result set. No action required now; monitor in production.
- **`functions/src/project-signals.ts:447-452`** — Same note for landing-pages `runReport`: no explicit limit and no `rowCount` capture. Functionally correct for current traffic volumes.
- **`functions/src/project-signals.ts:935`** — Host validation regex `[A-Za-z0-9.-]+` admits technically-invalid forms (leading/trailing dots, consecutive dots). All real hostnames from the config are owner-controlled, so this is low risk. A stricter regex could be filed as a future improvement if needed.
- **`functions/src/project-signals.ts:379`** (erosion) — Cyclomatic complexity of the file rose from 32→34 (worst function) and 157→177 (total). The increase is inherent to the hostName-partitioning logic added by this PR. No immediate action required; worth watching if the file continues to grow.
- **`functions/test/project-signals.test.ts:116`** (erosion) — Test file complexity rose from 94→95. Marginal, expected from the expanded multi-host test case.

### Dismissed

- **`functions/src/project-signals.ts:436-444`** — The referral and landing accumulation loops are structurally similar (`get-or-default-empty-array → push → set`). Extracting a helper would add indirection with no correctness benefit at this size.

### False-positive (security)

- **`functions/src/project-signals.ts:429-434`** (input-validation, auth, data-exposure) — Limit removal flagged as unbounded data ingestion. The code groups by host and `slice(0, 10)`s per host before writing to Firestore, so the stored data is always bounded regardless of API row count.
- **`functions/src/project-signals.ts:385`** (input-validation) — `propertyId` interpolated into URL. The property ID is caller-validated to `^\d+$` before `fetchGa4Live` is invoked; no injection path.
- **`functions/src/project-signals.ts:440,458`** (input-validation, red-team) — GA4 `sessionSource`/`landingPage` dimension values written to Firestore without sanitization. These are analytics metadata stored in the project signals collection, not rendered as HTML. The threat model (Firestore storage of analytics dimension strings) does not constitute an injection or XSS surface here.
- **`functions/src/project-signals.ts:397-399`** (data-exposure) — Error message embeds the property ID. Property IDs are non-secret numeric identifiers (public in GA4 UI); no exposure concern.

PR #2633, opened 2026-06-29
print: load cloud image archives whole-file (cache-first) to fix flaky offline @cache test

natb1:
Closes #2631

## Why

`print/e2e/viewer.spec.ts:483` (`viewer image archive loads from cache on second view @cache`) flakes in CI. The issue's stated hypothesis — a debounced position-save IndexedDB flush race — is disproven: for an anonymous user the position store is synchronous `localStorage`, and PR #1283's teardown `flushSave` already makes the position half reliable. Position is not the flake.

The real flake is the **image-archive offline load path**. Cloud image-archive was the one media type that did **not** route through whole-file caching. `view.ts` handed the renderer a bare URL string, which drove a range path (`CachedRangeReader` → `unzip(reader)`). `unzip` first calls `getLength()`, an **uncached** network `HEAD`. On the blocked offline second visit that HEAD aborts; the fallback then misses cache (the range path only ever wrote per-chunk entries via `putChunk`, never `putFile`), so the follow-up `fetch` aborts too and the viewer never renders — the test times out. It was intermittent because the blocked request was sometimes served from the browser HTTP cache, and the fire-and-forget chunk writes were not reliably durable before navigate-away.

## What

Route cloud image-archive through `resolveFileSource`, exactly like PDF and EPUB, and remove the now-dead range machinery.

- **Unit 1 — deflake.** `view.ts` image-archive case: `resolveSource: () => Promise.resolve(url)` → `() => resolveFileSource(url, spath)`. The offline second visit is now `getFile` → `unzip(buffer)`: both in-memory, zero network, no `getLength` HEAD and no per-chunk durability race. Flipped the `view.test.ts` canary to assert cache routing.
- **Unit 2 — strip dead machinery.** Removed `CachedRangeReader`, `fetchArchiveBuffer`, the string/range `init` branch, and the `storagePath` param from `createImageArchiveRenderer`. `init` now unconditionally `unzip(source)`. The renderer already had a whole-buffer branch, so this adds no new capability. Converted/deleted the tests for the removed paths.
- **Unit 3 — cleanup.** Removed the now-orphaned `getChunk` / `putChunk` / `chunkKey` from `media-cache.ts` (no remaining callers in `print/`) and their dedicated tests; kept the live whole-file `getFile` / `putFile` path and eviction coverage.

## Tradeoff

Whole-file load means a large `.cbz` fully downloads before first paint, losing the range reader's lazy-loading — matching existing PDF/EPUB behavior. The renderer already eagerly prefetches the next page, so the lazy-load benefit was largely unrealized, and offline reliability is the property the `@cache` test encodes. Per-chunk IndexedDB entries written by old clients become orphaned but harmless (nothing reads them; the LRU cache evicts on its own) — no migration needed.

## Verification

- `npx vitest run --project print --root .` — 578 passed, 1 skipped.
- `run-typecheck.sh --app print` — passed (proves the `createImageArchiveRenderer` signature cascade and no dangling references to the removed symbols).

The Playwright acceptance test cannot run locally (browser-version mismatch; CI is authoritative), and one green CI run cannot prove de-flaking. **Acceptance criterion 3 (no recurrence)** should be tracked as an observation over subsequent CI runs of `print/e2e/viewer.spec.ts:483`, not a one-shot check.

No Playwright retries added, no test weakened or skipped — the deleted tests cover genuinely-removed code paths.

### DIFF
--- a/print/src/media-cache.ts
+++ b/print/src/media-cache.ts
@@ -1,7 +1,7 @@
 /**
- * IndexedDB-backed LRU cache for print media files (PDFs, EPUBs, image archive
- * chunks). Thin wrapper around the shared `@commons-systems/idbutil/lru-blob-cache`
- * primitive. Whole files and range-request chunks share the same stores.
+ * IndexedDB-backed LRU cache for whole print media files (PDFs, EPUBs, image
+ * archives). Thin wrapper around the shared
+ * `@commons-systems/idbutil/lru-blob-cache` primitive.
  */
 import { createLruBlobCache } from "@commons-systems/idbutil/lru-blob-cache";
 
@@ -24,31 +24,10 @@ export const blobCache = cache;
 export const closeDb = cache.closeDb;
 export const clearCache = cache.clearCache;
 
-function chunkKey(storagePath: string, offset: number, length: number): string {
-  return `${storagePath}:${offset}:${length}`;
-}
-
 export function getFile(storagePath: string): Promise<ArrayBuffer | null> {
   return cache.getEntry<ArrayBuffer>(storagePath);
 }
 
 export function putFile(storagePath: string, data: ArrayBuffer): Promise<void> {
   return cache.putEntry(storagePath, data);
-}
-
-export function getChunk(
-  storagePath: string,
-  offset: number,
-  length: number,
-): Promise<Uint8Array | null> {
-  return cache.getEntry<Uint8Array>(chunkKey(storagePath, offset, length));
-}
-
-export function putChunk(
-  storagePath: string,
-  offset: number,
-  length: number,
-  data: Uint8Array,
-): Promise<void> {
-  return cache.putEntry(chunkKey(storagePath, offset, length), data);
 }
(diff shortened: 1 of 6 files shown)


natb1:
# Fix-checks summary — PR #2633

## Iteration 1

**Failed checks**: `test-integrity`

**Outcome**: `needs-human`

**Reproduced**: yes

**Reproduce command**: `bash .github/scripts/check-test-integrity.sh`

**Failure excerpt**:
```
Signal 2: net test-declaration removals detected
  DECL_ADDED=1, DECL_REMOVED=15, DECL_NET=-14
  Net removals across *.test.ts / *.spec.ts vs origin/main
```

**Why not caught**: The PR intentionally removed 15 test declarations that covered deleted symbols:
- 7 `it()` tests in `print/test/viewer/image-archive.test.ts` for `CachedRangeReader` / `fetchArchiveBuffer` / range-reader paths (deleted by Unit 2)
- 6 `it()` tests + 1 `describe()` in `print/test/media-cache.test.ts` for `putChunk`/`getChunk` (deleted by Unit 3)
- 1 `it()` replaced (net: 1 added, 15 removed = -14)

Restoration is impossible — the tested symbols no longer exist. `check-test-integrity.sh` has no exemption mechanism for deliberate dead-code cleanup.

**Required action**: Human sign-off to accept the net -14 test declaration count. Options for the owner:
1. Add an exemption to `check-test-integrity.sh` for this PR (e.g. a bypass comment or a per-PR allow-list), or
2. Relax the Signal 2 threshold to allow the PR's net count, or
3. Confirm the deletion is acceptable and merge with the check disabled or amended.

PR #2641, opened 2026-06-30
check-test-integrity: exempt dead-code-cleanup test removals from Signal 2

natb1:
Closes #2637

## What

Refine `check-test-integrity.sh` **Signal 2** so it no longer fires on
legitimate dead-code cleanup — a test removed because the symbol it exercised
was deleted in the same PR — while still firing on genuine test-weakening.

Signal 2 counts net removed test declarations and cannot, on its own,
distinguish:

- **test-weakening** — deleting a test whose subject code still exists (what the
  gate must stop), from
- **legitimate dead-code cleanup** — deleting a test because the code it
  exercised was removed in the same PR (mandatory cleanup, not weakening).

The motivating instance is **PR #2633** (for #2631): it removed
`CachedRangeReader`, `fetchArchiveBuffer`, `getChunk`, and `putChunk` from source
along with their now-orphaned tests (net −14 declarations) and could only land
via a human admin override-merge. All five changed files were **Modified** (no
whole-file deletions), so the existing basename co-deletion exemption — which
requires a co-*deleted* impl file of matching basename — did not apply.

## How

A second, import-based exemption runs alongside the existing basename exemption
and feeds the same `EXEMPT_DECL_REMOVED` accumulator (one correction applied
once, after both loops). For each changed test file F whose net declaration
count went negative:

1. Compute the **removed-import set** as a true `imports(origin/main:F) −
   imports(HEAD:F)` whole-file set difference. This is why a symbol merely
   re-imported on a reformatted line nets out — exactly the #2633 case where
   `createImageArchiveRenderer` survived on the same import line as the deleted
   `CachedRangeReader`.
2. F is exempt **iff** the removed-import named set is non-empty, contains no
   unverifiable namespace/default import, and **every** removed-import symbol is
   **absent** from the post-PR non-test source tree.
3. The existence check matches word-boundary declaration/export/re-export forms
   (never bare substrings, so `Cache` does not match `CacheManager`), grepping
   `HEAD` with `:(exclude)` pathspecs for the test globs.

Two conservatism axes are embedded as comments and bias toward **firing** on any
uncertainty: a still-present symbol, an unverifiable removed import, or an empty
removed-import set all leave the file non-exempt. A double-counting guard skips
files already basename-exempt so the single correction cannot over-subtract and
mask weakening elsewhere in the same PR.

The override-merge relief path remains documented behavior for the file-granular
cases this exemption deliberately does not cover.

## Tests

Ten new hermetic cases (g)–(p) in `test-check-test-integrity.sh` pin each
conservatism boundary and hole — the #2633 happy path, weakening still firing,
mixed gone/surviving symbols in one file, vacuous-true, substring false-match,
`import type`, namespace and default (unverifiable) imports, multi-line import
blocks, and re-export brace forms. The full suite is green: **54 passed, 0
failed** (28 pre-existing + 26 new assertions). It is the same script CI runs at
`.github/workflows/unit-tests.yml`.


natb1:
# QA Summary — Issue #2637 / PR #2641

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Co-deletion cleanup passes Signal 2 | shell-command | PASS |
| 2 | Test-weakening still fires, including mixed-file case | shell-command | PASS |
| 3 | Vacuous-true hole closed | shell-command | PASS |
| 4 | Full suite green — new adversarial cases plus all pre-existing cases | shell-command | PASS |
| 5 | Exemption logic stays diff-scoped vs origin/main | needs-human-judgment | (see below) |

**Counts: 4 PASS, 0 FAIL, 0 SKIP** (plus 1 already-satisfied — see below)

## Shell-command verification

```
bash .github/scripts/test-check-test-integrity.sh
Results: 54 passed, 0 failed, 54 total
All tests passed.
```

Cases (g)–(p) covering the import-based co-deletion exemption all pass alongside
the pre-existing 28 assertions (cases a–f).

## Bugs found

None.

## Disposition triage

Step 3.5 ran on 1 residue item (item 5, `needs-human-judgment`).
No `needs-main` items were filed.

| # | Title | Rationale |
|---|-------|-----------|
| 5 | Exemption logic stays diff-scoped vs origin/main | Code inspection of `check-test-integrity.sh` confirms every git invocation in the new import-based exemption path is diff-scoped against `origin/main...HEAD` or the named `origin/main`/`HEAD` revisions: candidate enumeration (`git diff --name-only origin/main...HEAD`), per-file filter (`git diff --unified=0 origin/main...HEAD -- $F`), old/new blob parse (`git show origin/main:$F` / `git show HEAD:$F`), and the existence check (`git grep HEAD -- :(exclude)` globs). No `HEAD~1` narrowing and no working-tree/unscoped scan, so CI behavior is identical regardless of branch history depth. The criterion is provably met by readable evidence; no code change and no human needed. |

Item 5 is dropped as PASS (absent from the escalation set).

## Deferred to office-hours

None.

natb1:
## Review: PR #2641 (`2637-check-test-integrity-exempt`)

**8 findings surfaced · 4 fixed (fix commit: 41750b48) · 2 refuted · 1 informational · 1 dismissed · 0 follow-ups**

---

### Fixed (code-review)

- **`:288` — IMPORT_AWK `from` keyword test is an unanchored substring match.** `$0 ~ /from/` fires on any member whose name contains `from` as a substring (e.g. `fromEntries`, `dateFromNow`). Replaced with `$0 ~ /from[ \t]+['"]/` on the opening-line classifier; the collecting-block terminator now closes on `}` alone, dropping the `|| $0 ~ /from/` clause entirely.

- **`:370` — `NAMED_EXPORT_FORM` `[^}]*` doesn't cross newlines.** The pattern `export[[:space:]]*\{[^}]*\b${X}\b[^}]*\}` fails to match a multi-line `export { … }` block — `[^}]*` does not consume newlines, so a symbol re-exported via a brace block spanning lines reads as absent and the file is wrongly exempted. Replaced with a two-phase `EXPORT_AWK` program that reads the post-PR non-test source and emits every brace-exported name (joining multi-line blocks), then checks exact membership with `grep -qxF`.

- **`:368` — `$X` interpolated into ERE patterns without escaping.** Import binding names containing `$` (valid in JS: `$store`, `data$`) become end-of-line ERE anchors inside `git grep -qE`, making the existence check impossible to match and silently keeping `all_absent=1`. Escaped `$` before interpolation.

---

### Required (security) — upheld and fixed

- **`:378-382` — File-granular exemption credits all removed declarations in a file regardless of which ones are tied to the gone symbols.** When the import-based loop exempts a file, it adds the full `F_DECL_REMOVED` count — including declarations for subjects whose imports *survived* the PR (kept live by remaining tests). Concrete bypass: a PR legitimately deletes `GoneSymbol` from source and drops its import, but co-removes an unrelated `Survivor` test while keeping `Survivor`'s import. `REMOVED_NAMED = {GoneSymbol}`, `GoneSymbol` absent → `all_absent=1` → `Survivor`'s test removal is swept into the exemption credit; Signal 2 stays silent. **Fixed by making the exemption symbol-granular:** the correction is now the count of removed declarations in `F` whose enclosing `it`/`test` block references at least one symbol from `REMOVED_NAMED`, rather than the flat `F_DECL_REMOVED`. Added discriminating test cases covering the inline-helper and survivor-import-kept paths.

---

- **`:367-376` — `input-validation-0`: ERE-escaping `$`-prefixed identifiers.** The finding is technically correct that `$` in an identifier name becomes an ERE end-of-line anchor. However, the proposed fix (`sed 's/[.^$*+?{}|\\()[\\]]/\\&/g'`) is itself empirically broken: the escaped pattern `const[[:space:]]+data\$\b` still returns no match because POSIX ERE `\b` defines word characters as `[A-Za-z0-9_]` only — `$` is never a word character, so `\b` immediately after `$` never fires. Upholding would apply a broken fix. Additionally, the codebase has zero `$`-suffixed exports; the exploit chain requires condition (a) a `$`-named exported symbol to exist, which is absent entirely. **Dropped.**

- **`:368-372` — `red-team-0`: `$`-prefixed identifiers silently bypass the existence check.** Same underlying mechanism. Two independent skeptics both refuted: zero `$`-prefixed named exports or imports exist anywhere in the codebase; no Svelte/SolidJS/Angular dependencies are present (the frameworks where this idiom is idiomatic); the project is React + TypeScript + Firebase throughout. The bug is latent with no exploitable attack surface. **Dropped (2/2 skeptics refuted).**

---

### Informational

- **`:367` — `for X in $REMOVED_NAMED` uses unquoted word-splitting.** The surrounding script uses the same unquoted-variable iteration pattern at lines 159 and 304 for similar constructs. Import binding names are valid JS identifiers and cannot contain whitespace, so this is consistent with the established style rather than a correctness gap. No change required.

---

### Dismissed

- **`:292-295` — `red-team-2`: awk collecting-block `$0 ~ /from/` fires on member names containing `from`.** The code-review-0 finding already addressed the `from` substring issue in the opening-line classifier (`:288`); the collecting-block terminator's `|| $0 ~ /from/` clause was removed in the same fix. This finding is subsumed by code-review-0 and is resolved. **Dismissed as duplicate.**

---

_No deferred follow-ups. No security follow-up issues filed._

PR #2672, opened 2026-06-30
budget-etl: fold report/patch/dump into one binary with subcommands

natb1:
Closes #2643

## What

Fold `budget-etl`'s `report`/`patch`/`dump` programs into the single
`budget-etl` binary, dispatched on the first argument
(`report`/`merge`/`patch`/`dump`), so a distributed plugin can run all of them
without a Go toolchain. The existing flag-only surface keeps working unchanged.

This is one migration PR of the epic merging `/budget` + `/budget-etl` into a
single distributed skill (parent #2642).

## How

- **Unit 1 — `patch.go`:** copy the patch symbols (`Spec`, `RemoveSpec`,
  `applySpec`, `runPatch`, predicate helpers) into the parent `package main`,
  plus a `runPatchCmd` flag wrapper on a dedicated `FlagSet`.
- **Unit 2 — `dump.go`:** extract the dump logic into `runDump` + a
  `runDumpCmd` flag wrapper on a dedicated `FlagSet`.
- **Unit 3 — dispatcher:** add an `os.Args[1]` switch at the top of `main()`.
  `patch`/`dump` route to their wrappers; `report`/`merge` set a forced mode and
  strip the subcommand word before flag parsing (`flag.Parse` →
  `flag.CommandLine.Parse(args)`). Forced-mode branches run after password and
  `DiscoverOpts` construction, reusing the parsed flags exactly as the legacy
  cascade does; the legacy flag-only path is preserved verbatim in the default
  branch. New `dispatch_test.go` subprocess tests cover patch/dump routing and
  legacy flag fall-through.
- **Unit 4 — tests:** move the patch tests + `testdata/` into the parent package
  and delete the duplicated `convertExportRulesForTest`/`dollarsToCentsForTest`
  helpers, rewiring the integration test to call the real `convertExportRules`
  (removing the documented divergence risk).

## Intentional transitional duplication

Go's `package main` cannot be imported, so the patch/dump logic folded into the
parent is **copied**, not moved — `cmd/patch/main.go` and `cmd/dump/main.go`
stay as working standalone binaries. This is deliberate: the `budget-etl` skill
still invokes `go run ./cmd/patch` / `go run ./cmd/dump`, and the issue keeps
`cmd/patch`/`cmd/dump` until they are retired in PR4. The one duplication this
issue targets for removal now — `convertExportRulesForTest` — is removed
(Unit 4).

## Backward compatibility

No consumer passes a bare first-arg word — every legacy invocation's first arg
starts with `-`, so the dispatcher routes all of them to the unchanged legacy
path. The `budget-parse-job` and `budget` skill flag forms are preserved.

## Verification

`go test ./...` is green for `projects/budget-etl` (parent, `cmd/patch`,
`cmd/dump`, and `internal/*`) locally. CI is authoritative
(`.github/workflows/unit-tests.yml` runs `go test ./...`). The release workflow
is unchanged — it still builds `-o budget-etl-… .` from `projects/budget-etl`.


natb1:
## QA Summary — Issue #2643 / PR #2672

### Results

| Item | Title | Result |
|------|-------|--------|
| 1 | Go test suite is green across budget-etl | ✅ PASS |
| 2 | Subcommand dispatcher recognizes report/merge/patch/dump | ✅ PASS |
| 3 | Legacy top-level flag surface preserved | ✅ PASS |
| 4 | convertExportRulesForTest removed; patch test uses real convertExportRules | ✅ PASS |
| 5 | Release workflow still builds one artifact per platform | ↗ planned-deferral (filed as follow-up) |

**Counts:** 4 PASS, 0 FAIL, 0 SKIP, 1 planned-deferral filed as follow-up.

No bugs found. No browser walkthrough (pure Go backend PR).

### Acceptance criteria

- ✅ `budget-etl report|merge|patch|dump <args>` each work as subcommands — confirmed by dispatcher arms in `main.go` (lines 28, 30, 32) and the subprocess tests in `dispatch_test.go` (green via `go test ./...`).
- ✅ Legacy flag forms (`--report`, `--input`, `--dir`, `--allow-uncategorized`, `--output`, `--plaintext`) still defined and the legacy cascade preserved; `TestDispatchLegacyFlagFallthrough` confirms runtime routing.
- ✅ `convertExportRulesForTest` is absent from the entire tree; `patch_integration_test.go:68` calls the real `convertExportRules`.
- ✅ `go test -count=1 -buildvcs=false ./...` green: parent + cmd/dump + cmd/patch + all `internal/*` packages pass.
- ↗ Release workflow builds one artifact per platform — see #2673 (deferred to post-merge verification).

### Disposition triage

| Item | Kind | Class | Verify | Rationale |
|------|------|-------|--------|-----------|
| 5 — Release workflow unchanged | needs-human-judgment | needs-main | n/a | planned-deferral: criterion #5 is documented as non-assertable at merge time; verifiable only against merged main / production release. |

### Filed follow-ups

- **#2673** — "qa: verify against main — release workflow still builds one artifact per platform" (newly filed; route: `human` → `main-qa` + `dispatch:office-hours`; blocked_by #2643)

### Phase log entry

failed: none — all 4 script-verifiable items passed; 1 planned-deferral filed as #2673 (needs-main, human route).

natb1:
## Review summary — #2643

### Fixed (879c6a2f)

- **`projects/budget-etl/patch.go:91`** — Collision check built `existingIDs` from ALL current rules before `removeByID` was known, so a spec could not atomically remove and re-add a rule with the same ID. Fixed: build `removeByID` first, then exclude to-be-removed IDs when populating `existingIDs`.
- **`projects/budget-etl/main.go:58`** — The `--report requires --allow-uncategorized` guard ran unconditionally even when the user invoked the `report` subcommand (where `--allow-uncategorized` would be redundant/surprising). Fixed: introduced `effectiveAllowUncategorized` (`true` when `--allow-uncategorized` is set or `forced == "report"`) and used it in the guard.

### Informational

- **`projects/budget-etl/patch.go:109`** — IDs in `spec.Remove.ByID` that do not match any existing rule are silently ignored. A typo in a rule ID would produce no error and no warning. Leaving as informational — silent no-ops are consistent with the current design philosophy and changing it is a separate UX decision.
- **`projects/budget-etl/dump.go:28`** — The local `summary` struct in `runDump` only exports `Transactions`, `Budgets`, and `Rules`, silently dropping `Statements`, `NormalizationRules`, `BudgetPeriods`, and `WeeklyAggregates`. This matches the documented scope of the `dump` subcommand but is worth noting for future completeness.

### Dismissed

- **`projects/budget-etl/patch.go:209`** — `predicateMatches` uses `!=` to compare `*float64` values for `MinAmount`/`MaxAmount`. In current call paths both pointers originate from JSON-decoded values (no arithmetic), so bit-identical comparison is correct. Dismissed: not a real bug in this context.

### False-positive (security)

- **`projects/budget-etl/patch.go:87–100`** (`input-validation`) — `Type` field in `add[]` rules not validated against the enum. The downstream categorization engine already handles unknown types gracefully and enforcement is an application-layer concern, not a security boundary here.
- **`projects/budget-etl/patch.go:144–151`** (`red-team`) — `validateAddRule` only validates `TransactionID` format. No injection surface exists; the field is stored as a string key, not executed.
- **`projects/budget-etl/patch.go:43–45`** (`red-team`) — Same-file check uses string equality rather than canonical paths. Valid concern in general, but the caller context (CLI tool, single user, no privilege boundary) makes path-canonicalization bypass a non-issue.
- **`projects/budget-etl/patch.go:57`** (`red-team`) — No file size limit on `--spec`. The tool runs as a CLI under the invoking user's own files; no DoS surface exists.
- **`projects/budget-etl/dump.go:34`** (`data-exposure`) — `runDump` decrypts and prints financial data to stdout by design. The user's own encrypted snapshot, decrypted by the user's own password, written to stdout at the user's explicit request is not a data exposure issue.

PR #2719, opened 2026-07-02
testing: harden post-deploy @hosting smoke gate + retry transient CDN 5xx/hangs

natb1:
## Problem

The Production Deploy workflow's `deploy-and-smoke` job runs post-deploy
`@hosting` smoke tests against freshly-released Firebase Hosting sites. It has
repeatedly reddened `origin/main` with a **transient** hosting error even though
the deploy itself was healthy (#2705: `landing`/`print` smoke failed on HTTP 503
and `Request context disposed`; re-probing right after returned 200). Each
episode latches the main-broken gate, files a `dispatch:main-broken` issue, and
stalls the queue — real cost for a non-defect.

Two root causes, both in the smoke phase (not the deploy-release call — that 500
is #1227, out of scope):

1. **The readiness gate declared the site ready too early.** `run-smoke-tests.sh`
   polled only the root document and broke on the *first* `200` containing
   `<script type="module"`. Firebase Hosting release propagation is not atomic
   across the edge, so a single good root response did not mean the new release
   had stably propagated — a subsequent request could still get a 503 or a stale
   version.
2. **The `@hosting` assertions had no per-request tolerance.** The shared
   `packages/config/*-smoke.ts` helpers issued one `request.head`/`request.get`
   per URL and asserted synchronously with no retry. A single mid-propagation 503
   — or a request that *hangs* through the propagation window until Playwright's
   test timeout disposed the test's `request` context (the `Request context
   disposed` symptom) — failed the assertion, the job, and reddened main.

## Fix

Treat post-deploy edge propagation as eventually-consistent within a bounded
window: confirm *stable* propagation before tests run, and make the deployed
`@hosting` assertions retry a transient 5xx **or a transient throw/hang** through
a bounded budget — while still hard-failing when the edge stays bad past that
budget, preserving the real deploy-integration guarantee.

- **Primary defense — stable-propagation gate (Unit 3, 4).** New
  `wait_for_stable_propagation` in `lib.sh` polls until `REQUIRED_CONSECUTIVE`
  (default 3) good observations in a row, each verifying **both** the root
  document (200 + `<script type="module"`) **and** a re-extracted content-hashed
  asset returns 200. Any miss resets the counter and logs a one-line reason; on
  timeout it hard-fails with the last status and `head -20` of the HTML.
  `run-smoke-tests.sh` now calls it instead of breaking on the first 200. Covered
  by a deterministic PATH-shim test (`test-run-smoke-tests.sh`, 10/10) registered
  in CI's `hook-tests` job.
- **Backstop — transient-aware, timeout-bounded retry wrapper (Unit 1, 2).**
  `fetchWithRetry` + `getWithRetry`/`headWithRetry` in
  `hosting-smoke-helpers.ts` retry **only** a transient signal (a 5xx response,
  or a throw whose message matches a transient/timeout shape). Each wrapper passes
  an explicit bounded per-request `timeout` (default 10s) so a hang throws a
  catchable `TimeoutError` *early* — while the request context is still alive —
  leaving budget to retry, which prevents the disposal rather than trying to retry
  an already-disposed context. A `<500` response or a non-transient throw fails
  immediately, unmasked, so a genuine header/content regression is not hidden. All
  three smoke modules route their GET and HEAD requests through the wrapper.
  Covered by a deterministic vitest test with an injected fake request context.

### Why a wrapper, not native Playwright `retries`

- **Precision.** The wrapper retries only on a transient signal; a genuine
  200-with-wrong-header regression fails immediately. Native `retries` would
  re-run the whole test on *any* failure, masking real regressions that land
  mid-propagation.
- **Deterministic testability.** The wrapper is unit-testable with an injected
  fake `APIRequestContext`; native `retries` is not.
- **Emulator-path safety.** The same `@hosting` tests also run against the
  zero-egress Hosting emulator, which never returns 5xx and never hangs — so the
  wrapper never fires there and needs no env-var gating. A blanket config
  `retries` would slow every emulator run.

### Budget arithmetic

Worst case (a single persistently-hanging request): `attempts × per-request-timeout
+ backoff = 4 × 10s + ~3.5s ≈ 43.5s`, comfortably below the 90s Playwright test
timeout, so the wrapper's own exhaustion produces a clear error *before* the
test-timeout disposal. All knobs are env-overridable
(`HOSTING_SMOKE_RETRY_ATTEMPTS`, `HOSTING_SMOKE_RETRY_DELAY_MS`,
`HOSTING_SMOKE_REQUEST_TIMEOUT_MS`; gate: `TIMEOUT`, `INTERVAL`,
`REQUIRED_CONSECUTIVE`).

### Deliberate tradeoff

The gate moves from "break on first 200" to N-consecutive, raising happy-path
latency by ~`(REQUIRED_CONSECUTIVE-1) × INTERVAL` (≈4s at defaults) — bounded and
intended, in exchange for not starting tests mid-propagation.

## Verification

- `npx vitest run --project packages/config` — 38/38 pass (includes the new
  `hosting-smoke-helpers.test.ts`: transient-503-then-200, budget-exhausted 503,
  transient throw/hang retry, persistent-throw rejection, non-retried 404 /
  non-transient throw, and bounded-`timeout` assertions).
- `test-run-smoke-tests.sh` — 10/10 pass (stable-from-first-poll, reset-then-recover,
  persistent 503, missing module script, failing asset probe).
- `run-typecheck.sh --app packages/config` — skipped: `run-typecheck.sh` probes
  for a `tsconfig.json` and packages/config has none (it ships `tsconfig.lib.json`
  / `tsconfig.app.json` instead), so it type-checked zero targets here. Vitest
  transpile-strips types, so as insurance the two new/changed `.ts` files were
  type-checked standalone with the `tsconfig.lib.json` options
  (`tsc --noEmit --strict`) — exit 0. (Pre-existing gap: the sibling `*-smoke.ts`
  files are likewise not reached by `run-typecheck.sh`.)

Manual / observe-in-production (require a real Firebase deploy + egress): the
next prod/preview deploy exercises the hardened gate + retrying assertions
end-to-end. A genuinely persistent 5xx still hard-fails after the retry budget;
the transient cases are covered deterministically at unit level.


natb1:
# Fix-checks summary — PR #2719

## Iteration 1

- **Failed checks**: `type-safety-sensor`
- **Outcome**: `fixed`
- **Reproduced**: yes
- **Reproduce command**: `git diff --no-color -U0 --diff-filter=d origin/main...HEAD -- '*.ts' '*.tsx' '*.js' '*.jsx' | .github/scripts/check-type-safety-escapes.sh --scan-stdin`
- **Failure excerpt**:
  ```
  ::error file=packages/config/test/hosting-smoke-helpers.test.ts,line=21::Net-new type-safety escape hatch (as <Type> cast).
  ::error file=packages/config/test/hosting-smoke-helpers.test.ts,line=50::Net-new type-safety escape hatch (as <Type> cast).
  ::error file=packages/config/test/hosting-smoke-helpers.test.ts,line=181::Net-new type-safety escape hatch (as <Type> cast).
  ::error file=packages/config/test/hosting-smoke-helpers.test.ts,line=191::Net-new type-safety escape hatch (as <Type> cast).
  ```
- **Why not caught**: `fakeResponse` and `fakeRequest` test helpers used `as unknown as Type` double-casts for minimal fake objects, and two assertions used `as number` to narrow an `unknown[]` element type. None had `type-safety-ok` suppressions.
- **Fix**: Added `// type-safety-ok: <reason>` to lines 21, 50, 181, and 191 in `packages/config/test/hosting-smoke-helpers.test.ts`. Commit: `98baa71a`

natb1:
# QA Summary — Issue #2713 / PR #2719

## Results

**5 items executed, 5 PASS, 0 FAIL, 0 SKIP**
**1 item deferred (needs-main planned-deferral, filed as follow-up)**

No bugs found. No browser walkthrough (no browser component — pure CI/config/shared-package changes).

## Item results

| # | Title | Result |
|---|-------|--------|
| 1 | Stable-propagation gate requires multiple consecutive good observations and probes a hashed asset | PASS |
| 2 | run-smoke-tests.sh is rewired to the stable gate (old first-200 loop removed) | PASS |
| 3 | Every @hosting smoke assertion module routes through the retrying request wrappers | PASS |
| 4 | Retry masks only transient 5xx/network errors, is budget-bounded, and does not swallow persistent failures | PASS |
| 5 | Readiness gate hard-fails with a clear error when the site never serves expected content within the timeout (no silent pass) | PASS |
| 6 | run-all-prod/preview-deploy-smoke.sh pass end-to-end against a real deploy | needs-main (planned-deferral) |

**Evidence (items 1 and 5):** `test-run-smoke-tests.sh` — 10/10 cases pass, including reset-then-recover, persistent 503 hard-fail, missing-module-script hard-fail, asset-probe failure hard-fail.

**Evidence (item 2):** `wait_for_stable_propagation "$BASE_URL"` is on line 36 of `run-smoke-tests.sh`; no old first-200 readiness loop remains.

**Evidence (item 3):** All three smoke modules (`reachability-smoke.ts`, `security-headers-smoke.ts`, `cache-headers-smoke.ts`) import and use `headWithRetry`/`getWithRetry`; no raw `request.head()`/`request.get()` calls remain.

**Evidence (item 4):** `isTransientError` in `hosting-smoke-helpers.ts` covers HTTP 5xx, ECONNRESET, and timeout shapes only; `<500` responses return immediately without retry. Attempts bounded at 4 (default), per-request timeout 10s. Vitest suite 38/38 passes.

## Disposition triage

Step 3.5 ran (residue list non-empty — one planned-deferral item).

| 6 | run-all-prod/preview-deploy-smoke.sh end-to-end | `needs-main` | n/a | planned-deferral: AC#5 requires a real Firebase deploy + live egress; only observable on the next prod/preview deploy |

Non-fixing pass: no `opus-fixable` items; the only residue was the planned-deferral `needs-main` item above, which was filed as a follow-up and dropped. Escalation set is empty → clean pass.

## Filed follow-ups

- #2726 — "qa: verify against main — qa-needs-main #2713: run-all-prod-deploy-smoke-sh-and-run-all-preview-deploy-smoke-sh-pass-end-to-end-against-a-real-deploy-across-changed-apps" (route: human → `main-qa` + `dispatch:office-hours`)

natb1:
## Review — `/review-fix`

Combined review pass (code-review quality + surface-gated security finders +
inline CodeQL/erosion scans) over the post-deploy `@hosting` smoke-test
hardening. Surface: `code` (app source + dispatch scripts). CodeQL: no open
alerts. Dependency audit: not run (no `package.json`/lock changes). Security
finders ran with no findings. 8 findings surfaced, 4 fixed.

Fix commit: `0a78de77`.

### Fixed

- **`.claude/skills/dispatch-propagate/scripts/lib.sh:3166,3190`** — `wait_for_stable_propagation`'s
  root-poll and asset-probe `curl` calls had no time bound, so a hung request
  could block past `TIMEOUT`. Added `--connect-timeout 5 --max-time 10` to both,
  well under `INTERVAL`, so a hang now fails fast into the existing
  retry/reset logic.
- **`packages/config/hosting-smoke-helpers.ts:96`** — `fetchWithRetry`'s
  resolved-response retry gate only caught 5xx; a transient `429` was returned
  immediately. Now retries `429` alongside 5xx.
- **`packages/config/hosting-smoke-helpers.ts:6-7`** — `isTransientError`'s doc
  comment cited a `packages/intentionsutil/scripts/backfill.ts` path that no
  longer exists. Replaced with an accurate description.
- **`packages/config/hosting-smoke-helpers.ts:13`** — the transient-error regex
  carried bare, unanchored `Timeout` and `exceeded` terms that could swallow an
  unrelated application error into a retry loop. **Correction to the applied
  fix:** the auto-fix dropped both bare terms outright, which broke the PR's core
  backstop — Playwright's real per-request-`{ timeout }` message is
  `apiRequestContext.get: Timeout <n>ms exceeded.`, which matched *only* via
  those two bare terms and none of the narrowed regex's kept phrases (`timed
  out`, `deadline exceeded`, `i/o timeout`). Verified empirically: an actual
  `request.get(url, { timeout })` firing produces `Timeout 400ms exceeded.` with
  `name: TimeoutError`, and `fetchWithRetry`'s catch branch gates the throw-retry
  path solely on `isTransientError(message)` (no `TimeoutError` name
  short-circuit). So the over-narrowed regex would have caused a mid-propagation
  hang — issue #2713's exact symptom — to rethrow without retrying. Replaced the
  removed bare terms with an anchored `Timeout \d+ms exceeded`, which restores
  Playwright coverage while still satisfying the over-breadth concern (no bare
  `exceeded`). Strengthened the vitest test with a positive case using the real
  Playwright message shape so this regression cannot recur silently (the prior
  test used `"Request timed out"`, which is Playwright's MCP-path error, not the
  APIRequestContext path this PR exercises). Suite now 39/39.

### Informational

- **`.claude/skills/dispatch-propagate/scripts/run-smoke-tests.sh:29-34`** — the
  new tradeoff comment documents only the happy-path latency increase, not the
  hard-fail-on-timeout behavior. FYI; no change.
- **`.claude/skills/dispatch-propagate/scripts/lib.sh:3175-3189`** — the
  content-hashed-asset discovery regex (script-src then stylesheet-href fallback)
  is implemented once here and again in the shell — a minor duplication. FYI; no
  change.
- **`packages/config/hosting-smoke-helpers.ts:60`** (erosion) — worst-function
  cyclomatic complexity in this file rose 6 → 15 (total 6 → 36) vs `origin/main`.
  Expected for a file that gained the retry/backoff machinery; surfaced for
  visibility, no change required in this PR.

### Dismissed

- **`packages/config/hosting-smoke-helpers.ts:20-22`** — `requestTimeoutMs()`
  reads `HOSTING_SMOKE_REQUEST_TIMEOUT_MS` via `envInt`, whose `parseInt(...) ||
  fallback` treats an explicit `0` as unset. Not a real defect: a `0` (or
  negative) per-request timeout would disable the bound, which is undesirable, so
  falling back to the default is the correct behavior — not a bug.

_No security-required, refuted, false-positive, deferred, or out-of-scope
findings._

PR #2722, opened 2026-07-02
office-hours: dispatch event-hook triggers for low-latency snapshot refresh

natb1:
Closes #2661

## What & why

Adds the **low-latency event layer** for the office-hours local snapshot, on top
of the existing local producer (`office-hours-snapshot/`, #2658) and the sibling
Nix systemd timer (#2660). The dispatch router's own events now launch a
best-effort, detached producer refresh so the dashboard reflects chain activity
promptly, without polling.

Two attach points matter for latency — a worker **parking** and **un-parking** —
because they change exactly what the Parked Issues panel shows and are the only
sub-tick-latency events in the chain. A per-tick `full` refresh covers everything
else (merges, phase-completes, and drains all resolve inside a tick, so observing
at tick-start captures them).

## How

A single reusable launcher, invoked at four call sites. It is **env-gated OFF by
default** (`OFFICE_HOURS_SNAPSHOT_ENABLED`), **detached** (`setsid … &`), **never
blocks** its caller, and **never aborts** a tick or hook — so the wiring is
byte-for-byte dormant on every current tick/hook until the snapshot layer is
explicitly enabled. This lets it ship independent of #2660 (no `blocked_by`): when
enabled without #2660 it runs the producer via `npx tsx`; #2660's Nix packaging
later overrides `OFFICE_HOURS_SNAPSHOT_CMD` to the compiled binary.

- **Unit 1** — new `office-hours-snapshot-launch` (+ auto-discovered unit test).
  The shared primitive: arg-validated (`full` | `parked-only`), env-gated,
  synchronous `npx`-presence precondition for the default command, per-scope
  `flock` serialization via a **bounded-wait `flock -w 30` inside the detached
  child** (so a `full` and a `parked-only` run may proceed concurrently while two
  of the same scope serialize, and the caller never waits).
- **Unit 2** — `dispatch-tick`: tick-start `full` refresh beside the existing
  detached side-effect launches. `full` (not `parked-only`) because only `full`
  runs the cores that carry queue-metrics, which the tick-start criterion
  requires.
- **Unit 3** — `dispatch-apply-office-hours`: `parked-only` refresh on the
  fresh-apply (park) path only, past the idempotency no-op so a repeatedly-firing
  hook does not re-launch.
- **Unit 4** — the two un-park hooks: `dispatch-stop.sh`'s
  `strip_office_hours_label()` (gated on the existing park→unpark
  `ISSUE_OFFICE_HOURS_PRESENT=yes` signal) and `dispatch-office-hours-strip.sh`
  (unconditional — already scoped to `<N>-*` worktrees on user engagement).

## Verification

Both auto-runnable `verify` blocks pass:

- `test-office-hours-snapshot-launch.sh` — 14/14 (gate-off no-op, arg misuse,
  gate-on `full`/`parked-only` launches, per-scope lock isolation, and a
  contention case pinning that the `flock` lives in the child so the caller stays
  fast).
- `test-dispatch-scripts.sh` — 4166/4166.

Manual (owner-machine) end-to-end and concurrency checks are noted in the plan's
Verification section for QA; the launcher is a no-op until
`OFFICE_HOURS_SNAPSHOT_ENABLED=1`, so it does not change any current behavior.

### DIFF
--- a/.claude/hooks/dispatch-office-hours-strip.sh
+++ b/.claude/hooks/dispatch-office-hours-strip.sh
@@ -47,4 +47,10 @@ fi
 gh_issue_remove_label_rest "$ISSUE_NUM" dispatch:office-hours >/dev/null 2>&1 \
   || echo "[dispatch-office-hours-strip] WARNING: gh_issue_remove_label_rest failed for issue #$ISSUE_NUM" >&2
 
+# Office-hours snapshot: un-park `parked-only` refresh (#2661). Fired
+# unconditionally — this hook already runs only inside an <N>-* worktree on user
+# engagement; env-gated OFF by default + flock-serialized + best-effort, so a
+# redundant refresh is harmless. Preserves the never-block-submission contract.
+"$SCRIPTS/office-hours-snapshot-launch" parked-only || true
+
 exit 0
--- a/.claude/hooks/dispatch-stop.sh
+++ b/.claude/hooks/dispatch-stop.sh
@@ -295,6 +295,11 @@ strip_office_hours_label() {
   # cross-phase accumulation that is the whole point of the ceiling.
   if [ "${ISSUE_OFFICE_HOURS_PRESENT:-no}" = yes ]; then
     clear_attempt_counter
+    # Office-hours snapshot: park→unpark `parked-only` refresh (#2661). Same
+    # transition gate as the counter reset — only on a genuine unpark, not on
+    # every advance/self-close. Best-effort; the `|| true` keeps the function
+    # total (the file runs `set -uo pipefail` with a `trap … exit 0 ERR`).
+    "$SCRIPTS/office-hours-snapshot-launch" parked-only || true
   fi
 }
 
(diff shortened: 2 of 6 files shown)


natb1:
# QA Summary — PR #2722 / Issue #2661

## Items executed

| # | Title | Lane | Result |
|---|-------|------|--------|
| 1 | Producer refresh launched at tick-start (env-gated, detached, best-effort) | shell-command | PASS |
| 2 | Park and un-park each trigger a parked-only fast refresh | shell-command | PASS |
| 3 | Concurrent timer + event runs serialize without a corrupt snapshot | shell-command | PASS |
| 4 | Merge / phase-complete / drain are NOT individually wired | shell-command | PASS |
| 5 | Full dispatch suite regression (call-site integration intact) | shell-command | PASS |
| 6 | End-to-end enabled run + live concurrency (owner machine) | deferred | needs-main |

**Counts:** PASS 5 / FAIL 0 / SKIP 0 / needs-main 1

## Bugs found

None.

## Disposition triage

Step 3.5 ran (1 residue item from triage).

| 6 | End-to-end enabled run + live concurrency (owner machine) | needs-main | n/a |

**Item 6** carries `planned_deferral: true` — the issue's Verification section marks the end-to-end enabled run and live concurrency checks as non-auto manual checks requiring owner-machine setup with `OFFICE_HOURS_SNAPSHOT_ENABLED=1` and a live producer. Disposition Workflow classified it `needs-main` (planned-deferral rule: not assertable at merge time, no defect to fix).

## Filed follow-ups

- Item 6 → **#2727** (newly filed, route: `human` — requires owner-machine setup + subjective judgment; labels: `main-qa` + `dispatch:office-hours`; blocked_by #2661)

PR #2775, opened 2026-07-05
tactic-graph-write-validation-hardening: symlink mode check, semantic shapes, tactic-body loss guard

natb1:
Hardens the graph write path per the node plan. Defense-in-depth: every gap addressed here currently fails elsewhere downstream; this makes each rule explicit at the boundary.

## Unit 1 — reject non-regular files in the fast-path guard

`.github/workflows/graph-fast-path.yml` guard step: the intentions/-only check is name-based, so a committed symlink (mode 120000) or gitlink (160000) at `intentions/x.md` passed it, leaving only validate-graph's frontmatter parse between it and a stamped SHA. Added a `git diff --raw origin/main...HEAD -- intentions/` dst-mode check that fails closed, in the same step and shell style.

## Unit 2 — schema semantic shapes and writeNode kind-change guard

`packages/intentionsutil/src/schema.ts`:
- `execution.pr`, each `execution.attempts` value, and `rounds.count` now require non-negative integers (new `requireNonNegativeInt` beside `requireNumber`).
- `office_hours.since` now requires `^\d{4}-\d{2}-\d{2}$` (new `requireDateString`) — the shape `graph-commit`'s park_write stamps via `date -u +%Y-%m-%d`.
- Live-store scan before tightening: 129 nodes (2 with execution, 1 with rounds, 10 with office_hours), zero violating values — the validator change is a no-op on existing data, no state fixes needed.

`packages/intentionsutil/src/store.ts`:
- `writeNode` now throws when rewriting an existing tactic file with `kind` changed away from `tactic` while the on-disk body is hand-authored (non-placeholder) plan content — previously the rewrite silently regenerated the placeholder body and discarded the plan. A placeholder-only body may still change kind freely; deliberate reclassification requires deleting or rewriting the file explicitly.
- `isPlainObject` exported from schema.ts for reuse.

Unit tests added for each new rejection and both kind-change paths (throw on hand-authored body, allow on placeholder).

Out of scope (recorded in the node): referential/layer rules stay validateGraph-only.

## Verification

- `npm test --prefix packages/intentionsutil` — 8 files, 158 tests, all pass
- `npx tsx packages/intentionsutil/scripts/validate-graph.ts intentions` — ok, 129 nodes
- `tsc -p packages/intentionsutil --noEmit` — clean
- Guard mode-check emulated locally against a real committed symlink at `intentions/scratch-link.md`: the `git diff --raw` awk filter flags it (`:000000 120000 ... A`). The plan's scratch-branch push only exercises the new check post-merge (on this branch the intentions/-only check fires first because non-intentions paths changed), so it is deferred to post-merge observation.

Graph node: tactic-graph-write-validation-hardening


natb1:
## QA summary — tactic-graph-write-validation-hardening (autonomous phase)

**Merge:** `git fetch origin main && git merge --no-edit origin/main` produced a merge
commit (102 files, mostly unrelated graph-node landings); pushed to
`tactic-graph-write-validation-hardening` (`65706437`). All 22 CI checks green
against that SHA (`gh pr checks 2775`).

### QA plan (from the intent, triaged)

The tactic promises three defense-in-depth guards. For each, I checked mechanism
correctness against real repo state/data, not synthetic fixtures mirroring the
implementer's own tests:

1. **Unit 1 — fast-path guard rejects symlink/gitlink under `intentions/`** — script-verifiable (guard shell logic), but the guard job itself only runs on `push` to `graph/**` branches, which this PR's branch is not — genuinely needs-main (see residue below).
2. **Unit 2a — schema tightens `execution.pr`, `execution.attempts.*`, `rounds.count` to non-negative integers, `office_hours.since` to `YYYY-MM-DD`** — script-verifiable against the live store.
3. **Unit 2b — `writeNode` throws on a kind change away from `tactic` that would discard a hand-authored plan body** — script-verifiable against real live-store nodes.

### What was validated, with real-data evidence

- **Unit 1 mechanism**, isolated in a scratch git repo (not a fixture — real `git diff --raw`/`awk` against real commits):
  - Committing a real symlink (`ln -s /etc/passwd intentions/evil.md`) → `git diff --raw main...feature -- intentions/` emits `:000000 120000 0000000 <sha> A␉intentions/evil.md`; the guard's `awk '$2 == "120000" || $2 == "160000"'` correctly matches on the new-mode field and would fail the job.
  - Committing a regular file only → the same awk filter emits nothing, confirmed exit 0 — no false-positive on the common case.
- **Unit 2a**, against the real store: `npx tsx packages/intentionsutil/scripts/validate-graph.ts intentions` → `ok — 186 nodes`, i.e. the plan's required pre-tightening scan (fix any live violations first) is a genuine no-op — no live node in `intentions/` violates the new integer/date rules.
  - Traced the sole writer of `office_hours.since`: `packages/intentionsutil/scripts/graph-commit`'s `park_write()` always stamps `date -u +%Y-%m-%d`, which is exactly the new regex's shape — the tightened schema cannot false-reject the only real write path for that field.
- **Unit 2b**, against a real live node (not a synthetic fixture): copied `intentions/tactic-graph-router-transitions.md` (a real tactic with a hand-authored plan body) into a scratch store and called `writeNode` with `kind: "strategy"` — it threw the expected `Refusing to change kind of "tactic-graph-router-transitions" from "tactic" to "strategy"` error, file left untouched.
  - Regression-checked the common case on the same real node: a same-kind rewrite (`phase: "qa"` only) preserves the body verbatim — no behavior change for the normal graph-commit write path.
- `npm test --prefix packages/intentionsutil` → 158/158 passing (8 files).
- `npx tsc -p packages/intentionsutil --noEmit` clean; `run-lint.sh` passes for `packages/intentionsutil`.
- Confirmed `isPlainObject`'s new export is not re-exported from `index.ts` — no public-API surface widening beyond the internal `schema.ts`↔`store.ts` need.

### Fixes made

None — no defects found.

### Residue

- **needs-main**: Unit 1's guard job only triggers on `push: branches: ['graph/**']`; this PR's branch name doesn't match, so the job has never actually run for this change, and no CI on this PR exercises it. My scratch-repo test proves the shell logic is correct, but does not prove the *actual GitHub Actions job* fails end-to-end. After merge: push a scratch `graph/**` branch that adds a symlink under `intentions/` (e.g. `intentions/x.md -> /etc/passwd`), confirm the "Graph Fast Path" guard job fails with the new "only accepts regular files under intentions/" error and a non-zero exit, then delete the scratch branch.

Disposition: pass (clean run, no fixes needed; one needs-main residue item recorded above).

natb1:
## Review disposition — tactic-graph-write-validation-hardening (terminal review gate)

**Disposition: clean — auto-merge armed (squash).** No Required findings; the three defense-in-depth guards are correct, well-tested, and non-regressive.

### Merge

`origin/main` advanced twice during review and was merged both times (clean, no conflicts). The second advance landed sibling **PR #2778** (`tactic-graph-commit-hardening`), which also edits `packages/intentionsutil/src/store.ts`. The two changes are orthogonal and coexist coherently in `writeNode`: #2778's `assertPathSafeId` refinement (relaxed `..`-substring → exact-id reject) and this PR's `assertNoTacticBodyLoss` body-loss guard. HEAD pushed as `ef7ba2d2`.

### What was reviewed (full `origin/main...HEAD` diff)

- **Unit 1 — fast-path symlink/gitlink guard** (`.github/workflows/graph-fast-path.yml`): the added `git diff --raw ... | awk '$2 == "120000" || $2 == "160000"'` correctly tests the *destination* mode (field 2 of raw output) so a committed symlink/gitlink under `intentions/` is rejected before `validate-graph` parses it. Regular-file additions (mode `100644`/`100755`) pass. Reuses the pre-existing `origin/main...HEAD` reference already relied on by the name-only check above it — no new ref-resolution risk.
- **Unit 2a — semantic shape tightening** (`schema.ts`): `requireNonNegativeInt` (whole, `>= 0`) applied to `execution.pr`, `execution.attempts.*`, `rounds.count`; `requireDateString` (`^\d{4}-\d{2}-\d{2}$`) applied to `office_hours.since`. Both build on `requireNumber`/`requireString`. No live node violates the new rules — `validate-graph` reports `ok — 219 nodes`.
- **Unit 2b — tactic-body-loss guard** (`store.ts`): `assertNoTacticBodyLoss` throws when a rewrite changes an existing tactic's `kind` away from `tactic` while its body is non-placeholder, preventing silent plan loss; a still-placeholder body reclassifies freely.

### Adversarial checks that passed

- **No live caller changes a node's kind in place**: `graph-commit` round-trips via `readNode` (kind preserved), `read-sensors` spreads the existing node (kind preserved), `write-node.ts` passes operator JSON — exactly the guarded path. So the new throw cannot break an existing autonomous flow.
- **At most one file read** in `writeNode`: `assertNoTacticBodyLoss` and `readExistingTacticBody` run in mutually exclusive `kind` branches.
- Awk destination-mode field indexing confirmed against the raw-diff format; deletions (dst mode `000000`) correctly ignored.

### Verification (merged tree)

- `npm test --prefix packages/intentionsutil` → **161/161** passing (8 files; +3 from #2778's merged store tests).
- `npx tsc -p packages/intentionsutil --noEmit` → clean.
- `npx tsx packages/intentionsutil/scripts/validate-graph.ts intentions` → `ok — 219 nodes`.

### Custody + freshness

- **Freshness**: this node's `execution.strategy_fingerprint` is `null` (bootstrap — stamping not started); null is never stale → gate passes.
- **Scope custody**: branch base `d9243d8b`. Node-file commits on `origin/main` since base are only `61b56d7d` (implement→qa) and `38a9cc3c` (qa→review, `qa-done` marker + qa-residue body note) — both are this node's own dispatch lifecycle transitions (the machinery that advanced it into review), not scope/plan re-authoring. Statement, benefit, conditions, serves, and plan units are untouched. Custody intact.

### Deferred (optional, non-blocking)

- **DRY the placeholder-body string** (`store.ts:46` writer vs `store.ts:68` guard both hardcode `` `# ${statement}\n` ``). Drift-prevention, not a bug: if the placeholder format ever changes in `writeNode` alone, the guard would stop recognizing placeholders and start throwing on legitimate reclassification of never-planned tactics. A shared `placeholderBody(statement)` helper would couple them.

PR #2752, opened 2026-07-04
intentionsutil: resolveAttention as a weighted term registry

natb1:
## Summary

Implements `tactic-calculated-attention` from the `tactic-graph-native-dispatch` subtree (intentions/tactic-calculated-attention.md), executed as the next ready on-path unit in strategy-graph-native-dispatch's round 1.

Restructures `resolveAttention` from a single additive-flow algorithm into a term registry composed as a weighted sum, per strategy clarification 11 on `strategy-graph-native-dispatch`:

- **authored** — the existing v2 boost/override flow, unchanged (`parent`/`serves` propagation, undecayed/undiluted).
- **signal** — structural: a node is on-path iff it (transitively) blocks — reachable by walking `blocked_by` in reverse, or inherits down a `parent` chain — a validates-terminal (a tactic with a `validates` edge to an unvalidated strategy, or such a strategy itself).
- **capture** — from the node's own `recovers` (strategies) or its serving strategy's `recovers` (tactics via `serves`): each resolved delegation contributes a normalized divergence/irreversibility axis score.

An authored `override` still pins the total absolutely — derived terms never touch it. `ResolvedAttention` gains an optional `terms` field exposing the per-term breakdown, and `renderFrontier` surfaces it as a `{term v, term v}` suffix when more than one term contributes, for explainability.

Manually verified against the live store: `tactic-align-init-skill`, `tactic-graph-commit-hardening`, and `tactic-graph-write-validation-hardening` (off-path — no `blocked_by` chain to a validates-terminal) rank below their on-path siblings in the same subtree, with no backlog flag anywhere in the store.

## Test plan

- [x] `npm test --prefix packages/intentionsutil` — 149 tests pass (existing suite + new signal/capture/composition tests)
- [x] `npx tsc --noEmit -p packages/intentionsutil` — clean
- [x] lint (`eslint src/`) — clean
- [x] Manual: `npx tsx packages/intentionsutil/scripts/frontier-view.ts` and a direct `resolveAttention` check against the live `intentions/` store confirm the documented on-path/off-path ranking for the tactic-graph-native-dispatch subtree

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2778, opened 2026-07-05
tactic-graph-commit-hardening: surface gh errors, no deterministic-failure retry, park_write atomicity, id-validation fix

natb1:
Graph node: tactic-graph-commit-hardening

Hardens `packages/intentionsutil/scripts/graph-commit` per the node plan (deferred robustness findings from the 2026-07-04 independent review of PR #2750; the in-scope conflict data-loss finding shipped earlier in PR #2751, whose fail-closed path this builds on).

## Unit 1 — surface check/API errors; stop retrying deterministic failures

- `await_checks()` no longer discards gh stderr (`2>/dev/null … || counts=""`). It captures stderr per poll and, after 3 consecutive gh failures (auth/rate-limit/network), dies with the captured error text instead of collapsing them into the generic ~180s timeout.
- Return states are now distinguished: `0` green, `1` timeout-pending (transient — caller rebases/retries), `2` a required check CONCLUDED non-success (deterministic).
- `try_land()` dies immediately on a concluded check failure — the content itself fails CI and every retry would reproduce it — instead of burning the remaining attempts and printing the misleading "main busy … retry later" message.

## Unit 2 — park_write atomicity, id validation, signal traps

- `park_write()` tsx helper is two-pass: pass 1 `readNode`s every id (fail fast before any write), pass 2 writes all — a mid-loop throw can no longer leave earlier ids mutated on disk, uncommitted.
- Id validation rejects only the exact ids `.` and `..` plus the existing `/` and `\` bans — `..` as a substring cannot traverse once separators are banned. Ids like `v1..v2-migration` now land end-to-end. Making that work also required sanitizing the scratch-branch name (git refs reject `..` sequences), mapping non-`[A-Za-z0-9-]` chars to `-`; the name is cosmetic and the PID suffix keeps uniqueness.
- `trap 'exit 130' INT TERM` so the EXIT trap (snapshot removal, remote scratch-branch delete) also runs on signal death.

The conflict/park flow from PR #2751 is untouched.

## Test harness committed

`packages/intentionsutil/scripts/test-graph-commit.sh` recreates the PR #2751 verification harness (bare origin + two writer clones, `gh`/`npx` PATH shims, `GRAPH_COMMIT_*` overrides) so it stops living in job-scratch dirs, extended with the two new cases the plan asks for: concluded-check-failure (gh shim returns `3 1` — immediate die, exactly 1 poll, no retry burn) and gh-hard-failure (shim exits 1 — die with the shim's stderr surfaced after exactly 3 polls). No network or node needed.

## Verification

- `bash -n packages/intentionsutil/scripts/graph-commit` — clean.
- `bash packages/intentionsutil/scripts/test-graph-commit.sh` — 15/15 assertions pass (happy path, idempotent re-run, non-overlapping auto-merge, overlapping conflict fail-closed park + snapshot preservation, concluded-failure no-retry, gh hard-failure stderr surfacing, pending-timeout transient retry, `..`-substring id end-to-end, `/` `\` `.` `..` rejections, no leftover `graph/**` scratch branches).
- `lint-prose-rules.sh` — pass.


natb1:
## QA — tactic-graph-commit-hardening (autonomous pass)

**Merge:** synced `origin/main` (clean fast-merge, no conflicts, pushed).

### QA plan (from the tactic's stated intent, not the implementer's checklist)

The tactic promises four consumer-visible behaviors:
1. `await_checks`/`try_land` surface real `gh` errors instead of collapsing them into the generic timeout, and stop retrying a *concluded* check failure.
2. `park_write` is atomic: a bad id among several fails before any write.
3. Id validation accepts `..`-substring ids (e.g. `v1..v2-migration`), rejecting only the exact ids `.`/`..`.
4. `INT`/`TERM` during the poll loop still runs cleanup (scratch-branch delete).

Triage: all four are script-verifiable end-to-end without a browser. Executed each independently against **real** `gh`, real `git`, and real `store.ts` — not just the PR's own `test-graph-commit.sh` harness (which shims both `gh` and `npx`/`store.ts`, so it can't catch a defect that lives on the other side of those shims).

### Validated with real-data evidence

- **Real `gh` + real API data**, sourcing the actual script and calling `await_checks` directly (no shim):
  - Real `HEAD` SHA on `origin/main` (genuinely green, 4/4 checks) → `await_checks` returns 0.
  - Real `gh api` 422 against a bogus SHA → captured stderr, 3 consecutive-failure die with gh's real message surfaced verbatim. Matches the documented contract exactly.
- **Real signal handling**, end-to-end against a throwaway bare-origin + real `git push`/`gh` (only `gh` unreachable, which is itself a real failure path, not a shim): `timeout --preserve-status -s INT 5 … graph-commit …` during the poll sleep → exit 130, and the scratch branch (`graph/sig-test-<pid>`) was genuinely deleted on the throwaway origin. Repeated for `SIGTERM` with the same result. (My first attempts used manual `kill -INT $PID`/`-$PGID` and appeared to hang — that turned out to be a quirk of manual signal delivery in this sandbox, not the script; `timeout`-delivered signals confirm the trap fires correctly.)
- **Real `park_write` two-pass atomicity**, running the exact embedded tsx snippet against real `store.ts` (not the harness's bash shim): a good id + a nonexistent id → the good id's `office_hours` stayed `null` (no partial write) and the process exited nonzero. A second run with two real, valid ids showed both get parked correctly with the tactic body preserved verbatim (real YAML round-trip, not the shim's flat-text stand-in).
- **The PR's own `test-graph-commit.sh`** (15 assertions) also passes as the floor.
- `bash -n` syntax check passes.

### Fix made: cross-file id-validation mismatch (opus-fixable, fixed and pushed)

**Found via the real-data pass, invisible to the checklist re-run** (same shape as PR #2752's capture-term bug): the tactic's Unit 2 claims the graph-commit id-validation relaxation "unblocks ids like `v1..v2-migration`." That's false as delivered. `write-node.ts` — the documented mandatory gate every writer goes through *before* calling graph-commit — routes through `packages/intentionsutil/src/store.ts`'s `assertPathSafeId`, which still rejected any `..` *substring*. So a node with such an id could never be created via the real workflow; graph-commit's own relaxed check was reachable only for ids that had (impossibly) already been written.

Confirmed directly: `writeNode(dir, {id: 'v1..v2-migration', ...})` via the real (non-shimmed) `store.js` threw before the fix, succeeds after.

**I expanded scope beyond the plan's stated "graph-commit only"** to `packages/intentionsutil/src/store.ts`, because store.ts *is* the gate — there's no in-scope-only fix that delivers the claimed capability. This relaxes a path-traversal guard (a security-relevant check) in a file the plan didn't name, so flagging it explicitly here for scrutiny on its own merits rather than folding it in silently:

- Changed `assertPathSafeId` to the identical policy already in graph-commit's bash: reject `/` and `\` unconditionally, reject `.`/`..` only as the *exact* id. This does not weaken the actual protection — once separators are banned, a `..` substring in a single path component can't traverse (it's never joined as `../..`, only used as `<dir>/<id>.md`). The existing `../evil` traversal test still throws (verified).
- Added a dedicated `describe("id path safety", …)` block in `store.test.ts` pinning the exact accept/reject boundary against real `store.ts`, so the boundary can't silently regress again (the harness's Case 8 tests the same acceptance but through the shim, so it doesn't protect this file).
- Full suite: `npx vitest run --project packages/intentionsutil --root .` → 155/155 pass (152 pre-existing + 3 new).
- `run-lint.sh` and `run-typecheck.sh --app packages/intentionsutil` both pass.
- Commit `8a3e8ef5`, pushed to `tactic-graph-commit-hardening`.

### Residue

- **needs_main (soak item, named by the plan itself as the ongoing verification)**: Unit 1's "concluded check failure → immediate die, no retry" is verified via the harness's `gh` shim (returns `"3 1"`) and via my own real-`gh`-green / real-`gh`-hard-failure checks, but a genuine *concluded non-success* on a real scratch SHA was not exercised against live CI in this pass (would require deliberately landing broken content on a `graph/**` branch). Per the plan's own Verification section ("live `graph-commit` use continues as the ongoing soak"): on the next real `graph-commit` run where a required check concludes non-success on the scratch SHA, confirm exit 1 with "a required check concluded non-success … not retrying" logged, and no retry-burn / no "main busy" message.
- Everything else in the tactic's four units is confirmed working against real data, not just satisfied per the implementer's own tests.

natb1:
## Review — tactic-graph-commit-hardening (terminal gate)

**Disposition: clean — auto-merge armed.** No Required findings; no deferred findings. Prior QA fix (`8a3e8ef5`, store.ts id-validation alignment) verified independently below.

**Merge:** synced `origin/main` (clean, pushed `97a67bb6`).

### Correctness review (surfaced, then adversarially refuted)
- **`await_checks` return-code split (0 green / 1 pending-timeout / 2 concluded-fail) + die-after-3-consecutive-gh-fails.** Logic sound: `gh_fails` resets on any successful poll so the die requires 3 *consecutive* real gh failures (auth/rate-limit/network), surfaced verbatim; a concluded non-success returns 2 and `try_land` dies without burning retries. Considered whether a required check concluding `cancelled`/`timed_out`/flaky could false-trigger the no-retry die: within a single `await_checks` call the polled SHA is fixed and PID-scoped, so it is never self-cancelled by a sibling push; a genuinely flaky conclusion dying immediately is the tactic's *stated intent* ("no deterministic-failure retry"), with recovery = re-run (rebases to a fresh SHA, re-stamps). Not a defect. **Refuted.**
- **`park_write` two-pass atomicity.** Reads every node before any write, so an unreadable id fails before mutating disk. Scoped to read failures (mid-loop write failures are out of scope and not claimed). Correct improvement. **Refuted.**
- **id-validation relaxation (allow `..` substring; reject `/`, `\`, and exact `.`/`..`) in both `graph-commit` and `store.ts`.** Security-relevant: independently confirmed both path-constructing consumers (`writeNode`, `readNode`) call `assertPathSafeId` before `join(dir, id + ".md")`. With `/` and `\` banned, a `..` substring in a single path component cannot form a `../` traversal segment. `../evil`, `a/b`, `a\b` still rejected. The two policies are now byte-identical, so graph-commit's relaxed check is actually reachable (write-node.ts routes through store.ts first). **Refuted.**
- **`INT`/`TERM` → `exit 130` trap** installed after the `EXIT` cleanup trap: signal death now runs `cleanup()` (snapshot removal + scratch-branch delete). Correct. **Refuted.**
- **Scratch-ref name sanitization** (`${IDS[0]//[!A-Za-z0-9-]/-}`): cosmetic ref-name only (PID gives uniqueness); node ids like `v1..v2-migration` produce a git-legal ref. **Refuted.**

### Verification (all green)
- `test-graph-commit.sh` — 15/15 (happy/idempotent/non-overlap-merge/overlap-park/concluded-fail/gh-hard-fail/pending-timeout/id-validation).
- `npx vitest run --project packages/intentionsutil --root .` — 155/155.
- `run-typecheck.sh --app packages/intentionsutil` — pass. `run-lint.sh` (incl. prose-rule lint) — pass. `bash -n graph-commit` — pass.

### Custody + freshness
- **Freshness:** `execution.strategy_fingerprint: null` (bootstrap, never stale) → gate passes.
- **Custody:** one commit touched the node file on `origin/main` since branch base `5a8beb6d` — `38a9cc3c`. Its node-file diff is a phase transition (`qa`→`review`), execution `attempts`/`markers` bookkeeping, and an appended `## main-qa residue` soak note. No scope-bearing edit (statement / Scope-per-unit / plan bodies / conditions unchanged). Custody intact.

Armed `--auto --squash`; the merge lands out-of-band and the next tick's sweep absorbs it.

PR #2790, opened 2026-07-07
graph-commit: add --prune support and --base compare-and-swap freshness check

natb1:
Implements tactic-graph-commit-prune-support: two units adding to the sanctioned graph write path (graph-commit) so it can express two operations it could not before.

## Unit 1 -- `--prune <id>` support

graph-commit's existence guard previously required `intentions/<id>.md` to be present on disk for every id, so a prune (a `phase: done` node deletion) could not land through graph-commit -- callers had to hand-orchestrate the `graph/**` fast-path push instead. `--prune <id>` (repeatable, mixable with ordinary positional ids) inverts the guard for that id: the file must be absent on disk but present in HEAD. An edit id and a prune id in the same invocation land in ONE commit.

## Unit 2 -- `--base <id>=<blobsha>` compare-and-swap freshness check

The read-modify-write flow (readNode dump -> patch -> write-node -> graph-commit) carried no record of the base the editor read; a stale dump relied on textual rebase luck to avoid clobbering concurrent state. `--base <id>=<blobsha>` (repeatable, or a manifest file of one pair per line) is optional and opt-in: after a fetch, graph-commit refuses the write if origin/main's blob for that path no longer matches what the caller read, before any local commit is made. Sessions passing no `--base` see no change in behavior.

## Verification

Adds `packages/intentionsutil/scripts/test-graph-commit.sh`, a functional harness (throwaway bare origin + writer clones + a `gh` check-run shim) with 5 cases: mixed edit+prune lands in one commit, the still-present-on-disk prune guard rejects, a fresh `--base` lands, and a stale `--base` refuses. All 5 pass locally.

```
bash packages/intentionsutil/scripts/test-graph-commit.sh
```

## Coordination note

PR #2778 (tactic-graph-commit-hardening) touches the same script and is still open; this tactic's plan anticipated the overlap ("Coordinate mechanically with #2778 -- whichever merges second rebases"), so `test-graph-commit.sh` is newly created here rather than extended -- it did not exist on this branch's base. The rebase reconciliation between the two harnesses is deferred to whichever PR lands second.

---

Graph-native context: implements node `tactic-graph-commit-prune-support` (serves `strategy-graph-native-dispatch`). Router will land the implement -> qa transition after this PR opens.


natb1:
## QA (autonomous, 2026-07-07)

**Scope:** independent user-acceptance QA of `--prune` (Unit 1) and `--base` compare-and-swap (Unit 2) against the node's plan and this PR's description.

**Merge reconciliation.** The branch was 6 commits behind `origin/main`, which had absorbed PR #2778 (`tactic-graph-commit-hardening`) — a genuine conflict on `packages/intentionsutil/scripts/graph-commit` and an add/add conflict on `packages/intentionsutil/scripts/test-graph-commit.sh`, exactly the coordination this PR's description flagged as "deferred to whichever PR lands second." Resolved both:

- `graph-commit`: kept this branch's `validate_id_chars()` refactor (shared between ordinary and `--prune` ids) but folded in #2778's id-validation bug fix (`.`/`..` rejected only as *exact* ids, not as a substring, so `v1..v2-migration`-style ids stay legal) — the pre-merge version of this branch still had the old buggy substring check. Also fixed the scratch-branch ref-name sanitizer to key off `ALL_IDS[0]` (not `IDS[0]`) so a pure-`--prune` invocation with no ordinary ids still gets a valid ref name.
- `test-graph-commit.sh`: the two PRs had independently authored non-overlapping harnesses (#2778's 8 cases: happy path, idempotent re-run, concurrent merge, conflict/park, concluded-check-failure, gh-hard-failure, pending-timeout, id-validation; this PR's 4 cases: prune+edit same-commit, prune-guard, base-fresh, base-stale). Merged into one 12-case harness rather than picking one side.

**Independent verification, real data:**
- Ran the merged `test-graph-commit.sh` end-to-end against a scratch bare origin + real writer clones (no mocks beyond a `gh`/`npx` PATH shim standing in for the GitHub API): **20/20 PASS**, covering every #2778 hardening case plus both new prune cases and both new base-freshness cases (fresh lands, stale is refused with the concurrent writer's content surviving on main).
- `bash -n` on both scripts, `npx vitest run --project 'packages/intentionsutil' --root .` (161/161 tests), `run-typecheck.sh --app packages/intentionsutil`, and the repo lint suite (`run-lint.sh`, including the prose-rule linter for the jq/echo trap) — all clean.
- Read the retry/rebase-conflict path (`land()` / `park_write()`): confirmed the plan's "snapshot restore must handle prune-id absence on retry" concern is satisfied by a simpler mechanism than literally described — `snapshot()` deliberately does not copy prune ids (documented rationale: the deletion is part of the committed diff itself, so a rebase replays it like any other commit content; there is nothing external to restore). Also confirmed that if a `--prune` id hits a genuine rebase conflict, the parking fallback correctly falls back to *not* deleting the node (the reset-to-origin/main brings the file back if it still exists there) and flags it `office_hours` for manual reconciliation, rather than silently completing a prune whose concurrent-edit context was never actually reconciled.
- Pushed the merge commit (`1ff30451`); all 18 required + informational CI checks are green on the merged SHA, including `acceptance`, `lint`, `preview-and-smoke`, `unit-tests`.

**Residue:** none. No opus-fixable defects found, nothing needs-main, nothing needs-human. The one plan-text deviation (prune-id snapshot handling) is a documented, verified-correct simplification, not a gap.

Recommending qa-done.

natb1:
**QA follow-up:** on advisor review, found the prune coverage above always paired `--prune` with an ordinary edit id, so the actual owed-prune-backlog shape (a pure `--prune` with `IDS` empty — no accompanying edit) was never exercised, nor was the `--base` manifest-file form. Added two more cases (pure-prune-lands, base-manifest-file) — harness is now 23/23 pass (`b69df94e`). All CI green on the updated SHA. Residue: none.

PR #2802, opened 2026-07-10
Self-hosted webmention endpoint for landing and fellspiral

natb1:
Implements the graph tactic `tactic-indieweb-audience` (serves `strategy-own-audience`): a self-hosted webmention receiver on the repo's existing Blaze Firebase project, advertised from both owned publication surfaces. Received mentions are stored, not displayed (display needs a moderation story and is out of scope).

## Unit 1 — webmention receiver function
- New `functions/src/webmention.ts` exporting an `onRequest` v2 function `webmention`, modeled on `feed-proxy.ts`. Per the W3C Webmention receiving rules:
  - POST only (405 otherwise); `application/x-www-form-urlencoded` body with `source` + `target` (400 on missing params, non-URL values, non-http(s) schemes, `source === target`).
  - `target` host must be `commons.systems` or `fellspiral.commons.systems` (400 otherwise).
  - No AppCheck — external senders cannot carry a token, so this endpoint is deliberately public. Abuse is bounded by a source-fetch byte cap, a 10s timeout, and `maxInstances: 2`.
  - Verifies synchronously: fetches `source` (redirect-following, per the receiving rules) and requires the body to reference `target`; 400 otherwise.
  - Stores `{source, target, receivedAt}` to `webmentions/prod/mentions/{id}`, where `{id}` is a sha256 of `source|target` (idempotent re-sends overwrite).
- Exported from `functions/src/index.ts`.
- `firestore.rules`: deny-all `match /webmentions/{env}/mentions/{id}` (only the admin-SDK function writes).
- Unit tests in `functions/test/webmention.test.ts` (param validation, host allowlist, source-verification success/failure, idempotent id).

## Unit 2 — endpoint advertisement and hosting rewrites
- `firebase.json`: `/api/webmention` rewrite on the `landing` and `fellspiral` targets, before the SPA catch-all.
- `rel=webmention` head link (absolute URL) in `landing/index.html` and `fellspiral/index.html`.
- E2e head-link smoke specs in `landing/e2e/` and `fellspiral/e2e/`.

## Notes for review
- SSRF residue: the receiver fetches arbitrary `source` URLs (redirect-following) to verify the mention. Per the plan, abuse-bounding is scoped to byte-cap + timeout + `maxInstances`; IP-range blocking of link-local/private targets (e.g. 169.254.169.254) was deliberately out of plan scope. Flagging for the review/security phase.
- Manual/QA is post-deploy: POST an empty body to the deployed endpoint expecting 400; a first-of-kind function deploy may need owner IAM.

## Verification
- `npx vitest run --project functions --root .` — 92 passed.
- `run-typecheck.sh --app functions` — passed.


natb1:
## Review disposition — CLEAN (auto-merge armed)

Combined `/review-fix` fan-out (code-review + security surface) run on this PR by path. Freshness gates green, merge-custody chain (implement -> qa -> review) verified unbroken against the merge-time tree, all 22 CI checks green.

**Required findings: none.** Correctness is clean and the unit suite (`functions/test/webmention.test.ts`) covers method/content-type/param/URL-scheme/host/self-mention rejections, fetch-failure, non-ok, no-link, the 202 store path, idempotency, and `mentionId` stability.

**Security — public unauthenticated endpoint.** The endpoint correctly restricts `target` to the two owned hosts, rejects non-http(s) schemes, verifies the source actually links to `target` before storing, denies all client Firestore access (`webmentions/{env}/mentions` `allow read, write: if false` — server SDK bypasses), documents the deliberate no-AppCheck decision, and bounds abuse with a 10s timeout, a 5 MB body cap, and `maxInstances: 2`.

One defense-in-depth observation (non-blocking, inert in the current deployment): unlike `feed-proxy` (which blocks redirects as an SSRF guard), the source fetch follows redirects and does not reject destinations that resolve to private/link-local IPs or the metadata endpoint. In the current Cloud Functions v2 (Cloud Run) deployment this is inert — no VPC egress connector (RFC1918 unroutable), the GCP metadata server requires a `Metadata-Flavor: Google` header this fetch never sends, and the fetched body is never reflected to the sender (the response is a fixed 202/400). It would only become relevant if a VPC egress connector is ever added; flagged here for future hardening rather than filed as a standing follow-up.

Custody: all product code authored by the dispatch implementer unit commits; the branch's merge commits integrated origin/main without touching any product file.

Marker: `dispatch:reviewed`.

PR #2807, opened 2026-07-10
tactic-recover-publishing-reading: audit-publishing instrument

natb1:
Graph tactic `tactic-recover-publishing-reading` (serves/validates `strategy-recover-publishing`).

## What

Adds `packages/intentionsutil/scripts/audit-publishing.ts`, a report-only instrument that produces the machine-verifiable half of the `strategy-recover-publishing` signal reading:

- Enumerates each owned surface's published pieces from its live RSS feed (`landing` -> `https://commons.systems`, `fellspiral` -> `https://fellspiral.commons.systems`).
- Verifies each piece is publicly readable with no account: unauthenticated GET -> HTTP 200 with the piece's title present in the prerendered body.
- Prints the platform-first attestation checklist only the owner can complete.
- Exit 0 when every piece passes; exit 1 when any fails (the failure list is the gap evidence). Fatal error naming the URL on an unreachable / non-200 / zero-item feed.

Pure core (`auditSurfaces`/`parseFeedItems`/`unescapeHtml`/`formatReport`) is separated from IO via an injected `fetchText`, so the test suite runs with no network.

## Boundaries

- Report-only: does NOT write `reading`/`gap` onto any node — the owner stamps the strategy node at office-hours.
- Deliberately NOT registered in `read-sensors.ts` (local-first, no-network by design; this fetches live production URLs).

## Verify

- `npx vitest run --project packages/intentionsutil --root .` — 226 passed
- `npx tsc -p packages/intentionsutil --noEmit` — clean

### DIFF
--- a/packages/intentionsutil/scripts/audit-publishing.ts
+++ b/packages/intentionsutil/scripts/audit-publishing.ts
@@ -22,6 +22,11 @@
 
 import { pathToFileURL } from "node:url";
 
+/** Narrow an unknown thrown value to a message string without a type cast. */
+function errMessage(err: unknown): string {
+  return err instanceof Error ? err.message : String(err);
+}
+
 // --- Surfaces --------------------------------------------------------------
 // Hardcoded owned surfaces. Values must match `landing/src/site-config.ts:6`
 // and `fellspiral/src/site-config.ts:6`.
@@ -130,7 +135,7 @@ export async function auditSurfaces(
       feed = await fetchText(feedUrl);
     } catch (err) {
       throw new Error(
-        `audit-publishing: failed to fetch feed ${feedUrl}: ${(err as Error).message}`,
+        `audit-publishing: failed to fetch feed ${feedUrl}: ${errMessage(err)}`,
       );
     }
     if (feed.status !== 200) {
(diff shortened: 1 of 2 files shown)


natb1:
## Review disposition — clean

Combined review fan-out (code-review; no security domain applies — a report-only
instrument doing unauthenticated outbound GETs, no secrets/inputs/writes) over
the 2-file diff (`audit-publishing.ts` + tests).

**Verdict: clean.** All CI green on HEAD; 12/12 unit tests pass. Pure core cleanly
separated from IO, fatal-vs-finding error handling matches the plan and
`code-style.md` (clear errors over fallbacks), entity-unescape order correct
(`&amp;` last), exit-code semantics sound. Verified the instrument produces an
honest reading against the actual live corpus — all four published titles
(landing: "Recovering Autonomy with Coding Agents"; fellspiral: "Disciplinary
Review Operations", "Scenes from a Hat", "The Surreal: Returning to the
Darkness") are plain-ASCII, so the no-account check passes on current data.

**Deferred (low severity, non-blocking, no current trigger):** the no-account
check `resp.body.includes(item.title)` compares the HTML-*unescaped* feed title
against the raw prerendered HTML body. `marked` escapes `&`/`<`/`>` in the H1
(and React SSR escapes them elsewhere), so a future published title containing
one of those characters (e.g. "Q&A", "Tips & Tricks") would false-negative —
flagging a genuinely readable piece as `no-account-readable: no` and exiting 1,
i.e. spurious gap evidence. Zero impact on the current corpus. Suggested fix when
touched: also match `escapeHtml(item.title)`, or unescape the fetched body before
the substring test. Recorded here as a deferred robustness follow-up (owner reviews
the report at office-hours, which catches such a case in the interim).

No blocking findings. Auto-merge armed.

PR #2809, opened 2026-07-10
resolveAttention: backward blocked_by critical-path propagation

natb1:
Implements `tactic-attention-critical-path-propagation` (graph-native tactic node).

## What

Extends the `authored` term in `resolveAttention`
(`packages/intentionsutil/src/attention.ts`) so an authored source-set flows
backward along `blocked_by` to a node's blockers, recursively — not only
downward via `parent`/`serves`. A boost/override on a blocked node now lifts
its whole critical path, per the 2026-07-07 `strategy-graph-drives-dispatch`
clarifications. Fixes the live defect where boosting
`tactic-attention-surface-velocity-pace` could not reach its blockers
`tactic-attention-surface-signal-types` / `tactic-attention-surface-graph-read`.

## How

- Widen `distributors(c)` with the reverse `blocked_by` relation (every node X
  with `c ∈ X.blocked_by` distributes to c), built from a `reverseBlockers`
  index like `computeSignalPath`.
- Convert the authored term from a memoized DFS to a monotone fixpoint: the
  widened relation admits legitimate mixed `parent`/`serves`/`blocked_by` cycles
  (e.g. a node blocked by a tactic inside its own subtree) that the old
  cycle-throw rejected. Unions only grow and override outputs are constant, so
  convergence is guaranteed; sorted-id sweeps keep the result input-order
  independent.
- Retain a narrow guard that still throws `IntentionSchemaError` on a pure
  `parent`-edge cycle (a self-ancestor that `validateGraph` does NOT catch —
  rule 15 rejects only `blocked_by` cycles) rather than silently converge it to
  0. This is a deliberate deviation from the plan's literal "drop the cycle
  throw" wording: dropping it entirely would break the existing pure-parent
  cycle-guard test (which Unit 2 did not ask to modify) and silently accept a
  malformed graph.

## Tests

Unit 2 adds five cases: two-hop backward flow, subtree inheritance off a
blocker, override backward flow, mixed-cycle convergence, and single-count
dedup across a `serves` + backward `blocked_by` edge.

## Verification

- `npx vitest run --project packages/intentionsutil --root .` — 236 pass.
- Typecheck passes.
- Manual: a hypothetical boost of 7 on
  `tactic-attention-surface-velocity-pace` lifts it and both transitive
  blockers by the identical amount; no other node moves.


natb1:
## QA — pass

Independent QA against the tactic's recorded intent and live data (not just re-running the implementer's own checks):

**Merged origin/main** into the branch (pulled in `tactic-graph-self-consistency-sweep`, blog h-entry work, and other unrelated main churn) and re-verified.

**Code review against intent** (`strategy-graph-drives-dispatch` 2026-07-07 clarifications): the backward `blocked_by` distribution, monotone fixpoint conversion, and narrowed pure-parent-cycle guard all match the recorded semantics. The PR's own deviation (keeping a throw on pure-`parent` cycles instead of dropping the cycle guard entirely, per its body) is correct — dropping it would silently accept a malformed graph that `validateGraph` does not catch.

**Test verification**: all 5 planned Unit 2 cases present and correct (two-hop backward flow, subtree inheritance off a blocker, override backward flow, mixed-cycle convergence, single-count dedup). Full suite: 263/263 pass. Typecheck: pass.

**Independent live-graph verification** (not the implementer's own manual check — ran it myself from a fresh script): patched a boost of 7 onto `tactic-attention-surface-velocity-pace` and called `resolveAttention` over the real `intentions/` graph.
- Baseline: `tactic-attention-surface-velocity-pace` = 4.1667, `tactic-attention-surface-signal-types` = 4.1667, `tactic-attention-surface-graph-read` = 4.1667
- With boost: all three = 11.1667 (delta = 7.0 on each, identical)
- Exactly 3 nodes changed total — no other node in the graph moved.

This confirms the critical-path propagation works correctly on the live graph, exactly as the tactic's rationale describes (boosting the surface-velocity-pace node now reaches its blockers).

**Residue found and fixed** (opus-fixable, fixed inline): the branch's own diff introduced 4 net-new `as <Type>` casts and 1 net-new non-null assertion, tripping `type-safety-sensor` on push of the origin/main merge. Fixed in d4757345:
- 4 `Map.get(id) as T` casts in the fixpoint loop/compose step replaced with a `mustGet` helper that throws a descriptive `IntentionSchemaError` instead of silently casting past `undefined` — these lookups are guaranteed by construction (`byId`/`authoredOutgoing` are keyed from the same `nodes`/`sortedNodeIds`), so a clear invariant-violation error is strictly better than a defensive cast (`code-style.md`).
- 1 non-null assertion (`let result!: ReturnType<...>`) is the standard definite-assignment-before-synchronous-callback idiom, already used elsewhere in the codebase (pre-existing, so unflagged there) — suppressed with a `// type-safety-ok:` comment explaining why.

All CI checks green after the fix (re-verified via `gh pr checks --watch`).

No needs-main residue, no needs-human residue. Transitioning to review.

natb1:
## Review — clean, merge-ready

Terminal review pass over the `attention.ts` backward `blocked_by` critical-path propagation diff (run inline at Opus within the graph-tick Workflow agent; no separate subagent fan-out available in this context).

**Merge-custody: intact.** The `implement → qa → review` chain is unbroken against the merge-time tree. The only post-implement code commit (`d4757345`, type-safety `mustGet` residue fix) is the QA session's own fix and landed at 15:54:57Z, before the QA-pass sign-off comment at 16:00:51Z — within the qa phase, not a post-signoff author edit.

**Correctness (verified):**
- The monotone fixpoint converges: the universe of `(src, amt)` pairs is finite and `amt` is constant per `src` (each source's own boost/override), so unions grow only toward a fixed bound — convergence is guaranteed and input-order-independent via the sorted-id sweep.
- `reverseBlockers` correctly maps `blocker → [blocked nodes]`, so `distributorIds(c)` pulls in the nodes `c` blocks and `c` (a blocker) inherits their source-set backward along `blocked_by` — matching the recorded rationale (boosting a blocked node lifts its blockers).
- The narrowed pure-`parent`-cycle guard is correct: it walks the single `parent` pointer per node and throws only on a self-ancestor repeat (which `validateGraph` rule 15 does not catch), while legitimate mixed `parent`/`serves`/`blocked_by` cycles converge via the fixpoint.
- Override nodes are correctly seeded constant and skipped in the sweep; boost nodes correctly re-include their own claim and inherit from distributors — semantics match the prior v2 behavior plus the new backward term.
- No dangling references to the removed `computeAuthored`/`onStack`/`stackPath`/`distributors` symbols.

**Tests:** the 5 added cases cover two-hop backward flow, subtree inheritance off a blocker (backward/downward interleave), override backward flow, mixed-cycle convergence (regression vs the old DFS throw), and serves+backward-`blocked_by` single-count dedup. Full suite and all CI checks green.

**Findings:** none Required. No deferred follow-ups.

Both freshness gates green (phase `review`; serving-strategy fingerprint unstamped/null). Auto-merge arming is not authorized this tick — holding clean and merge-ready for the human to merge.

PR #2811, opened 2026-07-10
shared packages low-severity sweep (tactic-review-lows-shared-infra)

natb1:
Seven confirmed low-severity findings from the 2026-07-05 code review of the
load-bearing shared platform packages (budget, print/audio, office-hours consume
them; no single strategy owns them). Graph node: `tactic-review-lows-shared-infra`.
Each unit is one commit; units are independent (different packages).

- **Unit 1 — crypto**: `worker.onerror` now calls `worker.terminate()` before
  nulling the reference, so the PBKDF2 thread is reclaimed instead of leaked.
- **Unit 2 — idbutil**: the open-request `onerror` guards its `dbPromise = null`
  with a promise-identity check, so a stale/late error can no longer clobber a
  freshly memoized promise from a subsequent open.
- **Unit 3 — firestoreutil**: `QueryBuilder` copies the constraints array on
  construction and returns a new builder per `where`/`orderBy`/`limit`
  (immutable-per-call), so queries branched off one base no longer
  cross-contaminate.
- **Unit 4 — authutil**: `signOut()` re-throws after its `logError` +
  `showAuthError` side effects, so an awaiting caller does not proceed as
  signed-out while the session is still live. Test updated to the new contract.
- **Unit 5 — firebaseutil**: `RECAPTCHA_SITE_KEY` (module-scope `requireEnv`,
  threw at import for non-App-Check apps) becomes a lazy `getRecaptchaSiteKey()`;
  the four App Check consumers (budget/print/audio/landing) call it. Test updated.
- **Unit 6 — router**: the vanilla history router now drives re-rendering off the
  shared location store (`subscribe` + `navigate`), so a store navigation and a
  vanilla navigation cross-notify and React `useLocation` subscribers stay in
  sync; reuses the store's existing listeners mechanism.
- **Unit 7 — mediautil**: the local-folder scan loop narrows its bare
  `catch { continue }` to the `getFile()` I/O failure, so a consumer `toItem`
  mapper bug surfaces instead of silently dropping the file from the library.

## Verification

- `npx vitest run --project packages/router` (+ crypto/idbutil/firestoreutil/
  authutil/mediautil/firebaseutil package suites): 389 passed.
- Consuming-app suites (budget/print/audio/landing/blog): 2324 passed.
- `run-typecheck.sh`: all matched targets pass (audio/budget/print/landing skipped
  by the runner due to pre-existing origin/main baseline errors, unrelated to
  this change; firebaseutil itself typechecks clean).

Graph-native tactic; no issue is closed by this PR.


natb1:
## QA (2026-07-10)

Merged `origin/main` into the branch (up through 714d0350, the 5-tactic
router-reconciliation absorb commit) and independently re-validated all 7
units against the tactic's stated intent, not just the plan's own
`verify` blocks.

**Found and fixed one residue item beyond the implementer's own checks:**
`fellspiral/src/firebase.ts` still imported the removed `RECAPTCHA_SITE_KEY`
named export from `@commons-systems/firebaseutil/config` (Unit 5 renamed it
to a lazy `getRecaptchaSiteKey()` and updated audio/budget/landing/print, but
missed fellspiral — also an App Check consumer). This was a real compile
break (`TS2724: has no exported member named 'RECAPTCHA_SITE_KEY'`), masked
in the implementer's own `run-typecheck.sh` run because fellspiral is
excluded there by the runner's "origin/main has pre-existing typecheck
errors" skip rule, not because it was actually clean. Fixed to match the
other four consumers (`cb6d5f7d`) and confirmed with a direct
`tsc --noEmit -p fellspiral` (clean) plus the fellspiral vitest project
(all passing).

**Per-unit validation:**
- Unit 1 (crypto): `worker.terminate()` now called before nulling the
  reference on `onerror` — thread reclaimed, confirmed by reading the diff
  against `crypto.ts:55-61`.
- Unit 2 (idbutil): the `onerror` null-out is now guarded by promise
  identity (`if (dbPromise === thisPromise) dbPromise = null`), mirroring
  the existing `settledByTimeout` pattern as instructed. Correct by
  inspection; no dedicated regression test for the stale-onerror race was
  added (minor coverage gap, not a functional defect — existing suite still
  passes).
- Unit 3 (firestoreutil): went further than "copy the array" — `QueryBuilder`
  is now fully persistent (`where`/`orderBy`/`limit` each return a *new*
  builder via a private `.with()`), so two queries branched off one base
  cannot cross-contaminate under any call pattern. No dedicated branching
  regression test was added either (same minor gap as Unit 2).
- Unit 4 (authutil): `signOut()` now re-throws after `logError` +
  `showAuthError`. Checked every caller in the tree
  (`audio/NavControls.tsx`, `blog/create-blog-app.ts`, `print/main.tsx`,
  `authutil/deferred-app-auth.ts`, scaffolding template) — all are
  `void signOut()` fire-and-forget, none chain on resolution, so the
  contract change is safe per the plan's own manual-check instruction.
- Unit 5 (firebaseutil): see residue above — now fully correct across all
  5 App Check consumers (audio, budget, landing, print, fellspiral).
- Unit 6 (router): `createHistoryRouter` now subscribes to the shared
  location-store instead of listening to `popstate` directly, and routes
  vanilla-router link clicks through `storeNavigate()`. Existing
  `router.test.ts` popstate-dispatch tests still pass unmodified, confirming
  the store's lazy popstate listener covers that path transparently.
- Unit 7 (mediautil): `catch` narrowed to just `fileHandle.getFile()`;
  `config.toItem(...)` now runs outside the try, so a mapper bug surfaces
  instead of being swallowed. Matches scope exactly.

**Verification run (fresh, independent):**
- `npx vitest run` across all 7 touched shared packages + fellspiral:
  430 passed, 5 pre-existing skips (unrelated build-artifact tests), 0
  failures.
- `npx vitest run` across budget/print/audio/landing/blog: 2332 passed, 9
  pre-existing skips, 0 failures.
- `.claude/skills/dispatch-propagate/scripts/run-typecheck.sh`: all
  targets pass.
- Direct `tsc --noEmit -p <app>` on fellspiral/audio/budget/print/landing
  (bypassing the runner's baseline-skip): no firebase/recaptcha errors on
  any of them.
- Full CI on the branch (post-fix): all checks green, including
  `acceptance` and `preview-and-smoke`.

No test-integrity issues (no test skipped/weakened). No needs-human or
needs-main items. Clean pass — advancing to review.

natb1:
## Review (2026-07-10) — clean, merge-ready hold

Terminal review pass over the 7-unit shared-packages low-severity sweep. All CI green, `mergeable: CLEAN`. Ran the finders inline at Opus (no subagent-spawn tool inside this Workflow agent).

**Merge-custody: intact.** The last branch commit `cb6d5f7d` (fellspiral RECAPTCHA fix, 15:51:06Z) landed *before* the QA sign-off comment (16:03:57Z) and the `qa->review` graph transition (16:06:28Z) — it was a within-QA residue fix, not a post-sign-off hand edit. No commits landed after QA sign-off.

**Per-unit verification (against the merge-time tree):**
- **Unit 1 (crypto):** `worker?.terminate()` before nulling on `onerror` — thread reclaimed. Correct.
- **Unit 2 (idbutil):** stale-onerror clobber guarded by promise identity (`if (dbPromise === thisPromise)`), with `dbPromise = thisPromise` assigned synchronously after `new Promise` so the guard holds. Correct.
- **Unit 3 (firestoreutil):** `QueryBuilder` made fully persistent (`where`/`orderBy`/`limit` return new builders via `.with()`). Checked both callers — `blog/src/firestore.ts:48` (captures the chained return in `bounded`) and `authutil/src/groups.ts` (pure chain) — neither relies on in-place mutation. Safe.
- **Unit 4 (authutil):** `signOut()` re-throws after `logError`+`showAuthError`. All callers are `void signOut()` fire-and-forget. Adversarial note (refuted, non-blocking): with the global `unhandledrejection` handler (`errorutil/global-handler.ts`), a sign-out failure now logs twice (once `sign-out`, once `unhandledrejection`). Benign observability noise — post-failure state (still signed in) is correct, no crash, distinct operation tags. Not a Required finding.
- **Unit 5 (firebaseutil):** `RECAPTCHA_SITE_KEY` const → `getRecaptchaSiteKey()` getter defers `requireEnv` past import; `app-context.ts` imports `firebaseConfig` at module scope, so non-App-Check importers no longer throw at load, while the getter still throws when actually called with the env missing. No silent App Check bypass. Correct.
- **Unit 6 (router):** re-render now driven off the shared location-store `subscribe()`; link clicks routed through `storeNavigate()` which does `history.pushState` + synchronous `emit()`. Matches the old `pushState`+`nav.navigate()` behavior including same-path clicks (store `navigate()` does not dedupe). Correct.
- **Unit 7 (mediautil):** `try` narrowed to `getFile()`; `config.toItem(...)` moved outside so a consumer mapper bug surfaces instead of being swallowed. Correct.

**Finders:** code-review over the diff plus a security-conscious pass on the auth (signOut), crypto (worker), and App Check (recaptcha) surfaces. No Required findings survived adversarial verification.

**Disposition:** clean review, both freshness gates green (phase == review; strategy fingerprint null/never-stale), custody intact. `reviewed` marker applied. Auto-merge arming not authorized this tick — **holding for human merge.** The PR is merge-ready.

PR #2812, opened 2026-07-10
budget-etl: accept overlapping same-month balance anchors (keep latest as-of)

natb1:
## Context

Graph tactic `tactic-budget-overlap-anchor-merge`, serving `strategy-recover-finance`.

`dedupStatementData` (projects/budget-etl/pipeline.go) hard-errored when two statements shared a StatementID (`{institution}-{account}-{YYYY-MM}`) with different balances. The 2026-07 monthly sync failed on mutually consistent overlapping exports. Per strategy clarification 2, overlapping evidence is surfaced, not gated.

## Change

When two same-StatementID entries disagree on Balance:
- Keep the entry with the later `BalanceDate` (as-of); a nil `BalanceDate` loses to a non-nil one.
- Log one reconciliation line naming both source files, both balances, and the delta (operator's machine — amounts fine there).
- Both-nil or identical-instant disagreements stay an error (no basis to choose).
- First-seen output order preserved; a later replacement takes the earlier survivor's position.

Equal-balance duplicate drop and caller signatures unchanged.

## Verification

```
go -C projects/budget-etl test ./...
```

Extended `TestDedupStatementData` with: later as-of arrives second (replacement), later as-of arrives first (kept), nil-vs-set BalanceDate (both orders), both-nil still errors, identical-instant still errors, order preservation across a reconciled replacement.

### DIFF
--- a/projects/budget-etl/pipeline.go
+++ b/projects/budget-etl/pipeline.go
@@ -58,39 +58,83 @@ func parseStatementDir(dir string, disc parse.DiscoverOpts) (parsed []parsedFile
 }
 
 // dedupStatementData deduplicates a slice of StatementData by StatementID,
-// preserving first-seen order. When two entries share a StatementID but have
-// different Balance values, it returns an error naming both source files and
-// balances so the caller can surface a clear failure rather than emitting
-// conflicting records. Equal-balance duplicates are silently dropped.
+// preserving first-seen order. Equal-balance duplicates are silently dropped.
+//
+// When two entries share a StatementID but disagree on Balance, they are
+// overlapping observations of the same account-month anchor. The collision is
+// resolved by as-of date (BalanceDate): the later observation is kept and a
+// single reconciliation line naming both source files, both balances, and the
+// delta is logged (log output stays on the operator's machine, so amounts are
+// fine there). A nil BalanceDate loses to a non-nil one. When there is no basis
+// to choose — both BalanceDate nil, or the same instant — the disagreement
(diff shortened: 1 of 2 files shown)


natb1:
## Review disposition — CLEAN, merge-ready

Terminal review pass over the diff (`pipeline.go` +60/-16, `pipeline_test.go` +94). No unresolved Required findings.

**Correctness** — `dedupStatementData` now resolves same-`StatementID` balance collisions by as-of date via the extracted `laterAnchor` helper: later `BalanceDate` wins, nil loses to non-nil, and both-nil / identical-instant disagreements stay a hard error. The `idx` map keyed to output positions correctly preserves first-seen order on in-place replacement, and chains correctly through three-way collisions (each incoming entry compares against the current survivor). Equal-balance duplicates stay silently dropped per spec.

**Tests** — the `TestDedupStatementData` table covers every case the plan named: later-as-of-second (replacement), later-as-of-first (kept), nil-vs-set both directions, both-nil error, identical-instant error, and surrounding order preservation. Genuine assertions, no weakening.

**Surface** — local operator Go CLI; logged reconciliation amounts are on the operator's machine (explicitly in scope per the plan). No auth/crypto/rules/external-I/O surface warranting a security finder.

**Checks** — all 22 CI checks green; `go -C projects/budget-etl test ./...` passes locally. Merge-custody intact (single implement commit, no post-qa author edit). Both freshness gates green.

Held at review with the `reviewed` marker for human merge.

PR #2818, opened 2026-07-10
audio player: cancel position-persist throttle + safe last-track removal

natb1:
Graph node: `tactic-audio-position-persist` (serves `strategy-recover-attention`).

Two verified playback-state defects (2026-07-05 review) in the owned audio player.

## Unit 1 — cancel the position-persist throttle on track change/stop

The `timeupdate` position-persist throttle (`audio/src/player.ts`) was never cancelled on track change or `stop()`. A timer armed while track A played could fire later and persist A's `currentTime` against the now-current track B, so the next session restored B to A's position. Added `cancelPositionThrottle()` and call it in `playTrack` (track change) and `stop()`.

## Unit 2 — stop, not restart, when removing the playing last track

Removing the currently-playing last-in-queue track wrapped to index 0 (`idx < queue.length ? idx : 0`) and force-played the first track from 0:00. Now stops playback when the removed slot has no following track.

## Tests

Added to `audio/test/player.test.ts`:
- throttle cancelled on track change (no stale-position persist)
- throttle cancelled on stop
- throttle still fires normally while a single track keeps playing (regression guard)
- removing the playing last track stops rather than wrapping to track 0

`audio` typecheck and the full `player.test.ts` suite (43 tests) pass locally.


natb1:
## QA summary (qa 2026-07-10)

**Scope**: Two verified playback-state defects in the audio player.

**Unit 1 — throttle cancellation on track change/stop.** Reviewed `cancelPositionThrottle()` and its call sites in `playTrack` and `stop()` (`audio/src/player.ts`). Confirmed the throttle var is cleared before the new track's `persistState(0)` runs, so a timer armed by the old track cannot fire against the new one. Traced `advanceOrStop` -> both branches (`playTrack`/`stop`) route through the cancel. `destroy()` already had its own `clearTimeout` (pre-existing); the two paths don't conflict (idempotent).

**Unit 2 — stop instead of wrap on removing the playing last track.** Reviewed `remove()`: `idx < queue.length` after splice correctly distinguishes "a following track shifted into the removed slot" (play it) from "removed slot was the tail" (stop, don't wrap to index 0). Confirmed the pre-existing "advances to next track when removing current [non-last]" test still covers the sibling branch, so both branches of the new conditional are exercised.

**Independent verification**: read the full diff and surrounding `player.ts` context (not just the new tests) to confirm the fix addresses the described root cause, not just makes the added assertions pass. Ran `audio` typecheck and the full `audio` vitest project (263 tests, 1 pre-existing skip) locally — all pass. Merged current `origin/main` into the branch (picked up an unrelated `intentions/` node edit only) and re-ran CI on the merge commit — all 17+ checks green.

**Residue**: none. Clean pass.

Applying `qa-done` and moving to review.

natb1:
## Review — clean (HOLD, not armed)

Terminal review pass for `tactic-audio-position-persist` complete. No blocking findings; the `reviewed` marker is applied on origin/main and the node stays at phase `review`.

**Verification**
- `npx vitest run --project audio` — 262 passed, 1 skipped (pre-existing conditional `skipIf(!hasDistBuild)`, unrelated).
- `run-typecheck.sh --app audio` — passed.

**Findings:** none confirmed. Both units match the plan exactly:
- Unit 1 — `cancelPositionThrottle()` correctly cancels the pending `timeupdate` throttle in both `playTrack` (track change) and `stop()`, so a timer armed by the old track can no longer persist its `currentTime` against the now-current track or a stopped queue. New tests assert the stale timer does not fire after change/stop while the live-position save still fires during normal playback.
- Unit 2 — removing the currently-playing last-in-queue track now `stop()`s + `persistState(0)` instead of wrapping to index 0 and force-playing track 0 from 0:00; the following-track-shifts-in case still plays the correct track. Covered by a new test.

**Merge custody:** implement -> qa -> review chain intact. The only commits touching `audio/` are the implement commit and a test-lint suppression, both before qa sign-off; later commits are pure origin/main sync merges with no code hand-edits.

**Disposition:** HOLD at review as a draft (auto-merge arming disabled this tick). Ready for human merge: `gh pr ready 2818 && gh pr merge --squash 2818`.

PR #2819, opened 2026-07-10
blogroll feed parser correctness + feed-proxy crash guard

natb1:
Graph tactic `tactic-blog-feed-parser-correctness` (serves `strategy-recover-discovery`). Fixes verified defects in the blogroll feed parser and the feed-proxy function (2026-07-05 review).

## Unit 1 — feed-fetch parser correctness
`packages/blog/src/blog-roll/vite-plugin-feed-fetch.ts`
- `decodeXmlEntities`: decode `&amp;` last (was first → double-decode) and handle hex/decimal numeric entities. WordPress `&#8217;`/`&#8230;` now decode instead of rendering literally.
- `xmlText`: strip `<![CDATA[...]]>` wrappers instead of shipping them verbatim.
- `selectAtomLink`: match `rel="alternate"` in either attribute ordering; exclude `rel="self"` from the fallback so the feed's own API URL is never used as a post link.

## Unit 2 — feed-proxy instance-crash guard
`functions/src/feed-proxy.ts`
- A non-AppCheck `verifyToken` error was re-thrown out of `handleFeedProxy`; the un-awaited `onRequest` wrapper turned it into an unhandled rejection that crashes the instance and drops concurrent in-flight requests. Now returns an explicit `500` with no error leak.
- Adds `X-Content-Type-Options: nosniff` to the proxied response (upstream Content-Type was reflected verbatim).

## Unit 3 — xmlns-strip whitespace
`packages/blog/src/blog-roll/parse-feed.ts`
- The xmlns-strip regex required a literal space; a Firefox-emitted `<feed\nxmlns=...>` was reported unavailable. Now accepts any whitespace.

## Verification
- `packages/blog` + `functions` vitest suites pass (49 tests across the three affected files, including new cases for reversed attribute order, CDATA, numeric entities, `rel=self` exclusion, the clean 500, the nosniff header, and newline-separated xmlns).
- Typecheck passes for both packages.


natb1:
## QA summary (2026-07-10)

Merged origin/main cleanly (no conflicts). Ran the full `packages/blog` (458 tests) and `functions` (92 tests) suites, plus a direct typecheck of `packages/blog` — all pass. All PR CI checks green.

**Verified against real third-party feed data** (not just the synthetic fixtures in the test suite): fetched live content from the three feeds configured in `FEED_REGISTRY` (`bastionland.com` — Blogger Atom, `newschoolrevolution.com` — python-feedgen Atom, `prismaticwasteland.com` — Squarespace RSS with CDATA) and ran them through both the build-time regex parser (`parseAtomFeedXml`/`parseRssFeedXml`) and the client-runtime DOM parser (`parseXml`). All three parse correctly end-to-end — title, canonical post URL, and published date all resolve as expected, including CDATA-wrapped Squarespace description/content fields and WordPress-style entity encoding.

**Found and fixed one gap beyond the stated scope:** Unit 1 fixed the rel="self" fallback bug in the build-time regex parser (`vite-plugin-feed-fetch.ts`'s `selectAtomLink`), but the identical defect was still live in the client-runtime DOM parser (`parse-feed.ts`'s `parseAtomFeed`, used by `AtomStrategy` via `feed-proxy`) — confirmed with a failing reproduction before fixing. Fixed by excluding `rel="self"` from the any-link fallback there too (`link[href]:not([rel="self"])`), matching the sibling fix's behavior, and added a regression test. Pushed as commit 5cf29472; all CI re-ran green.

**Investigated one false lead:** an initial probe against the real feeds through the client DOM parser (`parseXml`) failed in the `happy-dom` test environment with a `parsererror` (single-quoted XML declaration on the Blogger/python-feedgen feeds; a namespaced `<dc:creator>` element on the Squarespace feed). Reproduced the identical real feed content directly in a real Chrome tab via `DOMParser` — no error, correct extraction in both cases. This is a `happy-dom` XML-parsing strictness gap, not a production defect; no code change needed, and no test added (asserting behavior of a browser that doesn't run these feeds).

No residue items requiring `needs-main`, `needs-human`, or a follow-up. Clean pass.

Marking `qa-done` and advancing to review.

natb1:
## Review — clean (HOLD for human merge)

Graph-native review worker, phase `review`, node `tactic-blog-feed-parser-correctness`.

**Disposition: clean.** No confirmed blocking findings; the `reviewed` marker is landed on `main` (phase held at `review`, not armed — auto-merge is disabled this tick).

**Scope verified against the plan** (all 3 units + folded lows):
- Unit 1 — `vite-plugin-feed-fetch.ts`: entity decoder now decodes `&amp;` LAST and handles hex + decimal numeric entities (WordPress `&#8217;`/`&#8230;`); `stripCdata` removes CDATA wrappers; `selectAtomLink` matches `rel="alternate"` in either attribute order and falls back to the first non-`rel="self"` link. `parse-feed.ts` (DOM path) mirrors the `rel="self"` exclusion and widens the xmlns-strip regex to any whitespace (`\s+`), fixing Firefox `<feed\n  xmlns=...>`.
- Unit 2 — `feed-proxy.ts`: `verifyAppCheck` wrapped in try/catch → explicit `res.status(500)` for the non-AppCheck case (no more unhandled-rejection instance crash), and `X-Content-Type-Options: nosniff` added.
- Tests extended for reversed attribute order, CDATA, numeric entities, `rel="self"` skip, newline namespace, and the 500 path (asserting no `network timeout` leak).

**Checks:** `vitest --project packages/blog --project functions` — 550/550 pass.

**Merge-custody:** implement→qa→review chain unbroken. The DOM-parser `rel="self"` commit (`5cf29472`, 17:08) landed within the qa phase, before the qa→review sign-off (17:26) — qa fix lane, not a post-signoff hand-edit.

No deferred findings. No needs-main residue section in the node body.

To merge: `gh pr ready 2819 && gh pr merge --squash 2819`.

PR #2836, opened 2026-07-10
crypto: consolidate duplicated BENC crypto-core + fix worker buffer-transfer footgun

natb1:
Graph tactic **tactic-crypto-core-consolidate** (serves strategy-durable-owned-data).

`packages/crypto/src/crypto-core.ts` and `packages/crypto-core/src/crypto-core.ts` were two live, near-identical copies of the BENC primitives — the worker imported one and the main-thread fallback the other, so a KDF/format change could land in one path invisibly and corrupt the `.benc` round-trip.

## Unit 1 — one canonical crypto core

- `packages/crypto-core` is now the single source.
- Repoint `packages/crypto/src/crypto.ts` and the three `@commons-systems/crypto/core` consumers (`packages/crypto/test/crypto.test.ts`, `office-hours/test/snapshot.test.ts`, `budget/e2e/helpers.ts`) at `@commons-systems/crypto-core`.
- Delete the duplicate `packages/crypto/src/crypto-core.ts`; drop the `crypto/core` subpath export.
- Declare `@commons-systems/crypto-core` where it is now imported (crypto dep; office-hours/budget devDep).

## Unit 2 — worker buffer-transfer footgun

`postToWorker` transferred the caller's input `ArrayBuffer` to the worker, permanently detaching it. A retry on the same buffer — e.g. `decrypt(sameBuffer, correctedPassword)` after a wrong-password rejection — then hit the `isEncrypted` guard with `byteLength === 0` and threw the misleading "File is not in BENC encrypted format." instead of a password error.

- Clone the buffer before transfer so retries are safe by default (BENC files are small; copy cost is negligible).
- Document the guarantee on `BencCrypto.decrypt`.
- Regression test installs a faithful stub `Worker` (Node's `structuredClone` with a transfer list really detaches the transferred buffer), so it genuinely guards the clone — confirmed by a negative control (reverting the clone fails the test).

## Verification

- `packages/crypto` unit tests pass (10), covering an encrypt/decrypt round-trip via BOTH the main-thread fallback and the worker path.
- `office-hours` tests pass (46).
- Typecheck, lint, and knip pass across all affected workspaces.


natb1:
## QA summary (qa 2026-07-10)

Independent verification of both units, beyond reproducing the implementer's own checks:

**Unit 1 — canonical crypto core**
- Confirmed `packages/crypto/src/crypto-core.ts` is deleted and the `crypto/core` subpath export is gone; grepped the whole worktree for `crypto/core` and stale `crypto-core.ts` imports — none remain outside the canonical `packages/crypto-core/` package.
- `packages/crypto/src/crypto.ts`, `budget/e2e/helpers.ts`, and `office-hours/test/snapshot.test.ts` all now import `@commons-systems/crypto-core`; `@commons-systems/crypto-core` is declared as a dependency (crypto) or devDependency (budget, office-hours) where used.
- The Go exporter (`projects/budget-etl/internal/export/export.go`) doesn't import the TS module — it's an independent implementation sharing the documented BENC format/PBKDF2 parameters (600k iterations, 256-bit key) with `packages/crypto-core/src/crypto-core.ts`, which this PR left byte-for-byte unchanged. No cross-language drift introduced.
- `npx knip` on the full worktree: clean, no dangling exports.
- Ran `packages/crypto`, `office-hours`, and `packages/crypto-core` vitest projects independently: 10, 436 (combined), and included counts all green — matches CI.

**Unit 2 — buffer-transfer footgun**
- Read the fix: `postToWorker` now clones (`msgData.slice(0)`) before transferring, and `BencCrypto.decrypt`'s JSDoc documents the non-consumption guarantee.
- Read the new regression test: it installs a stub `Worker` that uses Node's real `structuredClone(..., { transfer })` to faithfully emulate detachment, so it's actually exercising the transfer semantics, not a mock that always passes.
- **Negative control (independently reproduced, not just trusted from the PR body):** temporarily reverted the clone in the worktree (`w.postMessage({ ...msg, id }, [msgData])` with no clone), reran the suite — the new regression test failed exactly as expected (`encrypted.byteLength` was `0` instead of `>0`). Restored the fix, reran — full suite green again (10/10). This confirms the test genuinely guards the fix rather than passing vacuously.

**CI:** all 20 checks green (typecheck, lint, dead-code/knip, unit-tests, acceptance, preview-and-smoke, rules-test, go-tests, nixos-build, darwin-build, CodeQL, etc.).

No UI-facing surface changed (pure library consolidation + internal buffer-handling fix), so no browser QA pass was needed beyond the exercised test suites and knip/typecheck/lint already covered by CI.

**Disposition:** clean pass, no residue. Proceeding qa -> review.

PR #2837, opened 2026-07-10
CI change-detection: transitive dep closure + root-file blind spots

natb1:
## What

Fixes two CI change-detection blind spots that let real changes ship untested/undeployed. Graph node `tactic-ci-change-detection-transitive` (serves strategy-autonomous-execution).

### Unit 1 — transitive internal-dep propagation
`resolve_dirty_apps` in `lib.sh` mapped each internal package to its *direct* declarers only. Internal packages are themselves workspaces, so a leaf change never reached transitive dependents: a `ds`-only change tested/deployed every app except fellspiral, which imports `@commons-systems/blog` which imports `@commons-systems/ds` but never declares `ds` itself. Now computes the transitive closure of the reverse-dependency graph.

### Unit 2 — root-file blind spots
- `lib.sh`: added `vitest.config.ts` to the mark-all-dirty root-file list. A PR editing only it resolved zero dirty apps, so run-unit-tests exited 0 and a broken test config merged green.
- `detect-changes.sh`: added a root `go.work` / `go.work.sum` arm. Those files sit under no discovered Go module prefix, so a go.work change previously skipped go-tests.

## Tests
Added fixtures to `test-dispatch-scripts.sh`:
- ds change marks blog + transitive fellspiral
- blog change marks fellspiral but not the leaf ds it consumes
- vitest.config.ts fan-out marks all workspaces

All pass; the go.work regex is anchored to repo root (nested `foo/go.work` does not match).

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/detect-changes.sh
+++ b/.claude/skills/dispatch-propagate/scripts/detect-changes.sh
@@ -45,11 +45,14 @@ if echo "$CHANGED" | grep -qE '\.(ts|tsx|js|jsx|mjs|cjs)$|^knip\.(json|jsonc|ts)
 fi
 # go-tests needs the Go toolchain. Set go=true when a changed file is under a
 # discovered Go module. list-go-modules.sh discovers module roots from go.mod
-# locations, so a new Go module needs no edit here.
+# locations, so a new Go module needs no edit here. The root go.work / go.work.sum
+# files sit under no module prefix, yet they define the workspace every module
+# builds in — a change to them must still run go-tests, so match them directly.
 GO_MODULE_PREFIXES=$("$(dirname "$0")/list-go-modules.sh" | sed 's|$|/|')
+GO_REGEX='^go\.work(\.sum)?$'
 if [ -n "$GO_MODULE_PREFIXES" ]; then
-  GO_REGEX=$(printf '%s\n' "$GO_MODULE_PREFIXES" | paste -sd'|' -)
-  if echo "$CHANGED" | grep -qE "^($GO_REGEX)"; then
-    echo "go=true" >> "$GITHUB_OUTPUT"
-  fi
+  GO_REGEX="$GO_REGEX|^($(printf '%s\n' "$GO_MODULE_PREFIXES" | paste -sd'|' -))"
+fi
+if echo "$CHANGED" | grep -qE "$GO_REGEX"; then
+  echo "go=true" >> "$GITHUB_OUTPUT"
 fi
--- a/.claude/skills/dispatch-propagate/scripts/lib.sh
+++ b/.claude/skills/dispatch-propagate/scripts/lib.sh
@@ -1958,9 +1958,16 @@ resolve_dirty_apps() {
       [ -n "${_seen[$cur]+x}" ] && continue
       _seen["$cur"]=1
       # If cur is itself an internal package other apps consume, its direct
-      # dependents are transitive dependents of root_pkg too.
-      if [ -n "${direct_dependents[$cur]+x}" ]; then
-        for nxt in ${direct_dependents["$cur"]}; do
+      # dependents are transitive dependents of root_pkg too. direct_dependents
+      # is keyed on the dependency SHORT name (e.g. "blog"), but cur holds a
+      # full app/workspace path (e.g. "packages/blog") — the same short-name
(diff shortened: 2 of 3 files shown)


natb1:
## QA summary (2026-07-10)

**Plan re-verified independently, not just re-run.** Both units checked against real repo data, not only the fixtures the implementer wrote.

### Unit 1 — transitive internal-dep propagation
Traced the actual repo dependency graph (`packages/blog` peer-depends on `@commons-systems/ds`; `fellspiral` depends on `@commons-systems/blog` but never on `ds` directly) and called `resolve_dirty_apps` directly against the live `packages/ds/src/index.ts` path.

**Found and fixed a real bug**: the closure's BFS looked up `direct_dependents[cur]` using `cur` as the full workspace path (e.g. `packages/blog`), but `direct_dependents` is keyed on the dependency short name (`blog`). Since every internal package in this repo lives under `packages/`, that key never matched anything, and the closure silently truncated at depth 1 — a `ds`-only change never reached `fellspiral`, i.e. exactly the blind spot this unit exists to close. The implementer's own fixture didn't catch it because it used a flat top-level `blog` dir (short name == app path by coincidence).

Fixed by re-deriving the short name at each BFS hop (`cur_short="${cur##*/}"`), mirroring the existing bridge used a few lines below. Added a fixture nesting the mid-tier package under `packages/` to match the real layout, and confirmed by hand: `packages/ds/src/index.ts` now correctly resolves `fellspiral` (and every other real multi-hop dependent: `authutil`→`firebaseutil`, `idbutil`→`local-first`/`mediautil`, etc.) as dirty.

### Unit 2 — root-file blind spots
Verified the `vitest.config.ts` fan-out and the `go.work`/`go.work.sum` regex directly: real repo has `vitest.config.ts` and `go.work` at root, both now match; nested `foo/go.work` correctly does not match (regex anchored to repo root).

### Test suite
Ran the full `test-dispatch-scripts.sh` suite (4199 tests) before and after the fix, and separately against origin/main directly. 3 failures (2x `dispatch-phase-model`, 1x `dispatch-refresh-rate-limits` regression) are pre-existing on origin/main, unrelated to this tactic's scope — confirmed by running the identical suite against a clean origin/main checkout. CI's own `hook-tests`/`unit-tests` jobs pass green on this PR.

### Disposition
Clean pass after one opus-fixable-class residue item was fixed in place (transitive-closure BFS key bug, commit 59cd9742). No needs-main or needs-human residue.

natb1:
## Review disposition — clean pass (HOLD for human merge)

Node `tactic-ci-change-detection-transitive` reviewed at phase `review`. Applied the `reviewed` marker to `execution.markers` on origin/main (graph-commit `cdaa3908`); phase left at `review` (auto-merge arming disabled this tick).

**Scope reviewed** (3 files): `lib.sh`, `detect-changes.sh`, `test-dispatch-scripts.sh`.

**Findings:** none blocking; no deferred follow-ups filed.

Verified behavior directly (the 42k-line suite exceeds the run window, so the affected functions were exercised in isolation under bash):
- Unit 1 — transitive internal-dep closure: a `packages/ds` change marks `fellspiral` via `ds → blog → fellspiral`, including the nested-`packages/` mid-tier layout where the dependency short-name and the workspace path diverge; propagation is upward-only (a `blog` change does not mark `ds`); BFS terminates on cycles via the seen-set.
- Unit 2a — `vitest.config.ts` added to the root-config fan-out; editing it alone now resolves the full workspace set.
- Unit 2b — `detect-changes.sh` `go=true` now fires on root `go.work` / `go.work.sum`; regex is properly anchored (`ggo.work`, `go.workfoo`, `config/go.work` correctly do not match) and still fires on module-prefixed files, with or without discovered module prefixes.

**Merge custody:** intact. The two branch commits are the implement (`ed70dabb`) and a within-QA fix (`59cd9742`, "qa:" prefix, landed before the `qa-done` sign-off) that added the short-name re-derivation and the nested-`packages/` fixture. No author edit after QA sign-off.

**Freshness gates:** phase still `review`, `office_hours` null, `strategy_fingerprint` null (never stale) — all green.

No needs-main residue section on the node body.

Deviation note: no subagent-spawn tool available in this worker context — finders/adversarial-verification were run inline.

To merge: `gh pr ready 2837 && gh pr merge --squash 2837`.

PR #2839, opened 2026-07-10
sidecar: cross-tab write coordination + surfaced persist failures

natb1:
Implements graph node `tactic-sidecar-cross-tab-safety` (serves strategy-durable-owned-data).

## Problem

The shared sidecar single-flight write chain (`packages/sidecar/src/factory.ts`) guaranteed no-clobber per-tab only, and swallowed persist failures:

- `writeSidecar` rewrote the whole file against a possibly-stale in-memory `cachedModel`, so two tabs on one folder each merged against their own stale model and the last write discarded the other's entries.
- `enqueueWrite`'s per-link `.catch` logged and resolved, so a caller never learned a disk persist failed (a full/revoked-permission folder reported saves as successful — silent data loss, against the clear-errors rule).

## Unit 1 — cross-tab coordination

`enqueueWrite` now read-merge-writes against the CURRENT on-disk state under a cross-tab Web Lock (`navigator.locks`). The patch is merged onto whatever another tab has since persisted rather than onto this tab's stale model, so two tabs on one folder no longer clobber each other. Web Locks is a progressive enhancement — where unavailable, the per-tab single-flight chain still serializes writes (never below the prior single-tab guarantee). A disk re-read that finds corruption suppresses the write to preserve recoverable bytes, matching the load-time corruption contract.

## Unit 2 — surface persist failures

The promise `enqueueWrite` returns now rejects when the disk persist fails, so callers learn a save did not reach disk, while the single-flight chain is kept recovered so one failed write cannot poison the next. `flushWrites` reflects the most recent settled write's outcome; a later successful write clears it.

## Verification

- `packages/sidecar` unit tests: two-tab read-merge-write preserves both tabs' entries; a forced write failure rejects both the caller's promise and `flushWrites`; a failed write does not poison the chain.
- Full `packages/sidecar`, `audio`, and `print` suites pass (the factory's consumers); typecheck + lint clean.

### DIFF
--- a/packages/sidecar/src/factory.ts
+++ b/packages/sidecar/src/factory.ts
@@ -24,6 +24,23 @@ export function serializeSidecar<TData>(data: TData): string {
   return JSON.stringify(data, null, 2);
 }
 
+/**
+ * Run `fn` while holding a cross-tab exclusive Web Lock named `name`, so the
+ * read-merge-write it performs is atomic against every other tab on the same
+ * origin holding the same-named lock — the cross-tab no-clobber guarantee.
+ *
+ * Web Locks is a progressive enhancement: where `navigator.locks` is
+ * unavailable (older engines, non-window contexts, the test runner) the
+ * callback runs directly. The per-tab single-flight write chain still
+ * serializes writes within the tab, so correctness degrades only to the
+ * pre-existing single-tab guarantee, never below it.
+ */
+async function withWriteLock<T>(name: string, fn: () => Promise<T>): Promise<T> {
+  const locks = globalThis.navigator?.locks;
+  if (!locks) return fn();
+  return locks.request(name, fn);
+}
+
 // ---------------------------------------------------------------------------
 // B. Factory options + return shape
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


natb1:
## QA summary (qa 2026-07-10)

Merged latest origin/main into the branch (no conflicts) and re-verified CI green on the merged tip.

**What was reviewed**

- `packages/sidecar/src/factory.ts`: read the full diff for both units — cross-tab read-merge-write under a Web Lock (Unit 1) and the enqueueWrite/flushWrites promise now rejecting on a disk persist failure instead of always resolving (Unit 2).
- Traced every downstream caller of `enqueueWrite`/the wrapper write functions (`cacheMetadata(Batch)`, `savePlayerState`, `savePlaylist` in audio; `cacheMetadata`, `PositionStore.save`, `AnnotationsStore.save` in print) to confirm the now-possible rejection is handled everywhere it's reachable — all call sites either sit inside a try/catch or already have a `.catch(...)` (several print-viewer hooks already called `reportError(...)` in their catch, which was previously dead code since the promise never rejected; this PR makes that error-surfacing path live).
- Confirmed the progressive-enhancement fallback (`withWriteLock` runs the callback directly when `navigator.locks` is unavailable) degrades no lower than the pre-existing single-tab guarantee.
- Ran `packages/sidecar`'s unit suite (14 tests, including the two-tab no-clobber test and the forced-write-failure rejection tests) — all passed.
- Ran the full `audio` + `print` suites (dependents of the sidecar factory) — 895 passed / 2 pre-existing skips, no regressions.
- Found one real coverage gap: Unit 1's read-merge-write re-reads on-disk state on every write, which added a new "on-disk corrupted at write time" branch (mirroring the existing load-time corruption contract) that had no test. Added a case (`packages/sidecar/test/factory.test.ts`, commit f967e8b3) that clobbers the fake disk mid-session and asserts the disk write is suppressed while the in-memory merge still lands. All CI checks green on the updated tip.

**Result:** clean pass, no residue outstanding. Moving to review.

natb1:
## Review disposition — tactic-sidecar-cross-tab-safety

**Clean review. Held at phase `review` as a draft (auto-merge arming disabled this tick) for the human to merge.**

### Fan-out
Review finders + adversarial verification run inline (no subagent-spawn tool available in this worker). No confirmed blocking findings; no Opus fix lane needed.

### What was verified
- **Cross-tab coordination (Unit 1):** `enqueueWrite` now read-merge-writes against the current on-disk model inside a cross-tab Web Lock (`withWriteLock` / `writeLockName`), so two tabs on one folder no longer clobber each other. Where `navigator.locks` is unavailable it degrades to the pre-existing per-tab single-flight guarantee, never below it. A write-time on-disk corruption is suppressed (preserving recoverable bytes) and folded into the in-memory model, mirroring the documented load-time corruption contract.
- **Surfaced persist failures (Unit 2):** the caller-facing promise now rejects on a disk-persist failure while the single-flight chain stays recovered (`pendingFailure`); `flushWrites` reflects the most-recently-settled failure. Every app call site already attaches `.catch` (`print/src/viewer/{useAnnotations,useBookmarks,useViewerController}.ts`, `audio/src/player.ts`), so the new rejection path introduces no unhandled-rejection regression.

### Checks
- All 22 CI checks green on head `f967e8b3`.
- `packages/sidecar` unit suite: 15/15 pass locally (6 new tests cover cross-tab no-clobber, successive writes, write-time corruption suppression, caller-facing rejection, drain-reflects-failure, and chain recovery).
- Merge custody unbroken: implement (13:13-13:15) → qa transition (13:23) → in-qa test addition (`f967e8b3`, 17:40, before sign-off) → qa→review sign-off (17:46). No post-sign-off author hand-edit.

### Deferred / follow-up
None.

To merge: `gh pr ready 2839 && gh pr merge --squash 2839`.

PR #2841, opened 2026-07-10
print: EPUB highlight + note annotations via CFI ranges

natb1:
## Context

`tactic-print-annotations` added highlight + note annotations to the PDF
viewer with open-format, owned-storage persistence. This extends the same
capability to EPUBs via epub.js CFI-range annotations, so
strategy-recover-knowledge's entry artifact covers both text formats print
renders. The stores, open annotation schema, `useAnnotations` hook, and
`AnnotationsPanel` are reused unchanged.

Graph tactic: tactic-print-annotations-epub (serves strategy-recover-knowledge).

## Unit 1 — EPUB selection capture + persistent highlights (`print/src/viewer/epub.ts`)

- Capture text selections via epub.js's `rendition.on("selected", cfiRange)`
  event; read the quote through `book.getRange(cfiRange)`. Because the
  selection lives inside epub.js's content iframe, the shared `useAnnotations`
  hook (which listens on the TOP document's `selectionchange`) can't see it,
  so `captureSelection` stashes the anchor and re-dispatches a top-document
  `selectionchange` to nudge the hook to re-read `getSelectionAnchor`.
- Render persistent highlights with `annotations.highlight(cfiRange, …,
  "annotation-highlight", …)` — a class/fill distinct from search's base
  highlights. `setAnnotations` reconciles the epub.js registry
  (remove-then-re-add) and never touches search highlights (tracked
  separately), so the two coexist. epub.js re-injects registered annotations
  on each view render, so highlights survive chapter navigation.
- A content `selectionchange` listener clears the pending anchor when the
  selection collapses (epub.js only emits `selected` for non-empty
  selections).
- Loosen `SelectionAnchor.page/offset/length` to optional (`print/src/viewer/types.ts`)
  — an EPUB anchor is CFI-only, matching the already-optional PDF-anchor
  fields on `Annotation`.

## Unit 2 — Shared annotations UI (no code change needed)

The store layer (`pickAnnotationsStore`, keyed on bare filename / mediaId),
the `useAnnotations` hook (keys on `annotation.position`), and
`AnnotationsPanel` (quote-only label, no page number) are already
format-agnostic and were wired for EPUB by `tactic-print-annotations`. Added
an end-to-end test over a CFI-positioned annotation (add → list → navigate →
delete) proving no PDF-specific assumptions remain.

## Verification

- `npx vitest run --project print --root .` — 645 passed | 1 skipped (pre-existing).
- Manual browser QA (deferred to qa phase): open a local-folder EPUB, select
  text, add a highlight and a note; the highlight paints and survives chapter
  navigation and reload; panel navigation jumps to the CFI; the folder's
  `.commons-print/index.json` carries the annotation as readable JSON; search
  and annotation highlights coexist in the same section. Note: the floating
  capture control positions off the top document's selection rect, which is
  empty for an iframe selection — for EPUB it currently anchors near the
  viewport origin rather than the selection. Functional capture is unaffected;
  flagging for qa to judge whether a positioning follow-up is warranted.

### DIFF
--- a/print/src/viewer/epub.ts
+++ b/print/src/viewer/epub.ts
@@ -1,6 +1,7 @@
 import ePub, { type Book, type Rendition, type Location, type NavItem } from "epubjs";
-import type { OutlineEntry, SearchableRenderer, SearchResult, SearchResponse } from "./types.js";
+import type { OutlineEntry, SearchableRenderer, SearchResult, SearchResponse, SelectionAnchor } from "./types.js";
 import { MAX_SEARCH_RESULTS } from "./types.js";
+import type { Annotation } from "../annotations.js";
 
 // epubjs ships incomplete type declarations. These narrow shapes describe the
 // runtime members we use that are missing from the .d.ts, following the
@@ -24,6 +25,10 @@ interface EpubAnnotations {
 // matches the panel <mark> color (viewer.css:380); active is visually distinct.
 const BASE_STYLES = { fill: "#fde68a" };
 const ACTIVE_STYLES = { fill: "#f59e0b", "fill-opacity": "0.5" };
+// Persistent annotation highlights. A calmer, semi-transparent yellow mirrors
+// the PDF `.annotation-highlight` fill (viewer.css:141, rgba(255,235,59,0.4)) and
+// reads as distinct from search's fully-saturated base fill.
+const ANNOTATION_STYLES = { fill: "#ffeb3b", "fill-opacity": "0.4" };
 
 export function createEpubRenderer(
   onError?: (err: unknown) => void,
@@ -47,6 +52,11 @@ export function createEpubRenderer(
   let _searchCfis: string[] = []; // all base-highlighted match cfis
   let _activeCfi: string | null = null; // currently-navigated match
   let _searchEpoch = 0; // monotonic guard against overlapping search() calls
+
+  // Persistent annotation state.
+  let _annotations: Annotation[] = []; // full list, pushed in by setAnnotations
+  let _annotationCfis: string[] = []; // annotation-highlight cfis currently registered
+  let _pendingAnchor: SelectionAnchor | null = null; // last live text selection, or null
 
   // Resolve a section's TOC label, falling back to a 1-based spine index.
   // Caller must `await book.loaded.navigation` once before invoking.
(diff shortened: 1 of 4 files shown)


natb1:
## QA — tactic-print-annotations-epub

Independently verified against the tactic's manual QA plan using the public Firestore-seeded EPUB (`gutenberg-3296`, The Confessions of St. Augustine) via the QA server, not just the implementer's own vitest coverage.

**Automated (floor):**
- `npx vitest run --project print --root .` — 645 passed, 1 pre-existing skip
- `run-typecheck.sh --app print` — passed

**Manual browser QA (live app, real EPUB, Chrome via claude-in-chrome):**
- Selected text in the rendered EPUB content; the shared `AnnotationCapture` toolbar (Highlight/Note) appeared, driven by the new `getSelectionAnchor()` on the EPUB renderer.
- Added a highlight: painted immediately with the calmer annotation fill, distinct from search highlighting; persisted to `localStorage` (this item isn't a local folder, so it uses the device-local store per `annotations.ts`) with `position` as a CFI string and no `page`/`offset`/`length` fields — confirms the loosened-to-optional PDF anchor fields.
- Navigated forward a chapter and back: highlight survived the chapter change without a manual re-apply hook, and the `Annotations` panel entry navigated the CFI position correctly on click (`goToPosition`).
- Full page reload: annotation reappeared in the panel and repainted on the correct chapter, sourced from persisted storage.
- Added a second highlight with a note via the "Note" affordance; panel showed the note text under the quote (quote-only label, no page number, as designed).
- Searched for a term ("Ambrose") elsewhere in the same section as an existing annotation highlight: search highlighting (bright) and the annotation highlight (calmer) rendered simultaneously without interference — confirms `_searchCfis`/`_annotationCfis` are tracked independently.
- Deleted the second annotation: its highlight and panel entry were removed while the first annotation's highlight and entry remained untouched; verified via `localStorage` that only the surviving annotation persisted.

All items in the tactic's Verification section pass. No residue. Transitioning qa -> review.

natb1:
## Review — tactic-print-annotations-epub

Terminal review pass. Finders ran **inline** (no subagent-spawn tool available in this environment — recorded deviation), covering correctness, cross-renderer parity, and test integrity over the 4-file diff (`print/src/viewer/epub.ts`, `types.ts`, and the two test files).

**Gates**
- Freshness: node phase `review`, `office_hours` null on a fresh origin/main read.
- CI: all checks green; PR MERGEABLE / CLEAN.
- Custody: implement -> qa -> review chain intact. Last branch commit 2026-07-10 21:36 UTC precedes the qa sign-off (2026-07-11 03:25 UTC); no post-qa author edit — no merge-custody break.

**Findings:** none Required; no material deferred items.

Reviewed the selection-capture wiring, the annotation reconcile, and PDF parity:
- `captureSelection` reads the quote via `book.getRange` and re-dispatches a top-document `selectionchange` so the shared `useAnnotations` hook re-reads `getSelectionAnchor` (the hook listens on the top document, which never sees an iframe-internal selection). Error path reports and yields no anchor.
- `reapplyAnnotationHighlights` removes only previously-registered annotation cfis (`_annotationCfis`) and never touches `_searchCfis`, so search and annotation highlights coexist — first apply removes nothing, deletion removes only the dropped cfi and re-applies survivors.
- `registerSelectionClear` nulls `_pendingAnchor` on collapse, but the note-commit path uses the hook's frozen `pendingRef.current`, so focusing the note input does not drop the capture — matching the PDF renderer's freeze semantics.
- `SelectionAnchor.page/offset/length` loosened to optional so a CFI-only EPUB anchor fits the same shape as the PDF anchor and `Annotation`; new methods land as optional `ContentRenderer` members with parity in `pdf.ts`.
- Tests are genuine and unweakened: CFI-only capture, getRange-reject, per-annotation highlight registration, first-apply-removes-nothing, delete-and-reapply, plus an end-to-end EPUB annotation flow in `annotations.test.tsx`.

**Disposition: merge-ready-hold.** Both gates green and custody intact, but this tick carries no fresh in-chat human auto-merge grant, so auto-merge is NOT armed. Applying the `reviewed` marker (state-only, phase kept at `review`) and surfacing PR #2841 for human merge.

PR #2881, opened 2026-07-16
Two-phase tick: pre-selection scope-stale sweep before worker-group spawn

natb1:
## Summary

Implements `tactic-tick-scriptable-then-spawn`: a dispatch tick moves ALL
scriptable, non-worker dispositions (including the scope-staleness demote)
into a phase that runs before worker-group selection, so a metadata-only
disposition (like demoting a scope-stale tactic back to `implement`) never
silently consumes a `SPAWN_N` launch slot.

Previously the scope-staleness demote ran only at worker-launch time
(`dispatch-graph-execute`'s exit-13 case), after `dispatch-select-tick` had
already spent a selection slot on the node — so a tick could end having
launched zero workers even with headroom available.

## Changes

- **Unit 1** — `listScopeStaleTactics` (`packages/intentionsutil/src/scope-sweep.ts`):
  a pure, offline-testable enumerator returning scope-stale `fix`/`qa`/`review`
  tactics (unparked, no live worker, stamped scope-fingerprint mismatch;
  missing stamp fails open).
- **Unit 2** — `dispatch-graph-scope-sweep` + its `list-scope-stale-tactics.ts`
  CLI: gathers live node ids, runs the enumerator, and demotes each stale
  tactic via the existing `demote-node-to-implement` primitive. Wired into
  `dispatch-select-tick`'s Step 1d, after the other reconcile/census sweeps
  and before the concurrency gate / selection.
- **Unit 3** — `dispatch-graph-execute`'s exit-13 case is now a documented
  post-sweep safety net (rare residual drift between the sweep and launch),
  and it now clears the node's reservation (previously missing, unlike the
  sibling exit-12 case), so a demoted node is freed for next-tick
  re-selection instead of staying counted against the budget.

## Test plan

```verify
npx vitest run --project packages/intentionsutil --root .
```

- `npx vitest run --project packages/intentionsutil --root .` — 483 tests / 28 files, all pass.
- `bash -n` clean on both touched/created shell scripts.
- Manual end-to-end verification (reproduce a scope-stale tactic and confirm
  phase-1 demote + phase-2 launch in the same tick) is left to the graph
  node's QA phase, per the plan's Verification section.


natb1:
## QA summary — escalated to office-hours

Backend/CLI-only change (`dispatch-select-tick`, `dispatch-graph-execute`,
`dispatch-graph-scope-sweep`, `packages/intentionsutil`) — no browser
component detected, no frontend/Firestore surface touched.

### Items executed

1. **intentionsutil vitest suite passes fresh after merge with origin/main** — `script-verifiable` — **PASS** (502 tests / 28 files, all green)
2. **New/modified shell scripts are syntactically valid** — `script-verifiable` — **PASS** (`bash -n` clean on all three)
3. **Sweep is invoked in `dispatch-select-tick` BEFORE the concurrency gate / selection** — `script-verifiable` — literal check **FAIL**, investigated and confirmed a false fail: the check's naive grep matched a doc-comment mention of "graph-select-target" at line 190, not an invocation. The real invocation call sites are lines 615/879; the sweep call is at line 496 — strictly before both. The ordering property the tactic depends on **does hold**. Disposition triage classified this **already-satisfied** (dropped as PASS, no code defect).
4. **`dispatch-graph-execute` case `13)` clears the reservation** — `script-verifiable` — **PASS** (`reservation_clear` found at line 227)
5. **`listScopeStaleTactics` is exported from intentionsutil and importable** — `script-verifiable` — **PASS** (`src/index.ts:51`)
6. **Live-tick end-to-end: phase-1 demote + phase-2 launch in one tick** — `needs-human-judgment`, `Flag: planned-deferral` — not run (see below)
7. **Live-tick: single-pass convergence, liveness guard, residual safety net** — `needs-human-judgment`, `Flag: planned-deferral` — not run (see below)

**Counts:** 4 PASS, 1 investigated false-fail (dropped as already-satisfied), 2 planned-deferral items requiring manual verification.

No browser component was detected, so the browser lanes (single-assertion / walkthrough) never ran.

### Disposition triage

- Item 3 → **already-satisfied**: the QA check's own grep target was wrong (matched a doc comment, not the real call sites); the code's ordering property holds as written. No fix needed.
- Items 6 and 7 → **needs-main** (planned-deferral, per the node's own `## Verification` section, which explicitly labels these checks manual). Both require running a real dispatch tick against live graph state via `claude agents --json` (`dangerouslyDisableSandbox`, live daemon socket) — unsafe to auto-run in an isolated QA session (risk of mutating production graph state / demoting real in-flight tactics).

### Why this escalated instead of filing a main-qa residue section

`/qa-main`'s autonomous verifier is browser-only (Claude-in-Chrome against
deployed prod) — it has no mechanism to run a dispatch tick or query the local
daemon. Deferring items 6/7 to a `main-qa` residue section would only route
them into a doomed autonomous `qa-main` attempt that fails the same way and
parks anyway. Escalating directly to office-hours now, with a recommendation
comment, avoids that dead-end round-trip.

No code changes were needed or applied this pass — everything script-verifiable
already passes; the remaining work is a manual live-tick confirmation, not a
fix.

natb1:
## Review summary

**Fixed**
- `dispatch-graph-scope-sweep:88` — the sweep excluded only *live* nodes from
  demotion, not *reserved* ones, so it could race `graph-select-target`'s
  claimed set (which excludes on reservation OR liveness) and demote a node
  whose reservation is still outstanding across a ci-waiting/spawn-failure
  tick. Fixed by sourcing `lib-reservation-ledger.sh` and skipping any
  candidate where `reservation_exists "$id"` before invoking the demote.
  Commit f9fa93d4.

- `list-scope-stale-tactics.ts:44` — flagged that the `--live` CLI parser
  splits each value on whitespace/comma, so a live node id containing
  whitespace wouldn't match verbatim. Refuted: node ids are never
  attacker-controlled — they're minted exclusively by `align-tactics` /
  `align-strategy` as `tactic-<slug>` / `strategy-<slug>`, and multiple other
  parts of the system (digest.ts's wildcard regex, worktree/branch naming)
  already assume a whitespace-free id. No real generation path in this
  codebase produces a whitespace-containing id, so this is a theoretical
  robustness nit, not an exploitable gap.

**Informational**
- `scope-sweep.ts:96` — `listScopeStaleTactics` reads `phase` and
  `office_hours` directly rather than through a shared accessor; noted for
  reference, no action needed.

**False-positive (security)**
- `dispatch-graph-scope-sweep:58` — flagged that an exit-0-with-empty-stdout
  from `claude_agents_list_all` is trusted as a definite "no live sessions"
  rather than treated like the UNKNOWN case. This is the intended contract
  (empty stdout + exit 0 is a genuine empty live set; exit 1 is UNKNOWN and
  is the case the script already guards against) — not a real gap.

No dependency, CodeQL, or erosion findings (no `package.json` changes; no open
CodeQL alerts for this PR; no erosion hits).

natb1:
## review · attempt 1
Review pass on PR #2881: 4 findings surfaced (code-review x2, red-team x2), 1
fixed. Fixed: dispatch-graph-scope-sweep now excludes reservation-holding
nodes (not just live ones) from demotion, matching graph-select-target's
claimed set — closes a race where the sweep could demote a node whose
reservation is still outstanding across a ci-waiting/spawn-failure tick.
Refuted: a --live CLI whitespace-splitting nit (node ids are never
whitespace-containing in this codebase's actual generation paths).
Informational/false-positive: 2 no-action notes. No deviation. failed: none.

PR #2890, opened 2026-07-18
store: make all node bodies durable (drop kind gate)

natb1:
Resolves the `tactic-nontactic-body-durability` design decision with the ratified **greenfield** contract: every node body is authoritative, durable content that `writeNode` preserves verbatim across rewrites, for **every** kind — not just tactics.

## Problem

`store.ts`'s `readExistingTacticBody` gated body preservation on `node.kind === "tactic"`, so `writeNode` regenerated every non-tactic body from `statement` on every write. Any durable content authored into a `strategy` / `kind` / `delegation` / `virtue` body was silently erased by the next `reconcile-graph`, `read-sensors`, park, or transition write. This conflicted with the 2026-07-09 kind-strategy body-function rule and blocked three consumers.

## Change

- `readExistingTacticBody` → `readExistingBody`: drop the kind gate. An existing on-disk body is preserved verbatim for every kind; only a brand-new file falls back to the generated `# ${statement}` placeholder.
- `assertNoTacticBodyLoss` → `assertNoBodyLoss`: generalize the tactic-only kind-change guard into a kind-agnostic invariant — throw if a rewrite would replace an existing non-placeholder body with a regenerated placeholder (defense-in-depth at the `writeNode` public-API boundary).

## Verification

- `store.test.ts` + `reconcile-graph.test.ts` + `write-node.test.ts`: 24 passed.
- Full `intentionsutil` suite: 496 passed.
- Two `store.test.ts` tests that asserted the old clobber behavior were inverted to assert the ratified preservation behavior (mechanism fix, not a signal weakening).

Honors the 2026-07-09 body-function rule and unblocks `tactic-calibration-event-registry`, `tactic-mount-schema`, and `tactic-graph-native-dispatch-fold` with no content relocation.


natb1:
## Review

**Fixed**
- `packages/intentionsutil/src/store.ts:71` — `assertNoBodyLoss` compared the rewrite's body against the *new-statement placeholder* instead of the preserved on-disk body, so it could never actually catch a body-preservation regression (and could false-positive on a lossless rewrite that happened to equal the placeholder). Now compares `body !== existingBody`. Fix commit: 1dba9493.
- `packages/intentionsutil/src/store.ts:12` — `writeNode`'s JSDoc still documented the pre-change "cosmetic render for every kind except tactic" behavior. Rewritten to describe the durable-body-for-every-kind contract. Fix commit: 1dba9493.
- `packages/intentionsutil/test/store.test.ts:276` — no test exercised the `assertNoBodyLoss` throw branch. Added coverage for a genuine body-loss regression and for the lossless case where the preserved body happens to equal the new-statement placeholder. Fix commit: 1dba9493.

**Informational**
_None._

**Dismissed**
_None._

**Deferred**
_None._

**Out-of-scope (security)**
_None._

No open CodeQL alerts on this PR ref. Dependency audit not applicable (diff does not touch `package.json` / `package-lock.json`). Full coverage — no throttling or partial-coverage this pass.

natb1:
## review · attempt 1
review: fixed 3 code-review findings in packages/intentionsutil/src/store.ts — assertNoBodyLoss was comparing against the wrong placeholder (fixed to compare against the preserved on-disk body), stale JSDoc corrected for the durable-body-for-every-kind contract, and throw-branch test coverage added. No security/deps surface (deps=false); no CodeQL alerts. No deferred findings. failed: none

PR #2891, opened 2026-07-18
office-hours-graph: guard selector staleness against origin/main

natb1:
## Problem

`office-hours-select.ts` is a deliberately offline oracle — it "reads only the
local `intentions/` store (no gh, no daemon, no network)." It reads whatever git
state the checkout sits at. Run from a **stale PR-branch worktree** (behind
origin/main), it ranks a park that a prior session already drained-and-cleared as
the #1 head, because that checkout's committed node file predates the clear. The
`office-hours-graph` launcher then boots an `/office-hours` session for a park
that no longer exists on main.

Observed live this cycle: `tactic-nontactic-body-durability` was already cleared
on origin/main (`office_hours: null`, phase `review`, PR #2890 green), yet the
selector — run from that node's own stale worktree — still ranked it #1.

## Fix

Add the freshness guard in the **scripted launcher**, not the skill (the check is
purely mechanical — no LLM judgment):

- `git fetch origin main` before selecting.
- `park_live_on_main <node-id>` authoritatively confirms the park on origin/main
  (`git show origin/main:intentions/<id>.md` → an `office_hours:` field that is
  not `null`).
- Queue-head mode walks the ranked `--list` top-down and launches the first node
  still parked on main, skipping already-cleared false positives (each logged to
  stderr).
- A new `cleared <node-id>` disposition reports an explicit-target node already
  cleared on main.

`office-hours-select.ts` stays a pure offline oracle; the guard lives once in bash
rather than being re-derived per session in `SKILL.md`. This is the structural
counterpart of the stop-gap guard added to the emulated office-hours prompt.

## Verification

`bash -n` clean. `park_live_on_main` exercised against real origin/main state:

- `tactic-nontactic-body-durability` (`office_hours: null`) → not-live (skipped) ✓
- `tactic-align-tactics-mechanical-floor`, `tactic-review-sitting-skill-generalization`
  (block-scalar `office_hours:`) → live ✓
- absent node → not-live ✓

So the launcher would now skip this cycle's false-positive head and advance to the
next genuinely-parked node.

Relates to the office-hours self-modification / clear-park design (for #2890).


natb1:
## Review

### Fixed
- `packages/intentionsutil/scripts/office-hours-graph:67` — `park_live_on_main`'s `git show`/`git fetch` ran at the ambient cwd instead of `$SCRIPT_DIR`, making the freshness guard invocation-directory-dependent. Fixed: `code-review-0`
- `packages/intentionsutil/scripts/office-hours-graph:99` — queue-head mode's `--list` read via unchecked process substitution swallowed a selector crash as an empty queue. Fixed: `code-review-1`
- `packages/intentionsutil/scripts/office-hours-graph:87` — targeted mode's `cleared $TARGET` fired for any `park_live_on_main` failure, conflating a genuine stale-worktree false positive with an absent/never-parked node. Fixed: `code-review-2`
- `packages/intentionsutil/scripts/office-hours-graph:68` — the park-state grep scanned the whole file (frontmatter + body), so a body line matching `^office_hours:` could be misread as park state. Fixed: `code-review-3`

All four consolidated into one edit — commit `9f52945c`.

### False-positive (security)
- `packages/intentionsutil/scripts/office-hours-graph:67` (`input-validation`) — probed whether a crafted node id could path-traverse the `git show origin/main:intentions/<id>.md` call (e.g. `intentions/../../../etc/passwd`). `git show` resolves paths against the repo tree, not the filesystem, so traversal outside the repo is not possible; no exploitable vector.

### Coverage
CodeQL: no open alerts for this PR. Dependency audit: skipped (diff does not touch `package.json`/`package-lock.json`). Erosion scan: no findings.

natb1:
## review · attempt 1
Review pass on PR #2891: 4 code-review findings fixed in one commit (9f52945c)
— scoped park_live_on_main's git calls to $SCRIPT_DIR (cwd-independence),
restricted the office_hours park test to the YAML frontmatter block, made
queue-head --list failures abort loudly instead of silently reporting empty,
and made targeted mode defer to the selector so absent/never-parked ids get
`not-parked` instead of a false `cleared`. One input-validation path-traversal
probe against park_live_on_main's git show call was investigated and found
non-exploitable (git show resolves against the repo tree, not the filesystem).
No CodeQL alerts, no dependency-audit surface, no erosion findings.
failed: none

PR #2886, opened 2026-07-18
Split /align-tactics model routing (Sonnet orchestrator + Opus decompose/plan subagent)

natb1:
Splits `/align-tactics` model routing so the decompose-to-signal judgment and
each claude-eligible tactic's plan-body authoring run on Opus even though the
orchestrator session itself runs on Sonnet, and pins `/align-strategy` to
whole-session Opus in frontmatter (its interview dialectic is non-delegable).

## Why

`strategy-token-economy` clarification 10 (2026-07-16) resolves the
align-family model routing: `/align-tactics` no longer runs whole-session on
Opus. Router-launched `/align-tactics` workers were authoring tactic plans on
Sonnet — the worker launches on Sonnet and its `Plan` subagent inherited that
model, so the highest-stakes-to-signal act (defining the tactic nodes) ran
cheap. `dispatch-graph-execute` already launches `/align-tactics` on Sonnet
(`ORCH_MODEL="sonnet"`, confirm-only, not re-implemented here); the residual
was the Opus decompose/plan subagent inside the skill, plus the
`/align-strategy` frontmatter.

## What changed

- `.claude/skills/align-tactics/SKILL.md` Step 3: the `Plan`/decompose
  subagent now launches with an explicit `model: opus`, independent of the
  Sonnet orchestrator session. The `Explore` reuse-hunt fan-out stays
  demotable to Sonnet/Haiku. A new Model routing note makes explicit that the
  decompose-to-signal judgment is the Opus subagent's work, and the Sonnet
  orchestrator transcribes/reconciles rather than rewriting plan substance.
- `.claude/skills/align-strategy/SKILL.md`: added `model: opus` to frontmatter,
  with a note that enforcement is intended-default-plus-measurement for
  `user-invocable` main-loop skills (confirmed honored only for `context: fork`
  skills), backed by `/dispatch-token-audit`'s by-node/by-phase attribution as
  the after-the-fact check.

## Verification

- `grep -q '^model: opus' .claude/skills/align-strategy/SKILL.md`
- `grep -qiE 'model:\s*opus' .claude/skills/align-tactics/SKILL.md`
- Manual/observe-in-production: after a router-launched `/align-tactics` round,
  use `/dispatch-token-audit`'s by-node/by-phase attribution to confirm
  `/align-tactics` orchestration ran on Sonnet while its plan-creation
  subagent ran on Opus, and that `/align-strategy` sessions ran on Opus.

Graph node: `tactic-align-family-opus-default`.

PR #2893, opened 2026-07-18
qa-fix: bail cleanly on Step 0 when node already office_hours-parked

natb1:
## Summary

qa-fix node-lane Step 0 bails cleanly (exit 0, no-op) when the target node is
already `office_hours`-parked at `origin/main` (tactic-qa-fix-office-hours-reentry-guard).

- The selector correctly gates *fresh* dispatch on `office_hours` being null,
  but an in-session resume — e.g. a stale self-scheduled `ScheduleWakeup`
  re-firing mid-session — bypasses the selector entirely and re-enters Step 0
  directly. Since parking a node sets `office_hours` without changing `phase`,
  such a resume would see `NODE_PHASE == "qa"` still true and could re-run
  qa-fix (re-merge, re-triage, re-post the QA summary) against a node already
  handed to a human for review.
- Observed live 2026-07-18 on `tactic-align-family-opus-default` (PR #2886):
  a stale wakeup re-fired after the Stop hook had already parked the node.
  That particular resume happened to recognize the work was done and took no
  action, but the skill had no mechanical guard against a less-conservative
  resume doing real damage.

## Units shipped

1. `.claude/skills/qa-fix/SKILL.md` — Step 0's node-target lane now parses
   `office_hours` from the already-fetched `$NODE_MD` right after the
   existing `NODE_PHASE` check. It greps for the literal unparked form
   `^office_hours: null$`; if absent (any parked form — nested block or a
   non-null inline value), it prints a clear message and exits 0. The legacy
   issue-branch lane (`dispatch:qa-done` label guard) is untouched.

## Verification

Prose-only per the plan (no automated test harness for skill markdown):

- Hand-constructed a `$NODE_MD` fixture with `phase: qa` and a nested
  (parked) `office_hours:` block — confirmed the guard's grep condition
  detects it and would skip with exit 0.
- Hand-constructed a `phase: qa`, `office_hours: null` fixture — confirmed
  it proceeds normally (no regression on the common case).
- Confirmed the legacy issue-branch lane's own `dispatch:qa-done` guard is
  untouched.


natb1:
## QA summary — tactic-qa-fix-office-hours-reentry-guard

No browser component (change is scoped to `.claude/skills/qa-fix/SKILL.md` prose/shell only). All items ran in the shell-command lane; no QA server, no browser session.

### Items executed

1. **Parked node (nested `office_hours:` block) triggers the skip** — script-verifiable — **PASS**
2. **Unparked node (`office_hours: null`) falls through** — script-verifiable — **PASS**
3. **Non-null inline `office_hours:` value also triggers the skip** — script-verifiable — **PASS**
4. **Guard block placed after NODE_PHASE check, before TARGET_KIND assignment** — script-verifiable — **PASS**
5. **Guard behaves correctly against a real parked node body on origin/main** — script-verifiable — **PASS**
6. **Whole-body grep is not anchored to frontmatter — potential false-negative** — needs-human-judgment (triage) — see disposition below

PASS: 5 / FAIL: 0 / SKIP: 0

### Disposition triage

- Item 6 — class **`opus-fixable`** (not aesthetic, verify: n/a). Rationale: real code defect at `qa-fix/SKILL.md` — the re-entry guard's `grep '^office_hours: null$'` ran against the *entire* node body (frontmatter + markdown), so a prose body line reading exactly `office_hours: null` would invert the intended bail and re-run qa-fix on a parked node. Anchoring the match to the YAML frontmatter block is the unambiguously-correct fix.

`needs-main` items: none this pass — nothing filed.

### Fixing pass — outcome

**Fixed this pass:** item 6 — anchored the `office_hours: null` check to the YAML frontmatter region (the block between the first and second `---` delimiters), so a prose-body line reading `office_hours: null` can no longer be mistaken for the node's actual state. Verified with three fixtures: a parked node (nested block) still exits 0; an unparked node (`office_hours: null` in frontmatter) still proceeds; and a parked node whose *body* contains a literal `office_hours: null` line now correctly exits 0 (the pre-fix false-negative case).

**Failed to land:** none.
**Deferred to re-QA:** none.
**Filed as follow-ups:** none.

This commit restarts CI; the chain re-derives `qa` and re-QAs the fixed build once CI passes.

natb1:

natb1:
## Review

### Fixed
- `.claude/skills/qa-fix/SKILL.md:202` — The re-entry guard's parked-check only grepped the top-level `office_hours: null` line, missing the squatter `attributes.office_hours` block form the canonical selection gate (`readParked` in `check-node-selection.ts`) also treats as parked. Reworked the guard to check both the first-class and squatter forms, using text-capture matching instead of a bare `grep -q` exit code. Fix commit: ff64ac1e.

### Informational
- `.claude/skills/qa-fix/SKILL.md:203` — The guard's early return is a bare `exit 0` with only a stderr message; no accompanying phase-log or marker write. Noted for reference — no action needed since this is the intended no-op re-entry path.

### Dismissed
_None._

### Deferred
_None._

### Out-of-scope (security)
_None._

natb1:
## review · attempt 1
Review pass: docs-only diff (single-file change to qa-fix's node-lane re-entry guard). Code-review finder found the guard's parked-check missed the squatter `attributes.office_hours` form the canonical selection gate also treats as parked; fixed to check both first-class and squatter forms with text-capture matching (commit ff64ac1e). One informational note (bare exit-0 return) left as-is, no action needed. No security findings (docs surface). failed: none

PR #2915, opened 2026-07-19
graph-commit: CAS-claimed landing lock to serialize concurrent land attempts (tactic-graph-commit-landing-lock)

natb1:
`graph-commit`'s `try_land()` runs up to `MAX_PUSH_ATTEMPTS` (default 5) of rebase → stamp (30-180s CI poll) → push to main. When two or more `graph-commit` invocations race to land, only one can win the final push, and the loser re-buys the full CI-stamp cost on every retry until it wins or exhausts its attempts. This was observed 3 times on 2026-07-19 — concurrent writers burned all 5 attempts with zero progress under pure landing contention on the single `main` ref, distinct from the already-handled content-merge case (`tactic-graph-commit-auto-serialization`).

This adds a CAS-claimed lock ref (`refs/graph/landing-lock`) with TTL-based steal that serializes the rebase→stamp→push critical section across concurrent invocations. A writer that can't claim the lock waits/backs off cheaply instead of burning a CI-stamp cycle it's very likely to lose. This is interim mitigation — it goes away when `tactic-graph-ref-split` restructures the ref layout to remove the contention at its root.

## Changes

- `packages/intentionsutil/scripts/graph-commit`: new landing-lock section (tunables, globals, `build_lock_commit`/`read_lock_payload`/`lock_claim_or_renew`/`lock_release`) wired into `try_land()`'s retry loop and `cleanup()`'s best-effort backstop. The lock lives at `refs/graph/landing-lock`, outside `refs/heads/`, so it never triggers the `graph/**`-branch fast-path workflow.
- `packages/intentionsutil/scripts/test-graph-commit.sh`: four new functional cases (contention is cheap — a waiting writer makes zero `gh` polls while blocked, then exactly one once it lands; dead-holder steal; live-holder wait with no premature steal; lock-ref hygiene) plus a new `blocked-green` shim mode to make the contention window deterministic, and an update to case 7's exhausted-attempts message assertion to match the reworded `N/M attempts` format.

## Verification

- `test-graph-commit.sh`: 31/31 passed (all pre-existing cases plus the four new lock cases).


natb1:
## QA summary — tactic-graph-commit-landing-lock (PR #2915)

No browser component was detected (diff touches only `packages/intentionsutil/scripts/graph-commit` and `test-graph-commit.sh`) — the entire QA plan ran in the shell-command lane.

### Items executed

1. **Landing-lock functional test suite passes (golden path)** — `bash packages/intentionsutil/scripts/test-graph-commit.sh` — **PASS** (31/31, including all 4 new landing-lock cases: contention-is-cheap, dead-holder steal, live-holder wait, lock-ref hygiene).
2. **Landing-lock ref never leaks to the remote** — `! git ls-remote --heads origin | grep -q 'landing-lock'` — **FAIL** at face value, but this is a false positive: the grep matched this PR's own working branch `refs/heads/tactic-graph-commit-landing-lock` (a coincidental substring collision), not the lock ref `refs/graph/landing-lock` under test. A precise check (`git ls-remote origin 'refs/heads/graph/**'` — 7 unrelated branches, none named landing-lock; `git ls-remote origin refs/graph/landing-lock` — empty, no lock currently held; exact `refs/heads/graph/landing-lock` match — none) confirms the real property holds, and the suite's own two "lock hygiene" cases already assert this exact property under a real bare-origin harness (both PASS in item 1).
3. **No lock-exhaustion events during real concurrent graph-commit use** — the node body flags this "not automatable in this harness," deferred to observing live invocation logs after the change is in active use.

**Counts:** 1 PASS, 1 FAIL (reclassified below), 1 deferred. No browser walkthrough ran (nothing needed it). No bugs found.

### Disposition triage

- Item 2 was assessed by Opus and reclassified **already-satisfied** — no code defect, no human input needed. See "Assessed by Opus (already satisfied)" below.
- Item 3 classified **needs-main** by the disposition triage (a planned-deferral: verifiable only by downstream observation, not at merge time). On this node-lane worktree, needs-main items normally become a `## needs-main residue` section on the tactic's own body, routing it to the autonomous `main-qa` phase handler. That routing is only used for items an autonomous, read-only web-browsing check can verify (a deployed public URL, element/text presence, console/network state). This item requires watching internal `graph-commit` invocation logs across a period of live concurrent use — not a browsable deployed surface — so per the conservative "uncertain → human" default it is **not** written as residue; it escalates to office-hours instead (see below).

- **Landing-lock ref never leaks to the remote** — No code defect. The finding shows the verification Command was a naive substring grep that collided with this PR's own branch name (`refs/heads/tactic-graph-commit-landing-lock`). Precise checks confirm the real property holds, and the test suite already asserts this exact property under a real bare-origin harness with both lock-hygiene cases passing (31/31). The criterion is provably met by passing tests; nothing to fix and no human needed.

### Deferred to office-hours

- **No lock-exhaustion events during real concurrent graph-commit use** — requires sustained observation of real `graph-commit` invocation logs after this change is in active use; not verifiable at merge time or via an autonomous browsing check. See the recommendation note left for the reviewer.

No opus-fixable residue existed this pass, so no auto-fix lane ran.

natb1:
## qa · attempt 0
failed: none (code-level).
test-graph-commit.sh: 31/31 PASS including all 4 new landing-lock cases.
1 QA-plan item reclassified already-satisfied (false-positive substring-grep check colliding with this PR's own branch name; real ref-namespace property confirmed clean).
1 item (downstream log-observation planned-deferral) escalated to office-hours — not autonomously browsable, no code change needed.

natb1:
# Fix-checks summary — PR #2915

## Iteration 1

- **Failed checks** — `hook-tests` (`packages/intentionsutil/scripts/test-graph-commit.sh`), 41 passed / 2 failed: `dead-holder steal` (case 30) and `live-holder wait` (case 31).
- **Outcome** — `fixed`
- **Reproduced** — yes (two distinct defects, one of which needed CI-equivalent conditions to surface)
- **Reproduce command** — `packages/intentionsutil/scripts/test-graph-commit.sh`, and for the second defect the same command under `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`

### Defect 1 — clone workdir collisions

Case 30 cloned into `$WORK/w10`, a directory already created and left dirty by
case 25 (`unrelated dirty tracked file`). `git clone` refused the non-empty
directory, so the following `git config user.email/user.name` calls never ran,
and `graph-commit` then refused to start on case 25's leftover diff:

```
fatal: destination path '/tmp/tmp.0V85kaXsx6/w10' already exists and is not an empty directory.
error: graph-commit: refusing to start — unrelated dirty tracked file(s) outside this call's node set:
 M intentions/t-dirty-preflight.md
 M packages/intentionsutil/src/store.js
```

Cases 29, 21 and 22 likewise re-cloned over `w8`/`w9`. Renumbered every
colliding workdir (`w18`–`w22`) so each case clones into a fresh directory.
Sentinel assertion strings are unchanged.

**Why not caught** — the collision only bites when the earlier case has already
run in the same `$WORK`, and the resulting `fatal:` lines were non-fatal to the
suite, so they read as noise rather than as the cause.

### Defect 2 — bare origin repo has no git identity (the CI-only failure)

`plant_lock` (cases 30/31) runs `git commit-tree` directly inside `$ORIGIN`, but
`$ORIGIN` is created with `git init --bare` and never configured with an
identity — unlike `$SEED`, which sets `harness@test` immediately after init.
CI runners have no global git identity, so `commit-tree` died there:

```
Author identity unknown
*** Please tell me who you are.
fatal: empty ident name (for <runner@...>) not allowed
fatal: : not a valid SHA1
```

`plant_lock` then produced an empty sha and `update-ref` planted nothing. Case
31 had no lock to wait on and landed in 1s instead of the planted 3s expiry —
and case 30 was passing **vacuously**, with no lock present to steal.

Configured the identity on `$ORIGIN`, and made `plant_lock` exit non-zero on an
empty sha so a missing lock is a loud failure rather than a silent vacuous pass.

**Why not caught** — this can never reproduce on a developer machine, which has
a global git identity. It reproduces exactly (42 passed / 1 failed, same case)
under `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`, which is how it
was confirmed before and after the fix.

### Fix

- `9c7c95df` — give every case its own clone workdir
- `0080ac3b` — give the bare origin repo a git identity (rebased to `5db27d01` after merging `origin/main`)

**Verification** — 43 passed / 0 failed locally both with and without a global
git identity; `hook-tests` green on CI, all 20 checks passing.

PR #2925, opened 2026-07-19
Fold strategy-graph-native-dispatch's superseded clarification chains into its body

natb1:
## Summary

Folds `strategy-graph-native-dispatch`'s 78 clarification entries: 66 settled
router-mechanism / superseded-in-place entries move into a new `## Router
Mechanism` body section (12 topic-organized subsections), each compressed in
the frontmatter `clarifications:` array to a short pointer of the form
`<question verbatim> — See body §<section>[ for the full mechanism]. Recorded
<date>.`. Per `kind-strategy`'s body-function rule (2026-07-09), settled design
notes belong in the node body, not the live clarification list — the strategy
had grown into "a design document wearing a strategy node." The remaining ~12
entries are left as full live entries: either genuinely undecided posture, or
settled doctrine legitimately out of this tactic's router-mechanism +
superseded-chain scope.

## Sections added under `## Router Mechanism`

- Phase Transitions & Fix State
- Fingerprint & Freeze
- Worktree Claiming & Liveness
- Pace, Backlog & Attention
- Serialization & Commit
- Recovery & Session Lifecycle
- Review & QA Disposition
- Execution Substrate
- Other Settled Mechanism

Each section synthesizes its topic's supersession chains into one
authoritative current-rule statement (with a brief history note for each
superseded link), plus standalone dated paragraphs for non-chain entries. Two
findings surfaced during the fold and are recorded in the relevant section
rather than fixed here, since they're out of this tactic's scope:

- A known defect (freeze re-surface over-sweeps the whole subtree instead of
  just stale-stamped children) — tracked as draft
  `tactic-freeze-resurface-stale-children-only`.
- An open discrepancy: entry 67's `/align-tactics` → `dispatch-plan` rename
  has not shipped (the skill directory is still `align-tactics`) — noted
  explicitly in §Other Settled Mechanism rather than stated as settled fact.

## Verification

- `npx tsx packages/intentionsutil/scripts/validate-graph.ts` — `ok — 403
  nodes` / `ok — prose refs: 0 unresolved (10 grandfathered by baseline)`.
- `npx vitest run --project packages/intentionsutil --root .` — 31 files, 571
  tests passed.
- Fingerprint re-stamp check (Unit 7): 40 open child tactics serving this
  strategy, all unstamped (`strategy_fingerprint` null) — nothing to
  re-stamp.
- Protected fields (`statement`, `rationale`, `success_signal`, `reading`,
  `gap`, `serves`) confirmed byte-for-byte unchanged from before this fold.


natb1:
Closing without merge: this tactic's fold content already landed on main out-of-band (the six Unit commits were pushed directly to main under SHAs 568a9a22…14063cea, no associated PR). This branch is now stale — its diff vs current main would revert unrelated later main commits, so it must not be merged. The node is being transitioned to done directly, since its work is already realized on main. See #2925 branch for the original per-unit history.

PR #2927, opened 2026-07-19
Thin qa-fix and review-fix SKILL.md bodies to the 500-line cap

natb1:
Thins the two worst over-cap phase-orchestrator SKILL bodies (`qa-fix` 1,563 and `review-fix` 1,172 lines) to under Claude Code's 500-line `SKILL.md` cap, per `strategy-token-economy` clarification 12's standup-cost lever (`tactic-thin-oversized-skill-bodies`).

The harness loads a skill's full `SKILL.md` on invocation and holds it for the whole session. Fixed procedure the orchestrator does not need in-context up front — detailed subroutines, reference tables, worked examples, edge-case enumerations — is moved into `references/*.md` files loaded only when the model Reads them (the standard Claude Code pattern). This is a pure structural relocation: no instruction content is dropped, every relocated section is linked from the body with an explicit `See references/<name>.md` pointer at its point of need.

## Changes

- `qa-fix/SKILL.md`: 1563 → 499 lines. Nine new `references/*.md` files carry the triage-subagent prompt contract, execution-lane detail, disposition-workflow schema, needs-main follow-up filing, the auto-fix lane, PR-comment composition, terminal-disposition escalation logic, node-lane target resolution, and the idempotency-preamble sidecar-stamp mechanics.
- `review-fix/SKILL.md`: 1172 → 434 lines. Seven new `references/*.md` files carry the node-lane re-keyed seams, inline-scan finder rosters, the disposition table, follow-up filing procedures, PR-comment composition, terminal-action sequencing, and per-finding schema/edge cases.

Every behavioral contract is preserved and reachable from the body: `qa-fix`'s disposition classes, attempt-capped auto-fix loop, `dispatch:qa-done` gate, and office-hours escalation; `review-fix`'s full fan-out (surface-conditional finders → dedup → classify → adversarial-verify → Opus fix fan-out → follow-up filing).

Scope boundary (per the tactic): only these two skills are thinned this round. The other five over-cap skills (`align-init`, `plan-issue`, `file-issue`, `align-tactics`, `align-strategy`) are out of scope — candidates for a follow-up if the pattern holds.

## Verification

- `awk 'END{exit (NR<=500)?0:1}' .claude/skills/qa-fix/SKILL.md` — passes (499 lines)
- `awk 'END{exit (NR<=500)?0:1}' .claude/skills/review-fix/SKILL.md` — passes (434 lines)
- Every reference file created is linked from its skill body (verified via grep)
[...]

PR #2939, opened 2026-07-22
Graph-write completion recipes: CAS-guard fix-checks and transition-node

natb1:
Implements the graph node `tactic-graph-write-recipes-base-cas`.

## Context

`graph-commit` does whole-file replacement of `intentions/<id>.md`. A
graph-write completion recipe that reads/mutates/writes the LOCAL on-disk node
file without first refreshing it from `origin/main` can silently clobber a
sibling frontmatter field that another writer already landed — no textual
conflict, no error. Observed 2026-07-22 on `tactic-thin-oversized-skill-bodies`
(PR #2927): a fix-push completion commit that should have touched only
`execution.fix.pushed_sha` also reverted a `blocked_by` link a prior iteration
had landed on `origin/main`.

The fixing primitive already ships: `graph-commit --base <id>=<blobsha>` is a
compare-and-swap that refuses the land if the node's blob moved on
`origin/main` after the base was pinned. `park-node` already uses it correctly
(fetch → resolve `FRESH_BLOB` → overwrite the local file from `origin/main` →
mutate → `graph-commit --base`). This PR brings the two callers that omitted
it up to the same standard.

## Changes

- **fix-checks node-lane completion recipe**
  (`.claude/skills/fix-checks/SKILL.md`) — the `apply-fix-state --record-push`
  + `graph-commit` recipe now refreshes the node file from `origin/main`
  before mutating it and passes `graph-commit --base`. Also fixed a
  pre-existing bug: the recipe referenced
  `.claude/skills/dispatch-propagate/scripts/graph-commit`, which does not
  exist — corrected to `packages/intentionsutil/scripts/graph-commit`.
- **`transition-node`**
  (`.claude/skills/dispatch-propagate/scripts/transition-node`) — added the
  same fetch/refresh/`--base` guard before its `apply-node-transition.ts` call
  and final `graph-commit` land. This transitively fixes every caller
  (`/implement`, `/qa-fix`, `/qa-main` node-lane completions).
- **Regression harness** (`packages/intentionsutil/scripts/test-transition-node.sh`)
  — modeled on `test-park-node.sh`: verifies a stale far-ahead worktree
  transition does not revert a concurrently-landed sibling field, and that a
  concurrent `origin/main` advance is refused (auto-parked) rather than
  silently clobbered. Wired into `.github/workflows/unit-tests.yml`'s
  `hook-tests` job.

`park-node` itself was already correct and is unchanged. Two same-class
omissions were confirmed but deliberately left out of scope for a future
audit (context-chunks/grounding-research/reading-review's bundled
new-node+edit `graph-commit` calls, and `demote-node-to-implement`) — see the
tactic node's rationale for detail.

## Test plan

- [x] `packages/intentionsutil/scripts/test-park-node.sh` — still green
- [x] `packages/intentionsutil/scripts/test-graph-commit.sh` — still green
- [x] `packages/intentionsutil/scripts/test-transition-node.sh` — new, green
      (verified to actually fail against a mutation-tested pre-fix copy of
      `transition-node`, confirming it's load-bearing)


natb1:
# Fix-checks summary — PR #2939

## Iteration 1

**Failed checks:** hook-tests, unit-tests

**Outcome:** flake

**Reproduced:** no

**Reproduce command:** `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`

**Failure excerpt:**

```
Test: select-tick on-main but primary checkout off-main → guard halts (exit 2)
  FAIL: guard-halt: exit 2 (the guard's exit)
    expected: '2'
    actual:   '0'
  FAIL: guard-halt: no selection decision line (tail not 'empty')
    expected: ''
    actual:   'empty'
  FAIL: guard-halt: disposition internal-error
    expected: 'internal-error'
    actual:   'empty'
  FAIL: guard-halt: skip_reason primary-checkout-not-on-main
    expected: 'primary-checkout-not-on-main'
    actual:   ''
Results: 3003/3007 passed, 4 failed
FAIL: test-dispatch-scripts.sh
```

**Why not caught:** The `select-tick` primary-checkout-guard-halt wiring test
in `test-dispatch-scripts.sh` failed identically in both the `hook-tests` job
(runs the file directly) and the `unit-tests` job (runs it via
`run-unit-tests.sh`'s script-suite loop) in this PR's CI run. It did not
reproduce across 4 local runs, including one through the CI-equivalent
`run-unit-tests.sh` path — 3007/3007 passed every time. This PR's diff
(`transition-node`, `fix-checks/SKILL.md`, `unit-tests.yml` job wiring, new
`test-transition-node.sh`) never touches `test-dispatch-scripts.sh`,
`lib.sh`, or the select-tick primary-checkout guard. The guard
(`assert_primary_checkout_on_main`) and this wiring test were added in a
tight recent commit cluster (`b8a1ba75`, `a18bf1c1`/#2929, `fe25242f`), and
the same test fixture already had one confirmed defect fixed once before
(`93a82322`: a shadowed case arm in the fake-git stub had made this exact
assertion vacuously pass) — an actively fragile, CI-environment-sensitive
hot spot unrelated to this PR's own changes.

**Flake issue:** `tactic-flake-select-tick-primary-checkout-guard-halt` (CREATED)

**Fingerprint:** `hook-tests — .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:22026`

PR #2938, opened 2026-07-22
graph-commit: resolve target repo from caller (cwd/-C), not script location

natb1:
## What

`graph-commit` (`packages/intentionsutil/scripts/graph-commit`) used to resolve the
repo it commits from **its own on-disk script location** — three directories up
from `SCRIPT_DIR` — never from the caller. Invoking it via a `../../../`-climbing
relative path from inside a worktree (as the emulated `clear-park` sequence did)
silently resolved to the **primary checkout's copy** of the script, so it staged
and "landed" in the wrong repo while printing a false success message. The
worktree edit stayed uncommitted behind a still-forbidden park state.

This PR:

- Adds a `-C <path>` (alias `--repo`) flag to `graph-commit`. `REPO_ROOT` is now
  resolved via `git -C <resolved> rev-parse --show-toplevel`, where `<resolved>`
  is `-C`/`--repo` when given, else the caller's cwd. `SCRIPT_DIR` is kept only
  for locating sibling scripts (`store.js`, `merge-node.ts`). A cwd/`-C` that
  isn't inside a git repo now dies with a clear message — never a silent
  fallback to the script's own checkout.
- Adds a fail-loud guard: when nothing is staged for a node id, the resolved
  repo's committed blob is compared against `origin/main`. Equal → benign
  (already landed out-of-band), proceeds as before. Differing → dies loudly,
  since under caller-derived resolution that state can only mean a mis-pointed
  `-C`/wrong checkout — this is exactly the scenario that used to print a false
  "landed".
- Updates the 4 internal callers that don't `cd` into their own target repo
  (`demote-node-to-implement`, `park-node`, `transition-node`,
  `reconcile-graph-merged`) to pass `-C "$REPO_ROOT"` explicitly.
- Fixes a stale path reference to `graph-commit` in `fix-checks/SKILL.md`.
- Extends `test-graph-commit.sh` with 4 new cases covering `-C` targeting from
  an unrelated cwd, the no-repo error, and both branches of the fail-loud
  guard, and wires the harness into CI (`unit-tests.yml`).
- Fixes a pre-existing test-authoring bug in `test-park-node.sh`'s
  concurrent-write-refusal case (its wrapper-install step left a tracked file
  dirty, tripping `graph-commit`'s unrelated-dirty-file guard and masking the
  refusal the case meant to exercise) — surfaced while re-running this PR's own
  Verification requirement that the harness pass; independently also fixed on
  `origin/main` in parallel and reconciled via merge.

## Test plan

[...]

PR #2937, opened 2026-07-22
Rename fix-conflicts skill to dispatch-conflict (zero behavior change)

natb1:
Implements `tactic-dispatch-conflict-rename`.

## Summary

Renames the `fix-conflicts` skill to `dispatch-conflict` with zero behavior change, per the dispatch-* skill-naming convention (strategy-graph-native-dispatch clarification 67). Split from tactic-dispatch-conflict-greenfield so its graph-native lane addition lands on an already-renamed file, in its own reviewable, mechanical PR.

**One design decision bounds this whole change:** only the *skill*/slash-command identity is renamed (`fix-conflicts` → `dispatch-conflict`). The dispatch **phase token** `fix-conflicts` is unchanged — it stays load-bearing across `dispatch-mark-complete`'s `--phase` argument, the `dispatch:fix-conflicts-attempt-<n>` label keyspace, `dispatch-phase-model`, `dispatch-preflight.sh`'s exemption list, `dispatch-ci-ready`, and the Stop hook's #831 non-advancement invariant.

- Unit 1: `git mv` the skill directory, rewrite frontmatter `name`/`description` and in-prose self-references.
- Unit 2: retarget `dispatch-recover-dispatched-phase`'s skill→phase map key (echoed phase-token value stays `fix-conflicts`).
- Unit 3: located the legacy worker-prompt emitter named in the plan — confirmed it was already deleted by `tactic-dispatch-legacy-rewire` before this tactic landed, so only a stale doc reference (`reference.md`) needed updating.
- Unit 4: swept remaining skill-name references across `qa-fix`/`review-fix` SKILL.md, `escalation-recommend.md`, `dispatch-tick`/`dispatch-graph-execute`/`provision-node-worktree` comments, the token-audit worker-classification alternation, `README.md`, and a stale intention-node path pointer — leaving every genuine phase-token site untouched.
- Unit 5: updated the recovery-map test's input skill tag, leaving the expected phase-token output unchanged.

## Test plan

- [x] `test -f .claude/skills/dispatch-conflict/SKILL.md` and old path gone
- [x] Frontmatter `name: dispatch-conflict`
- [x] `--phase fix-conflicts` and `dispatch:fix-conflicts-attempt` markers unchanged in the renamed skill
- [x] Recovery map routes `dispatch-conflict` → `fix-conflicts`
- [x] Full dispatch script suite: 2994/2994 passed

PR #2940, opened 2026-07-22
Harden cleanup_stale_worktree_processes tests against CI timing flake

natb1:
Fixes the intermittent CI flake in `test-pid-cleanup.sh`'s
`cleanup_stale_worktree_processes` test group (recurred on PR #2937 and
PR #2938, same failure signature both times, unrelated to either PR's own
diff).

The test harness synchronized on fixed 0.3s/0.5s sleeps between spawning a
background fixture process and scanning for it via repo-wide `ps` snapshots.
Under a loaded/throttled shared CI runner, `fork()`/`exec()` completion and
`ps` snapshot visibility aren't guaranteed to land inside that fixed window,
so the fixture process (or its forked child) could still be invisible when
the function under test ran its scan — a false failure unrelated to
`cleanup_stale_worktree_processes` itself.

Replaced the fixed sleeps with two polling helpers (`wait_for_pid_args`,
`wait_for_child_pid`) that wait for the actual condition — the fixture's PID
appearing in `ps -axo pid=,args=` with the expected path, or a child PID
appearing via `ps -axo pid=,ppid=` — up to a bounded 5s timeout, returning as
soon as the condition holds. Verified 3 consecutive local runs, 10/10 passing
each time, plus the merge with origin/main.


natb1:
## QA summary — PR #2940 (tactic-flake-pid-cleanup-stale-worktree)

No browser component detected (only `.claude/skills/dispatch-propagate/scripts/test-pid-cleanup.sh` changed) — all plan items ran in the shell-command lane; no QA server, no walkthrough, no GIF.

### Items executed (shell-command lane)

1. **Flaky test group passes when run directly** — PASS (`bash .../test-pid-cleanup.sh`, exit 0, 10 passed / 0 failed)
2. **Fix reliably passes across repeated consecutive runs** — PASS (5/5 consecutive runs, all exit 0, 10 passed / 0 failed each)
3. **Fixed sleeps replaced by polling helpers in target groups** — PASS (`wait_for_pid_args` / `wait_for_child_pid` defined and used at all 4 call sites inside the three `cleanup_stale_worktree_processes` test blocks; the remaining `sleep 0.3`/`0.5` lines belong to the out-of-scope `kill_worktree_processes` / `kill_tree` groups, not the target group — correctly scoped)
4. **Polling helpers are bounded and cannot hang** — PASS (bounded `for ((i = 0; i < attempts; i++))` loops, `return 1` on timeout; `wait_for_child_pid`'s timeout is handled by the caller via `|| CHILD_PID=""`, feeding the existing fail path)

**Counts: 4 PASS / 0 FAIL / 0 SKIP.** No bugs found by the shell-command lane.

### Disposition triage

Two items were deferred by the Step-2 triage as `needs-human-judgment` and routed through the disposition Workflow:

- **#5 — Timeout budget (5s / 100×0.05s) well-chosen for CI variance** — class: `opus-fixable`, verify: n/a. The classifier judged this resolvable from the diff with best judgment (not a subjective/aesthetic call). The gated fix-planner was invoked but returned an **empty unit list with no scope-deviation** — i.e. it found no actionable code change to make (the chosen 5s/100-attempt budget already appears reasonable against the diagnosed CI-throttling cause). Per the qa-fix auto-fix-lane contract, an opus-fixable item whose planner returns no units and no deviation is **not** auto-fixable this pass and escalates to office-hours as "planning failed" — this is not a bug per se, it's a judgment the auto-fix lane couldn't close out on its own.
- **#6 — No recurrence of the flake signature across subsequent CI runs** — class: `needs-main` (planned-deferral). Reason: absence-of-recurrence is a downstream/monitoring observation across future CI runs, not decidable at merge time from the diff alone.

`needs-main` items are normally filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set. **On this graph-native node lane, filing means appending a `## needs-main residue` section to the tactic's own body, which rides in the *next* state-only `transition-node` commit** — and this pass does not reach that commit (it escalates instead, on account of item #5). So **item #6 is recorded here for visibility only; it has not yet been committed to the graph.** A subsequent qa-fix pass — after office-hours resolves item #5 — will re-derive and commit it (the planned-deferral classification is deterministic, so nothing is lost by the delay).

### Terminal disposition: escalate to office-hours

No opus-fixable unit could be planned for item #5, so this pass escalates rather than fixes or completes. No `dispatch:qa-fix-attempt` label / attempt increment applies (per the qa-fix contract, a planning-failed escalation is not a bounded attempt).

---
**Update (human review):** items #5 (timeout-budget judgment) and #6 (planned-deferral no-recurrence check) were reviewed and marked skipped by the repo owner — no code change needed for #5, and #6 is not being tracked as a follow-up. The office-hours park was cleared and the tactic advanced qa → review directly.

natb1:
## qa · attempt 0
failed: none (4/4 script-verifiable items PASS: direct run, 5x repeat run, sleeps-replaced scoping check, bounded-loop-no-hang check).
changed: nothing this pass — no auto-fix applied.
escalated: item #5 (timeout-budget judgment) classified opus-fixable but the gated fix-planner returned zero units with no scope-deviation — nothing actionable to auto-fix. Escalating to office-hours as "planning failed".
deferred (not yet committed to graph): item #6 (needs-main, planned-deferral — absence-of-flake-recurrence is a downstream/monitoring observation) — will be recorded on a subsequent completing pass.


## review · attempt 1
## review · attempt 1
failed: none
found: 1 finding (code-review, Fixed) — bare `wait_for_pid_args` calls under `set -euo pipefail` would abort the suite unlabeled on timeout instead of a labeled FAIL.
fixed: appended `|| fail "..."` to the three bare `wait_for_pid_args` calls (3af97f43).
deferred: none.
security: no attack surface — test-only diff.

natb1:
## Review

**Fixed**
- `test-pid-cleanup.sh:192-193,293` — the new `wait_for_pid_args` polling helper was called bare under `set -euo pipefail`; a timeout would abort the whole suite with an unlabeled `set -e` exit instead of a labeled `FAIL`. Appended `|| fail "..."` to each of the three bare calls, mirroring the existing `|| CHILD_PID=""` graceful pattern used for `wait_for_child_pid`. (3af97f43)

**Required (security)**
_None._

_None._

**Informational**
_None._

**Dismissed**
_None._

**False-positive (security)**
_None._

**Deferred**
_None._

**Out-of-scope (security)**
_None._

**Security review**: no attack surface — test-only diff (every changed path is a test file).

PR #2921, opened 2026-07-19
graph: explicit single-node dispatch entrypoint (tactic-graph-explicit-node-dispatch)

natb1:
## Summary

Gives the graph-native dispatch pipeline an explicit single-node dispatch entrypoint — `dispatch <node-id>` — that reuses the existing `dispatch-tick` → `dispatch-select-tick` → `dispatch-graph-execute` pipeline instead of the direct-call workaround, and skips the autonomous pace-curve throttle and `MAX_WORKERS` ceiling while still honoring genuine token exhaustion.

- **Unit 1** — `graph-select-target --node <id>`: an explicit single-target selection mode, mutually exclusive with `--top`/`--pace-exempt-only`, reusing every existing claim (`reservation_exists`, `worktree_has_live_session`) and sensor (`sensor_gate`) gate verbatim. Distinguishes "present but gated" (reason echoed to stderr) from "absent entirely" (not found/done/parked/blocked/reviewed).
- **Unit 2** — threads a node-id-shaped positional through `dispatch-select-tick` and `dispatch-tick`: bypasses the pace-curve/MAX_WORKERS gate for an explicit node dispatch (the tactic's actual purpose — without this, the explicit path would be throttled identically to the ranked autonomous path), while still hard-stopping on genuine rate-limit exhaustion. A selectable-but-gated or entirely-absent node yields a `node-not-selectable` decision line rather than silently falling through to the JIT-reminder/main-broken aux triggers. Also fixes the header comment in `nix/packages/dispatch.nix`, which still described the deleted legacy issue-number behavior.
- Adds `test-dispatch-scripts.sh` coverage for the new `--node`/`NODE_ARG` paths at both the `dispatch-select-tick` and `dispatch-tick` layers (usage-error combos, not-selectable, exhaustion, and pace-gate bypass).

Routing through the existing pipeline (rather than calling `dispatch-graph-execute` directly) also resolves a known false-`exit-12` stale-selection failure: the existing `BRANCH == "main"` sync in `dispatch-select-tick` now runs before target selection whenever the tick runs from the main checkout, with no new sync code needed.

## Node

tactic-graph-explicit-node-dispatch

## Test plan

- [x] `test-dispatch-scripts.sh` — full suite, 2988/2988 passed
- [ ] Manual: `dispatch <node-id>` from the main checkout with the pace curve pinned to zero and live workers at/over `max_concurrent_workers` still spawns
[...]

PR #2952, opened 2026-07-23
dispatch-select-tick: sweep the reservation ledger before explicit-node dispatch

natb1:
Wires the existing reservation-ledger sweep into `dispatch-select-tick`'s explicit-node dispatch branch (the `elif [[ -n "$NODE_ARG" ]]` path added by `tactic-graph-explicit-node-dispatch`), so a stale reservation marker left by a dead session no longer refuses a human's explicit node dispatch with `node-not-selectable`.

## What changed

- **`dispatch-select-tick`**: after the branch's existing rate-limit-exhaustion early-exit, source `lib-claude-agents.sh` and `lib-reservation-ledger.sh` and call `reservation_sweep 1>&2 || true` — mirroring the autonomous (no-arg) block's own sweep-before-count pattern verbatim. The `--manual` branch is untouched (out of scope — its non-sweep behavior affects fan-out pacing math, not a hard refusal).
- **`test-dispatch-scripts.sh`**: new regression test in the explicit-node-id test group plants a stale reservation marker (owned by a session absent from the fake live-session registry), then asserts the tick's decision is a normal graph selection (not `node-not-selectable`) and that the marker is no longer owned by the dead session.

## Context

Surfaced by `/review-fix` on PR #2921 (`tactic-graph-explicit-node-dispatch`), disposed `deferred` at the time. Finalized by `/align-tactics` as `tactic-explicit-node-reservation-sweep-policy`: a stale dead-session marker silently defeats the whole point of explicit human dispatch overriding the autonomous pace curve, so the fix (option 1 — run the sweep) was chosen over merely documenting the gap.


natb1:
## QA summary — clean pass

No browser component in this PR (only `dispatch-select-tick` and its bash test
suite changed) — the plan was 4 script-verifiable checks + 1 human-judgment
policy item.

### Results

| # | Item | Result |
|---|------|--------|
| 1 | Full dispatch test suite passes | **PASS** — `test-dispatch-scripts.sh`: 3024/3024 passed, 0 failed, including the new "stale reservation reclaimed by sweep" test |
| 2 | Sweep call wired into the explicit-node branch | **PASS** — `source lib-claude-agents.sh` / `source lib-reservation-ledger.sh` / `reservation_sweep 1>&2 \|\| true` confirmed present inside the `elif [[ -n "$NODE_ARG" ]]` branch (lines 756–759) |
| 3 | New regression test asserts real behavior, not a tautology | **PASS** — the test plants a stale `dead-sess`-owned marker, then asserts BOTH the decision line is normal graph selection (not `node-not-selectable`) AND the marker is no longer `dead-sess`-owned afterward |
| 4 | Scope containment — `--manual`/autonomous branches untouched | **PASS** — diff is purely additive (10 + 35 lines, 0 deletions), confined to the explicit-node branch |
| 5 | Design judgment: sweep-vs-document policy ratification | **needs-main** (see Disposition triage below) |

PASS: 4, FAIL: 0, SKIP: 0.

No bugs found. No browser walkthrough ran (no browser component detected).

### Disposition triage

Residue: 1 item (#5), classified by the qa-fix disposition Workflow.

- **#5 — Design judgment: sweep-vs-document policy ratification** → class
  `needs-main` (planned-deferral: the acceptance criterion — human ratifies
  the option-1 sweep policy and its `\|\| true` non-fatal semantics — is
  documented as verified downstream, not at this PR's merge; per the
  planned-deferral rule this is `needs-main`, not `opus-fixable`, regardless
  of its subjective-policy character). `verify: n/a` (excluded from the
  skeptic fan-out, per the planned-deferral rule).

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and
dropped from the escalation set, so they do not park this PR. On this **node
lane**, filing means appending a `## needs-main residue` section directly to
the tactic's own body (`intentions/tactic-explicit-node-reservation-sweep-policy.md`)
rather than filing a GitHub issue — that section rides in this pass's
`transition-node` commit, and routes the tactic through `main-qa` once
`review` completes, where `tactic-main-qa-phase` owns verification (falling
back to an `office_hours` park if it turns out not autonomously verifiable).

No `opus-fixable` or `needs-human` residue this pass — nothing escalates.

No `already-satisfied` items this pass.

### Terminal disposition

Non-fixing, clean pass: escalation set is empty (residue was entirely
`needs-main`, already filed as body residue). Advancing the qa phase via
`transition-node`.

natb1:
## qa · attempt 0
failed: none — all 4 script-verifiable checks passed (full test suite 3024/3024,
sweep call wired into explicit-node branch, new regression test asserts real
sweep behavior not a tautology, `--manual`/autonomous branches untouched).
1 planned-deferral item (sweep-vs-document policy ratification) classified
needs-main and filed as a `## needs-main residue` body section — routes
through main-qa once review completes.


## review · attempt 1
failed: none — 9 findings surfaced (code-review + input-validation/secrets/red-team/security-review
finders), 6 actionable, 5 fixed in commit 6384de59: guarded the sweep's `|| true`
against a silent load failure, documented the sourcing's second effect on
`emit_graph_selection`'s `reservation_write`, made the new stale-reservation
regression test assert sweep-only evidence (was previously insensitive to the
sweep it claimed to cover — confirmed by mutation testing), and tagged the
explicit lane's reservation claim `origin=explicit` with a matching TTL-reclaim
sweep rule so a dead-without-retry claim can't strand forever. 1 finding
refuted (adversarial-verify), 1 false-positive, 1 out-of-scope (no follow-up —
not a net-new gap). No deviation; no deferred follow-ups filed.

natb1:

natb1:
## Review-fix summary

Combined finder fan-out (code-review, security-review, input-validation, secrets, red-team) over the diff (`dispatch-select-tick`, `test-dispatch-scripts.sh`) surfaced 9 findings; 6 were actionable and 5 fixes were applied in commit `6384de59`.

### Fixed

- **[code-review] `dispatch-select-tick:759`** — `reservation_sweep 1>&2 || true` silently swallowed exit 127 on a failed `source`, skipping the ledger sweep with no signal. Replaced with a guard that logs a clear stderr message on load failure while keeping the non-fatal contract (mirrors `dispatch-tick`'s existing guard pattern).
- **[code-review] `dispatch-select-tick:756`** — the sourcing added by this tactic has a second, undocumented effect: it's also what makes `emit_graph_selection`'s `reservation_write` call work on the explicit-node lane at all (previously an undefined-command no-op). Comment extended to record this, plus a regression test asserting the lane now writes a marker.

- **[input-validation] `test-dispatch-scripts.sh:23470`** and **[red-team] `test-dispatch-scripts.sh:23480`** (same underlying defect, two finders) — the new "stale reservation reclaimed by sweep" test was insensitive to the sweep it claims to cover: both assertions passed identically with the `reservation_sweep` call surgically removed (verified empirically by both adversarial skeptics via direct mutation-and-rerun). Root cause: the fake `graph-select-target` never consults the ledger, and `emit_graph_selection`'s unconditional `reservation_write` overwrites the marker's `session=` field regardless of whether the sweep ran. Fixed by capturing the tick's stderr and asserting on the sweep's own `reclaimed reservation ... (dead-session-stranded)` note, plus a second stale marker under a never-selected id to prove reclaim rather than incidental overwrite.
- **[code-review residue] `test-dispatch-scripts.sh:23480`** — same test-insensitivity defect, independently surfaced by the residue disposition pass. Already resolved by the fix above (same commit); no additional edit needed.
- **[red-team] `dispatch-select-tick:758`** — sourcing `lib-reservation-ledger.sh` unconditionally on the explicit-node lane mints a claim with no `origin=` tag, so `reservation_sweep`'s TTL-reclaim rule (which only applies to `origin=standalone`) can't reclaim it if the reserving session later goes dead without retry — a stranding gap the `--manual` lane doesn't have with the same shape (confirmed by adversarial verify: upheld). Fixed by tagging the explicit lane's claim `origin=explicit` and widening the sweep's TTL rule to reclaim `origin=explicit` markers the same way, plus correcting a stale comment and adding a regression test.

- **[red-team] `dispatch-select-tick:759`** — claimed `reservation_sweep 1>&2 || true` silently swallows a load failure with real exploit value. Skeptic verdict: refuted — the pattern is not novel (identical `|| true` sweep call already exists twice in this same file, one pre-dating this PR by commits), the new call site is human-invoked/single-shot (not a systemic autonomous-tick blackout like the sibling guarded call site it was compared to), and the finder itself flagged low confidence. Reasonable as a low-priority follow-up to align all three call sites, not a blocking defect in this diff.

### False-positive (security)

- **[red-team] `dispatch-select-tick:759`** — claimed the sweep is "unscoped, machine-wide" despite the lane consulting the ledger for exactly one node. Classified false-positive: `reservation_sweep`'s cost (one `claude_agents_list_all` round-trip) is identical to the sibling autonomous-block sweep call this mirrors, and is bounded/best-effort by design.

### Out-of-scope (security)

- **[red-team] `dispatch-select-tick:759`** — claimed the sweep flips the lane from fail-safe-refuse to fail-open-reclaim, enabling a double-dispatch race onto an already-claimed worktree. Classified out-of-scope: this is the same fail-open reclaim semantics the autonomous sweep path already has; not a new risk introduced by wiring the existing sweep into the explicit-node lane. No follow-up filed (not a meaningful net-new gap).

No deferred filings — every actionable finding was resolved within this PR.

PR #2961, opened 2026-07-23
office_hours.session_type: closed enum, soft ranking penalty, --type filter, 34-node backfill

natb1:
Implements node `tactic-office-hours-session-type`: adds a closed-enum
`office_hours.session_type` field (`requirement-discovery` |
`curriculum-review` | `other`, default `other`) to the parked-node record,
plus soft-penalty type-aware ranking and a `--type` filter in the
office-hours selector.

## Unit 1 — Schema

`SessionType` type + `SESSION_TYPES` const in `schema.ts`, non-optional
`session_type` field on `OfficeHours`, `validateOfficeHours` default/validate
via `requireOneOf`. Fixed the one runtime test-expectation break this
introduces in `schema.test.ts`.

## Unit 2 — Repair typecheck breaks

Added `session_type: "other"` to every direct `OfficeHours`-typed object
literal across the test suite (`office-hours.test.ts`, `coverage.test.ts`,
`router.test.ts`, `scope-sweep.test.ts`, `graph-census-debt.test.ts`,
`strategy-fingerprint.test.ts`, `check-node-selection.test.ts`) and one
production write path (`apply-fix-state.ts`'s `applyParkCheck`).

## Unit 3 — Ranking

`officeHours.ts` gains `SESSION_TYPE_PENALTY = 0.5`, `QueueMember.sessionType`,
and a soft penalized-rank computation (`requirement-discovery`/
`curriculum-review` parks rank at half their raw attention; `other` parks
are unaffected). Both `officeHoursQueue` and `selectOfficeHours` gain an
optional `sessionType` filter param.

## Unit 4 — CLI

`office-hours-select.ts` gains a `--type <value>` flag (validated against
`SESSION_TYPES`, mutually exclusive with a positional node-id), and the
`--list` output now includes a session-type column.

## Unit 5 — Tests

Added schema tests (accept/default/reject `session_type`) and ranking tests
(soft-penalty ordering, boosted overtake, exact penalized-rank value, type
filtering) in `schema.test.ts` and `office-hours.test.ts`.

## Unit 6 — Backfill

Set `session_type` on the 34 currently-parked nodes that already have a
natural type: `strategy-recover-attention` → `requirement-discovery`; the 33
`tactic-reading-chunk-*` / `tactic-dialog-review-*` parks →
`curriculum-review`.

## Test plan

- [x] `npx vitest run --project packages/intentionsutil --root .` — 618/618 passing (34 files).
- [x] `npx tsc --noEmit -p packages/intentionsutil/tsconfig.json` — only the pre-existing baseline error (`graph-census-debt.test.ts:108`, unrelated to this change).
- [x] `npx tsx packages/intentionsutil/scripts/validate-graph.ts` — ok, 362 nodes, 0 unresolved prose refs.

Graph node: `tactic-office-hours-session-type`


natb1:
# Fix-checks summary — PR #2961

## Iteration 1

**Failed checks:** type-safety-sensor (x2 runs)

**Outcome:** fixed

**Reproduced:** yes

**Reproduce command:** `.github/scripts/check-type-safety-escapes.sh` (run from repo root after `git fetch origin main`)

**Failure excerpt:**
```
::error file=packages/intentionsutil/scripts/office-hours-select.ts,line=128::Net-new type-safety escape hatch (as <Type> cast). Suppress with `// type-safety-ok: <reason>` if intentional.
::error file=packages/intentionsutil/scripts/office-hours-select.ts,line=134::Net-new type-safety escape hatch (as <Type> cast). Suppress with `// type-safety-ok: <reason>` if intentional.
```

**Why not caught:** The PR's own added code introduced two `as SessionType` casts in `office-hours-select.ts` (`SESSION_TYPES.includes(typeValue as SessionType)` and `sessionType = typeValue as SessionType`), which the diff-scoped `type-safety-sensor` check correctly flags as net-new escape hatches. `ReadonlyArray<T>.includes()` requires its argument to already be of type `T`, which is why the original code reached for a cast; this is avoidable, not an intentional escape hatch, so no suppression comment was warranted.

**Fix:** Replaced `SESSION_TYPES.includes(typeValue as SessionType)` with `SESSION_TYPES.find((t) => t === typeValue)`, using the `found === undefined` narrowing (TypeScript treats `process.exit(2)` as returning `never`, so control-flow analysis removes `undefined` from `found` after that branch) to assign `sessionType = found` with no cast needed. Commit `d67d33e9` (`fix: replace as SessionType casts with cast-free .find() in office-hours-select`), merged with `origin/main` at `39ac4d7d` and pushed.

Verified: `.github/scripts/check-type-safety-escapes.sh` exits 0 clean; `npx tsc --noEmit -p packages/intentionsutil/tsconfig.json` shows only the pre-existing baseline error at `graph-census-debt.test.ts:108`; `npx vitest run --project packages/intentionsutil --root .` — 619/619 passing.

natb1:
# QA Summary — PR #2961 (tactic-office-hours-session-type)

**Pass 3** (re-QA after pass 2's `parseSelectorArgs` fix landed and CI went green). No browser component detected (backend-only change to `packages/intentionsutil` + `intentions/*.md` data). All items ran in the shell-command lane.

## Items executed (7 total)

**Script-verifiable (4 PASS, 2 FAIL):**

1. **`--list` output column shift breaks downstream consumers** — **FAIL** (real, severe bug). `office-hours-graph` (`packages/intentionsutil/scripts/office-hours-graph:122-123`) parses `--list`'s tab-separated output positionally (`read -r score nid date`), matching the *old* 3-column format. The diff changed `--list` to 4 columns (`rank\tsessionType\tnodeId\tsince`) without updating this consumer. Confirmed by direct simulation and end-to-end run: `office-hours-graph` with no target misreads every one of 111 parked nodes' session-type string as the node id, fails every lookup, and reports "queue is empty" even though 111 nodes are genuinely parked. This is the graph-native `/office-hours` entry point. `office-hours-graph` itself has zero diff lines in this PR — the break is an unavoidable side effect of Unit 4's `--list` column-format change on an untouched consumer.
2. **End-to-end CLI selection by positional node id (pass-2 regression check)** — **PASS**. Confirmed fixed.
3. **`--type` filter over real graph data** — **PASS**. Partition exhaustive/disjoint (1+33+77=111).
4. **Malformed `--type` argv handling** — **FAIL** (minor). `--type` with no value produces `unknown --type "undefined"` — the literal string "undefined" leaks into the message. Still exits 2, no crash, no silent fallthrough.
5. **Backfill data integrity (34 edited `intentions/*.md`)** — **PASS**.
6. **Soft penalty ranking on live queue data** — **PASS**.

**needs-human-judgment (1, routed through disposition triage):**

7. `selectOfficeHours(nodes, target, sessionType)` silently ignores `sessionType` when `target` is set — library API design question.

**PASS/FAIL/SKIP:** 4 PASS, 2 FAIL, 0 SKIP (script-verifiable lane). 1 item routed through disposition triage.

## Deferred to office-hours

All 3 disposition items (item-1, item-4, item-7) — see Disposition triage below. This pass could not auto-fix them (attempt cap reached).

## Bugs found

- **item-1**: `office-hours-graph`'s `--list` consumer reads a stale 3-column format; the real 4-column output desyncs the node-id field, breaking the graph-native office-hours queue-head walk entirely (reports "empty" with 111 real parks).
- **item-4**: `office-hours-select.ts`'s `--type` error message leaks the literal string `"undefined"` when no value follows the flag.

## Disposition triage

Residue list was non-empty (3 items), classified by the Step 3.5 Workflow:

- **item-1** (office-hours-graph column-shift regression) — class: **opus-fixable**, aesthetic: false, verify: n/a. Rationale: confirmed in-tree, mechanical single-file fix (add the `sessionType` field to the `read`); no main/prod state or human intent needed.
- **item-4** (`undefined` leaking into `--type` error message) — class: **opus-fixable**, aesthetic: false, verify: n/a. Rationale: deterministic local code defect with an obvious fix (detect a missing/flag-shaped value and emit a distinct usage message); readable from source alone.
- **item-7** (`selectOfficeHours` silently ignores `sessionType` when `target` is set) — class: **opus-fixable** (downgraded from `needs-human`), verify: **Refuted** (2/2 skeptics refuted). Rationale: the JSDoc already documents and confirms the current behavior is intentional, and the repo's own `.claude/rules/code-style.md` ("prefer clear errors over defensive fallbacks... input validation at system edges... encouraged") directly answers the hardening question — add a guard that throws when both `target` and `sessionType` are passed, rather than escalate for a decision the codebase's own convention already resolves.

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR. No item classified `needs-main` this pass.

**Non-fixing pass:** all 3 `opus-fixable` residue items escalate to office-hours below — this pass hit the **qa-fix attempt cap** (`ATTEMPT_N=2` == `CAP=2`), so `plan_fix` was false and no fix plan was authored (a cap-reached case, not a planning failure). The next fixing pass can only run after the attempt-cap window resets (e.g. via a manual attempt-label reset), or a human resolves these directly.

## Filed follow-ups

None this pass — no item classified `needs-main`.

natb1:

natb1:
## qa · attempt 0
qa-fix pass 4: 13/13 script-verifiable items PASS (no browser component; backend-only change). Re-verified all 3 items Pass 3 escalated to office-hours — office-hours-graph's --list column-read (4-var), --type missing-value message ("undefined" leak), and selectOfficeHours's target/sessionType mutual-exclusion guard — all confirmed fixed in the current diff. Also re-confirmed the attempt-1 positional-targeting fix and the type-partition/backfill/soft-penalty behavior on live graph data (115 parks: 1 requirement-discovery, 33 curriculum-review, 81 other). Full suite 664/664, tsc clean (only pre-existing baseline error), validate-graph ok. No residue. failed: none.


## qa · attempt 1
qa-fix pass 2: 4/5 script-verifiable items PASS. 1 real bug found and fixed: office-hours-select.ts's positional-arg filter silently dropped the node-id target whenever --type was absent (typeIdx+1 computed unconditionally, defaulting to index 0); every bare `office-hours-select.ts <node-id>` call fell through to queue-head mode instead. Fixed by extracting argv parsing into a pure exported parseSelectorArgs and guarding the +1 exclusion on typeIdx !== -1; 8 new tests added. 1 needs-human-judgment item (SESSION_TYPE_PENALTY=0.5 tuning) classified needs-main by disposition triage (planned-deferral — only observable across repeated real /office-hours sessions) and filed as a `## needs-main residue` section in this node's own body. failed: item-2 (office-hours-select.ts positional bug) — fixed this pass.


## qa · attempt 2
qa-fix pass 3: 4/6 script-verifiable items PASS, 2 FAIL. item-1 (severe): office-hours-graph's `--list` consumer (packages/intentionsutil/scripts/office-hours-graph:122-123, a file untouched by this PR) reads a stale 3-column format — the diff's new 4-column `--list` output (rank/sessionType/nodeId/since) desyncs the node-id field, breaking the graph-native /office-hours queue-head walk entirely (reports "queue is empty" despite 111 real parks). item-4 (minor): `--type` with no value leaks the literal string "undefined" into the CLI error message. 1 needs-human-judgment item (item-7: selectOfficeHours silently ignores sessionType when target is set) was reclassified opus-fixable by disposition triage (2/2 skeptics refuted — the repo's own code-style.md convention on public-API guards already answers the question). All 3 disposition items are opus-fixable but this pass hit the qa-fix attempt cap (ATTEMPT_N=2 == CAP=2), so no fix plan was authored (cap-reached, not a planning failure) — escalating all 3 to office-hours. failed: item-1 (office-hours-graph --list column desync), item-4 (undefined leak in --type error), item-7 (selectOfficeHours API-contract hardening) — none fixed this pass, cap reached.


## review · attempt 1
review pass: 15 findings across 9 finders + CodeQL (0 alerts) + erosion (5). 4 actionable, 3 fixed in 433bb6e6, 1 deferred. Fixed: office-hours-select.ts --type=<value> form was silently dropped by the exact-token args.indexOf match, returning the UNFILTERED queue head with exit 0 (now normalizes --flag=value, rejects unknown --flags with exit 2, 6 new tests, live-verified); the stale 3-column --list contract at :29-30 that office-hours-graph:239 cited as authoritative (re-pointed at the canonical 4-column block by name, not line number); extracted formatQueueRow with a column-order test. Deferred: --type unreachable from office-hours-graph (arg 1 parsed as node id, no flag forwarding) -> draft tactic node tactic-office-hours-graph-type-passthrough landed on main in 9b873800. All 7 Required findings refuted by adversarial verify (5 red-team, 2 erosion) - notably the session_type oversight-evasion angle, refuted because no production write path sets the field (park-node omits it, defaulting to the unpenalized "other"). Out-of-scope: unescaped tab-delimited --list output vs unrestricted node-id charset (pre-existing id-validation gap). Also merged origin/main (branch was 81 behind); apply-fix-state.ts conflicted because main replaced --park-if-capped with check-cap/reset-attempt - took main's side, so that file drops out of the diff. 698/698 tests, tsc clean but for the pre-existing graph-census-debt.test.ts:108 baseline, validate-graph ok. failed: none.

natb1:
## Office-hours recommendation — `tactic-office-hours-session-type` (PR #2961)

Three residue items are pending; the qa-fix attempt cap (2/2) stopped the autonomous lane before a fix plan could be authored. Two are mechanical, one needs your decision. Suggested order: decide item-7 first (it's the only judgment call), then land all approved fixes in one commit.

### item-1 — fix first, this is a live regression on the branch

**File:** `packages/intentionsutil/scripts/office-hours-graph`, line 123.

```bash
  local score nid date
  while IFS=$'\t' read -r score nid date; do
```

Unit 4 of this PR changed `office-hours-select.ts --list` from 3 columns to 4 (`rank\tsessionType\tnodeId\tsince`; the contract is documented at `office-hours-select.ts:29`), but this consumer still reads 3 fields. `$nid` therefore receives the session-type string, every node lookup fails, and the queue-head walk reports "empty" while 111 nodes are parked. `office-hours-graph` is the graph-native `/office-hours` entry point, so the branch currently cannot pick a next node.

**Fix:** make it a 4-field read matching the new column order.

```bash
  local score sessiontype nid date
  while IFS=$'\t' read -r score sessiontype nid date; do
```

(Note the lowercase local name — `.claude/rules/` has a zsh path-clobber caution about loop variable names; `sessiontype` is safe.)

Grep the tree for other `--list` consumers: line 118 of `office-hours-graph` is the only invocation outside the selector's own header comments and `parseSelectorArgs`. So this one edit closes the item — no other positional readers of the old 3-column shape exist. Worth adding a comment at the read site pointing at the `office-hours-select.ts:29` column contract so the next column change is caught.

**Verify:** run `packages/intentionsutil/scripts/office-hours-graph` with no target and confirm it names a real parked node id rather than printing "empty".

### item-4 — small UX fix, no decision needed

**File:** `packages/intentionsutil/scripts/office-hours-select.ts`, `parseSelectorArgs`, lines 120–121 and the error at line 142.

When `--type` is the last argv token, `args[typeIdx + 1]` is `undefined` and gets interpolated verbatim: `office-hours-select: unknown --type "undefined" (expected: ...)`. Exit code is still 2 and nothing crashes; it just reads as though the user typed the word "undefined".

**Fix:** before the unknown-value branch at 137–142, detect a missing or flag-shaped value and emit a distinct message:

```ts
if (typeIdx !== -1 && (typeValue === undefined || typeValue.startsWith("--"))) {
  return { kind: "error", message: `office-hours-select: missing value for --type (expected: ${SESSION_TYPES.join(", ")})` };
}
```

Add a `parseSelectorArgs` unit case for bare trailing `--type` and for `--type --list`.

### item-7 — needs your explicit decision before any code changes

**File:** `packages/intentionsutil/src/officeHours.ts`, `selectOfficeHours` at lines 117–131; the behavior is documented at lines 103–113 and re-stated in the inline comment at 123.

Today, when `target` is set, `sessionType` is silently ignored. That is deliberate and documented, and the CLI already makes the combination unreachable (`--type` is rejected alongside a positional at `office-hours-select.ts:132–133`). The two adversarial skeptics reclassified this from needs-human to opus-fixable on the grounds that `selectOfficeHours` is exported from `packages/intentionsutil/src/index.ts` — a public API boundary — and `.claude/rules/code-style.md` encourages guards at such boundaries.

They're arguing a contract change, not a bug fix, so pick one explicitly:

- **(a) Keep silent-ignore.** Defensible: the JSDoc at 103–113 already spells the behavior out, the plan called it intentional, and no in-repo caller can reach the combination. If you take this, no code change is needed beyond optionally tightening the JSDoc wording at line 113 to say *why* (targeting is by id, type is a queue-ordering concern only) — then mark item-7 resolved-as-designed in the PR comment so a future review pass doesn't relitigate it.
- **(b) Add the guard.** Throw when both `target` and `sessionType` are non-undefined, add a unit test asserting the throw, and rewrite the 103–113 JSDoc block plus the 123 inline comment to describe the new contract. This is a breaking change to an exported function's contract — cheap now (no other callers), but it is the kind of thing worth deciding once rather than drifting into.

My read: (a) is the lower-risk close given the CLI already forecloses the combination and the behavior is documented rather than accidental — but this is exactly the call the escalation is asking you to make, so state it either way rather than leaving it implicit.

### Getting the PR merged

Items 1 and 4 are small, single-file, and low-risk — implement both directly in this session (or in one subagent), plus item-7 per your decision. Then either:

- **Push the fix commits directly.** The review/qa chain re-runs from CI green; this is the shortest path and avoids the attempt-cap machinery entirely.
- **Or reset the qa-fix attempt window** (clear/decrement the `dispatch:qa-fix-attempt-2` marker) and let the autonomous Opus lane implement all three, if you'd rather the chain own the fix. Only worth it if you want the automated re-QA record; item-1 is a one-line change you can verify by eye faster than a chain round-trip.

Before executing whichever disposition you choose, re-check `origin/main` — the fleet may have landed the `office-hours-graph` column fix out-of-band while this node sat parked, in which case item-1 is already closed and you should not re-commit it.

natb1:
# QA fixes applied — pass 3 residue (human-directed)

The pass-3 escalation parked this node with three `opus-fixable` items that the
attempt cap (`ATTEMPT_N=2 == CAP=2`) had prevented the autonomous lane from
fixing. All three are now fixed in `725277a5`, the park is cleared, and the
`dispatch:qa-fix-attempt-2` label is removed so the QA lane can re-enter at
attempt 0.

`origin/main` was merged first (the branch was 45 commits behind), then each fix
was applied against current main — item-1's read site had moved from line 122 to
line 245 in the interim.

## item-1 — `--list` column desync (severe) — FIXED

`packages/intentionsutil/scripts/office-hours-graph` read three tab-separated
fields from `office-hours-select.ts --list`, which Unit 4 changed to four
(`rank\tsessionType\tnodeId\tsince`). `$nid` received the session-type string, so
every node lookup failed and the queue-head walk reported an empty queue while
109 nodes were genuinely parked.

Now a four-field read, with a comment at the read site pointing at the column
contract documented in `office-hours-select.ts`.

**Verified:** the fixed read against live `--list` output yields real node ids
(`tactic-manual-path-reservation-sweep`, `tactic-dispatch-test-monolith-split`,
`tactic-office-hours-session-type`), each of which resolves to an existing
`intentions/*.md` file, across all 109 rows. `bash -n` clean.

## item-4 — `undefined` leaked into the `--type` error (minor) — FIXED

`parseSelectorArgs` now detects a missing or flag-shaped `--type` value before
the unknown-value branch and emits a distinct message.

**Verified:**

- `--type` → `office-hours-select: missing value for --type (expected: ...)`, rc=2
- `--type --list` → same message, rc=2
- `--type bogus` → `unknown --type "bogus"`, rc=2 (unchanged)

Two `parseSelectorArgs` unit cases added, one asserting the message no longer
contains the string `undefined`.

## item-7 — `selectOfficeHours` sessionType/target contract — FIXED (guard added)

Took the guard option rather than the documented-silent-ignore option. The
function is exported from the package index, and `.claude/rules/code-style.md`
asks for input validation at public API boundaries — silently dropping one of
two explicitly-supplied arguments hides a caller mistake behind a plausible
result. Passing both `target` and `sessionType` now throws; the JSDoc and the
inline comment were rewritten to describe the new contract.

No in-repo caller is affected: the CLI already rejects `--type` alongside a
positional node-id at argument-parsing time, so the throw is reachable only from
a non-CLI caller. One unit case added asserting the throw.

## Loop check

Not looping. Each pass surfaced distinct new defects rather than re-surfacing an
unresolved set — pass 1: two JSDoc/comment items; pass 2: the `typeIdx + 1`
positional-argument bug; pass 3: the column desync, the `undefined` leak, and the
API-contract question. Pass 3 re-verified pass 2's fix as still passing. That is
forward progress, so the attempt counter was reset rather than the node being
left parked.

## Verification

- `npx vitest run --project packages/intentionsutil --root .` — 36 files, 664 tests, all pass
- `run-lint.sh` — all lint checks pass (including the prose-rule linter)
- `validate-graph.ts` — ok, 377 nodes, 0 unresolved prose refs
- Full CI on `725277a5` — 23/23 SUCCESS

`run-typecheck.sh --app packages/intentionsutil` skipped locally on a
pre-existing `origin/main` typecheck baseline failure unrelated to this change;
CI's `typecheck` job passes.

## Note for the graph tooling

Clearing the park took two `clear-park` runs. The first landed only the additive
body section and silently dropped the `office_hours` removal, reporting success —
the known `graph-commit` layer-2 field-removal gap. The second run, with the
additive change already on main, landed the removal cleanly. Worth knowing that
`clear-park` is not reliably single-shot when the node also carries an unlanded
body addition.

natb1:

Scope edits absorbed since phase start (ee234a197d58b80e48122cb4c193559927eed90a..origin/main):
  44ff773a graph: clear office_hours park on tactic-office-hours-session-type (2026-07-25 second attempt — first clear-park run landed only the additive body section and dropped the office_hours removal (graph-commit layer-2 field-removal gap).)
  b4fe4af5 graph: clear office_hours park on tactic-office-hours-session-type (2026-07-25 qa-fix pass-3 residue resolved: office-hours-graph now reads the 4-column --list contract, --type emits a missing-value message instead of leaking 'undefined', and selectOfficeHours throws when target and sessionType are both supplied (commit 725277a5 on PR #2961, unit-tested, lint green). qa-fix attempt label reset to 0 to re-enable the QA lane.)
  56d6487f graph: park tactic-office-hours-session-type (/qa-fix: qa-fix attempt cap reached (ATTEMPT_N=2 == CAP=2) with 3 opus-fixable residue items pending (severe office-hours-graph --list column-desync regression, minor undefined-leak in --type error message, and an API-contract hardening decision) — no fix plan could be authored this pass; escalating to office-hours)

natb1:
# QA Summary — PR #2961 (tactic-office-hours-session-type)

**Pass 4** — fresh qa-fix pass (attempt-label window reset since Pass 3; the
three items Pass 3 escalated to office-hours were confirmed fixed in the
current diff — see below). No browser component detected (backend-only
change to `packages/intentionsutil` + `intentions/*.md` data). All 13 plan
items ran in the shell-command lane (no browser needed).

## Items executed (13 total, all script-verifiable)

1. **`office-hours-graph` resolves the correct node after the `--list` column change** — **PASS**. Re-verified the previously-severe Pass-3 FAIL: the consumer's `read -r score sessiontype nid date` (4-var) now correctly matches the 4-column `--list` output; `nid` resolves to a real node.
2. **No other consumer of `--list` output still parses 3 columns** — **PASS**. `office-hours-graph` is the only programmatic consumer repo-wide; it's on the 4-column contract.
3. **`--list` golden path over real graph data** — **PASS**. 115 well-formed 4-column rows, rank non-increasing, every node id resolves to a real file.
4. **`--type` filter partitions the live parked set exhaustively/disjointly** — **PASS**. 1 (requirement-discovery) + 33 (curriculum-review) + 81 (other) = 115, no leaks.
5. **`--type` with a missing value gives a clear message, never "undefined"** — **PASS**. Re-verified the previously-minor Pass-3 FAIL: `--type` as the last token, and `--type` followed by another flag, both emit "missing value for --type" with no "undefined" leak, exit 2. (An initial shell-variable word-splitting artifact in the test harness — zsh doesn't word-split unquoted `$VAR` like bash — produced a false FAIL on first attempt; re-run with explicit argv confirmed correct behavior.)
6. **`--type` with an unrecognized value is rejected with the allowed set** — **PASS**. `bogus-type`, `Curriculum-Review` (casing), and empty string all rejected, exit 2, message lists all three valid values.
7. **Mutually-exclusive argv combinations are refused** — **PASS**. `--type <t> <node-id>` and `--list <node-id>` both exit 2 with "mutually exclusive" (same shell-harness artifact as item 5 on first attempt; explicit argv confirmed correct).
8. **`selectOfficeHours` throws when given both `target` and `sessionType`** — **PASS**. Re-verified the hardened Pass-3 item 7: 6/6 `selectOfficeHours`-scoped vitest cases pass, including the mutual-exclusion throw.
9. **Queue-head selection + soft penalty on live data** — **PASS**. 34 penalized (session-typed) parks, 81 `other`; the visible top-5 `--list` rows are all `other` (consistent with the penalty lowering typed parks' effective rank), and typed parks remain present in the queue (soft, not hard, penalty).
10. **Positional node-id targeting still works (attempt-1 regression check)** — **PASS**. Targeting a `curriculum-review` node and an `other` node both resolve to that exact node regardless of type; a nonexistent id correctly returns `empty not-parked <id>` per the documented disposition contract.
11. **Un-backfilled parks default to `other`; graph still validates** — **PASS**. 81 parked-but-unbackfilled files (no `session_type:` line) exactly match the 81-row `other` bucket; `validate-graph` passes (382 nodes, 0 unresolved prose refs).
12. **Backfill touched exactly the 34 intended files, one line each** — **PASS**. `git diff --numstat` confirms +1/-0 per file across 34 files; 33 `curriculum-review` + 1 `requirement-discovery`; each line lands inside the existing `office_hours:` block at the correct indentation.
13. **Full suite / typecheck / graph validation reproduce a clean baseline** — **PASS**. `vitest run --project packages/intentionsutil`: 664/664 passing (36 files). `tsc --noEmit`: only the pre-existing baseline error (`graph-census-debt.test.ts:108`, unrelated to this change). `validate-graph`: ok.

**PASS/FAIL/SKIP:** 13 PASS, 0 FAIL, 0 SKIP.

## Deferred to office-hours

None. No residue.

## Bugs found

None this pass. (One non-code observation: piping `--list`'s per-line stdout into a truncating consumer like `head` triggers an unhandled EPIPE crash — but this write loop is unchanged in structure by this diff and would crash identically on `origin/main`; out of scope for this PR's QA.)

## Disposition triage

Not run — residue list was empty (all 13 items PASS), so the Step 3.5 disposition Workflow was skipped per the skill's own gating.

## Filed follow-ups

None this pass.

natb1:
## `/review-fix` — review pass

15 findings surfaced across 9 finders + 2 inline scans (CodeQL, erosion); 4 actionable; 3 fixed in `433bb6e6`; 1 deferred to a draft tactic node. No unresolved high-confidence Required findings.

**Note on the base:** this branch was 81 commits behind `origin/main` when the pass started. `origin/main` had replaced the `--park-if-capped` mode this PR edited with `check-cap` + `reset-attempt`, so `apply-fix-state.ts` conflicted; resolved by taking main's side (the PR's only edit there added `session_type: "other"` to a function main deleted). `apply-fix-state.ts` correctly drops out of the diff. Verified the two new `office_hours` literal writers main added — `graph-census-debt.ts:171` and `hold-node-decide.ts:163` — omit the now-non-optional `session_type` safely, since `validateOfficeHours` defaults absent values to `"other"`.

### Fixed (3) — commit `433bb6e6`

- **`office-hours-select.ts:126`** — `parseSelectorArgs` matched `--type` only via exact-token `args.indexOf`, so the `--type=<value>` form was dropped into `positionals` (it starts with `--`) and `sessionType` stayed `undefined`, silently returning the **unfiltered queue head with exit 0**. Now normalizes `--flag=value` → `--flag value`, rejects unrecognized `--`-prefixed tokens with stderr + exit 2, rejects a value on the boolean `--list`, and treats empty `--type=` as a missing value. Six unit cases added. Live-verified: `--type=curriculum-review` now launches the curriculum park (previously an `other` park); `--typ` exits 2.
- **`office-hours-select.ts:29`** — the header carried two contradictory statements of the `--list` contract (stale 3-column at :29-30, canonical 4-column at :40-41), and `office-hours-graph:239` cross-referenced the *stale* line as authoritative. Replaced the stale restatement with a pointer to the canonical block, and re-pointed the `office-hours-graph` comment by searchable name rather than a brittle line number.
- **`office-hours-select.ts:191`** — extracted `--list` row rendering into an exported pure `formatQueueRow(m: QueueMember): string`, with a test asserting exact tab-separated column order both as a whole string and as split fields.

### Deferred (1)

- **`office-hours-graph:56`** — `--type` is unreachable from the ergonomic entry point: `office-hours-graph` parses arg 1 as a node id and forwards no flags, so `office-hours-graph --type curriculum-review` sets `TARGET="--type"` and aborts under `set -e`. Real gap, but the tactic's plan scopes the change to "the office-hours selector" and never mentions `office-hours-graph` plumbing — a deliberate scope boundary. Filed as draft tactic node **`tactic-office-hours-graph-type-passthrough`** (landed on `main` in `9b873800`), carrying full provenance, scope, and verification. Node lane files draft nodes, not GitHub issues.

### Refuted (adversarial-verify dropped) (7)

All 7 `Required` findings were refuted by severity-scaled skeptics (2 for high-confidence, 1 below). Not fixed; rationale retained for the audit trail.

- **`officeHours.ts:45`** (`red-team-0`, 1 skeptic) — "oversight evasion by self-declared `session_type`". Refuted: no production write path sets the field. `park-node:243` constructs `office_hours` without it, so it defaults to `"other"` — the *unpenalized* tier. The attack needs hand-edited raw YAML, a generic "agent with commit rights can write false graph state" concern that applies equally to `blocked_by`/`phase`/`attention.override`.
- **`schema.ts:610`** (`red-team-1`, 1 refuted / 1 upheld) — `session_type` silently defaulted rather than preserved across a re-park. Code-level claim accurate, but did not clear the bar; the write-side default is the plan's explicit, reviewed decision (Unit 2: "Leave as-is").
- **`officeHours.ts:47`** (`red-team-2`, 1 skeptic) — penalty falls on the classes most needing a human. Refuted: contradicted by `strategy-attention-surface`'s 2026-07-23/25 clarifications, where the author *requested* deprioritizing exactly these classes.
- **`office-hours-select.ts:131`** (`red-team-3`, 2 skeptics) — the `--type=` parse bug, raised as a *security* finding. Refuted **as security** (local offline CLI, no trust boundary, operator is both typist and victim, impact self-correcting) while explicitly acknowledged as a valid quality nit — and it **was** fixed on that basis as `residue-1` above.
- **`office-hours-graph:239`** (`red-team-4`, 2 skeptics) — stale cross-reference, refuted as security; likewise fixed as a quality nit (`residue-0`).
- **`office-hours-select.ts:128`** / **`officeHours.ts:38`** (`erosion-0`, `erosion-1`) — complexity increases refuted as the proportionate, direct cost of the new feature rather than net erosion.

### Informational (3)

- `schema.ts:638` — total file complexity 193 → 194.
- `office-hours.test.ts:21` — 54 → 92 (added test coverage).
- `schema.test.ts:879` — 267 → 274 (added test coverage).

### Out-of-scope (security) (1)

- **`office-hours-select.ts:192`** (`input-validation`) — the `--list` record is emitted as unescaped tab-delimited text and now carries four positional columns parsed by `office-hours-graph:250` via `IFS=$'\t' read -r score sessiontype nid date`. `validateNode` only requires `id` be a non-empty string — no charset or whitespace restriction — so a node id containing a tab would desync the column parse. Pre-existing weakness in the id-validation contract that the diff widens rather than introduces; below the meaningfulness threshold for a security follow-up (no CodeQL/npm identifier to key on), so recorded here rather than filed.

### Scans

- **CodeQL** — 0 open alerts on `refs/pull/2961/head`.
- **npm audit** — not run (`deps=false`; no `package.json`/lock changes).
- **erosion** — 5 findings (2 refuted, 3 informational).
- Coverage complete: all 9 finders returned, 0 errored.

### Verification

- `npx vitest run --project packages/intentionsutil --root .` — **698/698 passing**, 37 files (691 before the fixes; 7 new tests).
- `npx tsc --noEmit -p packages/intentionsutil/tsconfig.json` — clean apart from the pre-existing `graph-census-debt.test.ts:108` baseline error this PR's body already documents.
- `npx tsx packages/intentionsutil/scripts/validate-graph.ts` — ok, 404 nodes, 0 unresolved prose refs.

PR #2966, opened 2026-07-23
reservation_sweep before the pause-sentinel short-circuit (heartbeat reaper dormancy)

natb1:
Implements `tactic-heartbeat-sweep-before-pause`.

## Problem

`dispatch-tick` honors the pause sentinel by exiting before it ever reaches
`dispatch-select-tick`. The autonomous `reservation_sweep` lives **inside**
`dispatch-select-tick`, so a paused heartbeat never reaps the reservation
ledger. In a long pause with no manual ticks (operating mode (c) of the
2026-07-23 cross-mode ledger-validity clarification on
`strategy-graph-native-dispatch`), dead-session orphan markers accumulate
until scheduling resumes.

Per that clarification, the pause sentinel gates worker *spawning*, never
ledger *bookkeeping*: the timer-driven heartbeat must reconcile the ledger
**before** honoring the pause. This is the pause-independent reaper
(migration step ii); the sibling `tactic-manual-path-reservation-sweep`
covers the manual-read path (mode (b)), and the autonomous selection path
already sweeps (mode (a)).

## Change

- `dispatch-tick` now sweeps the reservation ledger immediately before the
  pause-sentinel check, on the autonomous path only (`--manual` is
  unaffected — it already reaches `dispatch-select-tick`'s manual block).
  Sources `lib-claude-agents.sh` and `lib-reservation-ledger.sh` (guarding
  against a redundant re-source) and calls `reservation_sweep 1>&2 || true`
  so a sweep failure never blocks or fails the tick.
- `test-dispatch-scripts.sh`'s `dispatch-tick` pause-sentinel test group now
  copies the two lib files into the scratch tmpdir (so the tick's
  `SCRIPT_DIR`-relative `source` calls resolve for real instead of silently
  failing), and the existing "paused (sentinel present)" test plants a stale
  dead-session reservation marker and asserts it is reclaimed even though
  the tick still exits without scheduling.

## Verification

- `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — 3041/3041 passed.


natb1:
## QA summary (attempt 0)

No browser component detected (pure shell-script change to `dispatch-tick` /
`test-dispatch-scripts.sh`) — all plan items were script-verifiable or
needs-human-judgment; no browser walkthrough ran.

**Result: 7 PASS, 0 FAIL, 1 deferred → filed as a `needs-main` follow-up.**

### 1. Full dispatch script test suite passes — PASS
3041/3041 passed, 0 failed. The new assertion "paused: pre-pause sweep still reclaimed the dead-session marker" ran and passed.

### 2. A paused autonomous tick reclaims a dead-session reservation marker and still does not schedule — PASS
rc=0, pause announced, marker reclaimed, nothing spawned.

### 3. The sweep does not over-reap (live session survives; failed/empty registry read does not wipe the ledger) — PASS
Live session: marker survives. Registry command error: marker survives (fail-safe). Genuine empty `[]` registry with an aged-past-grace marker: reclaimed — this is expected `reservation_sweep` design (already relied on by the existing autonomous-path sweep call in `dispatch-select-tick`), not a regression from this PR (`lib-reservation-ledger.sh`/`lib-claude-agents.sh` are untouched by the diff).

### 4. The `--manual` path is unaffected — PASS
Marker untouched on `--manual`; the new block is guarded by `[[ -z "$MANUAL" ]]` (`dispatch-tick:268`).

### 5. A failing or missing sweep never blocks or fails the tick — PASS
Both an erroring `reservation_sweep` stub and missing lib files still leave the tick at rc=0 with the pause correctly announced.

### 6. Sweep output stays on stderr — tick's stdout contract is unchanged — PASS
Paused-run stdout is exactly the pause line; unpaused-run stdout is unchanged from before this PR. All sweep diagnostics land on stderr.

### 7. The `declare -f reservation_sweep` re-source guard is correct — PASS
Source order (agents then ledger) matches `dispatch-select-tick`'s convention; double-sourcing both libs is harmless.

### 8. Real timer-environment sweep effectiveness across live pause/resume cycles — deferred, filed as follow-up
Classification: needs-human-judgment, Flag: planned-deferral.

## Disposition triage

- **Item 8** — class: `needs-main`, verify: `n/a` (planned-deferral item, excluded from the skeptic candidate set). Reason: effectiveness depends on the live systemd-timer environment's ability to read the Claude session registry (`claude agents --json`), which a scratch scaffold cannot reproduce; verify by observing reservation-ledger size and worker health over the next few live pause/resume cycles.

No `opus-fixable` items — this pass fixed nothing. No `needs-human` items — nothing is deferred to office-hours.

## Filed follow-ups

Node-target lane: item 8 was **not** filed as a GitHub issue. A `## needs-main residue` section was appended to the tactic's own body (`intentions/tactic-heartbeat-sweep-before-pause.md`) instead, riding in the phase-transition commit. The reconciler routes this tactic `qa → main-qa` on completion; `tactic-main-qa-phase` owns the downstream verification (observing ledger size / worker health over live pause/resume cycles).

None.

natb1:
## qa · attempt 0
qa: 7/7 script-verifiable items PASS (full test suite re-verified 3041/3041;
pause-then-reap behavior, over-reap safety, --manual isolation, sweep-failure
tolerance, stdout contract, re-source guard all confirmed against the actual
scripts). 1 planned-deferral item (real timer-environment sweep effectiveness
across live pause/resume cycles) classified needs-main and filed as a
`## needs-main residue` section on the tactic's own body — no GitHub issue
filed on this node lane. failed: none.


## review · attempt 1
review: 10 findings surfaced (5 finders: code-review, input-validation, secrets,
red-team, security-review) across the pre-pause reservation_sweep diff; 6 fixed
(2 real security/robustness issues confirmed by adversarial verify — test-env
ledger/daemon isolation in tick_setup, and a silently-swallowed `|| true` on the
sweep call — plus 4 code-review residue items: sweep scoped to the paused branch
only, unchecked source calls now diagnostic-guarded, a pinned sweep clock for the
new test). 4 findings refuted by adversarial verify (sandboxed-tool-invocation
artifacts and a documented, tested pause-contract carve-out, not real production
exploits). No CodeQL alerts, no dependency changes, no erosion findings. Full
suite re-verified 3041/3041 after fixes. No deviation, no follow-ups filed.
failed: none.

natb1:

natb1:
## Review

Surface: `code`. 19 subagents (5 finders + 2 dedup + 1 classify + 6 verify skeptics + fix/residue passes). 10 findings surfaced, 6 fixed, 4 refuted. No open CodeQL alerts; no dependency changes; no erosion findings.

### Fixed (code-review residue)

- `test-dispatch-scripts.sh:23018` — `tick_setup` now exports `CLAUDE_AGENTS_CMD` / `DISPATCH_RESERVATION_DIR` stubs so no tick test reaches the real shared ledger/daemon.
- `dispatch-tick:265` — the pre-pause sweep block moved from an unconditional `[[ -z "$MANUAL" ]]` guard into the actual paused branch, so it fires only when the pause sentinel is honored, not on every autonomous tick.
- `dispatch-tick:269` — the two `source` calls now check their own success and emit a greppable stderr warning on failure instead of degrading silently.
- `test-dispatch-scripts.sh:23121` — the planted stale marker now pairs with a pinned `DISPATCH_RESERVATION_SWEEP_NOW_EPOCH`, removing a real-wall-clock dependency in the sweep-age check.

Fix commit: `93d0400b`.

- **`test-dispatch-scripts.sh:23002`** — `tick_setup` copied the two sourced libs into `TMPDIR_TEST` but never isolated `DISPATCH_RESERVATION_DIR` / `CLAUDE_AGENTS_CMD`, so the new pre-pause `reservation_sweep` in every tick test (not just the one paused test that set them) read/wrote the real shared reservation ledger and queried the real `claude` daemon. Adversarial verify (2/2 skeptics upheld) confirmed this against live filesystem state — a real production reservation marker was reclaimed by an unrelated local test run during verification. Fixed by isolating both inputs in `tick_setup` itself.
- **`dispatch-tick:275`** — `reservation_sweep 1>&2 || true` swallowed every non-zero status including "command not found" (exit 127) from an unloaded function, so a broken lib load degraded to a silent no-op. Fixed with an explicit `declare -f reservation_sweep` check that runs the sweep unsuppressed or emits a loud, greppable stderr line when the load failed.

Fix commit: `93d0400b`.

- `dispatch-tick:275` (input-validation) — claimed a sandboxed `claude agents --json` call could cause mass-reclaim of live reservations. Refuted: every wired production invocation of the autonomous tick path (the `dispatch-heartbeat` systemd unit, `dispatch-spawn-tick`) runs outside any Claude Code Bash sandbox by design; triggering the failure mode requires deviating from the documented invocation contract (running the raw script sandboxed without `--manual`), which is a pre-existing, already-documented hazard (`.claude/rules/sandbox.md`) generic to every `lib-claude-agents.sh` consumer, not a defect this diff introduces.
- `dispatch-tick:271` (input-validation) — claimed an unchecked `source` failure degrades with "no journal evidence." Refuted: reproduced the exact code shape in bash — a missing-file `source` and a "command not found" dispatch both write directly to stderr independent of `|| true`, which only swallows the exit status, and `dispatch-tick` is journald-supervised, so both lines land in the journal verbatim.
- `dispatch-tick:268` (red-team) — claimed the pre-pause sweep silently violates the pause-sentinel contract, framed as an "elevation of privilege" via PATH-forwarding to a systemd unit. Refuted: the adjacent code comment and a dedicated test both document this as the intended, reviewed carve-out (ledger bookkeeping is exempt from the pause, worker spawning is not); the PATH-resolved `claude agents --json` exec is pre-existing fleet-wide architecture already reachable on every unpaused tick, and the escalation path requires a precondition (attacker-controlled PATH at an already-unsandboxed call site) that already grants the attacker everything the exploit claims to gain.
- `dispatch-tick:269` (red-team) — claimed the `declare -f reservation_sweep` guard is an env-injectable trust decision enabling privilege escalation. Refuted: the systemd units that launch the autonomous tick forward only a sanitized `PATH`, never an arbitrary function definition; there is no `PassEnvironment`/`import-environment` mechanism anywhere in the codebase that could inject a shell function into the unit's environment. The guard is a code-smell (a cleaner unconditional-source idiom exists elsewhere in the same file) but not a materialized vulnerability.

### Informational / Dismissed

_None._

### Deferred / Out-of-scope

_None._ No follow-up issues filed — no meaningful out-of-scope or deferred findings this pass.

---
Verification: full suite re-run after fixes — 3041/3041 passed.

PR #2958, opened 2026-07-23
Fix playwright_install_with_deps retry: kill process tree before retrying

natb1:
Fixes the `playwright_install_with_deps` retry loop in `.claude/skills/dispatch-propagate/scripts/lib.sh` so a stalled attempt's whole process tree is killed before retrying — previously `timeout --kill-after=30` only signalled the direct `npx`/`node` child, leaving reparented `apt-get`/`dpkg` grandchildren alive to hold `/var/lib/dpkg/lock-frontend`, so attempt 2 died instantly on the lock instead of getting a real retry (observed on PR #2946 CI).

## Changes

- Added `wait_for_dpkg_lock`: a best-effort helper that waits (bounded, degrading safely when `flock`/the lock file are unavailable) for the dpkg frontend lock to release before a retry.
- Rewrote the install loop to background the bounded install, run a watchdog that kills the whole process tree via the repo's existing `kill_tree` (ps-walk based; not `setsid`/process-group kill, which the sandbox blocks) once the real per-attempt timeout elapses, then call `wait_for_dpkg_lock` before the next attempt.
- Added unit tests: 4 for `wait_for_dpkg_lock`, 1 new stall test proving the tree is actually killed and the retry is real; the 5 pre-existing `playwright_install_with_deps` tests still pass unchanged.

Full suite: `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — 3030/3030 passed, 0 failed.

Graph node: `tactic-playwright-install-retry-lock`.

Manual verification still needed on a real CI run that exercises a timeout (cannot be reproduced locally — needs genuine `apt-get`/`dpkg` mirror contention): confirm attempt 2 starts from a clean dpkg lock and actually re-runs `apt-get` rather than dying instantly, and check whether `wait_for_dpkg_lock`'s shared `flock` probe reads the root-owned lock or hits `EACCES` as the non-root runner (if so, file a follow-up to try `sudo flock`).

References:
- #2946: https://github.com/natb1/commons.systems/pull/2946

PR #2946, opened 2026-07-23
Inject bounded ancestry projection into every node-assigned session

natb1:
## Summary

Implements `tactic-node-ancestry-context`: injects a bounded ancestry projection (parent chain + `serves` chain up to virtue roots) as read-only decision context into every node-assigned session, via one owned primitive.

- **Unit 1** — `packages/intentionsutil/scripts/node-ancestry.ts` (+ tests): a BFS walk over `parent`/`serves` edges from a node, bounded by a hop cap (64), a per-ancestor clarification-title cap (20), and a total byte cap (24 KB), rendered as read-only Markdown. Cycle-safe via a visited set; dangling edges are skipped with a note, never throw.
- **Unit 2** — wires the primitive into `provision-node-worktree`: writes `.claude/ancestry-context.md` into every freshly-provisioned worktree, best-effort/non-fatal (a failed write warns to stderr but never blocks provisioning). Ignored via a new `.gitignore` entry so it's never accidentally committed.
- **Unit 3** — documentation-only: instructs every consuming session entry (implement, fix-checks, qa-fix, review-fix, qa-main, align-tactics, align-strategy, office-hours) to read or generate the projection near its existing node-read step, with a shared discipline clause: the projection is read-only decision context, the node body stays the sole work contract, and a perceived plan-vs-ancestry conflict parks to `office_hours` rather than silently absorbing scope.

## Test plan

- [x] `npx vitest run node-ancestry` (packages/intentionsutil) — 7/7 pass
- [x] `npx vitest run` (packages/intentionsutil full suite) — 600/600 pass
- [x] CLI smoke test against the real store (stdout + `--out` write) — both pass
- [x] `bash -n` on the edited `provision-node-worktree` — syntax OK
- [ ] Manual: run `provision-node-worktree` on a real tactic and confirm `.claude/ancestry-context.md` lands untracked in the worktree (deferred to qa/main-qa — requires a live worker session)

PR #2967, opened 2026-07-25
Fix playwright watchdog to not leak an orphan sleep child on fast-path cancel

natb1:
Fixes `playwright_install_with_deps`'s watchdog so it doesn't leak an orphan `sleep` child when the main body cancels it on a fast success or non-stall failure.

## Problem

The watchdog subshell in `.claude/skills/dispatch-propagate/scripts/lib.sh` ran `sleep "$timeout_s"` as a plain non-final command, so bash forks it as a real child process rather than exec'ing into it. On a fast success (or fast non-stall failure) the main body cancels the still-idle watchdog with `kill "$watchdog_pid"`. That SIGTERM reaches only the subshell process itself, not its forked `sleep` child — the `sleep` gets reparented to init and keeps running for up to the remainder of `timeout_s`. Impact was benign (holds no lock, ~0 CPU, self-terminates) but was process-hygiene debt flagged by code-review as a `Deferred` finding on PR #2958.

## Changes

- Watchdog subshell now backgrounds its own `sleep` (giving it a known PID) and installs a TERM trap that relays the fast-path cancel directly to that PID, instead of relying on the subshell process's own signal (which never reached the forked sleep). The stall-escalation path (`kill_tree`) is unchanged.
- Added regression test 7 in `test-dispatch-scripts.sh`: proves a fast-success run's watchdog sleep is actually killed (not left orphaned) via a selective sleep stub that records the watchdog's PID.

Full suite: `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — 3044/3044 passed, 0 failed.

Graph node: `tactic-playwright-watchdog-orphan-sleep`.


natb1:
## QA plan (triage) — RESOLVED

Node: `tactic-playwright-watchdog-orphan-sleep` — PR #2967

No browser component detected in this change (touches only
`.claude/skills/dispatch-propagate/scripts/lib.sh` and its test file) — all
plan items were `script-verifiable`.

### 1. Full bash test suite passes, including the new regression test 7
- Verdict: **PASS** — 3044/3044 passed, 0 failed.

### 2. Test 7 exists, is actually executed, and asserts the sleep PID is dead (not merely "tests pass")
- Verdict: **PASS** — test 7's selective sleep stub records the watchdog's PID
  and asserts liveness via `kill -0` after a fast-success run; both its
  assertions (`recorded`, `not orphaned (killed)`) PASS.

### 3. Test 7 is a true regression test — it fails against the pre-fix lib.sh
- Verdict: **PASS** — temporarily reverted the watchdog subshell to the
  pre-fix plain `sleep "$timeout_s"` in place, re-ran the suite: test 7's
  `not orphaned (killed)` assertion FAILs as expected (`expected: '0',
  actual: '1'`), proving the test pins the fix rather than passing
  vacuously. Restored the fix immediately after (`git diff` confirmed clean).

### 4. Stall/kill_tree escalation path is unregressed (test 6 and the whole-tree kill)
- Verdict: **PASS** — the diff against the merge base shows the
  post-`wait` stall/escalation logic (`now=`, elapsed guard, `kill_tree`)
  is byte-identical to pre-fix; test 6
  ("stall → all grandchildren killed (whole tree died)") still passes.

### 5. No other forked-child watchdog sites in the dispatch scripts were missed
- Verdict: **PASS** — grepped for `watchdog` / `kill "$..._pid"` /
  non-final `sleep "$..."` sites across `.claude/skills/dispatch-propagate/scripts/*.sh`.
  The only backgrounded-subshell-with-external-cancel watchdog pattern is
  `playwright_install_with_deps`; the other `sleep` sites (lib.sh:145,254,3223,
  run-pr-checks-wait.sh, wait-for-url.sh) are plain foreground retry-loop
  sleeps in the caller's own process, not a separate cancel-path target.

## Summary

5/5 script-verifiable items PASS. No residue. Clean autonomous QA pass.

natb1:
## qa · attempt 0
qa: 5/5 script-verifiable plan items PASS (full suite 3044/3044; test 7 confirmed a real
regression test by reverting the fix in-scratch and observing it FAIL; stall/kill_tree
escalation path unregressed; no other watchdog-orphan-sleep sites found in dispatch scripts).
failed: none


## review · attempt 1
## review · attempt 1
review: 5 finder passes (code-review, input-validation, secrets, red-team,
security-review) surfaced 8 findings; 4 fixed (code-review residue: TERM trap
now exits on cancel instead of falling through, trap installed before fork to
close the orphan window; red-team: bounded poll replaces flaky fixed-0.5s
grace in test 7), 4 red-team findings on the same trap mechanics refuted by
adversarial skeptics (impact overstated — kill_tree only touches the
already-condemned tree, no attacker-controlled trigger, empirically ~0.7%
reproduction under artificial contention, no live errexit-less caller exists).
Full suite re-run after fixes: 3044/3044 passed, 0 failed. No deferred/
out-of-scope findings — nothing filed.
failed: none

natb1:
## Review-fix summary

Surface: `code` (bash-only diff — `lib.sh` watchdog fix + `test-dispatch-scripts.sh` regression test). No dependency, CodeQL, or erosion findings (no open CodeQL alerts on this PR; `deps=false`; erosion scan returned no findings). 5 finder passes ran: `code-review` (own fix+residue lane), `input-validation`, `secrets`, `red-team`, `security-review` (own fix+residue lane).

### Fixed

- **`lib.sh:1719`** (code-review residue) — the TERM trap killed the inner `sleep` but never terminated the watchdog subshell itself, so a fast-path cancel could fall through into the elapsed-time check and (in principle) `kill_tree`. Fixed: the trap now sets a `cancelled` flag and the subshell `exit 0`s right after `wait "$sleep_pid" || true` when cancelled, leaving the stall/`kill_tree` escalation path untouched.
- **`lib.sh:1717`** (code-review residue) — the trap was installed one statement *after* `sleep &`, leaving an unguarded fork→trap-install window where a cancel could still orphan the sleep. Fixed: the trap is now installed *before* the fork (tolerant of the not-yet-assigned `sleep_pid`), plus a post-fork `cancelled` check closes the remaining sliver.
- **`test-dispatch-scripts.sh:5665`** (code-review residue) — test 7 detected a leaked orphan but never cleaned it up, so a failing run would leave a ~25s orphan `sleep` alive for the rest of the suite. Fixed: on a failed assertion the test now `kill -9`s any still-live recorded PID before `teardown`.
- **`test-dispatch-scripts.sh:5655`** (red-team, adversarial-verified as a real hardening gap) — the orphan-cancellation assertion used a fixed `0.5s` grace, which is a scheduler-dependent budget that could flake under load. Fixed: replaced with a ~5s bounded `kill -0` poll that breaks as soon as the PID disappears — a slow scheduler now costs latency instead of a false failure.

Fix commit: 55838474

Four `red-team` findings on the same trap mechanics were classified `Required` and sent to adversarial skeptics (2 skeptics for the higher-confidence findings, 1 for the others — all defaulted to "refuted" under uncertainty). All four were refuted and **not** fixed:

- **`lib.sh:1719`** — "trap has no `exit`, so SIGTERM during `kill_tree`'s escalation doesn't abort it." Both skeptics confirmed the mechanism but refuted the impact: `kill_tree` only ever signals the already-condemned `install_pid` tree, the surrounding code's own comments document that *not* aborting escalation is intentional (avoids leaving SIGTERM-ignoring grandchildren alive — the original PR #2946 bug), and the proposed unconditional `exit 0` fix would have reintroduced that exact bug.
- **`lib.sh:1719`** (trap stays armed after `wait` returns, referencing a possibly-reaped `$sleep_pid`) — refuted: no attacker-controlled trigger, the victim PID is kernel-chosen (not attacker-steerable), and the only realistic window (a couple of shell-builtin calls) is far too short for practical PID reuse.
- **`lib.sh:1717`** (fork→trap-install race, claimed reachable "on every attempt") — refuted empirically: 300 trials under normal conditions produced 0 orphans; 150 trials under artificially heavy CPU contention produced 1 (~0.7%). The reporter's own repro required widening the code's timing window to demonstrate it. Impact even if hit is a single self-terminating orphan, not a persistent leak.
- **`lib.sh:1720`** (fall-through hazard if a future caller sources `lib.sh` without `errexit`) — refuted: verified both actual production callers set `set -euo pipefail` before sourcing, `lib.sh` never disables it, and no in-repo caller of `playwright_install_with_deps` lacks errexit. The scenario is a chain of unmet hypotheticals with no live trigger today.

No deferred/out-of-scope findings — nothing to file as a follow-up.

Node target: `tactic-playwright-watchdog-orphan-sleep`.

PR #2918, opened 2026-07-19
graph-select-target: add --standalone mode for concurrency-safe manual/emulated ticks

natb1:
## Summary

Closes the double-dispatch race from 2026-07-07, where a manual emulated
router tick and the live `dispatch-claude-daemon` selected and dispatched the
same graph node simultaneously. `dispatch-select-tick` (the daemon's
selection entrypoint) is concurrency-safe on its own — it holds
`dispatch-acquire-lock` across a live-worker headroom check and the
resulting reservation-ledger claim — but `graph-select-target` itself
performed neither the lock acquisition nor the headroom check when invoked
directly, leaving any caller outside `dispatch-select-tick`'s wrapper (e.g. a
manual/emulated-tick launcher) with zero concurrency protection.

- `graph-select-target` gains a self-contained `--standalone` mode: acquire
  `dispatch-acquire-lock`, compute the same live-worker headroom
  `dispatch-select-tick`'s `--manual` branch uses (busy workers + reservation
  count vs `dispatch-target-workers --max`, failing open to `TOP=1` if the
  busy-worker read itself fails), clamp `--top` to the headroom, run the
  existing selection unchanged, claim each selected id in the reservation
  ledger, and release the lock via the script's `_cleanup` EXIT trap so every
  exit path releases it.
- Purely additive: `dispatch-select-tick`'s own call site is untouched, so
  there is zero regression risk to the live daemon chain.
- Adds test coverage: headroom-available claim, `HEADROOM == 0` degrade,
  lock-contended-by-a-live-foreign-session degrade, and a non-`--standalone`
  regression guard proving the default path is unchanged.

## Test plan

- [x] `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — 2985/2985 passed.

PR #2971, opened 2026-07-25
Split test-dispatch-scripts.sh into per-script test files

natb1:
## Summary

`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` was a
31,514-line monolithic test file and the single append target for every
dispatch-script feature branch. Concurrent branches therefore conflicted on
test text rather than on genuinely contended behavior (PR #2918 conflicted on
this file, was resolved, and re-conflicted within minutes). This PR splits it
into a shared fixture plus ~99 per-SUT sibling test files, matching the
existing convention of the 22+ other `test-*.sh` files already in the
directory.

- **Unit 1** extracted the shared harness (`assert_eq`, `report_results`,
  `setup`/`teardown`, PR/rollup builders, plus 7 cross-section helpers used by
  2+ output files) into a new `dispatch-test-fixture.sh`, verbatim.
- **Unit 2** generated ~90 per-SUT `test-*.sh` files from the remaining body,
  spliced one section into the existing `test-dispatch-phase-model.sh`, and
  deleted the monolith.
- **Unit 3** removed the deleted monolith's step from
  `unit-tests.yml`'s `hook-tests` job (the `unit-tests` job's
  `run-unit-tests.sh` glob auto-discovers every new file with zero wiring) and
  updated stale prose references, leaving the `qa-fix`
  runtime-matched marker pair untouched.
- **Unit 4** verified byte-exact line-multiset parity between the original
  monolith and the split output (the primary correctness gate — proves no test
  text was dropped, duplicated, or edited), fixed two boundary bugs the parity
  check surfaced (an annotation comment that had drifted inside the
  parity-checked region, and a dropped trailing `teardown` call), and
  confirmed every generated file executes cleanly with the full assertion
  count preserved.
- Merge conflicts with `origin/main` — which keeps appending to the monolith
  this branch deletes — have been resolved twice by porting the new content
  into its per-SUT home files, verbatim. This is the exact failure mode the
  tactic exists to eliminate, and it recurs on every tick until this PR lands:
  - 2026-07-28: a new `playwright_install_with_deps` regression test.
  - 2026-07-30: 178 lines added since merge-base `00ddb774`, re-homed into
    `test-lib-reservation-ledger.sh` (+34, the `origin=explicit` TTL reclaim),
    `test-dispatch-select-tick.sh` (+130/-1, `reconcile-graph-review-stall`
    wiring, the `run_sel_tick_err` helper, and the explicit-node reservation
    tests), and `test-mark-node-terminal.sh` (+15, the `/dispatch-conflict`
    Lane 3 dispositions). All 8 hunks applied by context with zero fuzz and
    zero rejects; each anchor was unique to exactly one split file.

**Unavoidable changed line — required callout.** The plan (`Step 4`, "Hard
rule: the extractor never modifies a body line") requires any genuinely
unavoidable net-new or changed line to be listed explicitly here and to show up
as exactly that one line in the parity diff. There is exactly one, introduced
in `d7fe169d`: `test-dispatch-complete-phase.sh:115`'s self-exclusion grep
(`--exclude='test-dispatch-scripts.sh'`) became
`--exclude='test-dispatch-complete-phase.sh'`.

That test asserts "only `dispatch-complete-phase` contains the `BFD4F2` hex"
and excludes *the test file itself*, which legitimately mentions `BFD4F2` in
its own fixtures and comments. When the section moved to a new home file, the
self-exclusion filename had to follow it or the test would false-fail on its
own fixture text. It still excludes exactly one file — its own — so the
assertion's scope and strength are identical. This is a mechanism-preserving
rename, not a weakening of the signal (`.claude/rules/test-integrity.md`).

No test was dropped, weakened, skipped, or rewritten — this is a pure
file-boundary move. Line-multiset equality rules out any dropped or edited
line, and because a duplicate would also break the multiset, it rules out
copy-instead-of-move as well.

**Plan verification note.** The plan's own line-parity `verify` block hardcodes
the monolith's body boundary as line 31510 (its size when the plan was
authored). `origin/main` has kept growing the monolith since — it is 33,162
lines as of the 2026-07-30 merge — so the plan's hardcoded boundary is stale
and its automated check false-fails. Re-running the same line-multiset check
with the corrected boundary (the monolith's actual current tail, `7,33158`,
not the plan-time line number) confirms parity is exact: 33,152 lines on both
sides, modulo only the one documented rename above.

## Test plan

- [x] Line-multiset parity: every line inside a moved-content sentinel across
      all output files is a byte-exact permutation of the original monolith's
      body (verified against `origin/main`'s copy).
- [x] Every generated `test-*.sh` file executes directly and exits 0.
- [x] The monolith is gone; the fixture is excluded from the `test-*.sh` glob
      by name; every generated file is executable.
- [x] `run-unit-tests.sh --pr-scripts` globs and runs every file in the
      directory, all green (2026-07-30: exit 0, 133 suites globbed, 0 failures).
- [x] Runtime assertion-count parity across the group: `origin/main` totals
      3664 assertions and this branch totals 3664, with 3664/3664 passing — so
      no assertion was lost, and none was moved-but-never-executed. All 12
      assertion names relocated by the 2026-07-30 merge execute and pass.
- [x] No executable reference (CI `run:` step, `bash`/`./` invocation) to the
      deleted filename survives; only historical prose comments remain.
- [x] CI (`unit-tests` job) discovers all new files via the existing
      `test-*.sh` glob with zero workflow changes beyond removing the deleted
      monolith's `hook-tests` step.


natb1:
## Conflict hold resolved; one pre-existing red check remains

The `provision-conflict` hold tracking this branch (`tactic-hold-conflict-dispatch-test-monolith-split`) is resolved and the `blocked_by` edge has been removed from `tactic-dispatch-test-monolith-split`, so the node is selectable again.

`origin/main` now merges clean into this branch. The 178 lines main had added to the monolith since merge-base `00ddb774` were re-homed verbatim into their per-SUT files (see the updated PR description). `PR Checks` is green; `run-unit-tests.sh --pr-scripts` is exit 0 with 133 suites and 3664/3664 assertions passing, and assertion-count parity against main holds exactly.

### The remaining `lint` failure is pre-existing, and its fix needs care

`Unit Tests & Lint` fails on the `prose` suite's `gh-rest-porcelain` rule. This is **not** caused by the merge resolution — the previous branch tip (`db774be7`, 2026-07-28) failed identically.

The rule inspects **net-new added lines** in committed `.sh` files. Because this PR *moves* ~30k lines out of the monolith, every moved line reads as net-new, so the rule now fires on text that has been on `main` unchanged for a long time. The five flagged lines are:

- `test-dispatch-find-pr.sh:128` — an `assert_eq` **label string**
- `test-dispatch-statements-scan.sh:543,577,617` — `echo "Test: ..."` **prose strings**
- `test-dispatch-stop-hook.sh:353` — hook text inside a **fixture heredoc**

None of them is a live `gh` call. All five are false positives.

### Constraint for whoever fixes this

**Do not edit, weaken, or delete any of the flagged lines.** They are moved content under the byte-parity gate (`.claude/rules/test-integrity.md`), and changing one breaks V3 line-multiset parity.

Two sound options, in preference order:

1. **Fix the rule** (preferred, and the greenfield answer): teach `lint-prose-rules.sh`'s `gh-rest-porcelain` check to skip matches inside string literals — `echo "..."` prose and `assert_eq` labels are not porcelain calls. That removes a whole class of false positive rather than papering over five instances, and it is the rule's actual intent.
2. **Suppress at each site** using the rule's own documented escape hatch, `# lint-allow: gh-rest-porcelain <reason>` on the immediately preceding line. If this route is taken, each marker is an unavoidable net-new line and **must** be listed explicitly in the PR description per the plan's Step 4 hard rule, and must appear as exactly those lines in the parity diff.

Option 1 touches no moved line at all and so leaves parity untouched; option 2 costs five documented net-new lines.

Leaving this PR in draft until that check is green.

natb1:
## Lint fixed at the rule; all 23 checks green, PR marked ready

`db9de584` fixes the `gh-rest-porcelain` rule rather than the five flagged sites, so no moved line was touched and V3 byte-parity is untouched (33,152 lines both sides, still only the one documented pre-existing rename).

**Correction to my previous comment.** I described `test-dispatch-stop-hook.sh:353` as hook text inside a fixture heredoc. That was wrong — the heredoc ends at line 343. Line 353 is a genuine `gh issue view` call, run against a stubbed `gh` on `PATH` because the test deliberately reproduces the pre-migration idiom to prove the migrated REST mechanism works. The rule was correctly detecting it. So four of the five were string-literal false positives; the fifth was real.

**Why the fix is not "skip string literals".** That would have handled only the four, and it carries its own weakening risk: `sh -c "gh issue list ..."` is a real invocation inside a string literal. Instead the fix corrects what "net-new" means. The rule's contract is net-NEW porcelain, but it implemented that as "any line on the `+` side". A relocated line appears in the diff both as a removal and an addition; a genuinely new call appears only as an addition. The rule now counts removals by exact content and lets one removal excuse one addition — which greens all five, including the real call at `:353` (exempt because it is relocated, not because it stopped being porcelain).

**Proof it did not go blind.** The exemption is a multiset match, and three new tests cover the limits:

- `15a` — a pure relocation is not flagged.
- `15b` — a genuinely new call riding alongside a relocation **is** still flagged.
- `15c` — two copies added against one removal still flags the surplus.

The pre-existing detection tests (`7`, `8`, `9a`, `13`, `14`) all still pass unchanged. `test-lint-prose-rules.sh`: 40/40.

This also means the rule stops re-firing on every future large move, which was the recurring cost here.

**Verification:** `run-unit-tests.sh --pr-scripts` exit 0, 133 suites, 3670/3670 — the 3664 moved-content assertions unchanged, plus the 6 added by tests 15a–c. All 23 CI checks green; `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

natb1:
## QA summary (attempt 3) — clean pass

No browser component in this PR's diff (only `.claude/skills/**`, `.claude/workflows/review-fix.js`, `.github/workflows/unit-tests.yml`), so every plan item ran through the shell-command lane. This is a re-QA pass after attempt 2's fixes landed and merged (df6952a2/645869f2); `origin/main` has not grown the monolith further since that merge (still 33702 lines), so no new post-merge re-homing work was needed this pass.

### Items executed: 14 (12 script-verifiable, 2 needs-human-judgment) — 14 PASS

1. **V1 — every `test-*.sh` runs standalone and exits 0** — PASS (scoped to this PR's ~99 split files; one unrelated pre-existing failure noted separately below).
2. **V3 — line-multiset parity vs freshly fetched origin/main** — PASS. Every non-blank line matches exactly except the one disclosed rename (`test-dispatch-complete-phase.sh`'s self-exclusion grep). One extra blank line found in the split output vs the monolith (2525 vs 2524 blank lines) — confirmed cosmetic only (zero non-blank/content difference), not chased to an exact origin given the effort-to-value ratio; no test text was dropped, duplicated, or edited.
3. **V2 — monolith gone, fixture outside glob, all files executable** — PASS.
4. **V4 — runner globs and runs every file, group green** — PASS. 173 suites globbed, 0 FAIL, every on-disk file confirmed present in the run log.
5. **V5 — runtime assertion-count parity** — PASS. 3713/3713, matching attempt 2's already-verified and explained baseline exactly (no change since).
6. **V6 — no executable reference to the deleted monolith survives** — PASS. Only historical prose comments remain.
7. **Every file syntactically valid bash and sources the shared fixture** — PASS. `bash -n` clean on fixture + all test files. Files not sourcing the fixture are all pre-existing siblings outside this PR's scope, plus one new companion file (`test-lint-prose-rules.sh`) that correctly uses the generic `test-helpers.sh` instead (unrelated to the monolith split, appropriately scoped).
8. **Foreign-section defect class — exhaustive banner-vs-filename SUT scan** — PASS. Two independent heuristics (banner-text tokenization, echo-banner tokenization) surfaced only false positives — all traced to pre-existing files outside this PR's diff, or to legitimate multi-SUT files already documented in the tactic's manifest (merged sections, helper+consumer pairs). No new absorption instances found.
9. **Adjudicate every banner-vs-filename candidate from item 8** — PASS (no genuine candidates survived triage; item 8 found nothing requiring relocation).
10. **Assertion-density outlier scan — second independent absorption detector** — PASS. Manually inspected the top outliers (`test-qa-fix-partition.sh`, `test-dispatch-sweep.sh`, `test-detect-changes.sh`, `test-align-tactics-tempref.sh`) — all are legitimate: either documented multi-SUT merges from the manifest, or small correctly-scoped files whose header/body match exactly.
11. **Every extracted section landed in exactly one file — no duplication** — PASS. The four attempt-2 relocated blocks (dispatch-flake-dedup, dispatch-flake-dedup-node, align-tactics-tempref, qa-disposition planned-deferral) each exist in exactly one file; their former hosts retain only prose header comments documenting the relocation, not duplicated test content.
12. **Fixture promotions defined exactly once, only in the fixture** — PASS. All 7 promoted helpers: fixture=1, test files=0.
13. **qa-fix runtime-matched marker pair still byte-identical** — PASS. The matched substring (`>>> partitionDispositions: sliced + eval'd by test-dispatch-scripts.sh (#1844) >>>`) is byte-identical in both `qa-fix.js` (as a comment) and `qa-fix-partition-probe.mjs` (as a string literal it matches against at runtime).
14. **Tactic manifest staleness and header/SUT prose accuracy (8-file sample)** — PASS. Sampled 8 files spanning small/large and old/newly-touched; every header accurately names its SUT and matches body content. The tactic-node manifest staleness (rows 81/85a, disclosed in attempt 2) remains an accepted historical-record caveat, not corrected.

### Notes (non-blocking, out of scope for this PR)

- **`test-lib-claude-agents-zsh-path-clobber.sh` fails standalone** (4/8 assertions) — confirmed byte-identical to `origin/main`'s copy and fails identically there. This file is untouched by this PR's diff; it is a pre-existing repo-wide issue unrelated to the monolith split, out of scope for this tactic.

Zero test logic changed. No fixes needed this pass.

Not run — the in-memory residue list was empty (no `fail` / `needs-human-judgment` / `main-gated-fail` items survived triage).

natb1:

natb1:
## qa · attempt 1
qa attempt 1: 8/9 plan items passed clean. Found and fixed a real defect:
test-dispatch-jit-calendar-import.sh had silently absorbed an unrelated
detect-changes.sh/list-go-modules.sh test section (added to the monolith
after the tactic's manifest, under a banner style the extraction algorithm
didn't recognize as a section boundary). Split it into a new
test-detect-changes.sh; also corrected 3 header comments
(test-dispatch-plan-io.sh, test-dispatch-mark-complete.sh,
dispatch-test-fixture.sh) so name/SUT/body agree. Zero test logic changed.
Re-verified: bash -n clean, standalone runs green, line-multiset parity
exact (same one pre-existing disclosed rename), full --pr-scripts suite
3694/3694 across 135 suites. Landed in c4c3aebc, merged + pushed at a09d467e.
failed: none remaining.


## qa · attempt 2
qa attempt 3: re-QA after attempt 2's fixes landed (df6952a2/645869f2);
origin/main has not grown the monolith further since (still 33702 lines), so
no new re-homing work was needed. 14/14 plan items passed clean: V1-V6
re-derived fresh (line-multiset parity exact modulo the one disclosed rename
plus one cosmetic stray blank line with zero content impact; 173 suites
globbed and green; assertion count 3713/3713 matching attempt 2's baseline
unchanged), plus a more exhaustive re-check of the recurring "foreign section
silently absorbed" defect class via two independent heuristics (banner-text
scan, assertion-density outlier scan) — found nothing new; the four attempt-2
relocated blocks confirmed present exactly once with no duplication; fixture
promotions unique; qa-fix marker pair byte-identical; 8-file header/SUT
sample all accurate. One unrelated pre-existing failure noted out of scope:
test-lib-claude-agents-zsh-path-clobber.sh fails identically on origin/main
(untouched by this PR). Zero test logic changed, zero fixes needed this pass.
failed: none.


## review · attempt 1
review: fixed a real CI-coverage regression — the `hook-tests` job's
unconditional `test-dispatch-scripts.sh` step was deleted with no unconditional
replacement for 7 suites whose SUT lives outside
`.claude/skills/dispatch-propagate/scripts/` (would merge green with zero
coverage). Wired all 7 unconditionally into `unit-tests.yml`. Also fixed a
`lint-prose-rules.sh` relocation-exemption scoping gap (a doc-file removal
could exempt a net-new porcelain call in a `.sh` file) with 2 new regression
tests, corrected 2 stale CI-vector comments, rewrapped one over-long line, and
fixed one stale example path in fix-checks/SKILL.md. 2 findings (a narrow
low-impact lint bypass, and a pre-existing duplication metric misfire)
refuted by adversarial-verify skeptics — no action. 1 finding (same root
defect as the lint-prose-rules.sh fix above) left Unverified after its
skeptic agent errored out; no follow-up filed since the fix already landed.
Landed in 43845603, merged + pushed.
failed: none.

natb1:
## Review

Combined review pass (code-review, security-review, cost/erosion/CodeQL) over the
123-file test-monolith split. 5 finder agents surfaced 5 raw findings (4 after
dedup) plus 9 code-review residue items; adversarial-verify ran on the 4
Required findings.

### Fixed (commit 43845603)

- **CI-vector regression** — the `hook-tests` job's unconditional
  `test-dispatch-scripts.sh` step was deleted with no unconditional replacement
  for suites whose SUT lives outside
  `.claude/skills/dispatch-propagate/scripts/` (`.claude/workflows/*.js`,
  `.claude/hooks/*.sh`, `packages/intentionsutil/scripts/*`). A PR touching only
  one of those SUTs would merge green with zero coverage of the guarding test —
  confirmed end-to-end by 2 independent skeptics (both upheld). Wired 7 orphaned
  suites (`test-qa-fix-partition.sh`, `test-align-tactics-tempref.sh`,
  `test-dispatch-stop-hook.sh`, `test-dispatch-office-hours-strip-hook.sh`,
  `test-fix-checks-cas-guard.sh`, `test-dispatch-chain-worktree-ratchet.sh`,
  `test-mark-node-terminal.sh`) as unconditional `hook-tests` steps in
  `.github/workflows/unit-tests.yml`, and corrected the now-false CI-vector
  comments in both probe files.
- **lint-prose-rules.sh relocation-exemption scoping gap** (same root defect as
  the Unverified finding below) — the `REMOVED_LINES` relocation-exemption index
  scanned every removed line in the whole diff with no file-type scoping, so a
  net-new raw `gh` porcelain call added to a `.sh` file could be silently
  exempted by an unrelated byte-identical line removed from a `.md` doc or a
  comment. Scoped removals to old-side shell scripts (`old_path_is_shell`,
  evaluated against the merge-base blob for deleted paths) and moved the
  relocation exemption to run after the allow-marker check. Added regression
  tests 15d (a doc-file removal must not exempt a `.sh` addition) and 15e (an
  allow-marked addition must not spend the relocation budget a genuine
  relocation needs).
- **Stale CI-vector comments** in `align-tactics-tempref-probe.mjs` and
  `test-align-tactics-tempref.sh` — both claimed the `test-*.sh` glob now
  covers them post-split, which is false (that glob is itself gated behind
  `RUN_PR_SCRIPTS`). Corrected to name the actual driver and the actual gate
  (the now-unconditional `hook-tests` step above).
- **reference.md line wrap** — re-wrapped the edited paragraph to the file's
  ~80-column convention.
- **fix-checks/SKILL.md stale example path** — the retired monolith path in a
  worked example was updated to a live per-SUT path, with a note that the
  quoted historical incident predates the split.
- **dispatch-test-fixture.sh unused-helper note** — `make_pr_union` /
  `make_pr_union_mergeable` were already unreferenced in the pre-split
  monolith. Rather than deleting them (which would need to survive this PR's
  own line-multiset-parity gate, a mechanism this PR relies on to prove the
  move is lossless), documented that they're pre-existing dead weight, not
  something this PR introduced.

- **lint-prose-rules.sh:145** — the relocation-exemption bypass is real but
  low-impact: the porcelain check already ships a trivial, always-available
  `# lint-allow:` opt-out, so this only defeats grep-discoverability of an
  opt-out nobody needs to hide behind; exploiting the bypass instead requires a
  self-directed, self-incriminating deception against the PR's own author.
  Refuted by 1 skeptic. (Incidentally hardened anyway by the scoping fix above,
  which now requires the old-side removal to come from a shell script.)
- **align-tactics-tempref-probe.mjs:32 (erosion)** — the flagged duplication is
  pre-existing: both probe files' shared sentinel-slicing boilerplate predates
  this PR (this PR only edited prose comments in the cloned region), and is the
  natural byproduct of splitting one monolith into independent per-SUT files —
  matching 25+ other files already in the directory. Refuted by 2 skeptics.

### Unverified

- **lint-prose-rules.sh:66 (input-validation)** — adversarial-verify errored
  (StructuredOutput retry cap exceeded, no valid verdict after 5 attempts)
  rather than reaching a verdict. This is the same underlying defect as the
  scoping fix above (unscoped `REMOVED_LINES`), already fixed in this PR's
  commit and covered by new test 15d — exactly the negative case this finding
  asked for. No follow-up filed since the fix already lands here.

### Informational

- **test-dispatch-phase-model.sh:92** — the appended MOVED block duplicates the
  file's own Case 1/Case 2 coverage. Left as-is: consolidating it risks
  conflicting with this PR's line-multiset-parity invariant (no test may be
  dropped or weakened, `.claude/rules/test-integrity.md`) — a human call, not
  auto-resolved here.

### Coverage

CodeQL: 0 open alerts on this PR. No dependency changes (`package.json` /
`package-lock.json` untouched), so no audit ran. `security-review` returned no
findings.

PR #2979, opened 2026-07-28
tactic-flake-fingerprint-stability: pin flake fingerprint stable-id precedence, add near-miss advisory

natb1:
Implements `tactic-flake-fingerprint-stability`.

`dispatch-flake-dedup-node` dedupes flake tracking nodes by an exact
fixed-string grep on a canonical `Fingerprint: <fingerprint>` line, where
`fingerprint = <failing-check-name> — <stable-id>`. The check name is
mechanical, but the stable id was chosen by the model implementing
`/fix-checks` from an open-ended precedence — and its `file:line` fallback was
itself unstable, since a line number drifts whenever an unrelated edit lands
above it. Two CI recurrences of the *same* failure could therefore mint two
different fingerprints, and dedup silently missed. This happened for real on
2026-07-22: one assertion failure in
`.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` produced
two flake tactics three minutes apart (both later pruned as non-defects).

## Unit 1 — pin the stable-id rule

In `.claude/skills/fix-checks/SKILL.md`, replaced the open-ended precedence
with a strict two-item rule: (1) the failing test name / assertion label
verbatim, (2) only if absent, the file path with **no line number**. Added an
explicit "never a line number/run id/timestamp/PR number" list, and a worked
example from the 2026-07-22 incident showing both original fingerprints
collapsing to the test-name form under the new rule.

## Unit 2 — narrow the blast radius of a near-miss

Added an advisory-only near-miss check to the node lane's `NONE` disposition:
before minting a new flake node, grep node bodies for the failing check name
alone (the mechanical half of the fingerprint). A hit means an existing node
shares the check but carries a different stable-id — surfaced as an advisory
accumulator note for a human to judge, never blocking or delaying filing (a
genuinely distinct flake under the same check is normal and expected). This is
a plain inline `grep`, not a new script — no new script surface, no new test
file.

## Out of scope

`dispatch-flake-dedup-node`'s matching logic is untouched — it's correct as
written; the defect was upstream, in what the caller fed it. No retroactive
normalization of fingerprints already recorded in existing nodes.

## Verification

```verify
bash .claude/skills/dispatch-propagate/scripts/lint-prose-rules.sh
```

Ran and passed. Manually re-checked the incident's two divergent fingerprints
against the new rule — both now collapse to the test-name form.

### DIFF
--- a/.claude/skills/fix-checks/SKILL.md
+++ b/.claude/skills/fix-checks/SKILL.md
@@ -379,9 +379,34 @@ atomic tempfile+`mv` write) so the park records `execution.pr`
         runs).** The fingerprint is `<failing-check-name> — <stable-id>`, where
         `<stable-id>` is chosen by this **fixed precedence** (use the first the
         failure excerpt provides):
-        1. the **test node id** — e.g. `path/to/test.spec.ts:LINE:COL test title`;
-        2. else the **file path:line**;
-        3. else the **CI workflow name**.
+        1. the failing **test name / assertion label exactly as the suite prints
+           it**, verbatim — a Jest/Mocha/vitest/pytest test title, a shell
+           test's assertion description, or any comparable human-readable label
+           the test runner itself emits. This is **not** a `file:line` pointer.
+        2. **only when the excerpt contains no such label**, the failing **file
+           path with NO line number** — e.g.
+           `.claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh`,
+           never `…:22026`. Line numbers drift whenever an unrelated edit lands
+           above that line in the file, so a line-number-bearing id
+           re-fingerprints the *same* failure differently across unrelated
+           commits — dedup then misses and mints a second tracking node for one
+           flake. That is why the `file:line` form is **disallowed**, not merely
+           a different-but-acceptable spelling.
+
+        **Never** include any of these in `<stable-id>`: a **line number**, a
+        **run id**, a **timestamp**, or a **PR number**. Each of them varies
+        across recurrences of one defect, so including one defeats dedup by
+        construction.
+
+        Worked example (2026-07-22 incident). One assertion failure produced two
+        divergent fingerprints under the old rule:
+        `hook-tests — .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh:22026`
+        (keyed on `file:line`) and
+        `hook-tests — select-tick on-main but primary checkout off-main → guard halts (exit 2)`
+        (keyed on the test name) — dedup missed and two nodes were minted for one
+        flake. Under this rule both collapse to the test-name form,
+        `hook-tests — select-tick on-main but primary checkout off-main → guard halts (exit 2)`.
+
         Read `<stable-id>` from the excerpt strictly by this precedence and
         **never paraphrase or summarize it** — the same flake must yield a
         byte-identical fingerprint string on every run, or dedup silently fails in
--- a/.claude/skills/fix-checks/SKILL.md
+++ b/.claude/skills/fix-checks/SKILL.md
@@ -545,6 +545,32 @@ atomic tempfile+`mv` write) so the park records `execution.pr`
                re-run CI — the head is simply missing a fix that already landed.
              - **`CURRENT`** — proceed with the node write below, unchanged.
 
+             **Near-miss advisory check (`CURRENT` only, never blocks).**
+             `dispatch-flake-dedup-node` matched nothing because it greps the
+             **full** `Fingerprint: <fingerprint>` line as a fixed string, so a
+             stable-id that diverges even slightly from an existing node's
+             spelling reads as `NONE` and mints a second node with no signal to
+             a human. Before writing, grep for the **mechanical half alone** —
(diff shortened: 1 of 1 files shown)


natb1:
## QA summary — tactic-flake-fingerprint-stability (PR #2979)

Doc-only change (`.claude/skills/fix-checks/SKILL.md`) — no browser component, no app dir, no Firestore surface touched.

**Items executed:** 10 total — 6 script-verifiable (shell-command lane), 4 needs-human-judgment (deferred to disposition triage).

**PASS / FAIL / SKIP:** 6 PASS / 0 FAIL / 0 SKIP (shell-command lane). No browser lanes ran (none needed).

## 1. Prose-rule lint passes on the changed file — PASS
`bash .claude/skills/dispatch-propagate/scripts/lint-prose-rules.sh` exits 0, no violations.

## 2. Diff scope matches the PR body — PASS
Only `.claude/skills/fix-checks/SKILL.md` changed; no script/test file touched, matching the "out of scope" claim.

## 3. Old three-item precedence fully removed — PASS
No residual "test node id" / "CI workflow name" / "file path:line" / ":LINE:COL" guidance remains.

## 4. `Never include` list enumerates all four forbidden components — PASS
Line number, run id, timestamp, and PR number are all present in the block (manual check — a naive line-based grep undercounts because two terms share each of two lines).

## 5. Worked example's two fingerprints collapse correctly — PASS
The test-name-keyed fingerprint string appears twice in the file: once as the original, once as the stated collapsed result.

## 6. Near-miss grep pattern is a strict prefix with the correct separator — PASS
`NEARMISS=$(grep ...)` pattern is `Fingerprint: <failing-check-name> — ` — the canonical prefix with the em-dash separator.

**Deferred to office-hours:** none — see disposition triage below.

## Disposition triage

Residue list (the 4 needs-human-judgment items) went through the Step 3.5 disposition Workflow. `result.dispositions` is empty — every residue item was classified `already-satisfied` and dropped as PASS.

- **Near-miss placeholder syntax matches the file's existing conventions** — the new block's placeholders (`<failing-check-name>`, `<tactic-id>`) use the same angle-bracket convention as the rest of the file, and the `intentions/tactic-*.md` glob matches every existing tactic node's actual filename convention.
- **Precedence rule reads unambiguously to a future agent** — the two-item precedence (test/assertion label, else bare file path) is a decidable first-match rule; "not a `file:line` pointer" (clarifying what a test label is) does not contradict the later `file:line`-disallowed paragraph, reinforced by the explicit "never include a line number" list.
- **Near-miss advisory is unmistakably non-blocking and its accumulator output is actionable** — "Still mint the new node exactly as below" plus "never blocks" in the heading make the non-blocking contract explicit; the cited `STALE-SUPPRESSED` / `STALE-HEAD-SUPPRESSED` accumulator-note precedent is real and matches the same prose-note pattern.
- **Near-miss check is attached to the correct disposition branch** — the check is nested inside the `CURRENT` sub-bullet of the stale-head guard, itself nested under the `NONE` dedup-disposition branch — the exact branch reached when dedup finds nothing and a new node is about to be minted. `NONE` and `CURRENT` name the same overall branch at different levels, not a mismatch.

No code defect found. No fixes applied, no follow-ups filed.

natb1:
## qa · attempt 0
failed: none

Doc-only change to `.claude/skills/fix-checks/SKILL.md` (stable-id precedence rule + near-miss advisory check). 6/6 script-verifiable checks passed; 4 needs-human-judgment items all disposed already-satisfied by the Step 3.5 triage Workflow (verified against file content, no code defect). Clean pass, no fixes, no follow-ups.


## review · attempt 1
failed: none

Docs-only diff (no attack surface). code-review surfaced 4 findings against the new stable-id precedence and near-miss advisory text: a missing terminal fallback tier, an advisory clause with no home in the Accumulator field spec, an unanchored near-miss grep glob, and a missing flake-tracker confirmation before naming a near-miss match. All 4 fixed in commit 65f28fd0. No deferred follow-ups, no security findings.

natb1:
## Review

### Fixed

- The two-tier stable-id precedence had no terminal fallback, so a label-less, path-less failure left `<stable-id>` undefined — restored a third tier (CI workflow/job name) and stated the precedence is total. (`.claude/skills/fix-checks/SKILL.md:382`)
- The near-miss branch's advisory note had no home in the Accumulator field spec — extended the Flake issue bullet with the advisory clause form so the spec and the Step 4 instruction agree. (`.claude/skills/fix-checks/SKILL.md:566`)
- The near-miss grep used a cwd-relative glob with no repo-root anchoring, so it could silently miss everything if invoked outside the worktree root — anchored it at `git rev-parse --show-toplevel`, matching the sibling script's convention. (`.claude/skills/fix-checks/SKILL.md:557`)
- Matching only the mechanical check-name half of the fingerprint could misidentify an unrelated node quoting a `Fingerprint:` line as a flake tracker — added a confirmation step to discard non-flake-tracker matches before naming them in the advisory. (`.claude/skills/fix-checks/SKILL.md:553`)

Fix commit: 65f28fd0

### Informational

_None._

### Dismissed

_None._

### Deferred

_None._

PR #2990, opened 2026-07-31
graph-commit: guard rebuild path against stale-checkout silent reverts (tactic-graph-commit-rebuild-snapshot-stale-revert)

natb1:
## Summary

Fixes a silent-revert defect in `graph-commit`'s "rebuild path"
(`ensure_intentions_only_base()`), which fires whenever a PR-branch worktree is
ahead of `origin/main` with non-`intentions/` changes — the normal shape of any
phase-worker worktree. That path used to `git reset --hard` to fresh
`origin/main` and then overwrite each node with the writer's on-disk snapshot
by content, not by diff. A writer whose checkout held a stale copy of a node
(ordinary: `origin/main` keeps moving during a session even though
`provision-node-worktree` merges it in at provisioning time) would silently
land that stale content over whatever had landed on `origin/main` in between,
with no conflict and no warning.

- **Unit 1** — `packages/intentionsutil/scripts/graph-commit`: guards the
  rebuild path with an implicit merge-base compare-and-swap. For every
  ordinary/prune id with no explicit `--base`, it compares the blob at the
  writer's merge-base against the blob at fresh `origin/main`; if they match,
  the snapshot is safe to lay down unchanged; if they diverge, it reconciles
  through the existing structural three-way merge (`run_merge_node()`) or
  parks loudly via `park_and_exit()` — never a silent overwrite. Also fixes a
  related defect: `check_base_freshness()` reconciled a stale explicit
  `--base` but only wrote the merged result to the working file, not to
  `$SNAP_DIR` — so the rebuild path (which re-materializes from `$SNAP_DIR`)
  was discarding that reconciliation. Adds five new `test-graph-commit.sh`
  cases (36-40) that reproduce the silent revert against the pre-fix script
  and pass after the fix.
- **Unit 2** — `packages/intentionsutil/scripts/demote-node-to-implement`: the
  one remaining phase writer with no compare-and-swap. Brings it in line with
  its siblings `park-node`/`transition-node`: fetches `origin/main`, refreshes
  the local node file before mutating, and passes `--base` at the
  `graph-commit` land. Adds a new `test-demote-node-to-implement.sh`
  CAS-guard harness (mirroring `test-transition-node.sh`) and wires it into
  CI (`.github/workflows/unit-tests.yml`).

## Node

tactic-graph-commit-rebuild-snapshot-stale-revert

## Test plan

- [x] `packages/intentionsutil/scripts/test-graph-commit.sh` — 51/51 passed
      (new cases 36, 37, 39, 40 verified to fail against the pre-fix script,
      reproducing the silent revert)
- [x] `packages/intentionsutil/scripts/test-transition-node.sh` — 3/3 passed
- [x] `packages/intentionsutil/scripts/test-park-node.sh` — 21/21 passed
- [x] `packages/intentionsutil/scripts/test-demote-node-to-implement.sh` — 2/2
      passed (new)
- [x] `.claude/skills/dispatch-propagate/scripts/test-graph-write-rollback.sh`
      — 5/5 passed
- [x] `npx vitest run --project packages/intentionsutil --root .` — 717/717
      passed
- [x] `npx tsx packages/intentionsutil/scripts/validate-graph.ts` — clean
- [x] `.claude/skills/dispatch-propagate/scripts/run-lint.sh` — clean


natb1:
## QA summary — PR #2990 (attempt 2 — verifying attempt-1's fix landed clean)

Attempt 1 fixed a bug confined to this PR's own new test harness
(`test-demote-node-to-implement.sh` Case 1 false-failing due to a missing
`app.slug` fixture field plus an always-fail merge shim) and stopped without
transitioning phase, per the skill's fix-finalize contract — a fixing pass
always defers final verification to a subsequent pass. This pass re-verified
the full checklist end-to-end against the current tree and live CI rather than
trusting the prior pass's marks. No browser component in this PR (pure backend
git/bash tooling + CI-wiring) — every item was script-verifiable except one
judgment item.

### Items executed — shell-command lane

| # | Title | Result |
|---|---|---|
| 1 | Diff scope is exactly the four claimed files | PASS |
| 2 | New regression harness passes deterministically (3 runs) | PASS (2/2 x3) |
| 3 | New harness genuinely covers the silent-revert scenario (non-vacuous) | PASS |
| 4 | `graph-commit`'s rebuild-path CAS guard is present and total in code | PASS |
| 5 | `check_base_freshness()` `$SNAP_DIR` sync fix is real | PASS |
| 6 | `demote-node-to-implement`'s CAS wiring matches its siblings | PASS |
| 7 | `test-graph-commit.sh` green, new cases present | PASS (68/68) |
| 8 | Sibling writer suites (`transition-node`, `park-node`, rollback) green | PASS (3/3, 22/22, 9/9) |
| 9 | intentionsutil vitest + `validate-graph.ts` pass | PASS (820/820, 523 nodes clean) |
| 10 | CI wiring for the new harness is correct | PASS |
| 11 | PR body claims match shipped diff (no false substantive claim) | needs-human-judgment → see disposition triage |
| 12 | QA data policy — not applicable (no frontend/Firestore surface) | PASS |

**Counts: 11 PASS / 0 FAIL / 0 SKIP** (item 11 routed to disposition triage
rather than a pass/fail mark). No browser component; no walkthrough ran. Live
`gh pr view` re-check: all 23 CI checks SUCCESS, including `hook-tests` (the
only check attempt 0 found red).

### Bugs found

None this pass. Attempt 1's harness-only bug (see its own summary above this
one, still on this thread) is confirmed fixed and stable: `test-demote-node-to-implement.sh`
passed 2/2 on all three consecutive runs in this pass.

### Disposition triage

One residue item went through the Step 3.5 disposition Workflow:

- **id 11** — "PR body claims match shipped diff" — class: `needs-main`
  (planned-deferral). The substantive claims (the CAS guard, the `$SNAP_DIR`
  sync, the `demote-node-to-implement` wiring) were all mechanically confirmed
  true by items 4-6 above. The only drift found: the PR body cites the new
  regression cases as "cases 36-40," but an unrelated `origin/main` commit
  independently inserted its own cases 36-47 into `test-graph-commit.sh` after
  this PR's body was authored, shifting this PR's own five regression cases to
  case numbers 48-52. All five exist under their current numbers and pass —
  this reads as stale prose from unrelated churn, not a functional gap, but it
  is not autonomously fixable code, so it was filed as a needs-main follow-up
  rather than treated as already-satisfied.

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped
from the escalation set, so they do not park this PR. No `opus-fixable` or
`needs-human` residue this pass, so nothing escalates to office-hours.

### Filed follow-ups

Node-target lane: filed nothing on GitHub. Instead appended a `## needs-main
residue` section directly to the tactic's own body
(`intentions/tactic-graph-commit-rebuild-snapshot-stale-revert.md`), landed via
a standalone `graph-tx` worktree + `graph-commit --base` (commit `c71fcc7e`,
verified present on `origin/main`) — the node-lane equivalent of a filed
follow-up. `Verifiability: MACHINE`; drains post-merge when `review → main-qa`
fires.

### Deferred to office-hours

None this pass.

natb1:
## qa · attempt 0
failed: item 2 — test-demote-node-to-implement.sh Case 1 (rc=1, matched CI hook-tests failure). Root cause 1: gh-shim green.json fixture missing id/app.slug fields required by graph-commit's await_checks() forgery filter (added by unrelated origin/main commit 10f9e91a merged in since attempt 0). Root cause 2 (surfaced after fixing #1): npx shim's always-fail merge-node.ts stub misreported Case 1's disjoint 3-way merge as unresolvable. Both fixed this pass: fixture gained id/app.slug per row; npx stub replaced with the real 3-way merge shim ported from test-transition-node.sh. Fixed via commit 07ebeb29 (merged 51baecd0). test-demote-node-to-implement.sh now 2/2 deterministic (3 reruns); test-transition-node.sh, test-graph-commit.sh, test-park-node.sh all unchanged/green.
residue: id 16 (fixture-builder duplication across CAS-guard harnesses) classified opus-fixable but no fix unit emitted (recommended action is a non-blocking follow-up, not a code change) — not addressed this pass, no follow-up filed, not blocking.


## qa · attempt 1
failed: none. Re-verified the full 12-item checklist end-to-end after attempt 1's harness-only fix landed and CI went fully green (23/23 checks, including hook-tests). All script-verifiable items PASS: diff scope exact, new test-demote-node-to-implement.sh harness deterministic (2/2 x3 runs) and genuinely covers the stale-writer-does-not-revert-concurrent-edit scenario, graph-commit's rebuild-path CAS guard and check_base_freshness SNAP_DIR sync confirmed present in code, demote-node-to-implement's fetch/refresh/--base wiring matches its siblings, test-graph-commit.sh 68/68 (regression cases now at case numbers 48-52, not the PR body's stale "36-40" citation — unrelated origin/main churn renumbered the suite), sibling writer suites and intentionsutil vitest/validate-graph all green.
residue: id 11 (PR body's stale case-number citation, 36-40 vs actual 48-52) — class needs-main (planned-deferral); filed as a `## needs-main residue` append to the node's own body via a graph-tx worktree + graph-commit --base (commit c71fcc7e), not a GitHub follow-up (node lane). Not a functional gap — all five regression cases exist and pass, only the citation text is stale. Drains post-merge via review -> main-qa.


## review · attempt 1
## review · attempt 1
findings: 9 surfaced (input-validation x2, red-team x6, code-review-residue x1), 3 actionable, 3 fixed.
fixed: (1) demote-node-to-implement had no restore guard on its new origin/main
refresh + --base CAS pin — an aborted demotion via graph-commit's fail-closed
park path (--base staleness) would have reverted a just-landed office_hours
park, and separately would discard any uncommitted node-body content a
far-behind worktree was holding. Ported park-node's TMPORIG/HEAD_BEFORE/
restore_node() guard pattern and added --expect to the graph-commit land.
(2) test-demote-node-to-implement.sh only covered the disjoint-merge and
absent-node paths — added a Case 2 driving the --base stale-diagnosis park
path via a graph-commit wrapper that lands a concurrent phase change,
asserting exit 1, the park + concurrent content on main, and a clean demoting
checkout. Fixed via commit a2fefd93 (message-text change in that commit also
required updating one pre-existing grep assertion in test-park-node.sh Case 5,
which named the old wording; not a weakened test — porcelain-empty assertions
there are unchanged and still pass).
refuted: 5 findings dropped by adversarial-verify/residue skeptics — a
path-traversal claim (unreachable: writeNode's assertPathSafeId blocks unsafe
ids before creation), a `=`-in-id CAS-corruption claim (fails loud via
graph-commit's die, not silent), a comment-accuracy nit on stale-base
behavior, a symlink-follow concern (impractical threat model), and a
code-review residue item flagging the refresh's use of the moving origin/main
ref (mirrors the established transition-node/park-node pattern this node's
own plan explicitly instructed replicating).
dismissed: 1 (FETCH_HEAD vs refs/remotes/origin/main side-effect commentary,
no actionable change).
verification: full verify block green (test-graph-commit.sh 68/68,
test-transition-node.sh 3/3, test-park-node.sh 22/22, new
test-demote-node-to-implement.sh 3/3, test-graph-write-rollback.sh 9/9,
vitest intentionsutil 820/820, validate-graph clean, run-lint clean).
deviation: none. no deferred filings, no security follow-ups.

natb1:

natb1:
## Landing note — operational finding while closing out attempt 2

QA verification itself was clean (see the qa-summary comment above). Closing
out the pass — landing the needs-main residue append and then running
`transition-node` to advance `qa → review` — surfaced a real, reproducible gap
in this PR's own new `ensure_intentions_only_base()` guard, distinct from the
findings-surfaced residue already triaged:

- The node's own body (Unit 1b step 3) describes an explicit skip: "skip the
  id entirely when `${BASE[$id]:-}` is non-empty," since `check_base_freshness()`
  already validated/reconciled an explicit `--base`. The shipped
  `ensure_intentions_only_base()` / `replay_snapshot_onto_base()` do **not**
  implement that skip — every id (base-supplied or not) always goes through the
  independent `merge-base HEAD "$MAIN_SHA"` fork-point comparison.
- This did not cause a silent revert (the defect this PR fixes) — it failed
  safe every time, exactly as designed. But it did produce two spurious
  outcomes when I ran `clear-park` and `transition-node` (both pass `--base`)
  from this PR's own long-lived worktree, whose `HEAD` already carries many
  merged-in `origin/main` commits from other concurrent sessions:
  1. `clear-park` computed a stale `merge-base` fork point, took the
     unconditional fast-path `cp`, and silently no-op'd ("no new changes to
     stage ... skipping the landing cycle") without actually clearing
     `office_hours` — even though the write was correct and the `--base` was
     fresh.
  2. Separately, `transition-node`'s call into `demote-node-to-implement` (on
     its own `--base`-guarded land) hit a genuine `list-entry removal vs.
     concurrent edit` park on `execution.markers`, racing against a needs-main
     residue commit this same session had just landed via a separate
     `graph-tx` worktree.
- Both recovered cleanly by re-running the same script (`clear-park`, then
  `transition-node`) from a **freshly created `graph-tx` worktree at
  `origin/main` tip** instead — `ensure_intentions_only_base` early-returns
  there (no non-intentions diff), so the whole fork-point computation is
  skipped and the safe in-place rebase path is taken. Final state verified:
  `office_hours: null`, `phase: review`, `execution.markers: [planned,
  qa-done]`, `execution.pr: 2990`.

This is not blocking — the PR's core defect (silent revert) is fixed and
verified, and every observed failure mode here was fail-safe (no-op or park,
never a bad land), consistent with the PR's own design goal. But it is a real
usability gap for exactly the callers this PR is meant to protect
(`clear-park`, `transition-node`, `demote-node-to-implement`), triggered by
the ordinary long-lived-worktree shape the node's own rationale describes.
Worth a follow-up to either implement the documented `${BASE[$id]:-}` skip in
`ensure_intentions_only_base()`, or to update the node's own text if the
skip was deliberately dropped during implementation for a reason not
recorded here.

natb1:
## Review pass

Ran the combined review-fix fan-out (built-in `/code-review low --fix` pre-stage,
plus input-validation / domain-sweep / red-team / security-review finders,
dedup, classify, adversarial-verify, and Opus fix). Fix commit:
a2fefd93cb086ddb1862cfa04d38d9debb2aa847.

### Fixed

- **red-team-5** — `packages/intentionsutil/scripts/test-demote-node-to-implement.sh:447` —
  the new CI suite only exercised a disjoint merge and the absent-node refusal;
  it never drove the `--base` stale-diagnosis park path. Added a Case 2 that
  races a concurrent `phase` change onto `origin/main`, asserting `demote`
  exits 1, the office_hours park and the racer's content land on main, and the
  demoting checkout's `intentions/` tree is left clean. Fixed in a2fefd93.

- **red-team-0** / **red-team-2** — `packages/intentionsutil/scripts/demote-node-to-implement:124,144` —
  the new `origin/main` refresh (line 124) and `--base` CAS pin (line 144) had
  no restore guard: `graph-commit`'s fail-closed park path (`check_base_freshness`
  → `park_and_exit`) writes, commits, and pushes an `office_hours` park to
  `origin/main` *before* returning failure, and the script's unconditional
  EXIT-trap rollback would have overwritten that just-landed park with stale
  pre-operation content — reverting a human-attention hold. Separately, the
  refresh had no pre-capture of the file's prior state, so an aborted demotion
  from a worktree carrying uncommitted node-body content (e.g. `## needs-main`
  residue) would discard it. Ported `park-node`'s `TMPORIG`/`HEAD_BEFORE`/
  `restore_node()` guard pattern (declines to restore when HEAD moved or the
  on-disk blob no longer matches what this script wrote) and added `--expect`
  to the `graph-commit` land. Fixed in a2fefd93.

- **input-validation-0** — `demote-node-to-implement:124` (path-traversal via
  unvalidated `$NODE_ID`) — refuted: `writeNode()` (`store.ts:52-54`) calls
  `assertPathSafeId()` at node-creation time and rejects `/`, `\`, `.`, `..`
  before any file is ever written; line 83's `git rev-parse
  origin/main:intentions/$NODE_ID.md` precondition also requires the exact
  literal path to already exist on `origin/main`. No reachable path to an
  unsafe id without first bypassing `writeNode()` entirely.
- **input-validation-1** — `demote-node-to-implement:144` (unvalidated `=` in
  `$NODE_ID` corrupting the `--base` CAS pair) — refuted: the mechanics are
  real (`assertPathSafeId` permits `=`; `add_blob_pair` splits on the first
  `=`), but the failure mode is fail-loud, not fail-silent — a corrupted pair
  yields a garbage blob sha that `git cat-file -p` cannot read, hitting
  `graph-commit`'s explicit `die` rather than a silent bypass.
- **red-team-1** — `demote-node-to-implement:140` (comment says `graph-commit`
  "refuses the write (exit 1)" on a stale base, but it actually attempts a
  three-way reconcile first) — refuted as a documentation nit, not a
  vulnerability: successful reconciliation on disjoint fields is the desired
  behavior, and a genuine same-field conflict is still caught and parked.
- **red-team-3** — `demote-node-to-implement:124` (plain `>` redirect follows a
  symlink at the node path) — refuted: the exploit chain requires a
  git-tracked symlink to survive this repo's own review pipeline and the
  symlink-free `writeNode()` creation path, then be selected for demotion —
  impractical against this tool's actual threat model; a defense-in-depth
  suggestion, not a demonstrated exploit.
- **code-review-residue-refuted-0** — `demote-node-to-implement:124` (the
  refresh re-resolves the moving `origin/main` ref instead of the pinned
  `$FRESH_BLOB`) — refuted by the residue skeptic gate: this exactly mirrors
  the established sibling pattern in `transition-node`/`park-node`, which this
  node's own plan explicitly instructed mirroring; not a new defect introduced
  by this diff.

### Dismissed

- **red-team-4** — `demote-node-to-implement:51` (`git fetch origin main`
  updates `FETCH_HEAD`; `refs/remotes/origin/main` update is an opportunistic
  side effect) — dismissed, no actionable change identified.

### Coverage

No partial coverage — all finders and the built-in `/code-review` pre-stage
ran to completion; no instrument failures. No deferred filings or security
follow-ups (`deviation=false`, `coverage_incomplete=false`).

natb1:
Citation correction, for future readers of this PR.

This PR's body says it *"Adds five new `test-graph-commit.sh` cases (36-40)"*. Those five cases are now at **48-52**, and were never at 36-40 by the time this merged: an unrelated commit on main inserted its own new cases 36-47 ahead of them. The body text above is left as-is deliberately — it is a historical record of what was true when it was written, not a live index.

The five cases are the Unit 1 / Unit 2 regression guards, and they are all green today (suite: 124 passed, 0 failed).

Rather than chase ordinals, `test-graph-commit.sh` now carries a comment block above that group recording this drift and establishing the convention: **cite cases by assertion text, never by ordinal.** Three in-file citations that used bare ordinals have been re-anchored accordingly. That change landed in #3136.

Tracked by `intentions/tactic-tactic-graph-commit-rebuild-snapshot-stale-revert-main-qa-regression.md`, now closed against the follow-up-doc-fix branch of its expected outcome.

PR #2999, opened 2026-07-31
sweep timer: stop test-suite unit-dir leak, add stale-unit self-heal (tactic-sweep-timer-unit-dir-leak)

natb1:
Implements intention node `tactic-sweep-timer-unit-dir-leak` (serves strategy-autonomous-execution, strategy-exercise-recovery-paths).

Stops the dispatch script test suite from clobbering the host's real systemd sweep/heartbeat units with the run's temp fixture paths, and gives the sweep timer the same stale-unit self-heal the heartbeat already has.

## Unit 1 — stop the script-level harnesses leaking into the host unit dir

Five harness files (`test-dispatch-schedule-reseed.sh`, `test-dispatch-schedule-convergence-reseed.sh`, `test-dispatch-schedule-target-reseed.sh`, `test-dispatch-spawn-tick.sh`, `test-dispatch-schedule-rate-limit-resume.sh` — the successors of the now-split `test-dispatch-scripts.sh` monolith) each already override `DISPATCH_RECOVER_UNIT_DIR`/`DISPATCH_RECOVER_SYSTEMCTL_CMD` but never `DISPATCH_SWEEP_TIMER_*`/`DISPATCH_HEARTBEAT_*`, so any test path reaching `ensure_sweep_timer`/`ensure_heartbeat_units` wrote to the developer's real `~/.config/systemd/user/dispatch-sweep-periodic.service`/`dispatch-heartbeat.service` and ran live `systemctl` against them. Adds the four missing env-var overrides (reusing the stub dir/systemctl the recover pair already sets up) to each harness's setup/teardown, plus a regression guard in `test-dispatch-spawn-tick.sh` that hashes the real host unit files before and after the suite and fails loudly on any change.

## Unit 2 — sweep timer stale-unit self-healer

`ensure_heartbeat_units` (`lib.sh`) self-heals a stale installed unit via `cleanup_stale_heartbeat_units` (disables + lets the caller rewrite when the installed `WorkingDirectory=` no longer matches the current main worktree); `ensure_sweep_timer` had no equivalent, so a corrupted sweep unit (e.g. pointing at a deleted temp fixture) stayed broken indefinitely. Adds `cleanup_stale_sweep_units`, mirroring the heartbeat helper, called from `ensure_sweep_timer` in the same position (after the hot-path content/active check, before the rewrite). Also adds `ensure_sweep_timer "$MAIN_WORKTREE" || true` to `dispatch-tick-recover`, next to the existing `ensure_heartbeat_units` call — closing the recovery asymmetry where only the heartbeat was re-asserted on tick recovery.

## Adaptation note

The plan's `Verification` section named `test-dispatch-scripts.sh`, which no longer exists — `tactic-dispatch-test-monolith-split` (this node's `blocked_by`) landed and split it into per-script files before this session started, exactly as the plan's own Unit 1 "Note on ordering" text anticipated. Ran the successor suites directly instead (all pass, including the new regression guard); see Verification below.

## Verification

Ran all six affected suites directly (the plan's stale `test-dispatch-scripts.sh` reference no longer exists post-split — see Adaptation note):

- `test-dispatch-schedule-reseed.sh`: 66/66
- `test-dispatch-schedule-convergence-reseed.sh`: 11/11
- `test-dispatch-schedule-target-reseed.sh`: 21/21
- `test-dispatch-spawn-tick.sh`: 21/21 (includes the new host-unit-leak regression guard)
- `test-dispatch-schedule-rate-limit-resume.sh`: 28/28
- `test-dispatch-spawn-sweep.sh`: 49/49

`bash -n` clean on `lib.sh` and `dispatch-tick-recover`. Confirmed the real `~/.config/systemd/user/dispatch-{sweep-periodic,heartbeat}.service` hashes are unchanged before/after the full local run.


natb1:
## QA plan (triage, attempt 0)

### 1. All six affected dispatch suites pass end-to-end — PASS
66/66, 11/11, 21/21, 21/21, 28/28, 49/49 — matches the PR's claimed counts. `test-dispatch-spawn-tick.sh` prints the new leak-guard PASS line.

### 2. Running the full suite does not clobber the operator's real systemd unit files — PASS
sha256 of both real host unit files identical before/after all six suites.

### 3. All five harnesses export and unset the four new env-var overrides — PASS
All 5 harnesses export and unset all 4 vars.

### 4. Env-var names match what lib.sh actually reads — PASS
`lib.sh:2748/2751` (sweep) and `lib.sh:3023/3026` (heartbeat) consume exactly these names.

### 5. `cleanup_stale_sweep_units` is defined and called from the correct position in `ensure_sweep_timer` — PASS
`lib.sh:2671` def, `lib.sh:2822` call — after the hot-path `return 0`, before `mkdir -p "$UNIT_DIR"`; mirrors `cleanup_stale_heartbeat_units` exactly.

### 6. `cleanup_stale_sweep_units` handles the three edge cases without aborting the caller — PASS
No-unit / match / diverge / disable-fails / no-WorkingDirectory all behave as expected; disable failure never propagates (rc=0 throughout).

### 7. End-to-end self-heal: a stale sweep unit is actually repaired by `ensure_sweep_timer` — PASS
Seeded stale unit → WARNING + disable --now → daemon-reload → enable --now → `WorkingDirectory=` rewritten to current main worktree. Real host unit untouched.

### 8. `dispatch-tick-recover` calls `ensure_sweep_timer` and stays non-fatal — PASS
Both calls `|| true`-suffixed, above the `Step 1: NOW` marker; `bash -n` clean on both files.

### 9. The leak guard is genuinely load-bearing (negative control) — PASS
Guard sits after all tests, before `report_results`, increments `FAIL` on mismatch — not decorative.

### 10. Remaining harnesses in the scripts dir are not still leaking — PASS
No residual leak risk found across all `test-dispatch-*.sh`.

### 11. Operator judgment: is the stale-unit WARNING actionable on-call? — ASSESSED SATISFIED (Opus)
- Classification: needs-human-judgment → disposition: already-satisfied

### 12. Operator judgment: is re-asserting the sweep timer inside `dispatch-tick-recover` the right call? — ASSESSED SATISFIED (Opus)
- Classification: needs-human-judgment → disposition: already-satisfied

---

**Result: 10 PASS, 0 FAIL, 0 escalated.** No browser component in this change.

## Disposition triage

Both `needs-human-judgment` items went through the Step 3.5 disposition Workflow. `result.dispositions` is empty — every residue item was classified **already-satisfied** and dropped as PASS. No `opus-fixable` or `needs-main` items exist, so no auto-fix lane and no follow-up filing this pass.

- **Operator judgment: is the stale-unit WARNING actionable on-call?** — The WARNING at `lib.sh:2691` mirrors the shipped `cleanup_stale_heartbeat_units` WARNING (`lib.sh:2957`) word-for-word in tone and level, names both the installed and current worktree paths plus the corrective action taken. `cleanup_stale_sweep_units` early-returns when no unit file exists, when `WorkingDirectory=` is empty, and when the installed path matches — it only emits on a genuine path mismatch, exactly the fault this change repairs. Non-user-facing internal operational log; no human intent needed.
- **Operator judgment: is re-asserting the sweep timer inside `dispatch-tick-recover` the right call?** — `dispatch-tick-recover:179` sets `MAIN_WORKTREE="$PROJECT_ROOT"`, the same derivation `dispatch-spawn-tick`/`dispatch-schedule-reseed` use, so the recovery path cannot flap a healthy unit. Heartbeat-before-sweep at `:187`/`:194` mirrors the pre-existing ordering (liveness unit first, GC timer second), both best-effort `|| true`. `ensure_sweep_timer`'s hot path is a byte-for-byte content compare plus one `systemctl is-active` check with an early return — the disable+rewrite only runs on genuine drift. Internal, revisable dispatch plumbing; no defect to fix, no user intent required.

---

**Terminal disposition: clean pass.** No residue escalates to office-hours.

_Plan authored by the Step 2 Opus triage subagent. Verdicts updated in place as each item resolved._

natb1:
## qa · attempt 0
failed: none

10/10 script-verifiable QA items passed (all 6 affected test suites green, host unit dir
untouched, cleanup_stale_sweep_units edge cases + end-to-end self-heal verified directly).
2 needs-human-judgment items (WARNING message clarity; ensure_sweep_timer placement in
dispatch-tick-recover) were assessed already-satisfied by the Step 3.5 Opus disposition pass
and dropped as PASS — no code defect, no human input needed. Clean pass, no residue.


## review · attempt 1
failed: none

Combined code-review + security-review pass (surface=code). 15 findings surfaced,
11 fixed (3 required-security + 8 code-review residue), 2 refuted by adversarial
verify, 1 false-positive, 1 out-of-scope (not filed — pre-existing gap shared by
ensure_heartbeat_units/ensure_recover_unit, not attacker-reachable). Key fix:
test-dispatch-tick-recover.sh never redirected DISPATCH_SWEEP_TIMER_UNIT_DIR/
_SYSTEMCTL_CMD despite the PR adding a real ensure_sweep_timer call reachable
from that suite — the same host-unit-leak class this tactic exists to close,
reintroduced by its own Unit 2 change. Fixed by redirecting those vars and
promoting the host-unit-hash leak guard into the shared test fixture (XDG-aware,
covers all five installable units + timers.target.wants + a systemctl-call
recorder). Also deduped cleanup_stale_sweep_units/cleanup_stale_heartbeat_units
into a shared helper and added missing unit tests for the new sweep self-healer.
No deviation — all fixes verified, all affected suites green, no host unit-dir
writes.

natb1:
## Review-fix disposition

Combined code-review + security-review pass over the diff (surface=code). 15 findings surfaced, 11 fixed, 2 refuted by adversarial verify, 1 false-positive, 1 out-of-scope. Fixes landed in ef07b18e.

- **input-validation-0** (`dispatch-tick-recover:194`): new unconditional `ensure_sweep_timer "$MAIN_WORKTREE" || true` call was reachable by `test-dispatch-tick-recover.sh` without redirecting `DISPATCH_SWEEP_TIMER_UNIT_DIR`/`_SYSTEMCTL_CMD`, so the suite could write/enable the real host sweep unit. Fixed: redirected both vars in `tr_setup`/`tr_teardown` and added the host-unit-hash leak guard to this suite. Commit ef07b18e.
- **input-validation-3** / **red-team-3** (`test-dispatch-spawn-tick.sh:26,264`): the host-unit-dir leak guard only hashed two `.service` files, hardcoded `$HOME/.config` instead of resolving `XDG_CONFIG_HOME` like `lib.sh` does, and had no coverage in `test-dispatch-tick-recover.sh` — the one suite that actually exercises the newly-added sweep-timer recovery call. Fixed: promoted the guard into `dispatch-test-fixture.sh`, XDG-aware, covering all five installable units + `timers.target.wants` symlinks + a PATH-prepended recording `systemctl` stub, re-checked via `report_results` and an `EXIT` trap; wired into `test-dispatch-tick-recover.sh`. Commit ef07b18e.

### Fixed (code-review)

- **residue-0**: `test-dispatch-tick-recover.sh` didn't redirect the sweep-timer env vars (same root cause as input-validation-0, confirmed independently by code-review) — added positive assertions that recover installs the sweep unit into the redirected dir.
- **residue-1**: leak guard hardcoded `$HOME/.config` instead of resolving XDG like `lib.sh` — now single XDG-aware guard in the shared fixture.
- **residue-2**: guard hashed the whole unit file, which the live dispatch chain's legitimate rewrites (e.g. `Environment="PATH=..."`) could flip — now fingerprints only `WorkingDirectory=`.
- **residue-3**: `cleanup_stale_sweep_units` had zero tests despite its heartbeat twin being tested — added 4 cases to `test-lib-systemd-units.sh` plus 3 end-to-end assertions in `test-dispatch-tick-recover.sh`.
- **residue-4**: `cleanup_stale_sweep_units`'s doc comment was spliced into the middle of `ensure_sweep_timer`'s header block — moved back to a clean split.
- **residue-5**: doc comment said "mirroring cleanup_stale_heartbeat_units above" but that function is defined below — corrected the direction.
- **residue-6**: `cleanup_stale_sweep_units`/`cleanup_stale_heartbeat_units` were byte-identical apart from unit names — extracted a shared `cleanup_stale_unit_pair` helper; both are now thin wrappers.
- **residue-7**: the guard's comment overclaimed its coverage (two `.service` files) vs. what it now actually checks — comment updated to match the fixture-wide guard (all five units, symlinks, systemctl-stub recorder).

All 11 fixes verified by re-running the affected suites (`test-dispatch-tick-recover.sh`, `test-dispatch-spawn-tick.sh`, `test-lib-systemd-units.sh`, and the four `dispatch-schedule-*` suites) — all green, no host unit-dir writes.

- **red-team-0** (`dispatch-tick-recover:194`): claimed any local run (including autonomous sessions) could clobber the live host sweep unit through this same gap. Refuted 1-1 (skeptic split; second-pass skeptic prevailed) — reproduced that the Bash-tool sandbox's write-allowlist and D-Bus/Unix-socket network-namespace isolation independently block both the file write and the `systemctl --user` call for any sandboxed session, and CI runs on an ephemeral GitHub-hosted runner with no live unit at all. The only surviving channel is a human running the suite directly in a non-sandboxed terminal with a real `$HOME` — real but much narrower than the finding's "any autonomous session" framing, and not unique to this PR's change (any unhermetic test writing to `$HOME` shares the property). Note: this refutation is about severity/exploitability, not mechanism — the same underlying gap is what input-validation-0/residue-0 fixed.
- **red-team-2** (`lib.sh:2692`): claimed `cleanup_stale_sweep_units`'s disable-before-rewrite ordering could cause a "self-perpetuating" DoS if the rewrite path failed. Refuted — the mismatch precondition isn't independently attacker-triggerable (it requires the main worktree's checkout path to actually move), the pattern is a deliberate mirror of the already-shipped, tested `cleanup_stale_heartbeat_units` (same best-effort, never-abort contract), the disarm is not permanent (the very next tick/reseed/worker-launcher call re-enables the timer once the underlying fault clears), and any fault severe enough to break the write path would identically break the sibling `ensure_heartbeat_units`/`ensure_recover_unit` calls from the same scripts, producing loud multi-subsystem warnings rather than a silent isolated failure.

### False-positive (security)

- **red-team-4** (`lib.sh:2680`): claimed `installed_workdir=$(sed -n 's/^WorkingDirectory=//p' "$service_path" | head -1)` mishandles a unit file with multiple `WorkingDirectory=` lines. `systemd` unit files never legitimately carry more than one `WorkingDirectory=` in the same section, and this helper only ever reads units this codebase itself writes (always exactly one) — no reachable path produces the claimed multi-line input.

### Out-of-scope (security)

- **input-validation-1** (`lib.sh:2747`): `ensure_sweep_timer` validates `main_worktree` only against unit-file syntax breakers (newline/space/double-quote/backslash), not that the path is an actual valid worktree. Classified out-of-scope — the same validation gap exists identically in the pre-existing `ensure_heartbeat_units`/`ensure_recover_unit`, so it isn't introduced by this diff; not meaningful enough on its own to warrant a standalone follow-up (below the filing threshold — no exploitable path from an untrusted `main_worktree` value, which is always internally derived, not attacker-supplied).

PR #3003, opened 2026-07-31
dispatch-reclaim-audit: match SYSLOG_IDENTIFIER, never render an unmeasurable sweep as zero reclaims

natb1:
## Context

`dispatch-reclaim-audit` reads reclaim events from the systemd user journal
with a unit filter, `journalctl --user -u dispatch-tick`. `-u` matches
`_SYSTEMD_UNIT`, but the ticks that emit reclaim lines run either as transient
`dispatch-reseed-*.service` units or as children of the durable heartbeat
under `user@1000.service` — neither carries `_SYSTEMD_UNIT=dispatch-tick`.
They do carry `SYSLOG_IDENTIFIER=dispatch-tick`. So the unit filter matched
nothing, and the audit reported zero reclaims against a true count of 14+ on
the host — while failing OPEN to zero rather than erroring, so the vacuous
result read as a clean, healthy ledger.

## What changed

**Unit 1 — match the syslog identifier, and make the journalctl call testable**

- `dispatch-reclaim-audit`: switched the journalctl filter from `-u
  dispatch-tick` (unit match) to `-t dispatch-tick` (syslog-identifier
  match) at the source string, the live read, and the `command -v` check.
  Added `JOURNALCTL="${DISPATCH_RECLAIM_JOURNALCTL_CMD:-journalctl}"`
  (mirrors `dispatch-schedule-reseed`'s `SYSTEMD_RUN_CMD` override shape) so
  the journalctl invocation is stubbable in tests. Updated doc comments and
  the stderr warning text to match.
- `dispatch-tick`: comment-only update of the operator-facing journal
  inspection example from `-u` to `-t dispatch-tick`.
- `test-reclaim-audit.sh`: added a journalctl-stub test case that models the
  real host (`-u dispatch-tick` → empty, `-t dispatch-tick` → the fixture
  reclaim lines) and regression-locks the filter choice.

**Unit 2 — make "could not measure" impossible to read as "zero reclaims"**

- Replaced the boolean `SWEEP_AVAILABLE` with two sentinel strings mirroring
  `dispatch-graph-main-red-sync`'s `UNKNOWN` shape:
  - `SWEEP_STATUS` (`ok` / `unavailable`) — was the sweep source readable.
  - `SWEEP_CROSSCHECK` (`ok` / `mismatch` / `unavailable` / `skipped`) — a
    second, identifier-**un**filtered journal probe over the same window,
    compared against the filtered count. `mismatch` means the identifier
    filter is undercounting — exactly this bug's failure mode.
- Surfaced both in the human report (three distinct NOTE lines) and JSON
  output (`status`, `crosscheck`, `crosscheck_journal_lines`,
  `crosscheck_filtered_lines`; `available` kept as a legacy-compatible
  derived key).
- Exit code 3 now signals "counts are not trustworthy" (`SWEEP_STATUS !=
  ok` or `SWEEP_CROSSCHECK == mismatch`) — the full report is still emitted
  before exiting, never suppressed.
- Extended `test-reclaim-audit.sh` with cases covering the fixture path, the
  healthy live-journal path, a journalctl error, and the undercount
  regression lock (journal holds 6 lines, filter matches 2 → `mismatch`,
  exit 3).

## Verification

- `test-reclaim-audit.sh` — 30/30 passed.
- `run-unit-tests.sh --pr-scripts` — all suites passed.
- `run-lint.sh --prose` — clean.
- `grep -rn 'journalctl.*-u dispatch-tick' .claude/` — no functional site
  survives.

Graph node: tactic-reclaim-audit-journal-unit-filter


natb1:
## QA summary — clean pass

No browser component detected (shell-script-only change under
`.claude/skills/dispatch-propagate/scripts/`). All 6 plan items were
script-verifiable; no `needs-browser` or `needs-human-judgment` items.

**Results: 6/6 PASS, 0 FAIL, 0 SKIP.**

### 1. The PR's own test suite passes end to end — PASS
`bash .claude/skills/dispatch-propagate/scripts/test-reclaim-audit.sh` → 30/30 passed, exit 0.

### 2. An unreadable journal reports "unavailable" and exits 3 — never a clean zero — PASS
`DISPATCH_RECLAIM_JOURNALCTL_CMD=/bin/false dispatch-reclaim-audit --json` → `status: "unavailable"`, `available: false`, all counts 0, exit 3. No silent healthy-looking zero.

### 3. Filter regression lock: `-t dispatch-tick` counts what `-u dispatch-tick` missed — PASS
Stub journalctl (`-u` → empty, `-t` → 6 fixture reclaim lines) → `status: "ok"`, `crosscheck: "ok"`, `dead_session_stranded_events: 4`, `live_worker_redundant_events: 2`, exit 0.

### 4. Cross-check detects an undercount: `mismatch` plus exit 3 — PASS
Stub journalctl (`-t` → 2 lines, unfiltered `--grep` → 6 lines) → `crosscheck: "mismatch"`, `crosscheck_journal_lines: 6`, `crosscheck_filtered_lines: 2`, exit 3 — the direct regression lock for the reported bug's failure mode.

### 5. Live host reality check: audit count agrees with a hand-run journalctl — PASS
Hand count via `journalctl --user -t dispatch-tick --since '7 days ago' | grep -ci 'reclaimed reservation'` = 49. Audit's `crosscheck_journal_lines` and `crosscheck_filtered_lines` both = 49, `status: "ok"`, `crosscheck: "ok"`, exit 0. Counts agree exactly.

### 6. No functional `-u dispatch-tick` site survives, and operator docs match — PASS
`grep -rn -- '-u dispatch-tick' .claude/` matches only inside `test-reclaim-audit.sh`, in comments/stub logic that deliberately models the old broken filter — no functional site. `grep -rn -- '-t dispatch-tick' .claude/` confirms the corrected filter at every site the PR body enumerates (`dispatch-reclaim-audit`'s source string, live read, doc comments; `dispatch-tick`'s operator doc comment).

No bugs found, nothing deferred to office-hours, no follow-ups filed.

natb1:
## qa · attempt 0
failed: none. 6/6 QA plan items passed (script-verifiable, no browser component). Confirmed: -t filter recovers reclaims -u missed; unavailable journal exits 3 (not silent zero); cross-check mismatch detected with exit 3; live-host audit count (49) agrees with hand-run journalctl; no functional -u dispatch-tick site survives.


## review · attempt 1
19 subagents: 12 raw findings, deduped to 8. 4 code-review residue findings
fixed (missing --until bound on both journalctl probes; asymmetric
cross-check comparison missing an `inconsistent` sentinel; cross-check
--grep literal not derived from the shared pattern constant; test never
asserted the stubbed journalctl argv) — commit dabfbb45. 7 Required security
findings (client-spoofable SYSLOG_IDENTIFIER provenance, unvalidated
journalctl-cmd override, cross-check unfiltered-probe/foreign-emitter
concerns, theoretical path-traversal in the worktree token) all refuted by
adversarial verification: no privilege boundary crossed, report-only
diagnostic with no automated consumer, live-host measurement showed zero
historical contamination. 1 false-positive (unquoted test heredoc, no
untrusted input) dismissed at classification. No deviation, no follow-ups.

natb1:
## /review-fix — combined review pass

19 subagents ran across finders → dedup → classify → adversarial-verify → fix →
residue. 12 raw findings surfaced, deduped to 8 distinct issues; 4 fixed, 4
security findings refuted after adversarial verification, 1 false-positive
dismissed at classification. No deviation; no out-of-scope follow-ups to file.

### Fixed

- **residue-0** — `dispatch-reclaim-audit:217` — Neither journalctl probe (the
  `-t dispatch-tick` read nor the unfiltered `--grep` cross-check) was bounded
  at the upper end with `--until "$UNTIL"`, so the two probes could scan
  slightly different windows than advertised. Fixed: both invocations now pass
  `--until "$UNTIL"`, plus a test assertion that every logged journalctl call
  carries it. — commit `dabfbb45`
- **residue-1** — `dispatch-reclaim-audit:230` — The cross-check comparison
  was asymmetric: only `journal > filtered` was handled as non-`ok`; a
  `journal < filtered` result (filtered count somehow exceeding the unfiltered
  probe) fell through to `ok` silently. Fixed: added a distinct `inconsistent`
  `SWEEP_CROSSCHECK` sentinel with its own stderr/report NOTE (deliberately not
  exit 3), documented, with test case D. — commit `dabfbb45`
- **residue-2** — `dispatch-reclaim-audit:225` — The cross-check probe's
  `--grep` argument hardcoded the reclaim-line pattern as a second inline
  literal instead of reusing the single `$ALL_RECLAIM_RE` constant already
  defined for the authoritative counts. Fixed: probe now derives from the same
  constant. — commit `dabfbb45`
- **residue-3** — `test-reclaim-audit.sh:198` — `JOURNALCTL_ARGV_LOG` was
  declared and populated by every stub invocation but never asserted on, so
  the test didn't actually pin the constructed argv (only the display
  string). Fixed: added assertions that logged invocations use `-t
  dispatch-tick`, never `-u dispatch-tick`, that both probes ran, and that
  every probe carries `--until`. — commit `dabfbb45`

### Required (security)

_None._ All security findings classified `Required` were refuted by adversarial
verification below — none required a fix.

Every `Required` security finding below was independently verified by a
skeptic instructed to build the strongest case for false-positive, defaulting
to `refuted` under uncertainty. All 7 were refuted — full rationale in the
workflow transcript; condensed here:

- **input-validation-1** — `dispatch-reclaim-audit:224` (unfiltered
  cross-check probe accepts arbitrary journal text). Refuted: requires
  same-user code execution to inject; narrowing the probe would reintroduce
  the exact silent-undercount bug class this diff exists to fix; impact is
  cosmetic (a stderr note), no automated consumer exists.
- **input-validation-2** — `dispatch-reclaim-audit:140`
  (`DISPATCH_RECLAIM_JOURNALCTL_CMD` read unvalidated). Refuted: every
  attack vector requires the attacker to already control the operator's
  shell/env/PATH, which is a strictly greater capability than anything the
  override grants; this is a human-run, non-automated diagnostic.
- **input-validation-3** — `dispatch-reclaim-audit:274` (worktree token
  parsed without strict allowlist, theoretical path traversal). Refuted:
  empirically confirmed the write path is unconditionally gated by
  `find_project_dir`/`basename`, which structurally cannot contain `/`, so
  no traversal-capable token can ever reach a write or a pre-existing read.
- **red-team-0** — `dispatch-reclaim-audit:217` (`-t` matches
  client-spoofable `SYSLOG_IDENTIFIER` instead of trusted `_SYSTEMD_UNIT`).
  Refuted: verified live on the host — zero contamination across the full
  retained journal (3,199/3,199 matching lines from trusted units); the
  claimed injection path (a worker session inheriting dispatch-tick's
  stream) is architecturally false — workers run under a separate daemon
  unit with no fd inheritance from dispatch-tick.
- **red-team-1** — `dispatch-reclaim-audit:140` (`SWEEP_SOURCE` hardcoded
  independent of an overridden `$JOURNALCTL`, "forged provenance"). Refuted:
  exploiting it requires operator-equivalent code execution already; the
  artifact is a human-read report, not a control-plane gate; the override
  shape mirrors an existing codebase convention
  (`dispatch-schedule-reseed:479`) and matches the tactic's own plan.
- **red-team-2** — `dispatch-reclaim-audit:230` (cross-check can't detect
  upstream journald rate-limiting/retention loss). Refuted: accurate but
  out of the diff's declared scope (the filter-flag bug class only); no
  adversarial threat model applies on a single-operator host; non-gating
  diagnostic.
- **red-team-3** — `dispatch-reclaim-audit:231` (one foreign-identifier
  line latches `mismatch` for the whole window). Refuted: requires
  same-user code execution; the report is never suppressed — it prints a
  loud, self-explaining warning by design; the condition self-heals as the
  window ages.

### False-positive (security)

- **red-team-4** — `test-reclaim-audit.sh:234` — flagged an unquoted
  heredoc in `make_journalctl_stub`. Dismissed at classification: test-only
  code with no untrusted input reaching the heredoc.

### Informational / Dismissed / Deferred / Out-of-scope (security)

_None._

### Coverage

All finders completed; no probe-wave short-circuit. `deps=false` (no
dependency-audit run), CodeQL returned zero open alerts, erosion scan found
nothing.

PR #2997, opened 2026-07-31
Attention: three-tier ranking floor — (tier, rank) lexicographic ordering, max-based blocking lift, strategy-main-health tier-3 migration (tactic-attention-tier-ranking)

natb1:
## Context

Ranking today is a single scalar: `resolveAttention` composes `authored`/`signal`/`capture`
terms into `ResolvedAttention.value`, and every consumer (dispatch selector, frontier
projection, office-hours queue) orders by that one number. Three defects followed:
defects/production incidents had no floor other than an ever-inflating boost number
(`strategy-main-health` at `boost: 100`, protected by a numeric >=100 write-path guard);
`attributes.bug_fix`/`attributes.security` marks were authored but completely inert (no
code read them — a `bug_fix: true` node with no boost resolved to rank 0, dispatched
*after* every ordinary feature node carrying a boost of 3); and a boost value had no scale
it belonged to (a value chosen mid-way in tier 1 would wrongly dominate if carried into a
higher tier).

This replaces the single scalar with a lexicographic `(tier, rank)` pair, tier outermost.
Tier 1 is the default (ordinary work), tier 2 is bug-fix/security marks or explicit lifts,
tier 3 is production/main issues. It also drops a superseded backward-flow doctrine: a
2026-07-07 design let a blocked node's boost flow additively into its blocker via
`blocked_by`, which was superseded 2026-07-13 (recorded failure: unrelated `blocked_by`
compounding silently overtook intentionally top-ranked nodes). The replacement is a
max-based precedence lift in the selector — a blocker's *selection precedence* lifts to at
least the max `(tier, rank)` of the nodes it blocks, recursive, max-based, never additive.

Recorded on `intentions/strategy-graph-drives-dispatch.md` (2026-07-18 / 2026-07-21
clarifications); implemented per the plan on `intentions/tactic-attention-tier-ranking.md`.

## What changed

- **`packages/intentionsutil/src/schema.ts`**: `Attention.tier` (a per-tier boost
  provenance tag, default 1); exported `ownTier(node)` helper; new `validateGraph` rules
  19 (mark shape: `bug_fix`/`security` must be boolean, `attributes.tier` must be 2 or 3
  when present) and 20 (a node's `attention.tier` must equal its own tier — a tier change
  forces a fresh boost selection rather than silently carrying a stale-scale value); rule
  18 rewritten from a numeric `>=100` dominance guard to a structural tier-3 dominance
  guard (only `strategy-main-health` may author an explicit `attributes.tier: 3`; the same
[...]

PR #3012, opened 2026-08-01
tactic-attention-boost-scripts: script the boost-to-top-rank attention-write operation

natb1:
## Context

Today, escalating work in the intention graph means hand-editing `attention.boost`
in a node's YAML frontmatter and eyeballing the result against the selector. The
recurring failure mode: `boost:` holds a node's OWN authored claim, but ranking
order is the COMPOSED rank `resolveAttention` produces by summing every distinct
source claim flowing into the node — a boost sized against the raw `boost:`
column of the current leader lands at #2, not #1. There is also a hard schema
ceiling (rule 18: no node's `attention.boost` may match or exceed
`strategy-main-health`'s live boost without the literal `ACK:
main-health-dominance` in its rationale), which a hand-edit discovers only when
`graph-commit` refuses the land.

This implements the ungated "boost to top rank" slice of `strategy-graph-drives-dispatch`'s
2026-07-21 clarification: a considered, author-invoked operation (never a
mechanical max+epsilon race) that shows the live ranking, computes the minimal
boost that tops it, and requires an author-supplied rationale.

Deferred, not in this PR: the tier-aware default and the distinct tier-change
script named in the tactic statement — no tier concept exists in code yet, and
`tactic-attention-tier-ranking` (PR #2997, unmerged and parked) has not shipped
the per-tier boost namespace. Both script headers state this explicitly.

## Unit 1 — pure boost planner

`packages/intentionsutil/src/boost.ts` (new): `planBoost(nodes, targetId, mode, opts?)`
computes, purely from the resolved graph, the minimal own-`attention.boost`
that tops the selector's candidate list (or reaches a named composed rank),
respecting the `strategy-main-health` exemption and the rule-18 ceiling. Uses
an integer binary search over the monotone `topsAt(b)` feasibility predicate,
self-checked at the boundary so a non-monotone graph throws rather than
silently returning a wrong minimum. Reports an honest `unreachable_reason`
(naming the claim-riding peers, or the search bound) when no boost tops the
ranking. `packages/intentionsutil/src/schema.ts` gains two small pure
extractions backing it: `kindIsNotGoalLayer` exported, and
`dominantMainHealthBoost`/`MAIN_HEALTH_DOMINANCE_ACK` factored out of
`validateGraph`'s inline rule-18 threshold read (no behavior change).

## Unit 2 — CLI

`packages/intentionsutil/scripts/boost-node.ts` (new): a thin CLI wrapping
`planBoost` with a preview mode (human-readable ranking table or `--json`) and
a `--write` mode. `--write` mandates a non-empty `--rationale` — this script
never synthesizes one — resolves the boost from `--boost` or the planner's
recommendation, and refuses (exit 4) a boost that would need the rule-18 ACK
unless `--ack` is passed, in which case it appends
`MAIN_HEALTH_DOMINANCE_ACK` to the rationale. Writes through the validated
`readNode`/`validateNode`/`validateGraph`/`writeNode` path.

## Unit 3 — landing wrapper, harness, CI

`packages/intentionsutil/scripts/boost-node` (new, executable): the
author-invoked entry point. Fetches origin/main, refreshes the local node file,
CAS-pins via an optional `--base`, writes via Unit 2's CLI, lands through
`graph-commit --base/--expect`, and post-land re-verifies the new ranking
against a fresh origin/main snapshot. A failed write or a stale `--base`
rolls back to the pristine origin/main content via an EXIT trap; a failed
post-land assertion (exit 6) reports the landed state without attempting any
rollback, since the write already landed. `packages/intentionsutil/scripts/test-boost-node.sh`
(new) is a 7-case functional harness (preview no-op, successful landing,
missing-rationale refusal, `--base` staleness, byte-identical rollback on a
`graph-commit` failure, and the rule-18 ACK gate), wired into
`.github/workflows/unit-tests.yml` immediately after the `park-node` CAS-guard
step.

## Verification

- `npx vitest run --project packages/intentionsutil --root .` — 741/741 tests
  pass (40 files), including 13 new `boost.test.ts` cases and the
  `boost-node.test.ts` CLI cases.
- `npx tsc -p packages/intentionsutil/tsconfig.json --noEmit` — clean.
- `run-lint.sh --app packages/intentionsutil --prose` — clean.
- `packages/intentionsutil/scripts/test-boost-node.sh` — 7/7 pass.
- `npx tsx packages/intentionsutil/scripts/validate-graph.ts intentions` — ok,
  468 nodes.

Manual verification (preview against the live graph, honest refusals, one real
landing observed end to end) is called out in the plan as a judgment check for
office hours / QA — not run autonomously in this session.

Graph node: `tactic-attention-boost-scripts`

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -236,6 +236,8 @@ jobs:
         run: .claude/skills/dispatch-propagate/scripts/test-dispatch-daemon-liveness.sh
       - name: Run park-node CAS-guard tests
         run: packages/intentionsutil/scripts/test-park-node.sh
+      - name: Run boost-node tests
+        run: packages/intentionsutil/scripts/test-boost-node.sh
       - name: Run transition-node CAS-guard tests
         run: packages/intentionsutil/scripts/test-transition-node.sh
       - name: Run graph-commit tests
--- a/packages/intentionsutil/scripts/boost-node
+++ b/packages/intentionsutil/scripts/boost-node
@@ -1,0 +1,437 @@
+#!/usr/bin/env bash
+# boost-node — the author's "boost this node to the top of the frontier"
+# operation, landed on main (tactic-attention-boost-scripts, unit 3).
+#
+# The bash landing half of the boost tool. The planner
+# (packages/intentionsutil/src/boost.ts) and the store-level CLI
+# (packages/intentionsutil/scripts/boost-node.ts) do all the thinking; this
+# script does the git: fetch origin/main, refresh the node from it, delegate the
+# mutation to the CLI's `--write` mode, land through `graph-commit`, then
+# re-read origin/main and assert the write actually achieved what it promised.
+#
+# It authors no node content — the CLI's readNode → validateNode → writeNode
+# path is the single validation gate — and it never edits any node but the one
+# named on the command line.
+#
+# Why a wrapper at all: the boost the author wants is a claim about the CURRENT
+# landed graph. Computing it against a stale worktree copy sizes it against a
+# frontier that no longer exists, and writing it from a stale copy reverts
+# whatever landed in between. So both the plan and the write run against
+# origin/main's content, and the land carries a `--base` compare-and-swap token
+# pinned to the blob that was read (same guarantee park-node gives).
+#
+# REPO_ROOT is derived from this script's OWN on-disk location, never cwd. A
+# caller invoking the main checkout's copy by absolute path from inside a
+# worktree therefore targets the main checkout — deliberate, and the reason the
+# `graph-commit -C "$REPO_ROOT"` below is mandatory rather than cosmetic.
+#
+# Usage:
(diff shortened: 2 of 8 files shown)


natb1:
Closing without merging — this branch's premise was superseded while it sat.

It implements boost-to-top-rank against the numeric main-health dominance
ceiling (`MAIN_HEALTH_DOMINANCE_ACK`, `dominantMainHealthBoost`). PR #2997's
tier model **deleted both symbols** from `main` and replaced numeric
boost-dominance with tier-3 authorship, so merging would reintroduce a removed
API. The merge conflicts in `packages/intentionsutil/src/schema.ts` (5 hunks)
are that supersession, not an incidental drift.

The node `tactic-attention-boost-scripts` is NOT abandoned. Its own statement
already anticipated this — "boost-to-top-rank (now) **and tier-change (gated on
the tier model)**" — and the tier-change half is exactly the boost-to-tier
conversion still owed. The node returns to draft to be replanned against the
shipped tier model rather than reworked on a base 118 commits stale.

Author-directed 2026-08-03.

PR #3013, opened 2026-08-01
test isolation: shared decision-log seam stops test fixtures leaking into production routing-decisions.jsonl

natb1:
## Context

`lib-decision-log.sh` resolves the routing-decision log **once**, at source
time (`.claude/skills/dispatch-propagate/scripts/lib-decision-log.sh:68`):

```
DECISION_LOG_FILE="${DISPATCH_DECISION_LOG_FILE:-${DISPATCH_DECISION_LOG_DIR:-$HOME/.local/share/commons-dispatch}/routing-decisions.jsonl}"
```

`test-dispatch-standdown.sh` set neither override, so its fixture runs
appended synthetic rows to the **production** routing-decisions.jsonl — the
same file the fleet's own defect detects read as evidence (it is the sole
detect for the bootstrap plan's bug N, and it carries the site records the
frozen-session and standdown sweeps are validated against). By plan time the
production log carried ~39 fixture rows tagged `"node":"tactic-some-node"`.

The actual defect class: decision-log isolation was opt-in per test file, so
omission was silent. This mirrors a prior incident
(`tactic-sweep-timer-unit-dir-leak`) where a test run rewrote a live systemd
unit's `ExecStart` to a `/tmp` path.

## Unit 1 — shared decision-log isolation seam

New `.claude/skills/dispatch-propagate/scripts/lib-test-decision-log-guard.sh`,
sourced by both shared test harnesses (`dispatch-test-fixture.sh`,
`test-helpers.sh`). At source time it unconditionally redirects
`DISPATCH_DECISION_LOG_DIR` into a fresh `mktemp -d` sandbox (mirroring the
existing `DISPATCH_GUARD_BIN_DIR`/`PATH` prior art) and defines
`dispatch_decision_log_guard_check`, an idempotent end-of-suite assertion —
modeled on the existing host-systemd leak guard — that the decision-log
**seam** (not the file's contents, which the live fleet legitimately appends
to concurrently) still points away from production. Wired into both harnesses'
`report_results()` and (for `dispatch-test-fixture.sh`) its `EXIT` trap.

## Unit 2 — fix two teardowns that re-exposed the production path

`test-dispatch-select-tick.sh` and `test-dispatch-tick.sh` each `export
DISPATCH_DECISION_LOG_DIR` in per-test setup and previously `unset` it in
teardown — after Unit 1, that `unset` discarded the harness default too,
re-exposing the production path for the window between a teardown and the
next setup. Both teardowns now restore the harness default
(`export DISPATCH_DECISION_LOG_DIR="$DISPATCH_TEST_DECISION_LOG_DIR"`) instead
of unsetting.

## Unit 3 — regression test

New `.claude/skills/dispatch-propagate/scripts/test-decision-log-isolation.sh`
(picked up automatically by `run-unit-tests.sh`'s `test-*.sh` glob):

- **Part A (dynamic)** — runs `test-dispatch-standdown.sh` and
  `test-dispatch-stop-hook.sh` under a redirected `HOME` with the inherited
  decision-log overrides stripped, and asserts neither writes a
  `routing-decisions.jsonl` (or its `.lock`) under that fake `HOME`.
- **Part B (static ratchet)** — every `test-*.sh` in the directory must either
  source one of the two harnesses, set a decision-log override directly, or be
  named in an explicit `KNOWN_UNISOLATED` allowlist (verified to still exist
  on disk), so a new suite can't silently regress this by omission.

## Out of scope

Purging the ~39 existing fixture rows from the production
`routing-decisions.jsonl` — that's a separate, optional operator cleanup; see
the node body's Verification section for the post-merge check
(`grep -c '"node":"tactic-some-node"' ~/.local/share/commons-dispatch/routing-decisions.jsonl`
should stop growing).

## Verification

- `test-decision-log-isolation.sh` — 157/157 assertions pass.
- Full dispatch shell suite (`run-unit-tests.sh --pr-scripts`) — all suites
  green, including the seven previously-passing suites this change touches or
  depends on (`test-dispatch-standdown.sh`, `test-dispatch-stop-hook.sh`,
  `test-dispatch-select-tick.sh`, `test-dispatch-tick.sh`,
  `test-lib-frozen-session-park.sh`, `test-lib-standdown-recheck.sh`,
  `test-dispatch-fleet-watch.sh`).
- `run-lint.sh` — clean.
- Reproduced the defect against pre-fix code (verify block 1 in the plan):
  `test-dispatch-standdown.sh` under a redirected `HOME` with overrides
  stripped writes a production-shaped `routing-decisions.jsonl`; post-fix it
  does not.

Graph node: `tactic-test-decision-log-prod-leak`


natb1:
## QA summary — tactic-test-decision-log-prod-leak (PR #3013)

No browser component detected (pure dispatch-fleet test-infrastructure change) — every triage item was `script-verifiable` or `needs-human-judgment`.

### Items executed

**Shell-command lane (9/9 PASS, 0 FAIL, 0 SKIP):**

1. New regression suite passes on its own terms — PASS (157/157 assertions)
2. The two originally-leaking suites (`test-dispatch-standdown.sh`, `test-dispatch-stop-hook.sh`) write nothing under a redirected HOME — PASS
3. The two teardown-fixed suites (`test-dispatch-select-tick.sh`, `test-dispatch-tick.sh`) leak nothing under a redirected HOME — PASS (both re-verified with the sandbox disabled after an initial run false-failed with `sync-failed` decisions — confirmed pre-existing on the origin/main baseline of the same files too, caused by this session's sandboxed network-namespace isolation blocking the suite's own simulated `git fetch`/`merge --ff-only`, not a PR regression)
4. The guard is not vacuous — fires on a genuine leak and is idempotent — PASS
5. A leak turns the owning suite's exit code non-zero — PASS
6. The static ratchet catches a newly-added un-isolated suite — PASS
7. Previously-green neighbouring suites stay green (7 suites) — PASS
8. Full dispatch shell suite green end to end (`run-unit-tests.sh --pr-scripts`) — PASS; new suite auto-discovered, guard lib correctly not run as a suite
9. No sandbox-dir litter / cross-suite scratch-dir interference — PASS

**Not walked (needs-human-judgment, per triage classification) — 4 items**, resolved by the Step 3.5 disposition workflow (see below):
10. The real production routing-decision log is untouched by a full QA run
11. `KNOWN_UNISOLATED` allowlist entries genuinely cannot reach `lib-decision-log.sh`
12. Seam-only assertion is the right contract (guard cannot see mid-run leaks)
13. Post-merge: production fixture rows stop growing (planned-deferral)

No bugs found. No browser walkthrough ran (no browser component).

### Disposition triage

- **id 13 — Post-merge: production fixture rows stop growing** → class `needs-main`, verify `n/a` (planned-deferral). Rationale: the PR's own Verification section documents this as an out-of-scope post-merge operator observation requiring elapsed fleet runtime after merge — not assertable at merge time.

`needs-main` items are filed as follow-up residue and dropped from the escalation set, so they do not park this PR. On the graph-native node lane this means a `## needs-main residue` section in the tactic's own body rather than a `blocked_by` GitHub issue — see **Filed follow-ups** below.

No `opus-fixable` items were found, so the auto-fix lane did not run. All residue resolved to `already-satisfied` or `needs-main` — the escalation set is empty, so this is a **clean pass**.

**Filed follow-ups:**
- id 13 → filed as a `## needs-main residue` section in `intentions/tactic-test-decision-log-prod-leak.md` (Verifiability: `WAIT` — awaiting this PR's merge plus elapsed fleet runtime), landed to `origin/main` via a standalone `graph-commit` from a scratch worktree (base `ee2422a`, commit `b46a7ce8`). It will be drained by `tactic-main-qa-phase` after `review → main-qa` fires post-merge.

**Assessed by Opus (already satisfied) — dropped as PASS:**
- **id 10 — The real production routing-decision log is untouched by a full QA run.** Provable from readable evidence: `"node":"tactic-some-node"` is a test-fixture id the live fleet never routes, so fixture-row growth is unambiguously test-attributable. The production log held 42 fixture rows (newest timestamp 2026-08-01T02:27:36Z) throughout this QA pass (~10:54–11:31 EDT), while the file's mtime showed the live fleet still actively appending non-fixture rows in that window. Zero fixture rows were added across the QA run.
- **id 11 — `KNOWN_UNISOLATED` allowlist entries genuinely cannot reach `lib-decision-log.sh`.** Verified at HEAD: only 5 suites actually source `lib-decision-log.sh`; the 7 allowlisted suites exec unrelated scripts, none of which reach it. The residual "can rot" risk is inherent to any allowlist and is bounded by the static ratchet plus the dynamic Part-A check.
- **id 12 — Seam-only assertion is the right contract.** Design-sufficiency sign-off on a non-user-facing, fully revisable test instrument. The documented rationale (live-fleet concurrent appends make content-diffing flaky, so the seam variable is the only stable assertion target) is sound on its face, and the acknowledged gap (a suite that repoints the seam mid-run and restores it before exit) is already covered dynamically for the known leakers by Part A's fake-HOME check.

### Sandbox artifact note

Items 3 and 6 initially false-failed under this session's default sandboxed Bash — a network-namespace isolation blocking, respectively, a test suite's own internal `git fetch`/`merge --ff-only` simulation and a write to the read-only `.claude/skills/` carve-out. Both were re-run with `dangerouslyDisableSandbox: true` and passed clean; the origin/main baseline (pre-fix files) was also spot-checked and shows the same sandboxed false-fail, confirming it is unrelated to this PR's change.

natb1:
## qa · attempt 0
failed: none
9/9 script-verifiable checks pass. 3 needs-human-judgment items resolved already-satisfied (production log confirmed unchanged during QA run; KNOWN_UNISOLATED allowlist premise verified; seam-only design sign-off). 1 needs-human-judgment item (post-merge fixture-row-growth check) resolved needs-main and filed as a `## needs-main residue` section in the node body (landed via graph-commit to origin/main, commit b46a7ce8) for the post-merge review->main-qa drain. No opus-fixable items, no bugs found, no auto-fix ran. Clean pass.


## review · attempt 1
failed: none
3 Required findings fixed and adversarially verified (lib-test-decision-log-guard.sh's FILE-seam-not-cleared leak, its OR-vs-precedence check logic, and the static ratchet's unanchored substring match), landed in commit c5db3ab9. 4 Required findings refuted by skeptics (not currently exploitable in this tree) and 3 classified false-positive/informational — all recorded on the PR comment. The code-review Lane-A instrument failed to invoke (`Skill code-review cannot be used with Skill tool due to disable-model-invocation`, confirmed by both its own receipt and an independent transcript-verification agent) — its payload was discarded, `coverage_incomplete=true`, and the unverified-instrument rule fires `deviation=true` unconditionally. security-review (Lane A) ran and was verified clean. Parking to office-hours for the missing code-review pass; the 3 fixed findings and their adversarial verification are not in question.

natb1:
## Review-fix results

Surface: `code`. Finders: input-validation, secrets, red-team, security-review (code-review Lane A instrument failed to invoke — see Partial coverage below).

- **`lib-test-decision-log-guard.sh:60`** (input-validation, red-team) — the guard's seam check treated `DISPATCH_DECISION_LOG_FILE`, `DISPATCH_DECISION_LOG_DIR`, and the resolved `DECISION_LOG_FILE` as three independently-sufficient OR clauses, but `lib-decision-log.sh:68` gives `DISPATCH_DECISION_LOG_FILE` strict precedence over `DISPATCH_DECISION_LOG_DIR`. An ambient/inherited `DISPATCH_DECISION_LOG_FILE` pointed at the production path would win at resolution time while the guard's DIR clause (always true, since the guard itself unconditionally redirects DIR) reported PASS — reproduced empirically by two independent skeptics. Fixed in c5db3ab9: the check now computes one `effective` value using the same FILE > DIR > default precedence, and fails closed when it or the direct `resolved`/`file_override` values equal production.
- **`lib-test-decision-log-guard.sh:42`** (secrets) — the guard's unconditional redirect only exported `DISPATCH_DECISION_LOG_DIR`, never clearing the higher-precedence `DISPATCH_DECISION_LOG_FILE` seam, so an ambient `DISPATCH_DECISION_LOG_FILE=<production path>` (e.g. left exported in a developer's interactive shell) would resolve to production even after sourcing the guard — reproduced empirically. Fixed in c5db3ab9: the guard now `unset`s `DISPATCH_DECISION_LOG_FILE` and repoints any already-resolved `DECISION_LOG_FILE` before exporting the sandboxed DIR.
- **`test-decision-log-isolation.sh:113`** (input-validation, red-team, secrets) — Part B's static ratchet matched isolation with four unanchored `grep -qF` substring searches over the whole file, so a comment or string literal merely naming the harness/env-var (not an actual `source`/assignment statement) would satisfy it. Confirmed live in this repo: `test-dispatch-daemon-liveness.sh` and `test-dispatch-heal-units.sh` passed the check on header-comment prose alone, with no real isolation. Fixed in c5db3ab9: anchored regexes matching real `source`/assignment statements, both suites added to `KNOWN_UNISOLATED`, and a new Part C dynamically proves every `KNOWN_UNISOLATED` entry leak-free under a redirected `HOME` on every run (not just asserted at review time).

Each Required finding below was independently re-examined by skeptics instructed to build the strongest false-positive case; none survived scrutiny as a currently-exploitable issue in this diff (all are reasonable hardening ideas for a follow-up, not blocking):

- **`lib-test-decision-log-guard.sh:60`** (input-validation) — exact-string denylist accepts a non-canonical spelling (e.g. trailing slash) of the production path as a PASS. Real as a code-reading matter, but no call site in the current tree ever sets a non-`mktemp`-derived value, and the actual isolation mechanism is the unconditional DIR redirect two lines above, not this check. (Split skeptic vote: 1 refuted, 1 upheld — ties default to refuted per the gate's fail-safe.)
- **`lib-test-decision-log-guard.sh:62`** (red-team) — the check's third clause reads the plain (non-exported) `DECISION_LOG_FILE`, which a re-exec'd child process would never see. The two suites named as the concrete attack path (`test-lib-standdown-recheck.sh`, `test-lib-frozen-session-park.sh`) were traced end-to-end: their decision-log writes happen entirely in-process via a `(...)` fork (which inherits plain vars) before their teardown unsets `DIR`, so the described leak window is never actually reached by any suite in this tree.
- **`lib-test-decision-log-guard.sh:50`** (red-team) — the guard is a one-shot, end-of-run sample rather than continuous. No suite in the current tree exhibits the hypothesized unset-then-restore-then-run pattern; the two suites that fork the real leak-capable children (`test-dispatch-tick.sh`, `test-dispatch-select-tick.sh`) re-export `DIR` in every per-test setup, so the end-of-run check is a backstop, not the load-bearing mechanism.
- **`lib-test-decision-log-guard.sh:41`** (red-team) — `mktemp -d` is unchecked. `mktemp` cannot practically fail in this repo's documented sandboxed-Bash or CI environments (both guarantee a writable `TMPDIR`/`/tmp`), and where it hypothetically could, `set -e` is already active at essentially every call site and would abort loudly rather than silently leak.

### Informational

- **`test-decision-log-isolation.sh:56`** (secrets) — on a detected leak, `check_suite_isolated` dumps the full contents of the routing-decision log/lock file to stdout for debugging. Test-only diagnostic output over developer-local fixture data; no action needed.

### False-positive (security)

- **`dispatch-test-fixture.sh:1413`** (input-validation, red-team) — the `EXIT` trap's `rm -rf "$DISPATCH_TEST_DECISION_LOG_DIR"` (mirrored in `test-helpers.sh:97`) was flagged as running on a possibly-unset variable; the variable is unconditionally assigned earlier in the same load-guard block, so this is not reachable.

### Partial coverage

The `code-review` instrument (Lane A, `/code-review max --fix`) could not be invoked — its own receipt and an independent transcript-verification agent both confirm it: `Skill code-review cannot be used with Skill tool due to disable-model-invocation`. Its payload was discarded entirely (not merged, not dispositioned, not credited). Per the review-fix skill's unverified-instrument rule this is treated as a false review, not a degraded one, and escalates unconditionally — this run is being parked to office-hours rather than marked complete. `security-review` (Lane A) ran and was verified; it reported no findings.

---
_Fixes landed in c5db3ab9. All findings above were adversarially re-examined; nothing was dismissed solely because the change is small._

PR #2835, opened 2026-07-10
analyticsutil: beacon-transport web-vitals + remove dead logEvent wrappers

natb1:
## Context

web-vitals (CLS/INP/LCP) report at page-hide, but firebase `logEvent`
internally awaits a dynamic-config fetch, so a tab hidden before that fetch
resolves loses the metrics and the first `page_view` — systematically censoring
the short/bounce sessions vitals exist to measure. Serves
`strategy-promote-progressive-detachment` via the GA4 export channel.

Graph node: `intentions/tactic-analytics-vitals-delivery.md`.

## Changes

**Unit 1 — beacon/flush delivery on page-hide.** Add GA4
`transport_type: "beacon"` to the `web_vitals` and first `page_view` `logEvent`
calls so gtag delivers via `navigator.sendBeacon`, which survives page-hide/
unload. Scope (qa rescope, author-ratified 2026-07-10): this covers the
post-init segment — sessions whose firebase init (dynamic-config fetch) has
resolved by page-hide, whose gtag request would otherwise be cancelled at
teardown. Tabs torn down before init resolves are still lost (`logEvent`
suspends awaiting init before gtag ever reads `transport_type`); that
residual is recorded as draft `tactic-analytics-preinit-vitals` in the
intentions graph.

**Unit 2 — remove unreachable wrappers; couple the traffic-type key.**
- Remove the `try/catch`+`logError` wrappers around the `web_vitals` and
  `page_view` `logEvent` calls. `logEvent` never throws for analytics failures
  (the SDK swallows them internally), so those catch branches could never fire —
  dead code per the clear-errors rule. The `applyTrafficTag` and
  `initAnalyticsSafe` wrappers (which guard genuinely-throwing calls) are kept.
- Share the traffic-type localStorage key as one constant
  (`@commons-systems/analyticsutil/traffic-type`), imported by both
  analyticsutil and the Playwright harness, replacing the comment-only coupling
  that let a rename on either side silently make CI smoke traffic count as
  organic.

## Test-integrity Signal 2 — escalation (path 2)

Removing the unreachable `logEvent` wrappers deliberately orphans the four tests
that mocked `logEvent` to throw (an impossible path). Deleting them is dead-code
cleanup, not test weakening — but the mechanical test-integrity gate counts 4
net declaration removals and fires Signal 2 (a blocking check). The gate cannot
distinguish legitimate cleanup from weakening; per `check-test-integrity.sh`
guidance path 2, this is escalated to office-hours for a **human
[...]

PR #3016, opened 2026-08-03
Collapse the blog prerender to the PageShell single-root injection path

natb1:
## Context

`packages/blog/src/prerender.ts` carried two hydration-injection paths (see
its prior header doc, `prerender.ts:41-44` and `:81-84`):

- The legacy path: regex/string `String.replace` injectors at marked
  template sites — `injectMain`, `injectInfoPanel`, `injectNav`,
  `injectFooter`, `injectHomeExtra`, `stripHomeExtra`. This is the class that
  produced the `tactic-blog-prerender-injection` defect (regex injection into
  HTML).
- The PageShell single-root path: `injectRoot` replaces one empty
  `<div id="${mount}"></div>` placeholder with the SSR-rendered `BlogPageShell`,
  which already carries nav, panel, hero, and footer.

`landing`'s blog already used the PageShell path; `fellspiral` remained on the
legacy path. This PR migrates fellspiral onto PageShell and deletes the legacy
injectors, removing the regex-injection bug class entirely.

## Unit 1 — Migrate fellspiral's blog prerender onto BlogPageShell

- `fellspiral/index.html` — collapsed `<body>` to the single `<div id="root">`
  mount + script tag, matching landing's template shape.
- `fellspiral/src/main.ts` — added the `shell` config to `createBlogApp`
  (`mount: "root"`, `wordmark: "fellspiral"`, `panelAriaLabel: "Info"`; no
  tagline/hero — fellspiral has neither), added the ds token/context-panel CSS
  imports, dropped the now-dead `@commons-systems/components/footer` import.
- `fellspiral/scripts/prerender.ts` — added the byte-identical `shell` config
  to `prerenderPosts`, dropped `footerHtml` (the shell's ds `<Footer/>` renders
  the same markup `@commons-systems/components/footer` re-exports).
- `fellspiral/package.json` — added `@commons-systems/ds`, `react`,
  `react-dom` (+ devDeps), matching landing's versions.
- New `fellspiral/test/prerender.test.ts` asserting the single-mount template
  and no legacy per-region markers survive, and that the client/server `shell`
  configs stay textually identical (the hydration byte-match contract).

## Unit 2 — Delete the legacy injectors and their template markers

- `packages/blog/src/prerender.ts` — deleted `injectMain`, `injectInfoPanel`,
  `injectNav`, `injectFooter`, `injectHomeExtra`, `stripHomeExtra`, their
  `if (shell) {...} else {...}` legacy branches, and the config fields that
  only fed them (`homeExtraHtml`, `footerHtml`, `panelHtml`, `stripHero`).
  `shell` is now required (not optional) on `PrerenderConfig` and
  `StaticPageConfig`. `injectRoot` and the `<head>` SEO injectors are
  untouched.
- `packages/blog/src/create-blog-app.ts` — one stale doc-comment fix (no code
  change): fellspiral is no longer the shell-absent default.
- `packages/blog/test/prerender.test.ts` and `prerender-static.test.ts` —
  updated to the single-path defaults; tests exercising legacy-only behavior
  (`homeExtraHtml`, `footerHtml`, `panelHtml`, `stripHero`) were deleted
  outright since that behavior no longer exists (per this repo's
  test-integrity rule); tests still asserting real behavior were rewritten
  against shell markup instead of legacy markup.

## Verification

```verify
npx vitest run --project packages/blog --root .
```

```verify
npx vitest run --project fellspiral --root .
```

Both pass (packages/blog: 29 files / 448 tests; fellspiral: 8 files passed / 2
skipped, 55 tests passed / 5 skipped). Note: the node plan's verify block used
`--project blog`, which errors ("No projects matched") — the vitest workspace
project id for that package is `packages/blog` (the workspace dir), not the
bare package name; a plan-text typo, not a code defect.

`npx tsc -p packages/blog/tsconfig.json --noEmit` and `npm run lint` (both
touched packages) are clean. `npx knip` shows no newly-dead exports.

Manual (preview deploy, not run in this session): load a fellspiral blog post
and the blog index, confirm no hydration mismatch warning in the console and
that nav, info panel, and footer all render correctly from the prerendered
HTML before JS runs (disable JS to check the prerendered shell).

Graph node: `tactic-prerender-single-injection-path`

### DIFF
--- a/fellspiral/index.html
+++ b/fellspiral/index.html
@@ -11,17 +11,7 @@
     <link rel="webmention" href="https://fellspiral.commons.systems/api/webmention">
   </head>
   <body>
-    <div class="page">
-      <header>
-        <h1>fellspiral</h1>
-        <nav><app-nav id="nav"></app-nav><button id="panel-toggle" class="panel-toggle" aria-label="Toggle info panel" aria-expanded="false">&#9656;</button></nav>
-      </header>
-      <div class="content-grid">
-        <main id="app"></main>
-        <aside id="info-panel" class="sidebar"></aside>
-      </div>
-      <footer><app-footer></app-footer></footer>
-    </div>
+    <div id="root"></div>
     <script type="module" src="/src/main.ts"></script>
   </body>
 </html>
--- a/fellspiral/package.json
+++ b/fellspiral/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@commons-systems/analyticsutil": "*",
     "@commons-systems/components": "*",
+    "@commons-systems/ds": "file:../packages/ds",
     "@commons-systems/errorutil": "*",
     "@commons-systems/authutil": "*",
     "@commons-systems/blog": "*",
@@ -26,9 +27,13 @@
     "dompurify": "^3.3.1",
     "firebase": "^12.9.0",
     "marked": "^15",
-    "missing.css": "^1.1.3"
+    "missing.css": "^1.1.3",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1"
   },
   "devDependencies": {
+    "@types/react": "^18.3.0",
+    "@types/react-dom": "^18.3.0",
     "vite": "^6.4.3"
   }
 }
--- a/fellspiral/scripts/prerender.ts
+++ b/fellspiral/scripts/prerender.ts
@@ -3,7 +3,6 @@ import { prerenderPosts } from "@commons-systems/blog/prerender";
 import { generateFeedXml } from "@commons-systems/blog/feed";
 import { generateSitemapXml } from "@commons-systems/blog/sitemap";
 import { FEED_REGISTRY } from "@commons-systems/blog/blog-roll/feed-registry";
-import { FOOTER_HTML } from "@commons-systems/components/footer";
 import appSeed from "../seeds/firestore.js";
 import {
   NAV_LINKS,
@@ -44,7 +43,15 @@ try {
     organization: ORGANIZATION,
     author: AUTHOR,
     relMe: REL_ME,
-    footerHtml: FOOTER_HTML,
+    // Single PageShell root. Must stay byte-identical to the `shell` config in
+    // fellspiral/src/main.ts, or the client's hydrateRoot mismatches. The shell
+    // renders the ds <Footer/> itself, so no footer markup is injected here (the
+    // ds Footer is the same markup @commons-systems/components/footer re-exports).
+    shell: {
+      mount: "root",
+      wordmark: "fellspiral",
+      panelAriaLabel: "Info",
+    },
     showHomeLink: true,
   });
 } catch (err) {
--- a/fellspiral/src/main.ts
+++ b/fellspiral/src/main.ts
@@ -1,7 +1,10 @@
 import "missing.css";
+import "@commons-systems/ds/tokens/colors.css";
+import "@commons-systems/ds/tokens/typography.css";
+import "@commons-systems/ds/tokens/spacing.css";
+import "@commons-systems/ds/tokens/effects.css";
+import "@commons-systems/ds/context-panel.css";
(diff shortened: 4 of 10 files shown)


natb1:
# Fix-checks summary — PR #3016

## Iteration 1

**Failed checks:** `test-integrity`, `type-safety-sensor`

**Outcome:** needs-human

**Reproduced:** yes (both, reproduced locally via `.github/scripts/check-test-integrity.sh` and `.github/scripts/check-type-safety-escapes.sh`)

**Reproduce command:**
```
bash .github/scripts/check-test-integrity.sh
bash .github/scripts/check-type-safety-escapes.sh
```

**Failure excerpt:**
```
ERROR: Test-integrity violation
Violations detected:
  - Signal 2: 12 net test declaration(s) removed

::error file=packages/blog/src/prerender.ts,line=65::Net-new type-safety escape hatch (as <Type> cast). Suppress with `// type-safety-ok: <reason>` if intentional.
```

**Why not caught:**

- `test-integrity` (Signal 2): Unit 2 of this tactic deletes the legacy
  per-region injectors (`injectMain`, `injectInfoPanel`, `injectNav`,
  `injectFooter`, `injectHomeExtra`, `stripHomeExtra`) from
  `packages/blog/src/prerender.ts`, and the PR correspondingly deletes the
  behavioral tests in `packages/blog/test/prerender.test.ts` and
  `prerender-static.test.ts` that exercised that now-deleted behavior (net
  −12 test declarations). This is legitimate dead-code cleanup, not test
  weakening — but the check's import-based co-deletion exemption
  (`.github/scripts/check-test-integrity.sh:179-510`) can't detect it: the
  deleted functions were never directly `import`ed by these test files (they
  were internal helpers reached only through `PrerenderConfig` option fields
  like `homeExtraHtml`/`footerHtml`/`panelHtml`/`stripHero`), so the test
  files' own import statements are unchanged old-vs-new and the exemption's
  removed-import set is empty. This is exactly the residual class recorded in
  `intentions/tactic-test-integrity-waiver.md` (first case PR #2835,
  `tactic-analytics-vitals-delivery`) — "behavioral tests of deliberately
  deleted behavior on a surviving symbol." That waiver channel is still
  `status: raw` (no check-side lookup implemented — confirmed by reading
  `check-test-integrity.sh` in full, no `execution.waivers` handling exists),
  so there is no mechanical path to green this check from code alone; it
  needs an author decision (approve the removal, or direct that the tests be
  restored instead).
- `type-safety-sensor`: false positive. The new doc comment at
  `packages/blog/src/prerender.ts:65` reads "Emitted **as** SoftwareApplication
  JSON-LD..." — the sensor's `as <Type> cast` rule
  (`.github/scripts/check-type-safety-escapes.sh:143-151`) only strips
  trailing `//` line comments, not `/** */` block comments (a documented
  limitation, see the script's own "Limitation (D2)" comment), so "as
  SoftwareApplication" inside the doc comment matches the cast pattern. Not a
  real type-safety escape hatch — no code change needed here, just a reword
  or a `// type-safety-ok:` suppression. Left unfixed this pass since the
  whole PR is blocked on the test-integrity decision regardless; a
  subsequent fix-checks pass will pick this up cleanly as a
  `reproduced == true` fix once that's resolved.

**Required action:** Author decision needed on the `test-integrity` Signal 2
block: either (a) approve the fellspiral-migration test cleanup as
legitimate (this is the exact "residual class" `tactic-test-integrity-waiver`
exists for — since the live check-side waiver lookup isn't implemented yet,
the currently-supported resolution is an author override-merge once review
is otherwise clean), or (b) direct that behavioral tests for the deleted
legacy injectors be restored instead (which would mean reverting Unit 2's
dead-code deletion). Separately, low-priority: `type-safety-sensor`'s
`prerender.ts:65` flag is a doc-comment false positive fixable in a later
fix-checks pass (reword to avoid "as `SoftwareApplication`" or add a
`// type-safety-ok:` suppression).

PR #3043, opened 2026-08-04
fix: park-node rollback restores pre-refresh local bytes, not origin/main

natb1:
## Problem

park-node's failure-rollback trap restored the node file from FRESH_BLOB
(origin/main's content), not the local checkout's own pre-touch bytes. When
the invoking checkout's HEAD lags origin/main for that path -- the exact
condition the fresh-origin/main invariant exists to handle -- the restore
wrote newer-than-HEAD content into the tree and left it modified, while
logging a rollback as having happened.

A leaked dirty `intentions/` file then breaks dispatch-tick's own
`git merge --ff-only` sync step fleet-wide for as long as it persists, with
no worker ever spawning. This fired twice on 2026-08-04 alone and again that
morning, and previously caused a ~7-hour fleet stall (2026-08-01).

## Fix

Port `clear-park`'s already-battle-tested `TMPORIG`/`ORIG_EXISTED`/
`restore_node()` idiom into `park-node` verbatim:

- Capture the node file's bytes before the origin/main refresh overwrites
  them.
- Move the `MUTATED` gate and EXIT trap ahead of that refresh, so a
  failed/partial refresh redirect is now rolled back too (previously it had
  no rollback at all).
- Reword the graph-commit-failure message so it states intent
  ("rolling back the office_hours write") rather than asserting an outcome
  the trap has not produced yet.
- Add a loud CRITICAL stderr line if the restore itself ever fails, naming
  the node and the path.

## Test coverage

Added a fault-injection regression case (test-park-node.sh case 22) that
clones the writer under test BEFORE a second writer lands a concurrent
advance on origin/main for the same node, so the local HEAD blob
deliberately differs from origin/main's -- the one construction that can
tell a correct pre-refresh-local restore apart from the pre-fix FRESH_BLOB
restore. The existing case 4 is structurally blind to this defect: its
fresh-clone construction makes its HEAD blob and origin/main's blob
identical, so a buggy restore and a correct one are byte-for-byte the same
there.

**Reproduction proof:** reverting park-node to its pre-fix state makes case
22 fail with a genuinely dirty working tree (`git status --porcelain`
non-empty); the fixed version passes. Full output pasted below.

Also hardened six existing rollback/no-op assertions from `git diff` (which
cannot see an index-vs-worktree mismatch or a leaked untracked `.rollback`
sibling) to `git status --porcelain`, matching what
`assert_clean_outside_ids` and `git merge --ff-only` actually consult, and
added a directory-wide assertion on the three clones that land nothing at
all.

## Verification

`test-park-node.sh`: 22/22 passed (was 21, +1 new case).
`test-transition-node.sh`: 3/3 passed, unaffected (blast-radius guard).
`test-graph-commit.sh`: 63/63 passed, unaffected (blast-radius guard).
`run-lint.sh`: all checks (eslint x2, prose-rule lint) passed.

Reproduction proof against the pre-fix park-node:

```
FAIL: park-node stale-local-HEAD restore (rc=1)
...
porcelain:  M intentions/t-restore-stale.md
```

(Case 4 also failed against the pre-fix code, but only because its
assertion was updated in this PR to match the new rollback message text --
not evidence of the tree-dirtying defect, which case 22 demonstrates
directly.)

## Out of scope (follow-ups, not filed in this PR)

- Porting the same fix to five sibling scripts that hand-roll the same
  restore-from-origin/main pattern (`transition-node`,
  `demote-node-to-implement`, `dispatch-graph-census`,
  `dispatch-graph-main-red-sync`, and the bulk variant
  `reconcile-graph-merged`), and the eventual shared-helper extraction.
- `.claude/skills/dispatch-propagate/scripts/test-graph-write-rollback.sh`
  has the identical `git diff` blindness and is not wired into CI at all.

Both are named in `tactic-park-node-rollback-dirty-tree-blocks-tick-sync`'s
own Context section as deliberately out of scope for this narrow,
high-urgency fix.


natb1:
## QA summary (PR #3043)

Pure backend bash-script change (`park-node`, `test-park-node.sh`) — no browser component, no Firestore surface touched. Plan authored by a single bounded Opus triage subagent; 13 items, 11 script-verifiable + 2 needs-human-judgment. All items resolved this pass.

**Items executed:** 13 (11 shell-command lane, 2 sent to disposition triage). **PASS: 13. FAIL: 0. SKIP: 0.**

## 1. New restore trap is installed before the origin/main refresh (MUTATED=1 ordering)
- Command: `grep -n -e 'trap ' -e 'MUTATED=0' -e 'MUTATED=1' -e 'TMPORIG' -e 'git show origin/main:intentions' packages/intentionsutil/scripts/park-node`
- Verdict: **PASS** — TMPORIG capture + `restore_node()` + `MUTATED=0`/EXIT trap all precede the fetch/pin-check; `MUTATED=1` (line 283) sits immediately before the origin/main refresh redirect (line 284).

## 2. The old FRESH_BLOB-based inline trap is fully removed
- Command: `grep -n -e 'FRESH_BLOB' -e '^[[:space:]]*trap ' packages/intentionsutil/scripts/park-node`
- Verdict: **PASS** — exactly one EXIT trap in the file; `FRESH_BLOB` used only for the `--base`/`--expect` compare-and-swap and the stale-diagnosis pin check, never as a restore source.

## 3. restore_node() restores from local pre-touch bytes, never truncates, and deletes when the file did not previously exist
- Command: `awk '/^restore_node\(\)/,/^}/' packages/intentionsutil/scripts/park-node`
- Verdict: **PASS** — restores from `TMPORIG` via cp-to-`.rollback`-then-`mv` (atomic, never truncates); failure path `rm -f`s the `.rollback` sibling and emits a CRITICAL line; `ORIG_EXISTED=0` branch deletes.

## 4. Reworded graph-commit-failure message and the new CRITICAL stderr line are present verbatim
- Command: `grep -n -e 'rolling back the office_hours write' -e 'was rolled back' -e 'CRITICAL' packages/intentionsutil/scripts/park-node`
- Verdict: **PASS** — present-tense message present, past-tense wording gone (0 occurrences), CRITICAL line names both node id and path on stderr.

## 5. Full park-node suite passes at 22/22 with the new case actually executing
- Command: `packages/intentionsutil/scripts/test-park-node.sh`
- Verdict: **PASS** — `passed: 22  failed: 0`, case 22's own PASS line present in the transcript.

## 6. Case 22 is not silently degenerate (its fixture guard did not fire)
- Command: `packages/intentionsutil/scripts/test-park-node.sh 2>&1 | grep -c 'fixture degenerate'`
- Verdict: **PASS** — 0 occurrences; R's local HEAD blob genuinely differed from origin/main's after writer S's concurrent advance.

## 7. Reproduction proof: reverting park-node makes case 22 fail with a dirty tree
- Command: revert `park-node` to base commit `92465612…`, run the suite, restore, re-confirm 22/22
- Verdict: **PASS** — reverted script: `passed: 20  failed: 2` (case 4 fails on message-text mismatch as expected; case 22 fails with `git status --porcelain` showing `M intentions/t-restore-stale.md` — the exact fleet-stalling condition). Fixed version restored byte-identical (`git status --porcelain` on the file empty) and suite back to 22/22.

## 8. The six hardened assertions really use `git status --porcelain`, and no rollback/no-op assertion still uses the blind `git diff`
- Command: `grep -n -e 'diff -- intentions' -e "status --porcelain -- intentions" packages/intentionsutil/scripts/test-park-node.sh`
- Verdict: **PASS** — zero `git diff -- intentions` occurrences; all six sites (cases 4, 5, 13, 17, 18, 19) use `status --porcelain`; directory-wide assertions present on the three clones that land nothing (D, G, N_CLEAR).

## 9. Case 22's assertions match what the real caller consults
- Command: `awk '/[Cc]ase 22/,0' packages/intentionsutil/scripts/test-park-node.sh`
- Verdict: **PASS** — asserts non-zero rc, the reworded message, and a directory-wide `git status --porcelain -- intentions/` empty; the graph-commit stub fails deterministically and unconditionally.

## 10. The stale `--base` compare-and-swap refusal (exit 3) still fires before any local mutation
- Command: `grep -n -e 'stale-diagnosis' -e 'exit 3' -e 'MUTATED=1' -e 'git show origin/main:intentions' packages/intentionsutil/scripts/park-node`
- Verdict: **PASS** — `exit 3` (line 280) precedes `MUTATED=1` (line 283); the refusal remains a pure no-op with nothing to roll back.

## 11. Blast-radius guards: sibling suites unaffected
- Command: `test-transition-node.sh && test-graph-commit.sh && git status --porcelain`
- Verdict: **PASS** — 3/3 and 63/63, tree clean afterward (aside from a pre-existing unrelated stray artifact at `.claude/agents`, unrelated to this PR).

## 12. Caller's-eye read: what dispatch-tick observes when park-node fails partway
- Classification: needs-human-judgment → disposition: **already-satisfied**
- Rationale: every failure path (pre-mutation stale-pin, refresh-redirect failure, graph-commit failure, restore-itself failure) either leaves the tree clean with honest present-tense messaging, or screams CRITICAL naming the node and path. No gap found.

## 13. Documented follow-ups are genuinely out of scope, not silently-broken siblings
- Classification: needs-human-judgment → disposition: **already-satisfied**
- Rationale: verified directly against the five named sibling scripts (`transition-node`, `demote-node-to-implement`, `dispatch-graph-census`, `dispatch-graph-main-red-sync`, `reconcile-graph-merged`) plus `test-graph-write-rollback.sh` — all genuinely still carry the same origin/main-blob restore defect class, matching the PR's out-of-scope claim exactly. This is an explicit, already-documented scope decision in the tactic's own node body (driven by the `.claude/skills/` auto-mode commit denial and the narrow live-outage framing), not new QA-discovered residue. No evidence any sibling is hotter-traffic than park-node itself.

## Disposition triage

Both `needs-human-judgment` items were sent to the Step 3.5 disposition Workflow, which independently re-verified each against the worktree source and classified both **already-satisfied** — dropped as PASS. `result.dispositions` is empty: no `opus-fixable`, `needs-main`, or `needs-human` items survive. No auto-fix lane, no follow-up filing, no office-hours escalation triggered by this pass.

- **Caller's-eye read: what dispatch-tick observes when park-node fails partway** — independently confirmed against the worktree source: TMPORIG/ORIG_EXISTED pre-refresh capture, atomic `restore_node()` with CRITICAL stderr messaging, single MUTATED-gated EXIT trap, `MUTATED=1` set after the exit-3 stale-diagnosis refusal, present-tense failure messaging, `FRESH_BLOB` used only for the compare-and-swap. Every failure path either leaves a clean tree or screams CRITICAL — a positive confirmation with no gap and no code change to author.
- **Documented follow-ups are genuinely out of scope, not silently-broken siblings** — verified all five named siblings still carry the old origin/main-blob restore idiom, so the PR's out-of-scope list is accurate and complete. The deferral is an explicit, documented scope decision in the tactic's node body; no evidence any sibling sees hotter traffic than park-node.

## Result

Clean autonomous pass — no residue requiring a fix, a follow-up filing, or office-hours escalation. Phase advances `qa → review`.

natb1:
## qa · attempt 0
failed: none

QA pass: 13-item plan (11 script-verifiable, 2 needs-human-judgment), all resolved PASS.
Verified: MUTATED-gate/trap ordering, FRESH_BLOB-restore removal, restore_node() atomicity,
reworded messages, 22/22 test-park-node.sh with case 22 non-degenerate, revert-based
reproduction proof (dirty tree on pre-fix code), git-diff→git-status--porcelain hardening,
stale-base exit-3 ordering, and blast-radius guards (test-transition-node.sh 3/3,
test-graph-commit.sh 63/63). Both needs-human-judgment items (caller's-eye failure-path
read; sibling-scripts out-of-scope accuracy) independently re-verified and classified
already-satisfied by the disposition workflow — dropped as PASS. No residue.


## review · attempt 1
failed: none

Review pass: 4 findings surfaced (input-validation, domain-sweep, red-team x3,
security-review, code-review — all wired finders ran, none failed). 2 upheld
Required findings fixed on Opus in commit e3f05a33: park-node's rollback trap
now declines to restore intentions/<id>.md when HEAD has moved or the node
blob no longer matches EXPECT_BLOB (graph-commit already committed/reset it
on a busy-main path), and the ORIG_EXISTED=0 branch restores a
tracked-but-absent-at-capture file via `git checkout --` instead of an
unconditional `rm -f`. 2 findings refuted by adversarial verify (path
traversal — MUTATED never set before the traversal-shaped lookup fails;
missing lock on concurrent same-checkout invocations — not this codebase's
architecture, and graph-commit's own CAS governs what lands). Built-in
/code-review low --fix: 0 findings. CodeQL: no open alerts. Erosion: none.
deps=false, dependency audit skipped. test-park-node.sh: 22/22 pass after fix
(no regressions). No deferred/follow-up filings needed.

natb1:
## Review

Combined `/review-fix` pass (code-review pre-stage + security/domain finder fan-out, dedup, classify, adversarial-verify, Opus fix). 4 findings surfaced, 2 fixed, 2 refuted. Built-in `/code-review low --fix` pre-stage found nothing (0 findings). CodeQL: no open alerts. Erosion scan: no findings. `deps=false` — dependency audit skipped.

- **`packages/intentionsutil/scripts/park-node:245`** (red-team) — the new EXIT trap restored `intentions/<id>.md` from the pre-refresh copy on ANY non-zero exit once `MUTATED=1`, with no check that `graph-commit` had already committed the park locally (or reset HEAD onto a fresher origin/main) before returning non-zero on a busy-main/push-contention path. That restore would overwrite a file that already matches the just-created local HEAD commit, reintroducing a dirty tracked file — the exact failure class this PR exists to close. **Fixed** in `e3f05a33`: `restore_node` now captures `HEAD_BEFORE` and declines to restore (printing a diagnostic instead) when HEAD has moved or when the node file no longer hashes to `EXPECT_BLOB`.
- **`packages/intentionsutil/scripts/park-node:236`** (red-team) — same root defect, `ORIG_EXISTED=0` branch: an unconditional `rm -f` assumed a file absent at capture time could only have been created by this script's own write, but if `graph-commit` had already committed it (tracking it at local HEAD) before failing, the `rm -f` would delete a now-tracked file — again a dirty (locally-deleted) tracked file. **Fixed** in `e3f05a33`: the same HEAD/blob guard applies, and when the path is tracked-but-absent-at-capture, restore uses `git checkout --` instead of `rm -f`.

Verified: `test-park-node.sh` 22/22 pass after the fix (no regressions).

- **`packages/intentionsutil/scripts/park-node:214`** (input-validation) — claimed the caller-supplied node id was interpolated into filesystem paths without validation (path-traversal risk). Refuted: `MUTATED` is only set to 1 after the `origin/main` blob lookup succeeds, so a traversal-shaped `NODE_ID` fails that lookup and exits before any restore path is reachable; `NODE_ID` also comes from internal orchestration reading the graph's own trusted node set, not external input.
- **`packages/intentionsutil/scripts/park-node:214`** (red-team) — claimed the pre-refresh capture takes no lock, so two concurrent `park-node` invocations against the same node in the same checkout could race on the fixed `.rollback` sibling path. Refuted: this requires two invocations sharing one checkout, which isn't this codebase's normal architecture (each session/worker has its own worktree); even granting the race, `graph-commit`'s own per-node lock and `--base`/`--expect` compare-and-swap governs what actually lands on `main`, and any stray residue is already detected by `assert_clean_outside_ids`.

### Coverage

No partial coverage — all finders and instruments (security-review, input-validation, domain-sweep, red-team, code-review) ran and were verified.

PR #3046, opened 2026-08-05
reconcile-graph-merged: pin the diagnosis-time base blob so a concurrent park survives

natb1:
Implements graph node `tactic-reconcile-park-clobber` (bug X) — all three units from its plan.

## The defect

`reconcile-graph.ts` rewrites the WHOLE node from its in-memory read, and
`reconcile-graph-merged` built its `graph-commit` argv with **zero `--base`
flags** — so `check_base_freshness()` short-circuited on its first line and the
sweep silently overwrote any edit that landed on `origin/main` between its read
and its commit.

Measured: 8 `office_hours` park erasures across 5 nodes in one 24h window, each
with `phase` left untouched.

## Units

- **Unit 1** — pin each planned edit id's PRE-apply **on-disk** blob via
  `hash-object -w` and thread it as `--base <id>=<blob>`. `-w` is mandatory:
  `check_base_freshness` resolves the base with `cat-file -p` and `die`s when it
  misses. The disk blob, not `origin/main`'s — `graph-commit` defines base as
  "the blob the writer read", and pinning `origin/main` would compute a spurious
  `ours` delta on untouched fields.
- **Unit 2** — roll back to `HEAD` rather than to a captured blob. Unit 1 makes
  `park_and_exit()` reachable, and on that path `HEAD` has already moved;
  writing pre-park bytes back would leave the tree dirty with stale, park-erased
  content — the residue that bricks `assert_clean_outside_ids` for every other
  writer. Precedent: `resolve-hold`'s `clean_node_file()`.
- **Unit 3** — three cases in `test-graph-write-rollback.sh`: `--base`
  construction, the end-to-end race against the REAL `graph-commit` (only `gh`
  stubbed), and the park-path rollback.

## Verification

- `test-graph-write-rollback.sh` — 8/8 pass.
- Cases 6 and 7 both verified **RED** with Unit 1's threading reverted.
- `npx vitest run --project packages/intentionsutil --root .` — 820/820.
- `run-lint.sh` — prose rules pass (non-vacuously, on the added lines).
- `run-unit-tests.sh` — all suites pass.

## Note on the test fixture

Case 7 needed care to avoid being vacuous. With the checkout left BEHIND
`origin/main`, `graph-commit`'s push is rejected, it rebases, git reports a
textual conflict, and **layer-2's field-level auto-merge preserves the park on
its own** — passing with or without the fix. The fixture therefore fast-forwards
HEAD onto the park commit, which is the shape production erasures actually took:
no conflict to rescue, so `--base` is the only thing that can catch the lost
update.

## Scope

Ruling 31 is not reopened — the sweep stays ungated on `office_hours` for its
phase/merge decisions. This is the write side only. `reconcile-graph.ts`,
`store.ts`, and `graph-commit` are unchanged. The identical zero-`--base` shape
in `reconcile-graph-review-stall:276-290` is deliberately out of scope; it is
carried by `tactic-reconcile-review-stall-base-pin`.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/reconcile-graph-merged
+++ b/.claude/skills/dispatch-propagate/scripts/reconcile-graph-merged
@@ -70,14 +70,28 @@ NOW=${GRAPH_RECONCILE_NOW:-$(date -u +%s)}
 # Enumeration is STRICT (`listNodesStrict`): a corrupt node file dropped by the
 # tolerant reader would silently never be reconciled, stranding its merged PR at
 # a stale phase. Abort the pass loudly instead.
+#
+# The id shape is validated HERE, at the point unvalidated frontmatter strings
+# first enter the sweep, because the tab/newline-delimited `read` loops below
+# cannot see the difference between a multi-line id and two records. See
+# validate_node_id() for why an id carrying `=`, whitespace, or a control
+# character is unsafe downstream (it silently re-keys graph-commit's
+# `--base <id>=<sha>` pin). Refuse the sweep rather than reconcile the node with
+# a broken or mis-keyed compare-and-swap base.
 OPEN_TACTICS="$( (cd "$REPO_ROOT" && node --import tsx/esm -e '
   const { listNodesStrict } = await import("./packages/intentionsutil/src/store.js");
   const open = new Set(["implement","fix","qa","review","main-qa"]);
+  const idOk = /^[A-Za-z0-9][A-Za-z0-9._-]*$/;
   for (const n of listNodesStrict("./intentions")) {
     if (n.kind !== "tactic") continue;
     if (n.phase === null || !open.has(n.phase)) continue;
     const pr = n.execution && n.execution.pr;
     if (pr === null || pr === undefined) continue;
+    if (!idOk.test(n.id)) {
+      process.stderr.write("unsafe node id " + JSON.stringify(n.id) +
+        " (must match /^[A-Za-z0-9][A-Za-z0-9._-]*$/)\n");
+      process.exit(1);
+    }
     process.stdout.write(n.id + "\t" + pr + "\n");
   }
 ') )" || { echo "reconcile-graph-merged: node enumeration failed" >&2; exit 1; }
@@ -161,6 +175,34 @@ fi
 STATES_FILE="$(mktemp)" || { echo "reconcile-graph-merged: mktemp failed" >&2; exit 1; }
 declare -A BASE_BLOB=()
 RESTORE_ON_FAILURE=0
+# HEAD as of the moment the rollback is armed — the exact commit the apply run's
+# disk mutation sits on top of, and the only commit `git checkout --` can be
+# trusted to restore to. Captured in Step 3; read by restore_node_files().
+HEAD_AT_ARM=""
+
+# A node id must be safe both as a single path component and as the KEY of
+# graph-commit's `--base <id>=<blobsha>` pair. add_blob_pair() splits the pair at
+# the FIRST `=` (packages/intentionsutil/scripts/graph-commit:369-378), so an id
+# containing `=` is silently re-keyed: `tactic-a=b` records the pin under the
+# truncated id `tactic-a` with the garbage sha `b=<real-sha>`. If a genuine
+# `tactic-a` is in the same batch its correct pin is overwritten, and
+# check_base_freshness then three-way merges against the wrong (or unresolvable)
+# base — defeating the very lost-update protection the pin exists to provide.
+# Whitespace and control characters are equally unsafe: the pin loop reads the
+# planned edit ids line by line, so a multi-line id would hash the wrong path and
+# leave the true id unpinned.
+#
+# Node ids are unvalidated frontmatter strings, so gate them here. Hard-error
+# (.claude/rules/code-style.md: clear errors over fallbacks) rather than skipping
+# the pin — a skipped pin is an UNPROTECTED write, exactly the failure this
+# compare-and-swap closes.
+validate_node_id() {
+  [[ "$1" =~ ^[A-Za-z0-9][A-Za-z0-9._-]*$ ]] || {
+    echo "reconcile-graph-merged: unsafe node id $(printf '%q' "$1") — ids must match ^[A-Za-z0-9][A-Za-z0-9._-]*\$ (no '=', whitespace, or control characters) to be a valid --base pair key" >&2
+    exit 1
+  }
+}
+
 # Roll the apply run's disk mutations back to the checkout's CURRENT HEAD — not
 # to a captured blob. Restoring captured bytes is wrong on graph-commit's park
 # path: park_and_exit() does `git reset --hard FETCH_HEAD`, lands an office_hours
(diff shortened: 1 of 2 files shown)


natb1:
# QA summary — tactic-reconcile-park-clobber (PR #3046)

## Items executed

6 plan items (5 script-verifiable, 1 needs-human-judgment).

**PASS: 5 / FAIL: 0 / SKIP: 0** (shell-command lane; no browser component detected, so no browser lanes ran).

1. **PR's own claimed test harness passes fresh in this checkout** — PASS. `bash .claude/skills/dispatch-propagate/scripts/test-graph-write-rollback.sh` → `passed: 8  failed: 0`, exit 0. Matches the PR's own claimed 8/8.
2. **New race case asserts on `office_hours` survival, not just exit code** — PASS. Case 7's assertion block checks `phase7 == "done"` (reconciler's own write landed) AND `grep -q 'concurrent park landed between the sweep read and its commit'` AND merge-evidence strings — not exit-code-only.
3. **Case 7's fixture is non-vacuous (fast-forward shape, not conflict-rescue shape)** — PASS. Confirmed sentinel guard (`.concurrent-landed`), delegation through `graph-commit.real`, and `git reset -q --mixed FETCH_HEAD` (the fast-forward-clobber shape, not a rebase/conflict shape that would let layer-2 auto-merge rescue the park independent of `--base`).
4. **Declared out-of-scope boundary respected** — PASS. `git diff --name-only origin/main...HEAD` shows exactly the two declared files; no hit on `reconcile-graph-review-stall`, `reconcile-graph.ts`, `store.ts`, or `graph-commit`.
5. **Base pinning writes to object DB and hard-errors on a missing pin** — PASS. Confirmed `git hash-object -w` (not bare `hash-object`), per-id `--base "$id=..."` threading with an explicit missing-pin hard-error branch, and `restore_node_files()` using `git checkout --` with zero remaining `restore_snapshot` references.

## Deferred to office-hours

**Item 6 — Rollback-to-HEAD is the right shared-state contract for the park path.** Not walked (needs-human-judgment classification). See Disposition triage below for the full resolution chain — this escalates via a scope-deviation, not a plain needs-human park.

## Bugs found

None. All script-verifiable checks pass; no defect in the code or the new test cases.

## Disposition triage

`needs-main` items are filed as `blocked_by` follow-ups (Step 3.6) and dropped from the escalation set, so they do not park this PR — **not applicable this pass** (zero `needs-main` items).

- **Item 6** (`Rollback-to-HEAD is the right shared-state contract for the park path`) — disposition class: **opus-fixable** (downgraded from the triage's `needs-human-judgment` by the adversarial skeptic fan-out; verify: **Refuted**, 2/2 skeptics). Both skeptics independently concluded the "needs a human ruling" framing was unwarranted: the HEAD-restore contract is already fully reasoned in `reconcile-graph-merged`'s own comments, matches an existing three-times-precedented convention (`resolve-hold`'s `clean_node_file()`), and rests on a hard mechanical criterion (`assert_clean_outside_ids` breaks under the alternative) rather than a live subjective tradeoff.
  - However, the gated fix-plan phase then hit a **scope-deviation** (`deviation: true`, `units: []`): the node's own acceptance criteria explicitly mandate the as-implemented behavior (`git checkout --` to current HEAD) and do not authorize changing it, so there is nothing for an Opus fix to build — "fixing" would mean deviating from the node's own spec.
  - Net effect: no code defect exists and no code change is warranted, but per this skill's Step 3.7 contract a fix-plan `deviation: true` is a permanent escalation regardless of the class label. This escalates below as a **quick sign-off**, not a bug report — office-hours can most likely confirm-and-clear in one pass.

No `already-satisfied` items this pass (`result.already_satisfied` is empty — item 6 was not dropped as PASS; it went through the deviation escape instead).

**Non-fixing pass**: Step 3.7 took the escalate-finalize path (scope-deviation, no attempt label — permanent escalation, does not count against the `dispatch:qa-fix-attempt-<n>` cap). No units were planned or landed this pass.

natb1:
## qa · attempt 0
qa pass 1: 5/5 script-verifiable items PASS (test harness 8/8, race-case assertions non-vacuous, out-of-scope boundary respected, base-pinning hard-error present). No code defect found. 1 needs-human-judgment item (rollback-to-HEAD design contract) escalated via scope-deviation: adversarial verify refuted the "needs human" framing (already precedented by resolve-hold's clean_node_file()), but the gated fix-planner declined to touch code that already matches the node's own acceptance criteria verbatim. Escalating for a quick office-hours sign-off, not a bug fix.


## review · attempt 1
review: 7 findings surfaced (input-validation×3, red-team×3, code-review residue×1), 2 upheld+fixed — unsafe node-id injection into graph-commit's `--base` pairing (`=`-delimiter mis-split silently skips the freshness check), and a false "write rolled back" claim on the busy-main/unpushed-commit failure path. 3 refuted by adversarial-verify (path-traversal and tracked-file preconditions gated elsewhere; code-review residue mislocated and non-reproducible), 2 out-of-scope (documented/intended `graph-commit` behavior, no follow-up filed). Fixed in commit 4223819f; harness extended to 9/9 (case 9 verified red pre-fix). No deviation, no coverage gaps. failed: none

natb1:
## Review-fix pass

Combined code-review + security-review fan-out over the pending diff
(`.claude/skills/dispatch-propagate/scripts/reconcile-graph-merged`,
`.claude/skills/dispatch-propagate/scripts/test-graph-write-rollback.sh`).
Built-in `/code-review low --fix` pre-stage ran clean (0 files touched). No
CodeQL alerts, no erosion findings, no dependency changes.

- **input-validation-1** (`reconcile-graph-merged:249`) — `--base "$id=$sha"`
  is split on the *first* `=` by `graph-commit`'s `add_blob_pair`, and node ids
  are not restricted from containing `=`. An id like `tactic-a=b` mis-splits,
  never gets a correctly-keyed `BASE` entry, and silently lands **unprotected**
  by the freshness check this PR exists to add. Upheld by adversarial-verify
  (code-traced, no privilege bypass required). Fixed in commit `4223819f`.
- **red-team-1** (`reconcile-graph-merged:178`) — `restore_node_files()`
  unconditionally reports "write rolled back," but on a busy-main failure
  after `graph-commit` has already committed locally (and `RESTORE_HEAD` is a
  documented no-op for the main checkout), `git checkout --` is a no-op against
  the same moved `HEAD` — the mutation stays as a stranded unpushed commit
  while the false "rolled back" message is printed. Upheld by adversarial-verify
  (traced the exact `graph-commit` call sequence). Fixed in commit `4223819f`.

Fix: gated node ids to a safe shape before pinning/building `--base`, and made
the rollback HEAD-aware so an un-landed `graph-commit` commit is discarded (or
refused loudly) instead of being falsely reported as rolled back. Harness case
8 made faithful to the busy-main scenario; new case 9 added (9/9 pass; case 9
verified red without the fix).

- **input-validation-0** (`reconcile-graph-merged:212`) — path-traversal via an
  unsafe node id reaching `hash-object -w`. Refuted: `writeNode`'s
  `assertPathSafeId` is the sole creation-time gate and already rejects `/`/`\`
  outright, so such an id can never reach disk through any legitimate path;
  reaching this would require hand-editing a tracked file at committer trust
  level, which discloses nothing a less-privileged actor couldn't already read.
  Defense-in-depth gap at best, not exploitable.
- **red-team-2** (`reconcile-graph-merged:212`) — `hash-object -w` (vs. the old
  `git rev-parse origin/main:path`) drops a "file must be tracked" precondition.
  Refuted: real as a code-level comparison, but the exploit requires an
  implausible compounded race (a genuinely untracked node simultaneously
  satisfying `kind=tactic`, an open `phase`, and a non-null `execution.pr` —
  fields normally only set by a prior committed `writeNode`+`graph-commit`
  cycle), and `check_base_freshness`'s own merge/park logic is an independent
  downstream safety net regardless.
- **code-review residue** (`reconcile-graph-merged:181`, mislocated — actual
  call site is `:273`) — claimed `git checkout --` restores from the index,
  not `HEAD`. Refuted: nothing ever stages an `intentions/*.md` path outside
  `graph-commit`'s own commit, so index and `HEAD` never diverge here.

### Out-of-scope (security)

- **input-validation-2** (`reconcile-graph-merged:249`) — notes that
  `graph-commit`'s `parse_blob_arg` also accepts a manifest-file-path form for
  `--base`, distinct from the repeated-flag form this PR uses. Not a defect in
  this diff; no follow-up filed (informational note about an unused code path
  in a dependency this PR doesn't touch).
- **red-team-3** (`reconcile-graph-merged:249`) — notes that threading `--base`
  newly activates `graph-commit`'s layer-3 `check_base_freshness` /
  `park_and_exit` path. This is the intended, documented behavior this node's
  plan calls for (a park absorbs a concurrent write via three-way merge rather
  than clobbering it) — not a defect; no follow-up filed.

### Coverage

Full coverage this run — no throttled probe wave, no unverified instrument, no
undispositioned Lane-A residue.

PR #3051, opened 2026-08-06
tactic-wait-calendar-release: WAIT node calendar release predicate (attributes.wait_until, tick sweep, re-arm in place)

natb1:
## Summary

Implements the calendar release predicate for the strategy's WAIT hold node
(node `tactic-wait-calendar-release`, 8-unit clean-session plan). A WAIT node
is a `kind: tactic` node, phase-null and `office_hours: null` while armed,
that holds a source node via the source's `blocked_by` for a bounded calendar
interval, releases itself to `phase: done` when `attributes.wait_until`
passes, re-arms in place on a repeat not-yet-observed verdict (attempt
counter surviving, source edge untouched), and escalates to the author via
the existing `park-node` primitive at a finite attempt cap
(`WAIT_ATTEMPT_CAP = 4`).

- `packages/intentionsutil/src/waits.ts` — the vocabulary module (id
  derivation, `wait_until` parsing, the `isWaitNode` canonical-id security
  check).
- `packages/intentionsutil/src/wait-sweep.ts` — the pure classifying
  enumerator (`due` / `waiting` / `capped` / `malformed`).
- `packages/intentionsutil/src/schema.ts` — Rule 21 (WAIT-node shape) and the
  census owed-prune exclusion for a released WAIT whose source is still open
  — closes the recorded residual risk where an owed-prune would permanently
  break the re-arm-in-place design.
- `packages/intentionsutil/src/router.ts` — excludes an armed WAIT from the
  draft-candidate loop (it is phase-null/office_hours-null by construction
  and would otherwise be misread as an undecomposed draft every tick).
- `packages/intentionsutil/scripts/wait-node-decide.ts` +
  `list-recheckable-waits.ts` — the pure decision half (NONE/REARM/EXTEND)
  and a thin TSV enumerator CLI.
- `packages/intentionsutil/scripts/arm-wait` + `release-wait` — the bash
  landers: fresh-`origin/main` refresh, CAS-guarded single `graph-commit`,
  rollback-on-failure, and (`release-wait`) a post-land re-read that
  re-verifies the write survived `graph-commit`'s layer-2 union before
  reporting success.
- `.claude/skills/dispatch-propagate/scripts/lib-wait-recheck.sh` — the
  per-tick sweep predicate (`wait_recheck_sweep`), modeled on
  `lib-stale-hold-recheck.sh` but kept in its own file: a hold is born
  parked and resolves by clearing two independent facts; a WAIT is born
  unparked and releases by one — merging them would couple two independent
  fail-safe postures.
- `dispatch-tick` — wired at both the paused branch and the normal path, so
  a WAIT releases on schedule whether or not dispatch is paused.

Also fixes a latent bug surfaced during implementation:
`wait-node-decide.ts`'s `--now` CLI default emitted millisecond-precision
ISO output, which always failed its own `WAIT_UNTIL_RE` validation — any
caller omitting `--now` would have `decideWait` throw unconditionally.

**Deliberately out of scope** (recorded in the node body): flipping
`qa-main`'s WAIT branch from parking to calling `arm-wait` — that's a
`.claude/skills/**` edit an autonomous worker cannot commit (auto-mode
classifier denial), so it's left as a follow-up for an interactive session
now that the mechanism and a tested manually-invocable producer exist.

## Test plan

- [x] `npx vitest run --project packages/intentionsutil --root .` — 951/951
- [x] `.claude/skills/dispatch-propagate/scripts/run-typecheck.sh` — clean
- [x] `npx tsx packages/intentionsutil/scripts/validate-graph.ts intentions` — ok
- [x] `packages/intentionsutil/scripts/test-arm-wait.sh` — 10/10
- [x] `packages/intentionsutil/scripts/test-release-wait.sh` — 7/7
- [x] `.claude/skills/dispatch-propagate/scripts/test-lib-wait-recheck.sh` — 78/78
- [x] `.claude/skills/dispatch-propagate/scripts/run-lint.sh` — clean
- [ ] Manual: a full arm → release → re-arm → cap cycle by hand against a
      real remote, and the router-exclusion check against a live
      `select-targets.ts` run — left for QA per the plan's own
      "Manual / judgment checks" section (node body).

Implements `tactic-wait-calendar-release` — all 8 units of the plan in the
node body on `origin/main`. No GitHub issue is associated with this graph
node; the implement→qa transition is driven by `transition-node --set-pr`,
not by a `Closes #N` line.


natb1:
# Fix-checks summary — PR #3051

## Iteration 1

**Outcome:** fixed

**Reproduced:** yes

**Reproduce command:** `.github/scripts/check-type-safety-escapes.sh` (after `git fetch origin main`)

**Failure excerpt:**
```
::error file=packages/intentionsutil/scripts/graph-census-debt.ts,line=144::Net-new type-safety escape hatch (as <Type> cast).
::error file=packages/intentionsutil/scripts/wait-node-decide.ts,line=235::Net-new type-safety escape hatch (non-null assertion !).
::error file=packages/intentionsutil/src/schema.ts,line=1171::Net-new type-safety escape hatch (as <Type> cast).
::error file=packages/intentionsutil/src/wait-sweep.ts,line=96::Net-new type-safety escape hatch (as <Type> cast).
::error file=packages/intentionsutil/test/schema.test.ts,line=1941::Net-new type-safety escape hatch (non-null assertion !).
::error file=packages/intentionsutil/test/schema.test.ts,line=1943::Net-new type-safety escape hatch (non-null assertion !).
::error file=packages/intentionsutil/test/waits.test.ts,line=47::Net-new type-safety escape hatch (non-null assertion !).
```

**Why not caught:** The diff-scoped `type-safety-sensor` (`.github/scripts/check-type-safety-escapes.sh`) flags any net-new `as <Type>` cast or non-null `!` assertion on lines the PR adds relative to `origin/main`, with no exception for casts that are provably safe by an adjacent runtime guard. Two sites (`graph-census-debt.ts:144`, `wait-sweep.ts:96`) cast `attributes.wait_for as string` immediately after an `isWaitNode(...)` guard that already checks the value is a string — safe, but invisible to the sensor and to the TS compiler (the guard is a plain `boolean`, not a type predicate). One site (`wait-node-decide.ts:235`) used `existing!` where the enclosing `disposition === "REARM"` branch already implies `existing !== undefined`, but through an intermediate variable TS can't trace. One site (`schema.ts:1171`) cast a caught error as `(err as Error).message` instead of following this package's established `err instanceof Error ? err.message : String(err)` convention (used ~12 other places in the same package). The remaining three sites (`schema.test.ts:1941`, `schema.test.ts:1943`, `waits.test.ts:47`) are pure false positives: the sensor's line-based scanner doesn't strip string/regex literal contents (a documented limitation, "Limitation (D2)", in the sensor script itself), so the `!` inside deliberately-invalid test fixture ids like `"Tactic_Source!"` and `"tactic-Bad_Id!"` reads as a non-null assertion.

**Fix:** Applied the repo's standard `// type-safety-ok: <reason>` suppression marker to the two safe `wait_for` casts and the three string/regex-literal false positives; restructured the `wait-node-decide.ts` REARM branch to `existing !== undefined && disposition === "REARM"` (removing the `!` assertion via narrowing instead of asserting past it); replaced the `(err as Error).message` cast in `schema.ts` with the established `err instanceof Error ? err.message : String(err)` pattern. No behavior change. Commit `422eda81` (merged with `origin/main` in `72b0e6c3`), pushed to `origin/tactic-wait-calendar-release`.

natb1:
# QA summary — tactic-wait-calendar-release (PR #3051)

Backend/tooling-only change (no browser component, no Firestore rules touched). Triage plan authored by a bounded Opus subagent; all 10 items executed via the shell-command lane (no browser walkthrough was needed).

**Counts:** 9 PASS, 0 FAIL, 0 SKIP, 1 residue (needs-human-judgment, planned-deferral → classified `needs-main`).

## Items executed

1. **Re-run the intentionsutil vitest suite fresh** — PASS (967/967, all 48 files; sandboxed run false-failed with a tsx IPC-pipe EPERM artifact, confirmed as a known sandbox artifact and re-verified sandbox-off).
2. **Re-run the three new bash harnesses (arm-wait, release-wait, lib-wait-recheck)** — PASS (10/10, 7/7, 78/78).
3. **Re-run typecheck, lint, and validate-graph against the real intentions store** — PASS (typecheck clean, lint clean including the prose-rule linter, validate-graph ok — 558 real nodes, non-vacuous).
4. **Static wiring hygiene** — PASS. `arm-wait`/`release-wait` are mode-100755 in the index; exactly one `graph-commit` invocation call site in each; write path is exclusively `write-node.ts`/`dump-node.ts`/`graph-commit`; `unit-tests.yml` wires `test-arm-wait.sh` and `test-release-wait.sh` exactly once each (both live under `packages/intentionsutil/scripts/`, which `run-unit-tests.sh`'s glob does **not** cover); `test-lib-wait-recheck.sh` lives under `dispatch-propagate/scripts/`, which the glob **does** cover, so it correctly has no separate CI step — no missing coverage, no redundant double-wiring.
5. **Fail-safe posture** — PASS. `wait_recheck_sweep` returns 0 with exactly one summary line and correctly-distinguished `status=` values (`repo-unresolvable`, `ok`, `enumeration-failed`) across three degraded-mode fixtures. Both `dispatch-tick` call sites are `declare -f`-guarded, never `|| true`'d, and correctly ordered (after `stale_hold_recheck_sweep`, before Step-1 selection). The `--now` CLI-default latent-bug fix produces a self-validating second-precision instant. `wait-sweep.ts` never defaults a missing/malformed `wait_until` to "now" (confirmed by code read).
6. **Router exclusion live-verified** — PASS. Against a scratch fixture (armed WAIT + blocked source + an unrelated ordinary draft control), `select-targets.ts` emits neither the armed WAIT nor its blocked source as a candidate; after flipping the WAIT to `phase: done`, only the source becomes a candidate; the ordinary draft is unaffected throughout.
7. **Full arm → release → re-arm → cap cycle** — PASS. Against a real throwaway clone + local bare remote (using the PR's own scripts, with `graph-commit`'s CI check-run polling stubbed the same way the PR's own `test-arm-wait.sh`/`test-release-wait.sh` harnesses do, since it requires a live GitHub host): mint (NONE) → release (phase done, `wait_attempts` unchanged, `source.blocked_by` byte-identical, ONE commit) → re-arm (REARM, attempts 1→2, `blocked_by` still byte-identical) → extend-while-armed (EXTEND, attempts unchanged) → two more release+re-arm cycles to reach the cap (attempts=4) → the real sweep correctly cap-parks (not releases) the WAIT with the verbatim `wait_reason`/`wait_recommendation`, landing `office_hours` on the WAIT only while the source's `office_hours` stays null and the source remains held.
8. **Office-hours gating** — PASS. Against the cap-parked fixture from item 7, `office-hours-select.ts --list` surfaces the WAIT id; against the same node flipped to armed (phase null, office_hours null) or released-but-not-re-armed (phase done, office_hours null), it does not appear. The source node never appears in any state.
9. **Census owed-prune exclusion** — PASS across 5 fixtures. A released WAIT whose source is present and not done is excluded from `donePresent`; the same WAIT with its source marked done, or with its source deleted, is included; an ordinary done tactic (control) is always included; an armed WAIT is excluded for the ordinary not-done reason.
10. **Production tick observation (post-merge only)** — residue, `needs-human-judgment` with `Flag: planned-deferral`. Classified `needs-main` by the Step 3.5 disposition Workflow (not `needs-human`) — it is machine-verifiable via the tick journal, just not before merge. See **Filed follow-ups** below.

## Disposition triage

`needs-main` items are appended as node-body residue (Step 3.6, node lane) and dropped from the escalation set, so they do not park this PR.

- **Item 10** — class: `needs-main` (planned-deferral). `verify`: n/a (planned-deferral items are excluded from the skeptic candidate set; no `verify_report` entry). Rationale (Workflow): "planned_deferral is true: the node body's own Verification section documents this as an observe-in-production check requiring two consecutive live dispatch-tick runs against the real journal, which cannot be asserted pre-merge. The mechanism itself was already verified against fixtures, so there is no defect for Opus to fix — only post-merge cadence observation remains."

No `opus-fixable` residue existed this pass (`result.fix_plan` is `null`), so the Step 3.7 auto-fix lane did not run. No `needs-human` items survived (item 10 reclassified to `needs-main`), so nothing escalates to office-hours.

## Filed follow-ups

Node-target lane: item 10 was appended as a `## needs-main residue` section directly to the tactic's own body (`intentions/tactic-wait-calendar-release.md`), landed via a standalone `graph-commit` from a scratch worktree (commit `cb2d0111`, verified present on `origin/main`) — no GitHub issue was filed. `Verifiability: WAIT` (awaited event: this PR merging to `main` plus at least two subsequent live `dispatch-tick` runs). It will be drained post-merge when `review → main-qa` fires, per the sequencing this node body itself documents.

## Bugs found

None.

natb1:
## qa · attempt 0
failed: none

9/10 QA plan items PASS (all 7 auto-verify commands re-run fresh; live fixture checks of the router exclusion, the full arm→release→re-arm→cap lifecycle against a real bare remote, office-hours gating, and the census owed-prune exclusion all confirmed). 1 residue item (production tick observation) is inherently post-merge-only; classified `needs-main` and filed as node-body residue (not office-hours) — see `## needs-main residue` in this node's own body.


## review · attempt 1
failed: none

20 findings surfaced, 3 actionable, 3 fixed (commit 0df9ffdb: unbounded WAIT horizon, EXTEND attempt-cap bypass, wait-recheck head-of-line starvation). 7 Required findings refuted by adversarial-verify, 3 out-of-scope (no follow-up filed — none met the meaningfulness bar), 7 informational (erosion metrics). No deviation. disposition=completed_with_fixes.

natb1:
# Review-fix summary

20 findings surfaced (Lane B: domain security reviewers, npm, erosion — no CodeQL/npm alerts this run). 3 actionable Required findings were fixed; 7 Required findings were dropped by adversarial verification; 3 security findings were judged out of scope; 7 Informational (erosion) findings need no action. No code-review (Lane A) findings and no Deferred entries this run.

## Fixed (security) — commit `0df9ffdb`

- **red-team-0** `packages/intentionsutil/scripts/wait-node-decide.ts:202` — unbounded WAIT horizon: `decideWait` accepted an arbitrarily distant `--until`, so a WAIT could stay armed indefinitely while excluded from both router candidate selection and census debt (invisible hold). Fixed by bounding the horizon (30 days), stamping `wait_armed_since`, enforcing the bound in schema Rule 21, and classifying over-horizon waits `capped` so the sweep escalates them to office-hours.
- **red-team-1** `packages/intentionsutil/scripts/wait-node-decide.ts:219` — the EXTEND disposition never touched the `wait_attempts` counter, so repeated EXTEND calls could hold a source blocked indefinitely without ever tripping the re-arm attempt cap. Fixed as part of the same horizon-bounding change (armed-age tracking now classifies too-long-armed waits `capped` regardless of attempt count).
- **red-team-3** `.claude/skills/dispatch-propagate/scripts/lib-wait-recheck.sh:424` — head-of-line starvation: candidates are processed in a fixed ascending-`waitId` order against a single shared attempt-cap counter, so one or more persistently-failing early candidates could indefinitely starve later due WAITs of a recheck slot. Fixed with a round-robin candidate cursor plus per-node failure backoff and office-hours escalation, held in a machine-local state file; 3 new regression cases added.

## Refuted (adversarial-verify dropped)

- **input-validation-0** `packages/intentionsutil/scripts/arm-wait:153` — missing slug-shape validation on the positional source-node-id before filesystem interpolation. Refuted: the first use of the id is a `git rev-parse` against an `origin/main` tree-ish, which fails closed on any traversal payload before any write occurs; callers only ever pass internally-known ids. Hardening nit, not exploitable.
- **input-validation-1** `.claude/skills/dispatch-propagate/scripts/lib-wait-recheck.sh:478` — concern that an unrecognized/empty enumerator class could fall through to a due-release write. Refuted: the TypeScript producer emits from a closed 4-value union and both writer and consumer enforce field shapes that can't shift TSV columns, so the scenario has no code path today.
- **auth-0** `packages/intentionsutil/src/schema.ts:1184` — schema Rule 21 permits an absent `wait_attempts` while the sweep classifies that state `malformed`-and-skip, raised as a potential silent indefinite hold. Refuted: the sole production writer always mints/repairs `wait_attempts >= 1`; the gap is reachable only via direct hand-edited frontmatter (privileged operator error), and the malformed-skip behavior is documented-intentional.
- **auth-2** `packages/intentionsutil/scripts/release-wait:290` — claimed the release re-assert doesn't re-check the attempt cap. Refuted: the enumerator's class union already bakes in the cap, `wait_attempts` cannot change between enumeration and reassert, and the reassert independently checks `office_hours` (which capped waits also set).
- **data-exposure-0** `packages/intentionsutil/scripts/wait-node-decide.ts:312` — reason/recommendation/body file content is landed with only `.trim()`, no redaction. Refuted: no current caller in this diff passes freshly-composed diagnostic text through `arm-wait`; the intended future caller isn't wired up yet, so there is no live exposure path today. Forward-looking design note for that future caller.
- **red-team-5** `packages/intentionsutil/scripts/arm-wait:99` — duplicate of input-validation-0 (missing source-id slug validation); refuted for the same reason (the `git rev-parse` gate).
- **red-team-6** `packages/intentionsutil/scripts/release-wait:103` — `--now` clock-override flag could in principle force an early release. Refuted: the only caller (`lib-wait-recheck.sh`) never passes `--now`; an actor able to invoke it directly already holds equivalent graph-write privilege through other means.

## Out-of-scope (security)

No follow-up filed for any entry in this bucket (Step 5 filed nothing this run).

- **red-team-2** `.claude/skills/dispatch-propagate/scripts/dispatch-tick:633` — concurrency-gate concern on the new `wait_recheck_sweep` call sites.
- **red-team-4** `packages/intentionsutil/src/router.ts:539` — WAIT exclusion applied only in the draft branch, not the executable-tactic loop.
- **red-team-7** `.claude/skills/dispatch-propagate/scripts/lib-wait-recheck.sh:461` — `wait_reason`/`wait_recommendation` read verbatim into park attributes.

## Informational (erosion)

- `packages/intentionsutil/scripts/graph-census-debt.ts:118` — net cyclomatic-complexity increase vs origin/main.
- `packages/intentionsutil/src/router.ts:465` — net cyclomatic-complexity increase vs origin/main.
- `packages/intentionsutil/src/schema.ts:750` — net cyclomatic-complexity increase vs origin/main (file total 216→235).
- `packages/intentionsutil/test/graph-census-debt.test.ts:155` — net cyclomatic-complexity increase in test file.
- `packages/intentionsutil/test/router.test.ts:15` — net cyclomatic-complexity increase in test file.
- `packages/intentionsutil/test/schema.test.ts:998` — net cyclomatic-complexity increase in test file (file total 304→340).
- `packages/intentionsutil/test/wait-sweep.test.ts:6` — net code-duplication increase across changed files (clone blocks 1→8, duplicated lines 7→80).

## Coverage

Partial: the result dump's size was not independently verified (byte-count check unavailable in this runtime); its content is otherwise complete.

packages/intentionsutil/scripts/arm-wait
```diff
@@ -0,0 +1,358 @@
...
+# fall back to the local copy (code-style.md).
+if ! git -C "$REPO_ROOT" fetch origin main >&2; then
+  echo "arm-wait: could not fetch origin/main to refresh $SOURCE_ID before arming" >&2
+  exit 1
+fi
+if ! SOURCE_BLOB="$(git -C "$REPO_ROOT" rev-parse "origin/main:intentions/$SOURCE_ID.md" 2>/dev/null)"; then
+  echo "arm-wait: intentions/$SOURCE_ID.md does not exist on origin/main — cannot arm a wait for a node that is not landed" >&2
+  exit 1
+fi
+if ! git -C "$REPO_ROOT" show "origin/main:intentions/$SOURCE_ID.md" > "$INTENTIONS_DIR/$SOURCE_ID.md"; then
```
natb1: **Refresh writes happen before `MUTATED=1`, so the documented "nothing is written" refusal paths leave a dirty tree.**

This line overwrites `intentions/$SOURCE_ID.md` with origin/main's bytes, and line 187 does the same for the wait file — but `MUTATED=1` is not set until line 246, and the EXIT trap only rolls back when `MUTATED == 1`. Every exit between the refresh and line 246 therefore leaves the overwritten file(s) behind with no rollback:

- line 188 / 195 (wait refresh failure, stray unlanded wait file)
- the second `run_decide` at line 199, which `exit`s 2 on wait-node-decide's own refusals
- line 209 (disposition vs. origin/main-presence disagreement)
- line 213 (`mktemp -d` failed)

Concrete scenario: `arm-wait tactic-foo --until <35 days out> ...`. wait-node-decide refuses (over-horizon) and arm-wait exits 2 — but `intentions/tactic-foo.md` has already been rewritten from origin/main. In the far-behind PR worktree the header says this script "typically" runs from, that is a modified tracked file, which then trips `graph-commit`'s `assert_clean_outside_ids` for every *other* unrelated node on the next write. The header's exit-code note ("Nothing is written on any of those") is not true as written.

`release-wait` gets this right: it takes a pre-refresh copy and sets `MUTATED=1` *before* the refresh write.

packages/intentionsutil/scripts/arm-wait
```diff
@@ -0,0 +1,358 @@
...
+# pre-mutation failure must not touch files this script never wrote. A failed
+# land must leave NO dirty intentions/*.md — otherwise graph-commit's
+# assert_clean_outside_ids guard trips for every OTHER unrelated node on the
+# next run.
+MUTATED=0
+# Bash traps REPLACE (not stack) on the same signal, so this single EXIT trap
+# owns the temp cleanup AND both files' rollback: restore each pre-existing file
+# from the immutable blob SHA captured before the mutation (not the moving
+# origin/main ref), and delete a wait file that was born fresh in this run.
+trap '
```
natb1: **The rollback trap is missing both guards `release-wait`'s `restore_node` has.**

1. *It restores origin/main's bytes, not the pre-refresh copy.* `SOURCE_BLOB`/`WAIT_BLOB` are origin/main blobs. `release-wait`'s header spells out why that is wrong: a FRESH_BLOB restore "leaves origin/main's content behind as a local modification in the far-behind checkout this script routinely runs from." Scenario: run arm-wait from a PR-branch worktree whose `intentions/<source>.md` differs from origin/main and let graph-commit refuse the CAS — the trap writes origin/main's bytes over the file, so `git status` reports it modified and `git merge --ff-only` is blocked. A rollback should restore the bytes the script *found*: capture a pre-refresh copy (as `release-wait` does with `TMPORIG`) and restore from that.

2. *It restores unconditionally after a graph-commit failure.* `release-wait` declines to restore once HEAD has moved or the file no longer hashes to what it wrote, because "graph-commit has failure paths that commit locally or `git reset --hard` before exiting non-zero, and putting the pre-refresh bytes back over those would be an uncommitted REVERT of an already-committed release." The same applies here: graph-commit commits locally and then fails on push, this trap writes the pre-arm blobs back — the arm is committed but the working tree now carries an uncommitted revert of it. Port the `HEAD_BEFORE` / `EXPECT_BLOB` guards.

packages/intentionsutil/scripts/wait-node-decide.ts
```diff
@@ -0,0 +1,404 @@
...
+  const existing = nodes.find((n) => n.id === waitId);
+  if (existing !== undefined && existing.office_hours !== null) {
+    throw new Error(
+      `wait-node-decide: "${waitId}" already carries a non-null office_hours — ` +
+        `the re-arm attempt cap already fired, so re-arming here would erase ` +
+        `that escalation; clear the park first`,
+    );
+  }
+
+  const disposition: Disposition =
```
natb1: **`decideWait` never checks that the node it found at the derived id actually *is* a WAIT node.**

Any pre-existing node at `waitIdFor(sourceId)` is classified `REARM` (phase done) or `EXTEND` (anything else). `isWaitNode` — the canonical-id check that `wait-sweep.ts`, `router.ts`, `graph-census-debt.ts` and `release-wait`'s re-assert all apply — is not consulted here, so an ordinary tactic that merely happens to be named `tactic-wait-*` gets hijacked as this source's wait.

Concrete scenario, using a node this repo already has: `arm-wait tactic-calendar-release --until ...` derives `tactic-wait-calendar-release`, a real work tactic. It is not `phase: done`, so the disposition is `EXTEND`; `arm-wait` then merges `attributes.wait_until` (plus a backfilled `wait_armed_since`) into that tactic's frontmatter, appends an `## Extend <ts>` stanza to its body, and adds a `blocked_by` edge from `tactic-calendar-release` to it. Because the EXTEND patch never adds `wait_for`, validate-graph Rule 22 stays inert and the corruption lands silently; the sweep never enumerates the node either, so the bogus block on the source is invisible.

Suggested guard: when `existing !== undefined && !isWaitNode(existing)`, throw — the id is taken by a non-wait node, and the caller should be told rather than have it rewritten.

packages/intentionsutil/scripts/graph-census-debt.ts
```diff
@@ -120,7 +121,34 @@ export interface CensusDebt {
...
+   * becomes unreachable (the wait can never escalate to the author); and
+   * because Rule 13 rejects a dangling `blocked_by`, the prune must also strip
+   * the source's `blocked_by` edge naming the wait, releasing the hold the
+   * wait exists to apply.
+   *
+   * The exception is narrow. A released WAIT whose source is GONE from the
+   * store, or whose source is ITSELF done, has nothing left to hold and is
+   * genuine debt — it stays in `donePresent` for the census to drain.
+   */
+  const isLiveRearmTarget = (n: IntentionNode): boolean => {
```
natb1: **`isLiveRearmTarget` omits the edge check that `listWaitCandidates` treats as load-bearing.**

The exemption is described as narrow — "a released WAIT whose source is GONE from the store, or whose source is ITSELF done ... stays in `donePresent`" — but there is a third dead case it lets through: the source is present and open, yet its `blocked_by` no longer names the wait. `listWaitCandidates` (wait-sweep.ts, step 3) explicitly emits nothing for that state ("edge already detached").

Concrete scenario: an author (or a hand edit / conflict resolution) drops `tactic-wait-x` from `tactic-x`'s `blocked_by` while the wait is in its released `phase: done` window. The wait now holds nothing and is enumerated by nothing, but `isLiveRearmTarget` still returns true, so it is permanently excluded from `donePresent` and the census can never collect it — un-drainable residue that grows one node per occurrence.

Mirror the enumerator: `return source !== undefined && source.phase !== "done" && source.blocked_by.includes(n.id);`

.claude/skills/dispatch-propagate/scripts/lib-wait-recheck.sh
```diff
@@ -0,0 +1,796 @@
...
+      # streak) and the failure bookkeeping (arm the backoff, escalate at the
+      # threshold) cannot drift apart between them.
+      writer=""
+      rc=0
+      if [[ "$cls" == "capped" ]]; then
+        # The park strings come from the node itself. An unreadable or empty
+        # pair KEEPS the WAIT: parking with an invented reason would hand the
+        # author an escalation that names nothing. No landing-lock push was
+        # spent, so this path does NOT consume the cap.
+        if ! attrs_json=$(_wait_recheck_park_attrs "$root" "$wait_id"); then
```
natb1: **The two `attrs-*` failure paths bypass the backoff/escalation bookkeeping, so this failure mode retries forever.**

Both early-`continue`s here (`attrs-unreadable` at 700-706 and `attrs-empty` at 710-716) bump `failed` and then `continue` — they never call `_wait_recheck_note_failure`, so `_WAIT_RECHECK_FAILS` stays at 0, no skip window is armed, and `escalate_at` is never reached.

Concrete scenario: a `capped` WAIT whose node cannot be read by `_wait_recheck_park_attrs` (a store read that throws, a permissions problem, a node landed before rule 22 existed with an empty `wait_reason`). Every tick, forever: the sweep enumerates it, classifies it `capped`, fails to read the strings, logs `resolve-failed (rc=attrs-…)`, and moves on — `status=ok`, `escalated=0`. That is exactly the "re-attempt every tick until the heat death of the fleet" state rule (e)'s header says the backoff and escalation exist to close, and the one class of failure where the wait is *guaranteed* to need a human.

Fix: route these two through the same shared failure block (`_wait_recheck_note_failure` + the `fails_now >= escalate_at` escalation), or at minimum call `_wait_recheck_note_failure` before the `continue`. Note the escalation path itself reads the same attributes, so it will also fail — the escalation for this class needs a fallback reason rather than `_wait_recheck_park_attrs`.

.claude/skills/dispatch-propagate/scripts/lib-wait-recheck.sh
```diff
@@ -0,0 +1,796 @@
...
+
+      # (a) Unsafe id. Counted only as a candidate — no other bucket claims it.
+      if ! _wait_recheck_id_is_safe "$wait_id" || ! _wait_recheck_id_is_safe "$source_id"; then
+        printf 'lib-wait-recheck: unsafe-id (%s / %s)\n' "$wait_id" "$source_id" >&2
+        _wait_recheck_log_decision "$wait_id" "$source_id" "$attempts" "$until_at" "$cls" "unsafe-id"
+        continue
+      fi
+      # Every id that got past the guard is still live in the store, so its
+      # state entry (if any) is worth keeping; anything else is pruned at save.
+      seen_ids["$wait_id"]=1
```
natb1: **A failure streak survives a successful re-arm, so a stale streak can escalate a healthy wait on its first fresh failure.**

`seen_ids` is recorded for *every* candidate that passes the id guard, including `waiting` ones, so the prune loop at the end never clears a wait that is still enumerated. And `_wait_recheck_clear_failure` is only reached from a successful write or a successful escalation — a wait that leaves the failing state by being re-armed (by `arm-wait` REARM, or by a human) is enumerated as `waiting` from then on and keeps its old `fails` count indefinitely.

Concrete scenario with the defaults (`FAIL_ESCALATE=6`): `release-wait` fails 5 times during a landing-lock storm; the storm clears and the author re-arms the wait, so it goes back to `waiting`. Weeks later it comes due, hits one transient failure — `fails` goes 5 → 6 and the sweep immediately parks it to office-hours and stops retrying, as if it had failed 6 consecutive times in this cycle.

Fix: clear the node's state when it is enumerated in a non-actionable class that means the previous cycle ended (`waiting`, or a `wait_attempts`/`wait_armed_since` newer than the streak), rather than only on a successful write.

PR #3054, opened 2026-08-09
graph: main-checkout session census + held alarm (tactic-blocked-session-invisible-to-census)

natb1:
Implements the intention node `tactic-blocked-session-invisible-to-census`
(graph-native tactic; no GitHub issue).

## Summary

The reap/health census classified sessions on `state: done` alone, so a
session in any other non-`working` state (`blocked`, `stopped`) was invisible
to every health probe. When such a session sat in the **main checkout**, it
held the shared tree dirty indefinitely — blocking `dispatch-select-tick`'s
`--ff-only` sync and every `graph-commit`. This is exactly what happened
2026-08-06: background session `f2416fda` (`name: sync-repair`, `cwd` = main
checkout) went `state: blocked` after an API error and sat there for 3 days,
stalling the whole fleet, while `HELD_FOR_DEBUG_COUNT` read `n=0` the entire
time because `blocked` isn't a terminal state, and because `sync-repair`
doesn't match the existing worker-keyspace name filter either. This PR closes
both blind spots — state and keyspace — and adds a bounded escalation path so
a future incident like this self-resolves instead of stalling forever.

Invariant preserved throughout: `state: done` is the discriminator for
REAPING, not for HEALTH (I22). No unit here reaps, kills, or parks a `blocked`
session — reaping policy for `blocked` is deliberately out of scope.

## Unit 1 — Registered-view, cwd-keyed session lister + full-state census

Adds `claude_agents_list_sessions_in_cwd_all <path>` and
`claude_agents_count_by_state` to `lib-claude-agents.sh`. Both query
`claude agents --json --all` directly (the REGISTERED view, not the ACTIVE
`--cwd` snapshot) with **no keyspace filter**, closing the blind spot that
hid `sync-repair` (it matches none of `^[0-9]+-|^tactic-|^strategy-`). The
cwd filter is exact-string equality, never a prefix/substring test. The
by-state census emits both an overall and a keyspace-partitioned
(`worker`/`other`) breakdown, plus a `terminal` roll-up reusing the existing
shared `CLAUDE_AGENTS_TERMINAL_STATES_JQ` enum (never re-listed, and
`blocked` is deliberately never added to it). 28 new test cases in
`test-lib-claude-agents.sh`.

## Unit 2 — Surface the by-state breakdown in the sweep census

Wires `claude_agents_count_by_state` into `dispatch-sweep`'s existing
`HELD_FOR_DEBUG_COUNT` log block: one `SESSION_STATE_CENSUS` log line per
row, `none (empty registry)` on an empty census, `UNKNOWN (daemon
unqueryable)` on rc 1 — logged distinctly from zero, same convention as the
existing metric. Observability only; nothing downstream branches on it.

## Unit 3 — `main-checkout-held` alarm

Adds a `main-checkout-held` kind to `dispatch-fleet-alarm`'s closed `KINDS`
enum, and a sixth predicate to `dispatch-fleet-watch` that fires only on the
**conjunction**: the main checkout tree is dirty/ahead of `origin/main` AND
at least one registered session with `cwd` == the main checkout is
non-`working` and idle past a grace period (default 1800s, via the existing
`_standdown_session_idle_s` helper — not re-implemented). The conjunction is
deliberate: an ordinary interactive human session sitting idle in the repo
root must never alarm on its own — only a dirty/stuck tree does. Fail
direction: an unreadable daemon or unreadable tree yields `unknown`, never
`clear` (a false `clear` would `--resolve` an already-open alarm). The alarm
body is identity-only (sorted `name:state` pairs + the checkout path — never
session ids, ages, or counts, which stay in the live reading); it recommends
inspecting the session's transcript first, never a blind `claude rm`.

## Unit 4 — Bound Decision A's defer in `dispatch-select-tick`

The tick's "Decision A" deferred while a `sync-repair` session was "live" in
the main checkout, but the check was unbounded: it read the ACTIVE-only view
(invisible to a `stopped`/`done` holder) and treated any non-`"stopped"`
status — which includes `blocked`, since a blocked session's `.status`
projects as `idle` — as "still live," forever. This made the existing
3-attempt `sync-broken` escalation cap structurally unreachable. Now the
tick reads the REGISTERED view via Unit 1's lister, defers only while a
session is genuinely `working`/`busy` or non-working-but-within-grace, and
otherwise falls through to the existing fetch/merge probe — letting a failed
merge bump the attempt counter and eventually reach `sync-broken`, the
correct escalation the incident could never reach. `DLOG_SKIP_REASON` names
the stuck session and state for the decision log. Daemon UNKNOWN still fails
open (unchanged); idle-UNKNOWN still fails safe (defers).

## Verification

```verify
.claude/skills/dispatch-propagate/scripts/test-lib-claude-agents.sh
.claude/skills/dispatch-propagate/scripts/test-dispatch-sweep.sh
.claude/skills/dispatch-propagate/scripts/test-dispatch-fleet-alarm.sh
.claude/skills/dispatch-propagate/scripts/test-dispatch-fleet-watch.sh
.claude/skills/dispatch-propagate/scripts/test-dispatch-select-tick.sh
.claude/skills/dispatch-propagate/scripts/run-lint.sh
```

All six passed locally (922 total assertions across the five suites, plus a
clean lint pass).

### Manual / observe-in-production checks (not run this session — require
`dangerouslyDisableSandbox: true` on a live daemon)

1. Census breadth, live: run `dispatch-sweep` and confirm `SESSION_STATE_CENSUS`
   lines sum to `claude agents --json --all | jq length`.
2. Keyspace ratchet, live: confirm a non-worker-named session (e.g. a router)
   appears under `keyspace=other`.
3. Alarm precision, negative case: an ordinary interactive session + clean
   main checkout → `main-checkout-held: clear`.
4. Alarm, positive case: dirty tree + stuck session past grace → `finding`,
   `tactic-fleet-alarm-main-checkout-held` minted; a second pass with a
   different idle age pushes no second commit (body-identity gate holds).
5. Decision A end to end: the next real stuck main-checkout session should
   log the attributed skip reason, bump the attempt counter across ticks,
   and latch `sync-broken` on the third.
6. I22 preserved: no `blocked` session reaped/killed/parked by any of this.

### DIFF
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-fleet-alarm
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-fleet-alarm
@@ -177,7 +177,7 @@ if ! source "$SCRIPT_DIR/lib.sh"; then
 fi
 
 # The closed enum of alarm kinds. Extending it is a deliberate edit here.
-KINDS=(tick-stale daemon-degraded busy-stall automerge-suppressed unclaimed-hold watch-unknown heal-fired heal-unknown)
+KINDS=(tick-stale daemon-degraded busy-stall automerge-suppressed unclaimed-hold main-checkout-held watch-unknown heal-fired heal-unknown)
 
 usage() {
   cat <<'USAGE'
@@ -186,7 +186,7 @@ usage: dispatch-fleet-alarm --kind <kind> --statement <text> --body-file <path>
        dispatch-fleet-alarm -h|--help
 
 kinds: tick-stale daemon-degraded busy-stall automerge-suppressed
-       unclaimed-hold watch-unknown heal-fired heal-unknown
+       unclaimed-hold main-checkout-held watch-unknown heal-fired heal-unknown
 
 Records an out-of-band fleet-health reading as the graph node
 tactic-fleet-alarm-<kind> (find-or-create; body refreshed on re-detection).
--- a/.claude/skills/dispatch-propagate/scripts/dispatch-fleet-watch
+++ b/.claude/skills/dispatch-propagate/scripts/dispatch-fleet-watch
@@ -2,7 +2,7 @@
 #
 # dispatch-fleet-watch — ONE-SHOT fleet-health watcher pass.
 #
-# Reads five independent fleet-health predicates, reports each one, and turns
+# Reads six independent fleet-health predicates, reports each one, and turns
 # every finding (and every unreadable input) into a graph node through
 # dispatch-fleet-alarm. Then it exits. It does NOT loop and it takes no deadline
 # argument: cadence belongs to the systemd timer that invokes it, not to the
@@ -30,6 +30,8 @@
 #   4. automerge-suppressed  open tactic-main-red-* latch held too long
 #   5. unclaimed-hold        a tracked hold blocking top-ranked work that no
 #                            session and no reservation claims
+#   6. main-checkout-held    the shared main checkout is dirty/unpushed AND a
+#                            stuck (non-working) session is sitting in it
 #
 # WHY TWO SNAPSHOTS, BOTH CAPTURED AT THE INPUT BOUNDARY:
 #   Predicate 3 reads the live session array through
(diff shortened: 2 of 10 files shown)


natb1:
Closing as part of Bundle 0 item 4 of the dispatch-rsi serialized-PR plan, per the author's ruling on 2026-08-23.

The clean half of this draft — the `dispatch-select-tick` and fleet-alarm portion, touching no contested file — was extracted and merged as #3101. What remains here is the contested half only, which the plan absorbs into the intention graph rather than adjudicating as a PR diff.

**The branch stays alive.** Nothing is lost: `tactic-blocked-session-invisible-to-census` is unchanged on the remote and this PR can be reopened at any time.

One finding from this draft's extraction is worth carrying forward, and is recorded on `main`: the `dispatch-sweep` test N10c **passed vacuously** — it asserts the `UNKNOWN (daemon unqueryable)` branch, which is exactly what a *missing* `claude_agents_count_by_state` also produces. A test that passes because its subject is absent makes a broken extraction look green, so that half was treated as a failure rather than a pass.

Note for anyone reading the split rationale: the plan's original premise for this exercise — "the conflict boundary and the collision boundary are the same line" — was found false and is corrected on `main`. Textual mergeability is not semantic independence.

PR #3056, opened 2026-08-10
check-node-selection: snapshot provenance as an explicit input, refuse-by-default on unprovable freshness (I17)

natb1:
Implements graph node `tactic-graph-execute-fresh-main-read` (graph-native
tactic; no GitHub issue) — carrier for invariant **I17** ("sync the main
checkout before every `dispatch-graph-execute`").

## The defect

`packages/intentionsutil/scripts/check-node-selection.ts`'s selection gate
resolved node state from a filesystem directory whose freshness against
`origin/main` was only guaranteed by ONE caller (`dispatch-select-tick`) doing
a prior fetch + ff-only merge — a precondition documented only as a prose
comment in a different file (`provision-node-worktree`) and enforced nowhere.
Every other caller, including the sanctioned manual `dispatch <node-id>` lane,
could read a stale snapshot and let a worker launch onto a node that was
actually parked (`office_hours`) on `origin/main`.

## Ratified design (2026-08-09 office-hours sitting)

Snapshot provenance becomes an **explicit input** to the selection predicate
rather than something it fetches itself — the predicate stays pure (no
`node:child_process`, ever) and returns a distinct `unknown-freshness` result
variant when freshness can't be proven. The default posture on unprovable
freshness is **refusal**; the one sanctioned human-override lane
(`dispatch --allow-stale <node-id>`) can proceed anyway, at the cost of one
flag, leaving an attributable decision-log record.

## Units

1. **Snapshot provenance as a required input** to `evaluateSelection`, with a
   new `unknown-freshness` result variant (`EXIT_UNKNOWN_FRESHNESS = 15`) —
   landed warn-only so no caller's behavior changed yet.
2. **Every snapshot-acquisition site captures and passes its own
   provenance.** New `lib-main-snapshot.sh`; `provision-node-worktree` stops
   reading the shared main-checkout working tree and materializes its own
   `origin/main` snapshot instead of depending on `dispatch-select-tick`'s
   side effect.
3. **`--allow-stale` wired end-to-end** (`dispatch` CLI → `dispatch-tick` →
   `dispatch-graph-execute` → `provision-node-worktree`), gate exit 15 routed
   at every consuming site (`refused <id> unknown-freshness`, no graph write),
   and a decision-log attestation recorded on a stale-allowed launch.
4. **The offline manual lane made reachable.** `dispatch-select-tick`'s sync
   probe bypasses `sync-failed` under `--allow-stale`, without touching the
   `sync-repair-pending` defer or the `sync-broken` latch — otherwise the
   override would be unreachable in its own motivating scenario (network
   down).
5. **Flip the default: unprovable freshness refuses.** The gate now returns
   exit 15 unless `--allow-stale` was given; a parked node read through an
   unproven snapshot now exits 15 (not 12) — both refuse the launch, 15 is the
   honest reason. `provision-node-worktree`'s fetch-failure handling degrades
   gracefully to an unattested local-ref snapshot (still refused by the gate)
   instead of an `exit 2` that `dispatch-graph-execute` would otherwise park
   as a node defect. `assert-node-selection` passes exit 15 through
   distinctly (a gap the plan didn't anticipate, closed during Unit 5 so
   `align-tactics`'s own exit-15 routing bullet, added in Unit 3, is actually
   reachable).

## Verification

All 7 `verify` blocks from the node's plan pass:

- `npx vitest run --project packages/intentionsutil --root .` — 44 files,
  896/896 tests.
- `test-provision-node-worktree.sh` — 49/49.
- `test-assert-node-selection.sh` — 13/13.
- `test-dispatch-derive-node-target.sh` — 27/27.
- `test-dispatch-graph-execute.sh` — 129/129.
- `run-typecheck.sh` — pass.
- `run-lint.sh` — pass (nix flake check + prose rules).

Plus the plan's non-`verify` checks:

- `test-dispatch-select-tick.sh` (not wired into CI, 1741 lines) — 233/233,
  including the two new offline-bypass cases.
- Purity re-assertion: no `child_process`/`execFile`/`spawnSync`/`execSync` in
  `check-node-selection.ts`; the "no process exit" doc comment intact.

## Note on plan-fidelity

One unanticipated gap surfaced only after Unit 5 flipped enforcement:
`assert-node-selection` mapped gate exit 15 into its own generic `exit 2`
catch-all instead of passing it through distinctly, which would have made
`align-tactics`'s own exit-15 bullet (added in Unit 3) unreachable through
that caller. Closed as part of Unit 5 rather than filed as a follow-up, since
it directly completes work already landed on this branch.


natb1:
# Fix-checks summary — PR #3056

## Iteration 1

- **Failed checks**: `nixos-build` (Build nixosConfigurations.nixos)
- **Outcome**: fixed
- **Reproduced**: yes
- **Reproduce command**: `curl -L -o WezTerm-windows-nightly.zip https://github.com/wez/wezterm/releases/download/nightly/WezTerm-windows-nightly.zip && nix hash file --sri --type sha256 WezTerm-windows-nightly.zip` (matched CI's reported "got" hash `sha256-s/Jfeer9ELfFkYnzXMMPMbOQ42nQ3lxe7khksDHaGks=`), then `nix/home/sync-wezterm.sh`
- **Failure excerpt**:
  ```
  error: hash mismatch in fixed-output derivation '/nix/store/yf5iiwqzksps6zj1gy7qrbqj7l45w407-WezTerm-windows-nightly.zip.drv':
           specified: sha256-bTvVHVpB8Mh6g2lF2RB9Egs2IApanVb5Z1R2M9UCZZ8=
              got:    sha256-s/Jfeer9ELfFkYnzXMMPMbOQ42nQ3lxe7khksDHaGks=
  error: Cannot build '/nix/store/wlx5sc0i5wg4fzqzg2hqxznrxcfgdymr-nixos-system-nixos-26.11.20260804.e72e4f2.drv'.
  ```
- **Why not caught**: unrelated to this PR's own changes (which only touch `packages/intentionsutil/scripts/check-node-selection.ts` / `dispatch-graph-execute` node-selection freshness). Upstream `wez/wezterm` published a genuinely newer nightly build (`WezTerm-windows-20260805-104032-4b1c3c15`) than the one pinned in `nix/home/wezterm-pin.nix` (`20260716-195552-76b606ec`) — not merely a re-upload of the same build's bytes, which ruled out the file's documented one-line hash-only refresh shortcut and required a full re-pin via `nix/home/sync-wezterm.sh`.
- **Fix**: ran `nix/home/sync-wezterm.sh` to regenerate all five pinned fields (`version`, `rev`, `srcHash`, `cargoHash`, `windowsZipHash`) against the current upstream nightly; commit `e921987afd69d9204186088d253a87ed43f4d4d5`.

natb1:
Closing as part of Bundle 0 item 4 of the dispatch-rsi serialized-PR plan, per the author's ruling on 2026-08-23.

The clean half of this draft — the node-selection and main-snapshot portion, touching no contested file — was extracted and merged as #3104. What remains here is the contested half only, which the plan absorbs into the intention graph rather than adjudicating as a PR diff.

**The branch stays alive.** Nothing is lost: `tactic-graph-execute-fresh-main-read` is unchanged on the remote and this PR can be reopened at any time.

One finding from this draft's extraction is worth carrying forward, and is recorded on `main`: merging this draft's auto-mergeable files would have **silently reverted** `nix/home/wezterm-pin.nix` on `main`. Git reported no conflict, because there was none — the merge simply carried an older version of a file the draft had no business touching. That is why the extraction procedure now diffs every auto-merged file that is unrelated to the tactic, rather than trusting a clean merge.

Note for anyone reading the split rationale: the plan's original premise for this exercise — "the conflict boundary and the collision boundary are the same line" — was found false and is corrected on `main`. Textual mergeability is not semantic independence.

PR #3068, opened 2026-08-11
Run node-lane merge chain on dispatch-tick's paused branch

natb1:
## Summary

The pause sentinel in `dispatch-tick` is documented (dispatch-tick:311-312) as
gating worker SPAWNING only, never ledger bookkeeping. But the node-lane merge
lane (`graph-auto-merge`) only ran inside `dispatch-select-tick`, which sits
past the pause's exit-0 short-circuit — so a reviewed, green node-lane PR could
not merge autonomously while paused, and `reconcile-graph-merged` never ran
either, leaving merged nodes stuck unreconciled.

This made the standing dispatch pause self-blocking: its own first resume
criterion required PR #3052 (node-lane, phase review, 23/23 green, mergeable
CLEAN) to merge, but the only lane that would merge it was disabled by the
pause itself.

## Change

Added the node-lane merge/reconcile chain to the paused branch of
`dispatch-tick`, before its `exit 0`: compute `OPEN_MAIN_RED` via
`dispatch-graph-main-red-sync` (failing closed to `UNKNOWN` on error), gate
`graph-auto-merge` on it exactly like `dispatch-select-tick` does, then run
`reconcile-graph-merged` unconditionally — mirroring `dispatch-select-tick`'s
own invocation style verbatim.

Out of scope, left unchanged: the issue-lane merge (`dispatch-auto-merge`),
spawning/selection/reseed-arming behavior, the `OPEN_MAIN_RED` gate itself, the
pace curve, `--manual` behavior, and the sentinel's location. Lifting the
current pause is a separate, attended `/rsi` decision.

## Test plan

- [x] `test-dispatch-tick.sh` — 144/144 passed (12 new assertions covering:
      merge+reconcile run on a paused/known-good tick; merge suppressed on a
      paused/main-red tick; merge suppressed when main-red-sync fails
      (fail-closed to UNKNOWN); existing sweeps and non-paused-path behavior
      unchanged)
- [x] `run-lint.sh` — clean
- [ ] Manual, after landing: confirm a reviewed node-lane PR merges on a
      heartbeat tick with no operator action while the sentinel is present,
      and that `journalctl --user -u dispatch-tick` still logs the paused
      no-scheduling line. Then re-measure resume criterion 1.

PR #3002, opened 2026-07-31
graph: bound pending CI on the autonomous dispatch path (tactic-autonomous-ci-pending-liveness-bound)

natb1:
## Context

The autonomous dispatch tick has a cap-and-escalate convention for every mechanical stuck state except one: a graph-native tactic whose draft PR's CI never reaches a verdict. Two independent surfaces observe pending CI every tick and do nothing but skip forever — `graph-select-target`'s `qa|review` selection gate, and `reconcile-graph-review-stall`'s sweep over reviewed-marker-excluded nodes (whose armed auto-merge can never fire). Neither has a counter, a hold, a park, or an operator surface.

Implements node `tactic-autonomous-ci-pending-liveness-bound` (2026-07-31 `/align-tactics` round).

## What this adds

- **Unit 1** — a fourth hold kind, `ci-pending-stalled` (slug `ci-stalled`), registered in `packages/intentionsutil/scripts/hold-node-decide.ts`'s `HOLD_KINDS`/`KIND_SLUGS`, with matching usage-string updates in `hold-node` and `resolve-hold`, and new `hold-node-decide.test.ts` cases.
- **Unit 2** — a SHA-keyed strike sidecar (`ci_pending_strike_bump`/`ci_pending_strike_clear` in `lib.sh`, `DISPATCH_CI_PENDING_STRIKE_CAP=8`) and the selection-surface producer in `graph-select-target`: below cap it emits a `ci-pending (strike n/8)` skip reason; at cap it lands a `ci-pending-stalled` hold via `hold-node`. The explicit `--node` lane and an unreadable PR are exempt. `dispatch-graph-execute`'s exit-10 intra-tick race is documented as deliberately uncounted (the sustained form is caught here).
- **Unit 3** — the same bound at `reconcile-graph-review-stall`'s independent observation point (reviewed-marker-excluded nodes never revisit the selector), sharing the sidecar and a hold slot budget shared 1-per-sweep with the existing conflict-hold route.

## Test plan

- [x] `.claude/skills/dispatch-propagate/scripts/run-unit-tests.sh --pr-scripts` — all suites pass
- [x] `npx vitest run --project packages/intentionsutil --root .` — 722/722 pass
- [x] `.claude/skills/dispatch-propagate/scripts/run-typecheck.sh --app packages/intentionsutil` — clean
- [x] `.claude/skills/dispatch-propagate/scripts/run-lint.sh --prose` — clean
[...]

PR #2904, opened 2026-07-19
graph-tick: tick-owned label-free auto-merge of reviewed node-lane PRs

natb1:
Implements `tactic-graph-tick-node-lane-auto-merge`: the tick reconciler owns a single label-free, CI-validated auto-merge of a reviewed node-lane PR, keyed off the node's `reviewed` execution marker (not the gh `dispatch:reviewed` label).

## Problem

No owned code actually merged a reviewed node-lane PR. `transition-node`'s worker-invoked arm step ran `gh pr ready` then the legacy `dispatch-auto-merge`, which is gated on the `dispatch:reviewed` gh label — a label the node lane deliberately never writes (`reviewed` is a graph execution marker, not a label). Result: a reviewed node-lane PR was readied and mergeable but never merged, requiring a human to merge by hand (observed on PR #2859).

## Change

- Added `.claude/skills/dispatch-propagate/scripts/graph-auto-merge`: a new tick-owned, marker-keyed reconciler. Each tick it enumerates `kind:tactic` nodes at `phase:review` whose `execution.markers` includes `reviewed`, and merges the PR label-free when it is OPEN + `mergeable == MERGEABLE` + CI `passing` + the tactic-scope fingerprint is fresh (fail-closed on scope-stale or a missing stamp — held, not merged). Honors the same `auto-merge` config opt-out kill-switch as the legacy `dispatch-auto-merge`.
- `transition-node` no longer arms or performs the merge itself: on a clean review completion it lands the `reviewed`-marker graph-commit, refreshes the scope stamp, and prints `review-complete <id> (merge deferred to tick)`.
- `dispatch-select-tick` calls `graph-auto-merge` each tick, gated on the same main-known-good condition (`OPEN_MB` empty) as the existing issue-lane auto-merge, running before `reconcile-graph-merged` (which absorbs the merge to `done`/`main-qa` on a later tick, honoring its grace window).
- Updated `review-fix/SKILL.md`'s node-lane doctrine to describe the new split.

The legacy label-gated `dispatch-auto-merge` and its issue-lane tick call are unchanged.

## Verification

- `bash .claude/skills/dispatch-propagate/scripts/test-dispatch-scripts.sh` — 2976/2976 passed.
- `bash .claude/skills/dispatch-propagate/scripts/run-lint.sh` — all lint checks passed.

PR #3073, opened 2026-08-12
Unblock the /dispatch-ladder end-to-end run

natb1:
Unblocks the `/dispatch-ladder` end-to-end run.

`/dispatch-ladder <node-id>` landed as PR #3072 (merge commit `c0a66d49`). Its
unit tests are green and **its end-to-end run has never happened**. A sweep of
the graph for tracked defects on the ladder's execution path found several that
would derail that run before it could prove anything — several of them producing
a **false halt on a healthy node**, the worst outcome for a first e2e run because
it looks like the ladder is broken when it is not.

This PR fixes the ones not already in flight, so the e2e run can produce a real
verdict. No dispatch labels, no `dispatch:` lane — these are graph nodes, not
issues, so there are no `Closes #` lines.

## Units

| Commit | What | Node |
|---|---|---|
| `a3451bf9` | `graph-auto-merge` honours `blocked_by` | closes `tactic-graph-auto-merge-blocked-by-gate` |
| `4a4be10d` | `dispatch-ladder-advance` verifies the launch actually registered; names a terminal holder | `tactic-spawn-handoff-noverify-phantom-stall` (ladder lane only) |
| `5cf5d32d` | `graph-auto-merge` syncs a behind PR branch instead of merging it | closes `tactic-graph-auto-merge-up-to-date-gate` |
| `bc2136c9` | `dispatch-ladder-run` runs `terminal_without_disposition_sweep` | new tactic to be filed |
| `b0657c98` | An orphaned check run classifies stale, not pending forever | closes `tactic-orphaned-check-run-pins-pending-ci-guard` |

### Why each one blocks e2e

- **`blocked_by`** — the ladder calls `graph-auto-merge --node <id>` as its own
  merge step, and the Step 2 enumeration never read `blocked_by`. A ladder node
  with an unsatisfied blocker merged ahead of it.
- **Launch verification** — the script emitted `launched … exit 0` on the sole
  evidence that `claude --bg` exited 0, through a `--no-verify` spawn. Any
  condition that makes that exit 0 without a session registering produced a
  phantom, which `dispatch-ladder-await` then correctly read as "no session",
  halting the whole run `stalled` on a healthy node while the ledger held the
  claim for the full 300s TTL.
- **Up-to-date gate** — `MERGEABLE` means only "no textual conflict", so a green
  PR far behind `main` could merge and turn `main` red, tripping the main-health
  gate and suppressing the entire merge lane. During a multi-hour e2e run `main`
  will move.
- **The sweep** — on a host whose heartbeat is stopped (exactly the
  bootstrap-deadlock case the ladder exists to serve), a node-lane escalation
  left a HELD job nothing would ever park, and the node unselectable.
- **Orphaned check runs** — a node pinned this way keeps the reconcile pass
  quiet, so the driver burns its entire `--ci-wait-s` budget and halts naming CI
  — a false halt on a PR that is actually green.

## Verification

Every suite was re-run by the orchestrating session, not just by the implementer:

| Suite | Result |
|---|---|
| `test-graph-auto-merge.sh` | 84/84 (was 49) |
| `test-lib-gh-rest.sh` | 334/334 (was 321) |
| `test-graph-commit.sh` | 87 passed (was 83) |
| `test-dispatch-ladder-advance.sh` | 35 passed |
| `test-dispatch-ladder-run.sh` | 144/144 (was 131) |
| `test-dispatch-tick.sh` | 181/181 |
| `test-lib-reservation-ledger.sh` | 92/92 |
| `run-unit-tests.sh --pr-scripts` | All unit test suites passed |

All affected suites are already hand-wired in `.github/workflows/unit-tests.yml`,
so no workflow change was needed. `run-unit-tests.sh` globs only
`.claude/skills/dispatch-propagate/scripts/test-*.sh`, which is why that matters.

The orphaned-check-run fix was additionally confirmed against the live incident
sha `74548a2b` (#3068): 21 `success` + 1 `neutral` + 1 null-conclusion row now
returns `failing` where it previously returned `pending`.

## Judgment calls worth a reviewer's attention

- **`tactic-graph-auto-merge-up-to-date-gate` was implemented over an open
  blocker, knowingly.** It carries `blocked_by:
  tactic-graph-tick-node-lane-auto-merge`, which is not `done` — it sits at
  `phase: main-qa`, office-hours parked, because needs-main item 9b never fired
  in production. Its *code* landed in merged PR #2904. The edge is a bookkeeping
  park, not a missing dependency. **The edge was not cleared**; no node file was
  edited here.
- **The orphan rule is narrowed to NULL-conclusion rows only.** The node body's
  plan, taken literally, would have STALE-ified any non-completed row under a
  concluded suite — which catches the #2457 desync (`in_progress` + `success`)
  and would flip genuinely green PRs to failing. Narrowed, and pinned by a
  regression test.
- **The verdict cache now stores terminal verdicts only.** Caching a `pending`
  verdict would let the orphan rule be shadowed by the entry the orphan itself
  produced, for as long as the cache directory lives. *An earlier revision of
  this body said "permanently"; that was wrong* — the only producer is
  `dispatch-select-tick`, which `mktemp -d`s the directory and `rm -rf`s it on
  EXIT, and `dispatch-ladder-run` unsets the var before every reconciler call,
  so the shadow could never outlive one tick. The rule stands on the lifecycle
  *contract* (this helper does not own the directory) rather than on today's
  only caller. Cost: repeat queries on one in-flight sha within a single tick
  now re-fetch — one extra REST call per *distinct* parent suite; an
  all-completed sha pays zero.
- **Launch verification was extended to the `conflict-lane` branch**, which the
  plan scoped out — it prints `launched` off the same unverified spawn and had
  the identical hole.
- **A terminal holder still halts 13.** Auto-releasing another session's claim is
  a policy act; the driver may sequence, never gate. Only the token changed, so
  `claude rm <session-id>` becomes discoverable.
- **A sweep-library load failure is fatal in `dispatch-ladder-run`**, deliberately
  unlike `dispatch-tick`, which logs and ticks on. The tick retries in a minute;
  for a one-shot detached driver a skipped sweep is the wedge itself.

## Deliberately out of scope

- `tactic-graph-execute-fresh-main-read` — PR #3056 is open with the fix written,
  stalled only by the orphaned-CI defect this PR fixes. **The ladder still
  inherits the stale-selection-store hole** (`dispatch-ladder-advance:137` fetches
  but never ff-merges); the e2e run must be reported with that caveat.
- `tactic-autonomous-ci-pending-liveness-bound` — PR #3002 open but `CONFLICTING`,
  12 days stale, owns `graph-select-target`.
- `tactic-reconcile-review-stall-base-pin` — most contended file in the tree.
- `tactic-pending-merge-phase` — deletes `execution.markers`, which two units here
  key off. When it is done it must also update `dispatch-ladder-await`'s
  `reviewed` check.

## Still open after this

- `tactic-spawn-handoff-noverify-phantom-stall` — the tick's budget path and the
  probe-correctness half remain.
- `tactic-select-target-held-done-label` — only token *naming* remains; its anchor
  drifted and its proposal was substantially delivered by the
  `worktree_occupancy_state` extraction.

## Review outcome

`/code-review high` raised 7 findings. All 7 are now resolved.

Two were fixed during the review pass (`9d458baa`): `check_suite_concluded` is
memoised per poll — the required contexts share one suite, so the poll was
paying a REST call per row for an answer it already had — and
`dispatch-ladder-run` now logs the new `synced #N` line instead of journalling
it as "merged nothing". A third was a wrong comment of mine, corrected in
`d92903a7`.

The remaining four were flagged but left unpatched at the time, each because
patching it would have reversed a deliberate choice. Each has now been worked
through and fixed on its own terms:

- **`failing` relocated the pin rather than removing it.** An orphan routes the
  node to `/fix-checks`, whose Step 3 ran `gh pr checks --watch` with no timeout
  and blocked on exactly the row that never concludes. `failing` remains the
  honest classification — `passing` would let auto-merge land a PR whose CI never
  reported. Fixed in `083939b0` by bounding the watch at `PR_CHECKS_WATCH_S`
  (1800s) *and* adding a pending arm, since bounding alone would have been worse:
  the post-watch parse exits 1 only for the `fail` bucket, so a bounded watch
  would have fallen through to exit 0 — a false green. The pending arm resolves
  the head sha and delegates to `dispatch_ci_verdict_rest`, so the orphan rule
  has no third implementation. New suite, 14 cases.
- **The up-to-date gate had no cap.** Sync → CI restarts → `main` moves → sync
  again; self-limiting per *head*, not per *PR*. Fixed in `f3fcdbce` with a
  stateless counter: the PR's own multi-parent commit count *is* how many times
  `main` has been merged into the head, read from GitHub with nothing to persist
  or race. At `GRAPH_AUTO_MERGE_SYNC_MAX` (default 3) it emits
  `held <id> (sync-cap: <n> syncs)` — a hold, because a thrashing PR is a
  person's call, not a gate failure.
- **Clearing the reservation on an *unknown* daemon read.** This was called a
  judgment call on fail-safe direction; on re-examination it was simply pointing
  the wrong way. `verify_agent_registered_under` returned 1 both for "the
  registry answered and nothing is there" and for "could not ask", and the
  destructive response fired on both — inverting the fail-safe the *pre-launch*
  gate applies, where `unknown` counts as occupied. Fixed in `13f60b5b`: a new
  `verify_agent_registered_under_state` reports `registered|absent|unknown`, and
  only `absent` releases the claim. `unknown` retains it and throws
  `launch-unverifiable`, letting it age out under `reservation_sweep`'s TTL
  rather than releasing on a non-observation.
- **The sweeps' contribution to `--max-run-s` overshoot.** Fixed in `2577c67c`:
  the deadline is re-checked after both sweeps and before the advance, so the
  driver never *begins* new work past its deadline, and the sweep gets smaller
  per-pass caps through the library's existing tunables. A budget bound, not a
  park-policy change — the sweep runs every pass, so a lower cap catches up
  rather than dropping work. The same commit documents what was previously
  discoverable only by watching it happen: the sweep is fleet-wide and takes no
  node filter, deliberately, because it exists for the bootstrap-deadlock case
  where no heartbeat is running at all.

**The verdict cache stays terminal-only — accepted, with the cost measured.**
The reviewer was right that the "permanently shadowed" rationale was false: the
cache is per-tick (`dispatch-select-tick` mktemps it and `rm -rf`s it on EXIT,
and `dispatch-ladder-run` unsets the var before every reconciler call). The
residual cost is roughly 2–3 refetches per in-flight PR per tick, which is
accepted rather than deferred. No code change; the rationale is corrected in
`d92903a7`.

The unit tests prove the branches; they do not prove the ladder. The seven live
e2e steps run after this merges.


.claude/skills/dispatch-propagate/scripts/lib.sh
```diff
@@ -804,22 +844,72 @@ dispatch_ci_verdict_rest() {
...
+  # lowercase enums, so uppercase every entry; orphaned rows are replaced
+  # wholesale by the already-uppercase COMPLETED/STALE pair.
+  local adapted
+  adapted=$(jq --argjson stale "$stale_json" '
+    map(
+      if ((.status // "") != "completed")
+         and ((.conclusion // "") == "")
+         and ((.check_suite.id // "" | tostring) as $sid
+              | $sid != "" and ($stale | index($sid)) != null)
+      then {status: "COMPLETED", conclusion: "STALE"}
```
natb1, line 906: **Medium — the STALE→`failing` verdict routes the node into a lane that cannot act on it, and whose first step blocks on the very row that never concludes.**

An orphaned row now classifies `failing` (via `dispatch_classify_rollup`'s `STALE` arm), so the fix interrupt fires — `_gate_maybe_interrupt`, or `reconcile-graph-review-stall` for a `pending-merge` node — and `/fix-checks` launches.

Two problems in that lane:

1. `/fix-checks` Step 3 (`fix-checks/SKILL.md:276-284`) runs `run-pr-checks-wait.sh <pr>`, whose first act is `gh pr checks "$pr_number" --watch` (`run-pr-checks-wait.sh:66`). `--watch` blocks until *every* check concludes, with no timeout. The orphan row is precisely the one that never concludes, so the wait never returns: the pin moves out of `graph-select-target`'s `pending-ci-guard` and into a live worker session that burns its Bash timeout and its slot.
2. There is no failing check for the Step 4 reproduce subagent to map to a command. The header comment promises "the fix lane's normal budgeted re-push (a new head sha, fresh checks, orphan gone)", but the outcome fix-checks actually reaches on a green-everywhere PR is *generic no-repro*, which explicitly **pushes nothing** (`fix-checks/SKILL.md:357-359`). No new head sha, no fresh checks — the node cycles the fix lane until its attempt budget parks it.

If the intent is "force a fresh head sha", the orphan wants a lane that actually re-pushes, not the generic red-CI route.

.claude/skills/dispatch-propagate/scripts/lib.sh
```diff
@@ -804,22 +844,72 @@ dispatch_ci_verdict_rest() {
...
+  }
+
   local verdict
   verdict=$(dispatch_classify_rollup "$adapted")
 
-  # Memoisation miss: persist the verdict for subsequent ticks before printing.
-  if [[ -n "$cache_file" ]]; then
+  # Memoisation miss: persist TERMINAL verdicts only (see the header note) before
+  # printing.
+  if [[ -n "$cache_file" && "$verdict" != "pending" ]]; then
```
natb1, line 920: **Medium — the reason given for never caching `pending` does not hold, and the cost it buys is real.**

The header says a cached `pending` would leave the orphan rule "permanently shadowed by the very cache entry the orphan produced". But the cache is not long-lived: `dispatch-select-tick` creates it fresh per tick and deletes it on exit —

```
DISPATCH_CI_VERDICT_CACHE=$(mktemp -d "$CLAUDE_JOB_DIR/tmp/ci-verdict.XXXXXX")   # :296
export DISPATCH_CI_VERDICT_CACHE                                                  # :300
trap '_dlog_select_emit; rm -rf "$DISPATCH_CI_VERDICT_CACHE"' EXIT                # :304
```

and `dispatch-ladder-run` `unset`s it before every reconciler call. So a `pending` entry can survive at most one tick; the worst case it could ever have caused is a one-tick delay in noticing an orphan, never a permanent shadow.

What the change does cost is the intra-tick memoisation this helper exists for. Several consumers query the same head oid within one tick (`dispatch-ci-ready`, `dispatch-reconcile-ready`, `graph-select-target`, `graph-auto-merge`, `dispatch-context-pack`, `reconcile-graph-review-stall`). Every repeat query on a *pending* sha — i.e. every in-flight PR, the common case — is now a full paginated `check-runs` fetch **plus** one `check-suites/<id>` GET per distinct pending suite, where it used to be a file read. On a fleet with N in-flight PRs that is several times the REST spend per tick, every tick.

If a bounded refresh is genuinely wanted, an mtime-based TTL on the pending entry would keep the intra-tick memo while still recomputing.

packages/intentionsutil/scripts/graph-commit
```diff
@@ -1312,6 +1312,31 @@ rebase_in_progress() {
...
+# orphan local commit while the graph store's only write path stayed shut
+# (tactic-orphaned-check-run-pins-pending-ci-guard).
+#
+# Deliberately three-valued in effect: an UNREADABLE suite is a non-observation,
+# not evidence of an orphan, so a failed/empty lookup returns 1 and the caller
+# keeps polling. That is the safe direction — a non-observation can only cost
+# the caller the existing pending timeout (rc 1), never a green verdict.
+#
+# Uses `gh api ... --jq` so gh runs jq on its in-memory JSON (shell-json.md).
+check_suite_concluded() { # <suite-id>
```
natb1, line 1343: **Medium (API cost) — this lookup is unmemoised and runs per pending row per poll.**

`await_checks` calls `check_suite_concluded` inside its `while read` loop, so one REST call per *pending required row* per poll. With the defaults (`CHECK_POLL_SECONDS=7`, `CHECK_TIMEOUT_SECONDS=180`, `MAX_PUSH_ATTEMPTS=5`) that is ~26 polls per attempt × 5 attempts; while all four required contexts are still pending that is up to ~500 extra `gh api` calls per `graph-commit` invocation, on top of the ~130 check-run polls — roughly 4× the green-wait's whole API budget, paid on every normal slow CI wait, fleet-wide.

The four required contexts routinely share **one** check suite — this PR's own fixture is the real shape (`test-graph-commit.sh`: all four rows carry `check_suite.id` 85480333626) — so a per-poll memo keyed on suite id collapses 4 calls to 1 with no behavioural change. It must be reset each poll (a `not concluded` answer cached across polls would hide the transition this predicate watches for) and a *failed* lookup must not be cached at all, since that is a non-observation.

Second, smaller point on the same call: it runs as a subprocess inside a `while read … done <<<"$counts"` loop, so it inherits the here-string as stdin. `lib-frozen-session-park.sh` drains its candidate rows into an array specifically to stop `park-node` eating loop stdin; the cheap equivalent here is `</dev/null` on the `gh api`.

Both are applied in the working tree alongside this review.

.claude/skills/dispatch-ladder/scripts/dispatch-ladder-advance
```diff
@@ -224,10 +269,81 @@ fi
...
+if [[ "$KIND" == "strategy" || "$PHASE" == "align-tactics" ]]; then
+  LAUNCH_CWDS=("$PROJECT_ROOT" "$NODE_WT")
+else
+  LAUNCH_CWDS=("$NODE_WT" "$PROJECT_ROOT")
+fi
+
+launch_unverified() { # <spawn-cwd>...
+  # Release the claim: a spawn that produced no session must not hold the node
+  # for the handoff TTL while nothing works it.
+  reservation_clear "$NODE_ID" || true
```
natb1: **Medium — this releases the claim on an UNKNOWN daemon read, inverting the fail-safe the pre-launch gate applies 180 lines above.**

`verify_agent_registered_under` cannot tell "no session" from "could not ask". `claude_sessions_under` returns 1 for a daemon that is unreachable, for whitespace-only output, and for an uncorroborated `[]` (`lib-claude-agents.sh:590-606`), and the verify loop treats every one of those the same as an empty registry. So a daemon hiccup lasting through the verify budget (5 attempts × 3 s, twice over — once per candidate cwd, so ~24 s, not the ~12 s the comment above states) produces `throw launch-unverified` **and clears the reservation**, while a real worker may be booting into that worktree right now.

That is the opposite of the policy this same script applies before the spawn, where the comment is explicit that `unknown` folds toward occupied and must never manufacture an intervention (`:169-173`), and of `worktree_occupancy_state`'s own contract. The consequence is the duplicate-worker shape: the operator follows the stderr advice, re-runs, the still-unregistered first worker leaves the pre-launch check reading `free`, and a second session is spawned into the same worktree.

A minimal fix that keeps the intended behaviour for the definite case: probe once for definiteness (`claude_sessions_under "$cwd" >/dev/null`; rc 1 = unknown) and keep the reservation when the read was not definite — still throwing, but without dropping the claim on no evidence.

.claude/skills/dispatch-propagate/scripts/graph-auto-merge
```diff
@@ -302,6 +361,66 @@ while IFS=$'\t' read -r id pr parked; do
     continue
   fi
 
+  # ---- Up-to-date gate: sync a behind branch instead of merging it ----------
```
natb1, line 389: **Medium — the gate has no cap and no bypass, so on a main that moves faster than a CI cycle a PR can sync forever and never merge.**

"Repeated syncing is self-limiting: each sync changes the head oid, so the same head is never synced twice" is true of a *head*, but not of a *PR*. The loop is: sync → CI restarts on the new base (10-20 min) → CI green → next tick compares against a main tip that has moved in the meantime → `diverged` → sync again. Nothing bounds that; the only exit is main standing still for a whole CI cycle.

The fleet's own merge rate is the input here, and the effect is not just a stalled PR: because this is the *only* autonomous merge path, throughput collapses to roughly one merge per CI duration (whichever PR happens to win the race), and every loser pays a full CI run per tick-cycle in Actions minutes. That is a large behavioural change to ship without an escape hatch.

GitHub's own "require branches to be up to date" has exactly this property, which is why it is normally paired with a merge queue that serialises and batches. Worth considering at least one of: an attempt cap after which the sweep merges anyway (the main-health gate above already catches the red-main outcome this is defending against, and `dispatch-diagnose-main` + `dispatch-graph-main-red-sync` already own the recovery), or a sync budget per node recorded like `execution.attempts.conflict`, or ordering candidates so only the highest-precedence behind PR is synced per sweep.

.claude/skills/dispatch-propagate/scripts/graph-auto-merge
```diff
@@ -85,10 +104,13 @@
 # node --import tsx/esm needs the npm cache (see .claude/rules/sandbox.md). This
 # script sets NOTHING sandbox-related; that is the caller's responsibility.
 #
-# Stdout protocol: exactly `merged #<pr> (<id>)` per PR merged, and
-# `held <id> (<reason>)` per hold, where <reason> is one of `office-hours`,
-# `missing-stamp`, or `scope-stale`; NOTHING for a plain skip (an ineligible
-# PR is silently passed over).
+# Stdout protocol: exactly `merged #<pr> (<id>)` per PR merged,
+# `synced #<pr> (<id>)` per PR whose branch was updated against main instead of
```
natb1, line 123: **Low — the new `synced` line is invisible to `dispatch-ladder-run`, whose reconcile pass only greps `^held` and `^merged #`.**

`dispatch-ladder-run:578-591` matches `^held $NODE_ID ` and `^merged #`, and everything else falls into the `else` arm that logs `no-merge` with the detail "graph-auto-merge merged nothing for $NODE_ID". A `synced #N (<id>)` therefore reads, in the journal and in the state file, as if nothing happened — when in fact the PR's head was rewritten and CI restarted from zero. The driver's own protocol enumeration in its header (`dispatch-ladder-run:126-130`) lists only `held` and `merged #N`, so it is now stale too.

That matters because a ladder run that ends here halts `idle` (exit 10) once `--ci-wait-s` runs out, and the SKILL's exit-10 row sends the reader to "the PR's checks" without any hint that this run is what reset them.

Added an `elif grep -q "^synced #"` arm that logs the line (no change to loop control) plus the header entry, in the working tree alongside this review.

.claude/skills/dispatch-ladder/scripts/dispatch-ladder-run
```diff
@@ -646,6 +678,25 @@ while :; do
...
+  # an escalating phase leaves a job that `dispatch-self-close` HOLDs for want
+  # of a `node-terminal` marker (dispatch-self-close:48-101, and :70-77 for why
+  # it fails toward HOLD), which dispatch-ladder-await then reads as
+  # `done-held` → `throw <id> held-session` (dispatch-ladder-await:322-334),
+  # with the node left unselectable because worktree_has_live_session is
+  # name-keyed. Nothing in this process would ever clear that. So run the
+  # sweep the tick runs, unchanged — it consumes the escalation reason and
+  # parks the node to office_hours, turning a silent wedge into a halt a person
+  # can read. NO park policy of the driver's own: sequencing, not gating.
+  terminal_without_disposition_sweep 1>&2
```
natb1: **Low — the sweeps are unbounded work added before `check_deadline`'s next chance to fire.**

`check_deadline` runs at the top of the pass; both sweeps run after it and before the advance. `terminal_without_disposition_sweep` parks up to `DISPATCH_TERMINAL_DISPOSITION_PARK_MAX` (default 2) nodes per call, each under `timeout` `DISPATCH_TERMINAL_DISPOSITION_PARK_TIMEOUT_S` (default 120 s), plus two daemon queries, a `find` over the transcripts root, and up to `1 + park_max` `git fetch`es. So a single pass can spend ~4 minutes on *other* nodes' parks before this node's advance runs, and `--max-run-s` can overshoot by that much per pass.

Also worth noting in the SKILL prose: this makes a driver documented as "walk ONE tactic node" take park actions across the whole fleet. That is defensible on the heartbeat-stopped host it exists for (and `reservation_sweep` is already fleet-wide), but it is a wider blast radius than the skill's opening line implies, and an operator running `/dispatch-ladder` on one node will not expect two unrelated nodes to land in the office-hours queue as a side effect.

.claude/skills/dispatch-propagate/scripts/graph-auto-merge
```diff
@@ -302,6 +382,99 @@ while IFS=$'\t' read -r id pr parked; do
...
+      # to persist, sweep, or race, and correct across processes and hosts.
+      #
+      # Two honest caveats:
+      #   - a merge commit the PR AUTHOR made by hand also counts. That is
+      #     correct: it is still a sync of main into the head.
+      #   - `per_page=100` bounds the listing (gh_api_array does not paginate),
+      #     so a PR with more than 100 commits UNDER-counts and fails OPEN,
+      #     permitting another sync. Better than failing closed on a large PR.
+      rc=0
+      SYNC_COUNT=$(gh_api_array "repos/{owner}/{repo}/pulls/$pr/commits?per_page=100" \
```
natb1, line 466: **medium — the sync cap counts merges this gate never made, so a long-lived PR is held the first time it falls behind.**

`SYNC_COUNT` is every merge commit in the PR, but `commit-merge-push` runs `git merge origin/main` on the node branch at the end of **every** implement unit, and again on the review-fix and qa-fix fix passes (`.claude/skills/dispatch-propagate/scripts/commit-merge-push:188-196`). Each of those that is not a fast-forward leaves a merge commit on the head.

Concrete scenario: a 4-unit tactic lands 4 per-unit merges of `origin/main` while main is moving, then reaches `phase: review` with the `reviewed` marker. The very first tick that finds it behind reads `SYNC_COUNT=4 >= SYNC_MAX=3` and emits `held <id> (sync-cap: 4 syncs)` — a hard `throw` in `dispatch-ladder-run` (`:606`) — even though this gate has never synced the PR once and no CI thrash has occurred. The autonomous merge path is then closed for that PR until a person intervenes, and the bigger the PR the more certain it is.

The header calls out the author-hand-merge case as acceptable, but the mechanical per-unit merges are not an edge case — they are the normal shape of every multi-unit PR. A discriminator would keep the cap honest: GitHub's `update-branch` merge commit is `Merge branch 'main' into <head>`, while `commit-merge-push`'s is `Merge remote-tracking branch 'origin/main' into <branch>`, so filtering the count on the commit message (or on the `web-flow` committer) counts only this gate's own syncs.

Not auto-fixed: the counting rule is documented as deliberate, and the discriminator depends on GitHub's exact message format — your call.

.claude/skills/dispatch-propagate/scripts/run-pr-checks-wait.sh
```diff
@@ -60,10 +80,14 @@ if [[ -n "$output_file" ]]; then
...
-# here only means some check ended non-green; the authoritative verdict is parsed
-# from --json below, so watch's exit status is deliberately ignored.
-gh pr checks "$pr_number" --watch > /dev/null 2>&1 || true
+# Wait for all checks to complete (--watch blocks until done), bounded so a row
+# that never reports cannot pin this call forever. A non-zero exit here only
+# means some check ended non-green, or that the bound fired; the authoritative
+# verdict is parsed from --json below, so watch's exit status is deliberately
+# ignored. Falling out of the bound is not a green signal — the pending arm
+# after the fail-bucket parse below handles it.
+watch_bound="${PR_CHECKS_WATCH_S:-1800}"
```
natb1: **medium — the 1800s default outlives the 600s Bash-tool ceiling the docstring cites as the reason for the bound, so the orphan case this change targets still returns no verdict.**

The header says the bound exists because "this script is invoked by a model through the Bash tool, whose own ceiling is 600s, so an unbounded watch burns the whole tool call and returns no verdict at all" — but 1800 > 600, so on exactly the failure mode being fixed the tool call is killed first and nothing below line 90 ever runs.

Concrete scenario: `/fix-checks` Step 3 runs `run-pr-checks-wait.sh <pr>` (`.claude/skills/fix-checks/SKILL.md:279`) on a PR carrying an orphaned check run. `gh pr checks --watch` blocks forever on that row, `timeout 1800` will not fire for 30 minutes, and the Bash tool aborts at ≤600s. The session gets no check summary, no exit code, and none of the new orphan classification — the same wedge as before the change, just now with a `timeout` in front of it.

A default under the ceiling (e.g. `540`) makes the bound reachable on the path it was written for; `PR_CHECKS_WATCH_S` still lets a detached caller ask for longer. `test-run-pr-checks-wait.sh:281-282` asserts the default value, so it would move with it.

Not auto-fixed: 1800 is an explicit, tested constant — lowering it is your call, not a review edit.

.claude/skills/dispatch-propagate/scripts/graph-auto-merge
```diff
@@ -302,6 +382,99 @@ while IFS=$'\t' read -r id pr parked; do
...
+  fi
+  # `continue` inside a `case` continues the enclosing `while` (a `case` is not
+  # a loop) — correct and intentional. Every non-up-to-date arm ends in one, so
+  # a candidate can never both `synced` and reach the merge PUT.
+  case "$CMP" in
+    ahead|identical)
+      # ahead: main's tip is an ancestor of the head → up to date, merge.
+      # identical: head == main tip → up to date (degenerate). Fall through.
+      ;;
+    behind|diverged)
```
natb1, line 410: **low — folding `behind` into the sync arm mutates a PR whose commits are already on main, and leaves it permanently unmergeable.**

`behind` means the head is an *ancestor* of the main tip: the PR has no commits main lacks. Sending it to `gh_pr_update_branch_rest` merges main into that head, which fast-forwards the branch onto main's tip and leaves an open PR with an **empty diff**.

Concrete scenario: a node's branch commits reach main out of band (a hand push, or a resolve-hold that pushed `HEAD:main`). The PR is still `OPEN` and `MERGEABLE`, CI is green, the stamp is fresh — so this arm syncs it. Next tick the compare answers `identical`, the gate falls through to the squash-merge, and GitHub refuses a PR with nothing to merge — `HARD_ERROR=1`, i.e. `exit 1`, i.e. `halt 11 throw` in `dispatch-ladder-run:604` on every subsequent tick. The header's plan ("reconcile-graph-merged absorbs the genuinely-already-merged case on a later tick") is defeated by the branch mutation this arm performs first, since the reconciler keys off `mergedAt`, which stays null.

A plain `continue` for `behind` (skip, sync nothing, merge nothing) leaves that absorption to the reconciler as documented.

Not auto-fixed: `test-graph-auto-merge.sh:569-577` (n2) pins the current behavior, so this is a deliberate-behavior question rather than a review edit.

packages/intentionsutil/scripts/graph-commit
```diff
@@ -1312,6 +1312,51 @@ rebase_in_progress() {
...
+# 180s timeout across 5 attempts that is ~4x the green-wait's whole API budget.
+# The cache is reset at the top of every poll iteration in await_checks: a
+# suite's status is a statement about a moment, and a `not concluded` answer
+# cached across polls would hide the very transition this predicate watches for.
+# A FAILED lookup is not cached at all — it is a non-observation, not an answer.
+declare -A SUITE_CONCLUDED_CACHE=()
+check_suite_concluded() { # <suite-id>
+  local suite_id="$1" status
+  [[ -n "$suite_id" && "$suite_id" != "-" && "$suite_id" != "null" ]] || return 1
+  if [[ -n "${SUITE_CONCLUDED_CACHE[$suite_id]:-}" ]]; then
```
natb1: **low — the memo misses on a cached empty status, so the per-poll fan-out it prevents comes back. (Fixed in the working tree.)**

A successful lookup whose payload carries no `.status` caches the empty string (`--jq '.status // ""'`, line 1358), but `-n` reads an empty cache entry as a MISS. Concrete scenario: the four required contexts share one suite (as the fixtures do) and the suite read returns a payload without `status` — every one of the four rows then re-runs `gh api` in the same poll, which is exactly the 4x budget blow-up the memo was added to stop.

Applied fix: test presence rather than non-emptiness.

```suggestion
  # PRESENCE, not non-emptiness: a successful lookup that returned no `.status`
  # at all caches the empty string, and an `-n` test would read that answer as a
  # cache MISS and re-query the same suite for every remaining row sharing it —
  # exactly the per-row fan-out the memo exists to prevent. A failed lookup is
  # still never cached (it returns before the assignment below).
  if [[ -n "${SUITE_CONCLUDED_CACHE[$suite_id]+set}" ]]; then
```

PR #3074, opened 2026-08-12
Implement the 2026-08-12 session-evaluation round

natb1:
Implements all six draft tactics recorded by the 2026-08-12 `/align` round
(`bc897a5e`), by hand on one branch. Evaluation now fires per phase, findings
merge into one ledger node per distinct finding carrying summary metrics, and
one audit instrument serves both the fleet and a single node.

Each unit was implemented in its own subagent at the plan's recommended model,
then reviewed, verified and committed here.

## What landed

| Unit | Tactic | Commit |
|---|---|---|
| 1 | `tactic-measured-impact-schema` | `3ab2fac2` |
| 2 | `tactic-eval-finding-ledger` | `3ab2fac2`, `59396586` |
| 3 | `tactic-ladder-per-phase-evaluation` | `0b139586` |
| 4 | `tactic-audit-instrument-scoping` | `f9af1a69` |
| 5 | `tactic-audit-cache-efficiency-lens` | `f9af1a69` |
| 6 | `tactic-audit-permission-friction` | `f9af1a69` |

No unit was dropped.

## Premises this PR falsified

Three tactic-body premises were wrong or narrowed by exploration. Stating them
plainly, because the nodes read as satisfied otherwise:

1. **`prune-unreferenced` does not exist.** No such script, and no
   unreferenced-node query anywhere in the repo. `graph-commit --prune` is a
   content-blind deletion primitive with no exemption logic. The exemption
   landed in the owed-prune census (`graph-census-debt.ts`) instead, following
   the live `isCensusNode` precedent.
2. **The ladder's evidence was not pre-instrumented.** The node said "already
   exists, nothing new to instrument." `log_event` wrote five keys with
   `elapsed_s` / `window_s` / `await_repolls` buried inside the free-form
   `detail` string, and **no per-phase session id exists at all** —
   `dispatch-graph-execute` reuses `--name "$id"` every phase. The figures are
   promoted to numeric fields; scoping resolves as `--node` plus `LAUNCH_EPOCH`.
3. **Permission approval round-trips are not derivable.** A transcript records
   denials and blocks, never an approval nor the wall-clock spent at a prompt.
   Lens 12 measures rejections, classifier denials, policy blocks and sandbox
   overrides only, and says so rather than presenting a denial count as an
   approval count. Recorded in the tactic node body, so the gap lives in the
   graph and not only here.

Two further premises were falsified *favorably* during implementation:

- The ~1000-line jq program needed **no change** for scoping — there is exactly
  one point where the input set is chosen.
- `err_signature`'s PATH-collapse **does** merge the per-worktree variants of
  the worktree-isolation refusal. The near-duplicates are six genuinely distinct
  refusal reasons in the hook's own message, not path variation. A test now
  locks the collapse in. (A real fragmentation problem does exist elsewhere:
  auto-mode denials embed a per-call `Reason:`, so they fragment across
  `top_signatures`; documented as a caveat on the lens.)

## Not verified by this PR

The ledger's done-when criterion that **"open ledger-entry count does not grow
faster than entries retire"** is **not verified here**. It needs at least two
evaluation windows and cannot be demonstrated by this change. It remains a
design expectation.

The per-phase evaluator's **cost is also unmeasured**. It starts at `sonnet`,
matching the sibling one-off jobs; rsi condition 14's "no extra budget cost"
clause was retired on the expectation that the measured cost becomes the unit's
acceptance. That measurement needs a real ladder run.

## One unrelated fix, deliberately included

`9706c70a` re-registers the lifecycle sensor. `LIFECYCLE_SENSOR_NAME` must equal
`strategy-graph-native-dispatch`'s `success_signal.sensor` verbatim — it is the
registry key — and the `/align` round at `56039748` appended a park-cause clause
to the node prose without updating the constant, so the sensor silently stopped
being read. Main stayed green because pushes touching only `intentions/` never
trigger the intentionsutil suite; the failure surfaces only on a PR that also
touches that package, which this one does.

The park-cause reading that clause names is **not implemented**. Re-registering
restores the three readings that are implemented, which is strictly better than
reading nothing. The gap is recorded in a comment rather than papered over by
trimming the clause — deciding between implementing the reading and revising the
prose is an author call, not this PR's.

## Verification

- `npm test --prefix packages/intentionsutil` — 975 passed
- `npx vitest run --project office-hours --project packages/intentionsutil --root .` — 1409 passed (what CI selects)
- `validate-graph.ts intentions` — ok, 615 nodes, 0 unresolved prose refs
- `test-aggregate-usage.sh` — 257 passed
- `test-dispatch-ladder-run.sh` — 155 passed
- `test-dispatch-eval-finding.sh` — 62 passed
- `test-decision-log-isolation.sh` — 204 passed
- `run-lint.sh`, `run-typecheck.sh` — pass

CI caught one defect local runs missed: the new ledger suite failed
`test-decision-log-isolation.sh`'s Part B ratchet, which only runs inside the
full `run-unit-tests.sh` sweep. Fixed in `395eea48` by sourcing the shared
`test-helpers.sh` — which pulls in the decision-log guard that does the real
sandboxing — rather than by adding the suite to `KNOWN_UNISOLATED`.

## Node disposition

The six tactics stay `phase: null` until this merges. Transitioning them to
`done` before merge would mark nodes complete whose implementation is unmerged —
the integrity defect `graph-census-debt.ts` warns about. They will be
transitioned through `graph-commit` after merge, with `mergedAt` /
`mergeCommitSha` recorded in `execution.completion`.


packages/intentionsutil/scripts/read-sensors.ts
```diff
@@ -1293,18 +1299,29 @@ export function makeIntentionStoreSensor(
...
+ * `graph-commit`, which owns every write under `intentions/`. So there is a
+ * KNOWN WINDOW between the two landings in which the two strings differ and this
+ * sensor is de-registered. That is expected, not a defect — but it must be
+ * closed, and the check for it is read-sensors' UNREGISTERED-SENSOR COUNT (the
+ * `skipped (unregistered sensor)` figure and its named tail), not the readings
+ * count. The readings count moves for unrelated reasons on every run and will
+ * not show this at all; the unregistered count rises by exactly one while the
+ * window is open and falls back when the node prose lands.
  */
 export const RSI_SENSOR_NAME =
```
natb1, line 1318: The rsi sensor is de-registered by this change. `RSI_SENSOR_NAME` is rewritten here, but `intentions/strategy-recursive-self-improvement.md:1063` still carries the old prose ("the rsi-plan.md metrics section — … rendered by render-rsi-plan.ts each iteration, …"). The registry matches the constant against the node prose character-for-character, so on merge `strategy-recursive-self-improvement` stops getting a reading entirely and moves into read-sensors' `skipped (unregistered sensor)` tail — the pause/backlog/parked/worktree/token segments this PR keeps alive are all silently lost.

The doc comment added just above acknowledges the two-commit window and says it must be closed, but no matching `graph-commit` of the node prose is in this branch. Land the node-prose edit before/with the merge, and confirm the unregistered-sensor count returns to its prior value (it rises by exactly one while the window is open).

(The sibling `LIFECYCLE_SENSOR_NAME` was verified to match `strategy-graph-native-dispatch` — including the curly apostrophe — so only this one is open.)

packages/intentionsutil/scripts/attribute-spend.ts
```diff
@@ -0,0 +1,253 @@
...
+ *   outspending dispatch.
+ * - `>=`, not `>`: a rival that has merely CAUGHT dispatch is already the
+ *   condition the strategy says to review.
+ * - An all-zero window flags nothing. With no spend measured, every rival ties
+ *   dispatch at 0 and `>=` would fire on every empty aggregate.
+ */
+export function spendDeviation(spend: WorkflowSpend[]): SpendDeviation | null {
+  const dispatch = spend.find((s) => s.workflow === "dispatch");
+  const rivals = spend.filter((s) => s.workflow !== "dispatch" && s.workflow !== "other");
+  const measured = spend.some((s) => s.priceProxyUsd > 0);
```
natb1: The measured-window guard counts `other`, so the flag false-fires on a window with no attributed spend at all. `attributeSpend` always returns all four rows, so for a window whose sessions are all unmapped skills (dispatch 0, office-hours 0, rsi 0, other 90), `measured` is true via `other`, and `bigger` then collects office-hours and rsi because `0 >= 0`. The CLI prints `SPEND-DEVIATION FLAG: dispatch (0%) does not outpace office-hours (0%), rsi (0%)` — a review trigger naming rivals that spent nothing. That is exactly the all-zero case the guard is documented to suppress; the honest reading of that window is "`WORKFLOW_SKILLS` needs extending", which the `other` row already says.

```suggestion
  if (dispatch === undefined) return null;
  const measured =
    dispatch.priceProxyUsd > 0 || rivals.some((r) => r.priceProxyUsd > 0);
  if (!measured) return null;
```
(then drop the now-duplicated `if (!measured || dispatch === undefined) return null;` line below)

Applied in the working tree, with a regression test in `test/spend.test.ts`.

.claude/skills/dispatch-propagate/scripts/dispatch-eval-finding
```diff
@@ -0,0 +1,804 @@
...
+  absent)
+    # First occurrence: mint. recurrence_count starts at 1 and first_seen is
+    # stamped once, here, and never rewritten.
+    if ! origin_main_ref_ok; then
+      log "origin/main does not resolve in $REPO_ROOT (unfetched clone, or a pruned/stale remote-tracking ref) — refusing to mint $ID without a trustworthy pre-write blob"
+      exit 69
+    fi
+    PRE_EXISTED=0
+    [[ -e "$NODE_FILE" ]] && PRE_EXISTED=1
+    PRE_BLOB="$(origin_blob || true)"
```
natb1: The mint path can silently reset a landed `recurrence_count` to 1. `classify()` maps EVERY `readNode` failure to `absent` — a half-written `intentions/<id>.md`, a schema-invalid hand edit, a node whose new required field the local checkouts schema rejects, or a file this checkout has not fetched — not only a genuinely new finding. On any of those the `absent` branch mints a fresh node with `recurrence_count: 1`, `write-node` overwrites the file, `graph-commit` lands it, and the entrys whole payload (the recurrence figure, `first_seen`) is gone to git history. It hits hardest exactly where divergence 2 says it matters most: the entries that have recurred many times.

`PRE_BLOB` is already read here, so the fix is to treat the LANDED blob, not the classification, as the authority on existence: if `origin/main` carries `intentions/<id>.md`, this is a recurrence this checkout cannot read, and the local file is what needs repairing — refuse with a clear error rather than minting over it.

Applied in the working tree (guard + refusal message naming the `git checkout origin/main --` remedy); the now-unreachable restore-from-blob arm in the mint rollback was dropped with it. `test-dispatch-eval-finding.sh` stays 77/77.

.claude/skills/dispatch-ladder/scripts/dispatch-ladder-run
```diff
@@ -396,28 +453,126 @@ write_state() {
...
+EVAL_LAUNCH_PHASE=""
+EVAL_LAUNCH_EPOCH=""
+EVAL_SPAWNED_KEY=""
+
+spawn_phase_eval() {
+  local phase="${1:-}" since="${2:-}"
+  [[ -n "$phase" && -n "$since" ]] || return 0
+  local key="$phase@$since"
+  [[ "$key" != "$EVAL_SPAWNED_KEY" ]] || return 0
+  EVAL_SPAWNED_KEY="$key"
```
natb1: The dedup key is claimed before the spawn is attempted, so a FAILED spawn consumes the owed evaluation. `EVAL_SPAWNED_KEY="$key"` runs ahead of the `dispatch-spawn-job` call below; when that call exits non-zero (`spawn-failed`) the phase is recorded as already evaluated, so `halt()` — the one place a run pays what it owes — skips it. Concretely: the implement phase reaches its boundary, the spawn fails transiently (daemon busy, registration check times out), the run later halts on that same phase, and nothing ever evaluates it, which is the "most defect-rich runs produce no review" hole the halt-side spawn was added to close.

Not applied: `test-dispatch-ladder-run.sh`s "a spawn that fails is a warning" case asserts exactly one `eval spawn-failed` event, so moving the assignment into the `rc == 0` arm is a deliberate behaviour change for the author to make (or to close as intended — a failed spawn is not retried on purpose), not a review fix.

natb1:
## Review dispositions (`/code-review high`)

Four findings. Three fixed with regression tests, one already reported in the PR body.

**1. `RSI_SENSOR_NAME` vs the node's `success_signal.sensor` — known, reported, not fixable here.**
This is item 1 of the PR body. The two halves land through different mechanisms — code by PR, node prose by `graph-commit` — so there is no atomic version. The window is measured, not predicted: the unregistered-sensor count goes 57 → 58. Closing it is the first step of the follow-up graph round, and the exact replacement text is already in the code so the graph write can copy it verbatim.

**2. `spendDeviation`'s measured-window guard counted `other` — fixed** (`4d4d9696`).
A window whose whole spend sat in the unattributed remainder read as "measured", and then dispatch (0) tied every rival (0) under `>=` and the flag fired naming workflows that had spent nothing. The guard now reads the attributed rows only. Regression test added; it fails against the old predicate.

**3. `classify()`'s catch-all `absent` could mint over a landed ledger entry — fixed** (`d1a339cf`).
`absent` covers every `readNode` failure, not only a new finding, so an unfetched or half-written local file would overwrite a landed entry with `recurrence_count: 1` — the count reset divergence 2 exists to prevent. The landed blob now decides existence; a pass refuses with exit 1 and names the `git checkout origin/main --` that repairs the local copy. The now-unreachable restore-from-blob arm in the mint rollback went with it. New case (14) fails if the guard is dropped; existing case (3) fails if it over-refuses.

**4. `EVAL_SPAWNED_KEY` claimed before the spawn was attempted — fixed** (`08ba5447`), against the reviewer's "author call" recommendation.
The deciding fact is provenance: `spawn_phase_eval` does not exist on `origin/main` at all — it was introduced by this branch's `0b139586`. So this is a defect this PR ships, not one it inherits, which puts it in scope. A failed boundary spawn evaluated nothing, yet claimed the key, so `halt()` — whose stated job is to pay the evaluation a halted run owes — deduped against that claim and paid nothing.

The key is now claimed on an answer (`spawned`, `deduped`, an unrecognized answer, or no executable spawner — a retry cannot make a binary executable). `rc != 0` claims nothing, so `halt()` retries exactly once; `halt` is the one terminal path and runs once, so the budget is one retry, not a loop. The reviewer's blocker — "`test-dispatch-ladder-run.sh` asserts exactly one `spawn-failed` event" — was the *old* behaviour asserted; that assertion now reads 2, and a new case asserts the retry actually lands the evaluation.

### Verification on `4d4d9696`

`test-dispatch-eval-finding.sh` 83/83 · `test-dispatch-ladder-run.sh` 185/185 · `run-unit-tests.sh` all suites · `run-lint.sh` all checks · `tsc --noEmit` clean · `validate-graph` ok, 623 nodes, 0 unresolved prose refs.

Each fix was mutation-tested: reverting it fails the new test, and inverting the eval-finding guard fails the pre-existing mint case too.

